PR #15, opened 2015-12-11
More ACL rules

mscifo:
These changes were specifically made to allow AWS Lambda to be able to send requests to an ES cluster without having to open the entire cluster to the outside world.  

Currently, requests to an ES cluster (not an AWS managed ES cluster) from AWS Lambda may come from a large unknown pool of IP addresses.  By allowing access via an API Key that is made available to an AWS Lambda, you can allow access to just that Lambda.  Until AWS Lambdas can be run inside a VPC, this is a sufficient workaround.

Also, if your ES cluster is behind an AWS ELB, the addition of support for creating a rule using a subnet and retrieval of the remote address from the X-Forwarded-For header allows the host based rule parameter to still work since ELB attaches the X-Forwarded-For header to requests and also lives inside a known subnet (assuming your ELB is inside a VPC).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,8 @@ No need to spin up a new HTTP proxy (Varnish, NGNix, HAProxy) between ES and cli
 
 #### Flexible ACLs
 Explicitly allow/forbid requests by access control rule parameters:
-* ```hosts``` a list of origin IP addresses
+* ```hosts``` a list of origin IP addresses or subnets
+* ```api_keys``` a list of api keys passed in via header X-Api-Key
 * ```methods``` a list of HTTP methods
 * ```uri_re``` a regular expression to match the request URI (useful to restrict certain indexes)
 * ```maxBodyLength``` limit HTTP request body length.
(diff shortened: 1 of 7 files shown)


sscarduzio:
Nice one @mscifo! Very useful. I'm a bit scared about the X-Forwarded-For header, as it would allow anybody to bypass the ACLs with an HTTP header. What's your view about it?

mscifo:
I had the same thoughts.  Unfortunately, without support for the X-Forwarded-For header, any client that attempts to connect to an ES cluster that lives behind a load balancer would fail since the requesting IP would always be that of the load balancer.  You could mitigate this by using an API key which is now supported via this PR, but some might not want to go that route.  Do you have any other suggestions on how to deal with this use case?

sscarduzio:
Security wise, the X-Forwarded-For header is on the same level of having an API key (shared secret) with the exception that guessing an IP is infinitely easier than to guess one of your API keys.

On the practical side, allowing by X-Forwarded-For is really handy because you don't need to configure any API key, and in some use cases can be very annoying to do (e.g. Kibana).

One could greatly mitigate the security impact of X-Forwarded-For by configuring the security groups (firewall) to only accept connections from the ELB netmask. In that case, the security level would be almost like configuring the allowed hosts rule.

The above is valid only if the ELB overwrites any incoming X-Forwarded-For header in all circumstances. I assume this is the case, should double check.

So my opinion is that yes, we should accept X-Forwarded-For as an **optional** feature. That is:
* default ACL rule behaviour: don't even look at the X-Forwarded-For header.
* If expressively enabled in a rule configuration: we try to apply the rule first to the host in the header, if the rule doesn't match, retry to apply the same rule using the origin host in the connection as usual.

I think it's also important that we warn people in the README.md about the security aspect we just identified, and we should add a suggestion to configure the firewall to accept connections only from the load balancer as above to compensate.

@mscifo If you can add the extra ACL configuration toggle to enable the feature, I can fix the doc. What do you say? :)

PR #29, opened 2016-03-23
Problem with Indices Rule

paolociccarese:
Hi Simone,
thank you for sharing the plugin! It is turning out to be useful for some of our projects. I just had a small issue when trying to restrict indexes access. I tested this only with Elasticsearch v.2.2.0 but I've looked over the older APIs and I would expect that to be ok for previous versions as well.

I am sharing back to seek for validation as well.

So this is what I have done:

Changed reflection call from Field 'indices' to method 'indices' as the method 'getDeclaredField' gets the fields, regardless of their accessibility but only for the current class, not any base classes that the current class might be inheriting from. And the field 'indices' is private and inherited. Changed it to access the public method instead. By doing that I can now use 'getMethod' and get to the public methods of the superclasses.

### DIFF
--- a/src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/IndicesRule.java
+++ b/src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/IndicesRule.java
@@ -12,7 +12,8 @@ import org.elasticsearch.plugin.readonlyrest.acl.blocks.rules.Rule;
 import org.elasticsearch.plugin.readonlyrest.acl.blocks.rules.RuleExitResult;
 import org.elasticsearch.plugin.readonlyrest.acl.blocks.rules.RuleNotConfiguredException;
 
-import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


sscarduzio:
Hi @paolociccarese, thanks for contributing!
The PR looks very interesting, can you share a bit of background on how you discover this? Under what circumstances would the current code present the issue you fixed?

paolociccarese:
Sure @sscarduzio, I've installed a clean instance of Elasticsearch v.2.2.0 and I've created an index as example (my index). The idea was to be able to limit the access to that index to a GET for any request coming from a machine other than localhost.

I've used your basic configuration:
```
readonlyrest:
    enable: true
    response_if_req_forbidden: Sorry, your request is forbidden.
    access_control_rules:

    - name: Accept all requests from localhost
      type: allow
      hosts: [127.0.0.1]

    - name: Accept only GETs from other host
      type: allow
      methods: [GET]
      indices: [myindex]
```

I've tried to access the index from a different machine and the GET for my index was constantly turned down. 

After some debugging it looked like the invoke for 'indices' was always returning an array with a singe 'null' value. That triggered some more digging and the code detailed in the above comment. Unless I missed something else, I believe it goes down to the fact that java.lang.Class.getDeclaredFields() method returns an array of Field objects including public, protected, default (package) access, and private fields, but excludes inherited fields. And the 'indices' field is private and inherited by ActionRequest. However, the indices() method is public.

The ACLTest was mocking that call so it was passing fine with the previous code as well.

With the same approach, I can now invoke the public method indices(String... indices) and rewrite on the fly the query indexes to do custom aliasing and exclude those indexes the caller does not have access to so that, if the resulting indexes collection is not empty, I can still provide results without triggering an exception... However, I am not sure yet on how to streamline this last part to be helpful in a more general way.

sscarduzio:
Cool: I tested this and basically now we can resolve the indices behind aliases! Very nice and elegant, kudos @paolociccarese!
Keep it coming Paolo, if you need more features for your project let me know! Nice to see this thing is useful :rocket:

PR #106, opened 2016-10-07
tests: swap arguments of assertEquals according to JUnit API

maitai:
The method `assertEquals` has the signature `void assertEquals(Object expected, Object actual)` and therefore the first argument should always be the one that is expected in order to avoid confusing messages such as this one: `expected:<actualValue> but was:<expectedValue>`


sscarduzio:
Yep, makes sense mate :)

Issue #111, opened 2016-10-18
How to deal with url/_search?


teebu:
I know we're setting permission for each index, but this bypasses everything, and returns all indices. 

From this, you can do any query you want. 

`_search?q=message:hellotest1` for example, will give you data for all indices with that field, even if its not in the allowed ones.

sscarduzio:
Thanks for reporting this @teebu.

Just tried how the plugin sees this use case:
```
checking request:{ action: indices:data/read/search, OA:172.17.0.1, indices:[], M:GET, P:/_search, C:, Headers:[]}
```
looks like no indices are associated with this action, so looks like we're touching no indices at all (rather than all of them).

I'm starting to believe this approach of inbound index checks in requests is quite limited, ideally we'd need to (also) analyse/filter the results of the queries. Easier said than done, but I'll have a look.

For now, you can patch the hole with a `url_re` rule, or forbidding the access to the index `<no-index>`

teebu:
A general forbid may be enough.

```
    - name: forbid no index access
      type: forbid
      indices: ["<no-index>"]
```

Then I remembered about the explicit index stuff that ES requires you to disable for logstash to work.

sscarduzio:
Ok @teebu, I got a possible solution.

I'm able to detect if the request at hand is of type "SearchRequest" and if I see it has no reference to a particular index (like the generic `/_search` above). In this case, instead of parsing it as `<no-index>`, I parse it as "_all" index.
 
If one really wants this all indices search to work, you need to add "_all" as allowed index. 

A phase two will convert the absence of indices in a search request as "give me all the indices I'm allowed to see", but this requires #110 .

Will this work for you?

sscarduzio:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/releases/tag/v1.12.0-rc1_es-v2.3.3

teebu:
So, it will block `_all` by default, since it's not on the allowed list, im ok with that.

sscarduzio:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/releases/tag/v1.12.0-rc1_es-v2.4.1

sscarduzio:
this is addressed in 1.12.1

PR #115, opened 2016-10-20
Fixing #110 and #111

sscarduzio:
Fix for:

1. #111 Even if indices rule is set, _search?q=message:hellotest1 would return results from all indices.
2. #110 reverse indices wildcards + aliases support
3. Adding a few colored logs 

⚠️ This code is experimental, I tested it with success on the following multi-user Kibana + Logstash configuration:

```yml
readonlyrest:
    enable: true
    ssl:
      enable: true
      keystore_file: "/elasticsearch/plugins/readonlyrest/keystore.jks"
      keystore_pass: readonlyrest
      key_pass: readonlyrest

    response_if_req_forbidden: Forbidden by ReadonlyREST ES plugin

    access_control_rules:

    - name: "::LOGSTASH::"
      # auth_key is good for testing, but replace it with `auth_key_sha1`!
      auth_key: logstash:logstash
      type: allow
      actions: ["indices:data/read/*","indices:data/write/*","indices:admin/template/*","indices:admin/create"]
      indices: ["logstash-*", "<no-index>"]

    - name: "::KIBANA-SRV::"
      # auth_key is good for testing, but replace it with `auth_key_sha1`!
      auth_key: admin:passwd3
      type: allow

    - name: "::RO+ DEVELOPER::"
      auth_key: ro+:dev
      type: allow
      kibana_access: ro+
      indices: ["<no-index>", ".kibana", "logstash-*", "default"]

    - name: "::RO DEVELOPER::"
      auth_key: ro:dev
      type: allow
      kibana_access: ro
```
I'll release a build on this one, so hopefully @teebu and/or others can validate!
@teebu what ES version do you want the build for?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -6,14 +6,20 @@ then
     sed -i.bak 's|<elasticsearch.version>.*</elasticsearch.version>|<elasticsearch.version>'"${1}"'</elasticsearch.version>|g'  pom.xml && rm pom.xml.bak
 fi
 
-mvn clean package || exit 1
-
+PLUGIN_VERSION=`grep '<!-- plugin_version' pom.xml | sed 's|</b>|-|g' | sed 's|<[^>]*>||g' |xargs`
 ES_VERSION=`grep '<elasticsearch.version' pom.xml | sed 's|</b>|-|g' | sed 's|<[^>]*>||g' |xargs`
 
 sed  -i.bak  "s/^\(elasticsearch.version=\).*/\1$ES_VERSION/" src/main/resources/plugin-descriptor.properties && rm src/main/resources/plugin-descriptor.properties.bak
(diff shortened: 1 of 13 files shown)


nitram4:
Testing this release but, having issues when combining with auth_key_sha1 nad Group function. Only working when not using any limitations. 

This is not working, not accepting password
```
    - name: User
      type: allow
      indices: ["<no-index>", ".kibana*", "logstash-*", logstash-app-*, .marvel-es-1-*,  "default", ".marvel*"]
      kibana_access: ro+
      auth_key: xxxx:xxxx

```
but it works with this :

```
    - name: Admin
      type: allow
      kibana_access: rw
      auth_key: XXX:XXXX
```
and this works also with auth_key_sha1.

Any limitation to this?
Thanks

PR #145, opened 2017-01-06
[PROXY-ISSUE] fix X-FORWARDED-FOR ip check ignored

seokhong:
* Test Environment
  * ElasticSearch : 2.4.3
  * ReadOnlyRest : 1.12.2
  * AWS ELB Proxy with X-Forwarded-For header.

We found a bug that HostsRule ignores ELB's X-Forwarded-For header.
When X-Forwarded-For header's IP does not match, HostsRule retry matching with ELB IP Address.

HostsRule must return the result of X-Forwarded-For header's IP match whether it returns true or false.

For example, we allow host 10.20.30.0/24 and ELB's IP exist in this subnet.
When we try to access other machine with 10.30.30.1, then X-Forwarded-For header's IP does not match. But, HostsRule ignore result and try to match ELB's IP and return success.

### DIFF
--- a/src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/HostsRule.java
+++ b/src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/HostsRule.java
@@ -91,10 +91,7 @@ public class HostsRule extends Rule {
 
     if (acceptXForwardedForHeader && xForwardedForHeader != null) {
       // Give it a try with the header
-      boolean attemptXFwdFor = matchesAddress(xForwardedForHeader, null);
-      if (attemptXFwdFor) {
-        return true;
-      }
+      return matchesAddress(xForwardedForHeader, null);
     }
     for (String allowedAddress : allowedAddresses) {
       if (allowedAddress.indexOf("/") > 0) {


sscarduzio:
Hey @seokhong, thanks for (very well) reporting and fixing this 👍 

The proposed fix would basically let the `accept_x-forwarded-for_header: true` change the behaviour of `hosts` rule to exclusively look into the X-Forwarded-For header. Which is good!

But now you got me thinking.. Ideally, a user *should* be able to restrict both the LB ip/range **and** also declare a range of allowed X-Forwarded-For. This is currently not possible because `hosts` looks **either** the origin ip address **or** the header (when the `accept_x-forwarded-for_header` flag is true).

### Proposal
How about `x-forwarded-for_header` becomes a rule of its own? This way, you can do something like:

```yml
- name: 1
   type: allow
   x-forwarded-for: [10.20.30.0/8] # your clients' IP ranges
   hosts: [10.10.10.0/24] # (optional, if you know the range or ip) your load balancers' IP ranges
```
What do you think? Would this work for you?

kwon37xi:
How about the following solution.
```
type: allow
hosts: [10.10.0.0/16]
```

when x-forwarded-for IP returns false -> just return false.
when x-forwarded-for IP returns true -> validate proxy(ELB, ..) IP and when it returns true -> return true
when when x-forwarded-for IP returns true -> validate proxy(ELB, ..) IP and when it returns false -> return false

In conclusion, the request is allowed when (x-forwarded-for IP && proxy's IP).

sscarduzio:
@kwon37xi your solution is less invasive and simpler, but my question is: do we always know the IP range of our load balancers?

As far as I know, when you set up an ELB in AWS you're given host name, but not a guaranteed-immutable internal IP or IP range. 

As an example, this is how I'd personally go about deploying this under ELB.

1. Elasticsearch nodes are segregated in a VPC (unreachable from internet or other VPCs)
2. Create the ELB and add it to the same Security Group with the Elasticsearch nodes.

NB: Now I don't care what's the IP/range of the ELB because I know the ELB is the only thing that can possibly ever talk to my ES nodes!

3. In ReadonlyREST I can safely ignore the originating IP addresses and only look at the X-Forwarded-For.

In this common scenario: @seokhong solution works (ignores the originating IP if the flag is on), and my solution would also work (new specific rule that only cares about  x-forwarded-for).
But the proposed one (originating ip && x-forwarded-for always in range) would render this setup impossible.

seokhong:
@sscarduzio I think your solution is better.

But I have a question. As you said, we can control ELB's access by using security group. If we don't use cloud, LB's IP is fixed, then we can set firewall policy.

Why should we care about LB's IP in plugin? Maybe I miss something. Could you please let me know some case?

Thanks!

sscarduzio:
@seokhong maybe you want to have a read-only access for external clients (which will access through the LB), and read-write for other back-end services? 

Or maybe you want clients connecting through the LB to only see one specific index?

sscarduzio:
BTW Amazon tells you the IP range of your VPCs, so in AWS environment you could reason in the dual way knowing the VPC range rather than the ELB IP.
vpc ip

seokhong:
@sscarduzio In that case, we can use two or more rules like:

```
- name: allow read-only access for external clients
   type: allow
   methods: [GET]
   hosts: [10.10.10.0/8] # load balancers' IP ranges
- name: allow read and write access for back-end services
   type: allow
   methods: [GET, POST, DELETE, PUT, OPTIONS, HEAD]
   accept_x-forwarded-for_header: true
   hosts: [10.10.10.0/8] # back-end services' server IP
```

I just wonder there's some case which we should control **both** LB's IP **and** client's IP.

BTW, your solution is more readable and intuitive. It will be nice to support that feature. 👍

sscarduzio:
@seokhong yep. BTW: I'd not rely on HTTP methods to distinguish between read and write ops. It's kinda true, but not guaranteed that whatever arrives from a GET is read only in ES.

That's the reason why I introduced the `actions` rule: this way we can make way more granular decisions AFTER elasticsearch has parsed the HTTP request and mapped it to an action.

There's a list of ES actions you can use:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/wiki/Supported-Rules#actions-and-apis

sscarduzio:
I released 1.13.0 which includes a new rule "x_forwarded_for" rule that accepts an array of  IP or ranges exactly like "hosts", but looks inside the "X-Forwarded-For" header instead.

seokhong:
@sscarduzio Thanks for your advice, hard work and plugin. 👍
I will add `actions` rule to improve my ES's security.

PR #186, opened 2017-02-24
FIX/License headers from previous pull request

innotech-research:
Now that I can run the full build, I note that I broke the licenseMain and licenseTest phases.

This pull request fixes the license headers so that the build process completes successfully.


sscarduzio:
Thanks @innotech-research!

PR #187, opened 2017-02-26
Feature / enhanced search logging

innotech-research:
Adds boolean configuration setting: searchlog (default: false)
Generates detailed information about each search including the query and number of hits


src/main/java/org/elasticsearch/plugin/readonlyrest/IndexLevelActionFilter.java
```diff
@@ -119,7 +127,15 @@ public int order() {
                 .thenApply(result -> {
                     if (result == null) return null;
                     if (result.isMatch() && result.getBlock().getPolicy() == Block.Policy.ALLOW) {
-                        chain.proceed(task, action, request, listener);
+                        if (conf.searchLoggingEnabled && SearchAction.INSTANCE.name().equals(action)) {
```
sscarduzio: what if we get a indices:data/read/msearch?

src/main/java/org/elasticsearch/plugin/readonlyrest/IndexLevelActionFilter.java
```diff
@@ -143,4 +159,51 @@ private void sendNotAuthResponse(RestChannel channel) {
...
+                ActionListener<SearchResponse> baseListener,
+                RequestContext requestContext) {
+            this.action = action;
+            this.searchRequest = (SearchRequest)searchRequest;
+            this.baseListener = baseListener;
+            this.requestContext = requestContext;
+        }
+        
+        public void onResponse(SearchResponse searchResponse) {
+            logger.info(
```
sscarduzio: Hey, this handler looks like the perfect place to implement document level  and field level security 👍

sscarduzio:
Logging needs some love, this is a beginning. Thanks @innotech-research!

src/main/java/org/elasticsearch/plugin/readonlyrest/IndexLevelActionFilter.java
```diff
@@ -143,4 +159,51 @@ private void sendNotAuthResponse(RestChannel channel) {
...
+                    "search: {" +
+                    "ACT:" + action +
```
sscarduzio: Can we print here also the request id (from RequestContext)? So one can correlate the incoming request log and compare? this would be vital for tracing back problems with index rewriting.

innotech-research:
OK, I've added the id to the log.

At the moment this won't log msearches, that's OK for now as I can prevent a user from performing that action.The urgent one for me is auditing normal searches which can now be done.

Also, as you noted this pattern can be used for lots of things including fine grained authorisation. This may be the start of a whole new area of the code base. If this happens, it may be best to move the listener out into a class of its own in a new package - but I didn't want to appear to presume the way that you might wish to evolve.

My next priorities are to investigate persistent settings to see if the ACLs can be changed on the fly instead of needing a server restart, and also to look at the msearch logging.

sscarduzio:
>My next priorities are to investigate persistent settings to see if the ACLs can be changed on the fly instead of needing a server restart

Investigating in how to do this is something I postponed quite long. There's a ES 5.x native way to let the ES Settings framework refresh your settings periodically, but - guess what - it would only work in ES 5.x onwards. it would be nice to have something that works well in all versions and with a stable, home-brewed API.

In general, I'm trying to limit the contact surface with ES code as much as possible, because they constantly break everything, tired of losing time every time a new version comes up.

sscarduzio:
BTW could you please comment here? http://forum.readonlyrest.com/t/call-for-docs-delegated-authentication-configuration-examples/35

PR #190, opened 2017-03-04
Feature/better integration tests

coutoPL:
This branch contains ldap integration tests. Also could be treated as base to write integration tests of other Readonly REST plugin features.

`LdapIntegrationTests` class is a jUnit test case. Before running tests docker containers with external dependencies are being created (eg. container with LDAP). After their successful start gradle `assemble` task is invoked to create Readonly REST plugin zip file. Next, ES container (with version from gradle.properties) is being started with ROR plugin installed. 

`RestClient` should be used to call ES rest api. 

Moreover, `build.gradle` has disabled original jarhell check (because test dependencies should not be taken into consideration). But custom jar hell check was created and is invoked before tests.


src/test/java/org/elasticsearch/plugin/readonlyrest/utils/containers/ContainerException.java
```diff
@@ -0,0 +1,31 @@
+/*
+ *    This file is part of ReadonlyREST.
+ *
+ *    ReadonlyREST is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
```
coutoPL: won't fix

sscarduzio:
Great job @coutoPL as usual!

PR #200, opened 2017-04-04
Order sync rules: auth, inspect, mutate

diegonc:
This PR implements the order suggested in the forum thread.

1. Auth related rules - those that implement UserRule interface
2. Rules that inspect the request - all the remaining ones
3. Rewrite rules - currently, only IndicesRewrite

I hope all the rules are in the appropiate set 😄 

The first commit also implements a test case to ensure the established order is observed.

Unfortunately, due to the previous implementation using a hashmap,  there's a slight chance the test case would pass even if implementation under test doesn't guarantee the required order.


sscarduzio:
Well done ordering the rules @diegonc 👍  
Don't worry about the test, this PR + @coutoPL 's upcoming one will wrap up the sync/async ordering issue too.

PR #202, opened 2017-04-05
Feature/external roles provider

coutoPL:
Added:
- external roles provider (I will create doc for it in next commit)
- async authorization interface
- async authentication interface
- cached authorization decorator
- cached authentication decorator
- async adapter for sync rules and rules collector method refactoring


coutoPL:
ok, I give up today.
Todo:
- update readme
- find out what causes not deterministic test failures

sscarduzio:
+2,476 −485  <-- lol @coutoPL dropped the 💣
I will review this PR ASAP 👍

coutoPL:
I'm still fighting with tests. I'm not sure where is a problem. Tests pass when I run them all in intellij, but fail (testcontainers ones) when I run in with gradle. I'm trying to figure out what causes this problem and how to workaround it. 

So, review the PR, but wait for the fix (hope I find one) and don't merge it yet, please.

coutoPL:
ok, ready to review and merge

sscarduzio:
Hey @coutoPL, I had a look at this.

Starting from the standpoint that ReadonlyREST's main strength is the ease of configuration, that is, it should be very difficult to achieve misconfigurations (one of the main causes of vulnerability), I have the following worry:

Although I understood what the roles service does, as a ReadonlyREST user I'd be pretty confused by the introduction of the concept of "role" in general.
 
We already offer top-level concept for "user" and "group". So, is there room for "role"?
The issue is that "role" is semantically (partially) overlapping with "group".

How would you see the option of consolidating "role" into "group"? So we'd be introducing an external "groups" service and so forth?

coutoPL:
yes, you're right. My role and already existing groups are pretty the same. 
So, this is naming problem. I'm going to fix it today.

coutoPL:
yes, user can be passed as query param or as header value. Now groups service have to return JSON and this is only requirement. Then user can define JSON Path to get list of groups (list of strings) - so response body should not be strictly defined.

coutoPL:
`proxy_auth` is back compatible, so if user does not want to overwrite header, old version is still available. Eg.

New version with custom header (which is defined in `proxy_auth_configs` section):
```proxy_auth:
        - proxy_auth_config: "proxy1"
          users: ["*"]
```

Old version (still available):
```proxy_auth: ["*"]
```

coutoPL:
And we have to separate 2 concepts:
- external authorization (not implemented yet - see `ExternalGroupsAuthorizationAsyncRule` class)
- external roles provider (with groups checking by ROR plugin)

sscarduzio:
I still don't understand this part

```yml
      groups_provider_authorization:
        - user_groups_provider: "GroupsService"
          groups: ["group3"]
```
Why should we specify groups here, since the groups service returns the groups for the user?

coutoPL:
in here we define groups which have access to the resource (eg. index). At the moment ROR knows only user identifier, so asks external service provider to give a list of groups of the user. When external groups provider returns groups, ROR can check intersection of predefined groups and returned groups. Is it's empty - user is not authorized, if not - he is.

sscarduzio:
The way I had imagined this working was:

* we configure 3 ROR groups: g1, g2, g3, g4 
* The authentication resolves the user = john
* The groups service says john belongs to groups  g3 and g4
* ROR will execute the rules block defined in g3 and g4. 

Question: in what order g3 and g4 blocks are executed?

In the proposed configuration model, when you add the `groups: g3` you resolve the above ambiguity by limiting the intersection to the single g3 group. 
The ambiguity is resolved only because  you list a single group. So shouldn't it be `group: g3` ?

coutoPL:
Sorry, maybe I missed sth, but I don't think I understand your proposal.
Could you create example config to show your case?

sscarduzio:
Of course! Sorry I was super unclear, here is an example.

```yml
   [...]
    - name: blah
      type: allow
      methods: GET
      proxy_auth:  ["*"]
      groups_provider_authorization:
        - user_groups_provider: "GroupsService"
          groups:  ["team1", "team2"]

    users:
    - username: alice
      groups: ["team1"]
      indices_rewrite:["logstash-*", "@user_$1"]
      
    - username: bob
      groups: ["team1", "team2"]
      indices_rewrite: ["logstash-*", "other-teams_$1"]

```

A request comes, and the user is "bob". 
The service says "bob" belongs to group "team1" and "team2".
What if GroupService says the user bob belongs to team1 AND team2?

sscarduzio:
Oohh shit now I understand what this is.. Regular groups are totally uninvolved.
The "groups" clause under `groups_provider_authorization` just happens to have the same name!

It's just a condition about a field in the JSON response!

coutoPL:
yes, yes ... in your example we user and his groups are statically declared in ROR configuration. My intention was not use groups from `users` section. Maybe we should somehow distinguish between those two. 

Users section provides statically defined users and roles. In my example in readme, user is provided by reverse proxy (`proxy_auth`) and groups by external groups provider.

sscarduzio:
yep gotcha. Maybe we could have "allowed_groups" or something like that, instead of referencing the same "groups" clause? My head had a buffer overflow reading the same string in another place. 😅

coutoPL:
recurrency can be dangerous ;) 
Maybe it means, that examples should more real, than "group1"-like. 
Agree, "allowed_group" seems to be better choice. But if we decide to change it, it'd be good to change also `groups` from `ldap_auth` to same name.

sscarduzio:
oh it's already out in a previous feature, better leave it like it is for now.
Let's reserve this kind of renaming for a major release. The configuration schema needs a good deal of rationalisation, features came up like mushrooms guided by the creativity of the users and contributors like you. 

Dude I'm merging this. Thank you! 🏅

coutoPL:
OK, great. This week it should also be ready ldap refactoring. So, stay tuned :)

sscarduzio:
That is great news, looking forward to updating this thread!
 http://forum.readonlyrest.com/t/external-auth-feature/60/12

PR #207, opened 2017-04-11
Feature/splitting ldap auth

coutoPL:
Splitted `ldap_auth` feature into separate rules:
- `ldap_autentication`
- `ldap_authorization`

Moreover, added:
- ldap logging on debug level
- caching at rule level


src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/LdapAuthorizationAsyncRule.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private final LdapClient client;
+  private final Set<String> groups;
+
+  public static Optional<LdapAuthorizationAsyncRule> fromSettings(Settings s,
+                                                                  List<LdapConfig> ldapConfigs) throws ConfigMalformedException {
+    return fromSettings(RULE_NAME, s, ldapConfigs);
+  }
+
+  static Optional<LdapAuthorizationAsyncRule> fromSettings(String ruleName, Settings s,
```
coutoPL: won't fix

src/test/java/org/elasticsearch/plugin/readonlyrest/rules/LdapAuthAsyncRuleTests.java
```diff
@@ -51,11 +49,9 @@ public void testRuleSuccessfulCreationFromSettings() {
     LdapConfig config1 = mockLdapConfig("ldap1");
     LdapConfig config2 = mockLdapConfig("ldap2");
     Settings blockSettings = baseExampleBlockBuilder()
-      .put("ldap_auth.0.name", "ldap1")
-      .putArray("ldap_auth.0.groups", Lists.newArrayList("group1", "group2"))
-      .put("ldap_auth.1.name", "ldap2")
```
coutoPL: fixed

sscarduzio:
Nice! Makes sense :)

How about we simplify this:

```yml
      ldap_authentication: 
          - name: "ldap2"  
            cache_ttl_in_sec: 60
```
into 

```yml
      ldap_authentication:  "ldap2"  
```
And the cache configuration we can set under "ldaps"?

You know what I'm trying to do here: making this as concise and explicit as possible..

coutoPL:
At the moment there are two levels of caching. Rule level (example above) and ldap client level (in  `ldaps' config). 

So, maybe I will implement also simplified version (from your example), but the "extended" version should also be available due to rule level caching feature. What do you think?

sscarduzio:
Yes that would be awesome. 
Also, for the README example let's have a single caching layer.

I'd really would like to brag like a bitch in forthcoming presentations about how concise we manage to be in ROR being super thoughtful and using convention over configuration. 
On that note, let's show as commented out any non-mandatory conf option like time outs, SSL etc!

coutoPL:
ok, fixed ...

and I've done some ldap client implementation refactoring ... now, when user uses only `ldap_authentication` there is no requirement to specify `search_groups_base_DN` attribute. So, as you wanted, configuration is simplified.

src/test/java/org/elasticsearch/plugin/readonlyrest/rules/LdapAuthenticationAsyncRuleTests.java
```diff
@@ -0,0 +1,112 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package org.elasticsearch.plugin.readonlyrest.rules;
+
+import com.google.common.collect.Lists;
```
coutoPL: fixed

src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/LdapAuthorizationAsyncRule.java
```diff
@@ -0,0 +1,97 @@
...
+
+  private final GroupsProviderLdapClient client;
+  private final Set<String> groups;
+
+  public static Optional<LdapAuthorizationAsyncRule> fromSettings(Settings s,
+                                                                  LdapConfigs ldapConfigs) throws ConfigMalformedException {
+    return fromSettings(RULE_NAME, s, ldapConfigs);
+  }
+
+  static Optional<LdapAuthorizationAsyncRule> fromSettings(String ruleName, Settings s,
```
coutoPL: won't fix

sscarduzio:
Brilliant. I like this too, thanks @coutoPL! Great stuff as usual 👍

PR #209, opened 2017-04-20
Feature/external authentication

coutoPL:
Added:
- external authentication feature
- service (external authentication service and groups provider service) level caching


sscarduzio:
Yep, very nice, right in time!

sscarduzio:
Thanks @coutoPL! 🚀

PR #210, opened 2017-04-22
Configuration fixes

coutoPL:
I've just realised that configuration created in my previous PR could be better (it worked, but could cause error). This is fix for my bad configuration design.


sscarduzio:
OK makes sense, merging!

PR #214, opened 2017-05-09
Feature/es modules preparing

coutoPL:
In this PR changes I've tried to prepare ROR project to introduce ES dependent modules for each supported version of ES. Core module of ROR should be ES independent. 

Current status:
- all ES dependent code was moved to `org.elasticsearch.plugin.readonlyrest.es53x` package
- packages refactoring
- separated rule description creation (aka settings) from rule execution 
- all tests are ES independent


README.md
```diff
@@ -83,6 +83,8 @@ readonlyrest:
 
 readonlyrest:
     enable: true
+    verbosity: error # don't log successful request
```
sscarduzio: Why did this move to global? Is it still available per-block?

src/main/java/org/elasticsearch/plugin/readonlyrest/settings/ConfigMalformedException.java
```diff
@@ -15,10 +15,13 @@
...
  */
 
-package org.elasticsearch.plugin.readonlyrest.acl.blocks.rules;
+package org.elasticsearch.plugin.readonlyrest.settings;
 
 public class ConfigMalformedException extends RuntimeException {
   public ConfigMalformedException(String msg) {
     super(msg);
   }
+  public ConfigMalformedException(String msg, Throwable t) {
```
sscarduzio: We should unify the concept of "Config" with the "Setting", let's choose one and stick with it

src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java
```diff
@@ -17,48 +17,22 @@
...
+  private final SearchlogRuleSettings settings;
 
-  public static Optional<SearchlogSyncRule> fromSettings(Settings s) {
-    try {
-      return Optional.of(new SearchlogSyncRule(s));
-    } catch (RuleNotConfiguredException ignored) {
-      return Optional.empty();
-    }
+  public SearchlogSyncRule(SearchlogRuleSettings s) {
+    shouldLog = s.isEnabled();
```
sscarduzio: Could you explain what is the advantage of separating the match logic from the settings? Now for each rule we have two classes.. There must be some advantage I'm not seeing

build.gradle
```diff
@@ -50,7 +50,7 @@ repositories {
 esplugin {
     name project.properties['pluginName']
     description 'Safely expose Elasticsearch REST API'
-    classname 'org.elasticsearch.plugin.readonlyrest.wiring.ReadonlyRestPlugin'
+    classname 'org.elasticsearch.plugin.readonlyrest.es53x.ReadonlyRestPlugin'
```
sscarduzio: If I understood correctly, when I will have implemented the equivalent package for `es52x`, I will just need to change this line and `bin/build.sh` again, right?

README.md
coutoPL: I've just checked verbosity implementation and I see a problem with it. Verbosity at 'readonlyrest' level was defined but not used. IMO this settings should be taken into consideration when no rule was matched. Moreover, verbosity functionality can be implemented without introducing new rule - when we treat verbosity as block setting, not specific rule, there is no need to mutate request context. I've changed the implementation to meet these criteria. 

But IMO naming is not so good. Current `check` function:
```public CompletableFuture<BlockExitResult> check(RequestContext rc) {
    logger.debug("checking request:" + rc.getId());
    return FuturesSequencer.runInSeqUntilConditionIsUndone(
        blocks.iterator(),
        block -> {
          rc.reset();
          return block.check(rc);
        },
        (block, checkResult) -> {
          if (checkResult.isMatch()) {
            if (Verbosity.INFO.equals(block.getVerbosity())) {
              logger.info("request: " + rc + " matched block: " + checkResult);
            }
            if (checkResult.getBlock().getPolicy().equals(BlockPolicy.ALLOW)) {
              rc.commit();
            }
            return true;
          } else {
            return false;
          }
        },
        nothing -> {
          if (Verbosity.INFO.equals(settings.getVerbosity())) {
            logger.info(ANSI_RED + " no block has matched, forbidding by default: " + rc + ANSI_RESET);
          }
          return BlockExitResult.noMatch();
        }
    );
  }

At the moment when verbosity level is set to INFO all matching is logged and no block match is logged also. When user sets verbosity to ERROR logging is disabled.

Maybe I missed sth, but verbosity values are not so descriptive (for me).

File: src/main/java/org/elasticsearch/plugin/readonlyrest/settings/ConfigMalformedException.java (hunk 2 above)
Comment (coutoPL) [inline src/main/java/org/elasticsearch/plugin/readonlyrest/settings/ConfigMalformedException.java]: ok, done

File: src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java (hunk 3 above)
Comment (coutoPL) [inline src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java]: Yes, we have 2 classes - settings and rule. Settings class of rule is responsible for parsing and validating. Rule class responsibility is checking if request is matched or not. Before each rule has too many responsibilities. Now matching logic is better exposed and easier to read.

This is from architecture point of view. But there are also other advantages. Eg. I saw, that you are preparing ROR plugin to be able reload configuration without restarting ES. Now, when settings are separating from rules, we can simply try to create new settings on reload request and if sth fails, the reloading attempt does not affect working rules and previous settings. 

There is no coincidence that you had doubts when you were reading `SearchlogSyncRule` class. As you can see, now this class is not so useful - it just matched all requests. Due to splitting previous rule class into parts with single responsibilities, there is easier to reason about rule as a request matcher. Same example was VerbosityRule - it should be refactored (and now it is).

What do you think? Have I convinced you?

File: build.gradle (hunk 4 above)
Comment (coutoPL) [inline build.gradle]: I've changed it to `es`. Next step (I think in next PR) will be moving code from this package to `es53x` module. And there will be also `es52x` module, etc. Each module will have own `build.gradle` file. Then `bin/build.sh` script will have `./gradle assembly [...]` for each module.

File: src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java (hunk 3 above)
Comment (sscarduzio) [inline src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java]: Yeah, but  you see, I kinda still miss the one-rule = one-class simplicity.

But now I think I understood why you split the old rule in the "RuleActionListener" + "RuleSettings" + "Rule" triplet.  Tell me if it's true! 😄 

So, first you noticed that the "onRequest/onError" part of the old Rule heavily depend on "instanceof" of ES-specific classes.

Therefore you split the "onResponse/onError" part from the "match" part and you put the first inside the es-dependent package.

The "match" part instead is not es-dependent because of "RequestContext" abstraction. So that could go together with generic code.

Now you are left with two aspects of the same thing: "RuleActionListener" implementors vs "Rule" implementors that cannot live in the same package. But both need to use the settings, so  you invented the RuleSettings interface.*

PS: *actually RuleActionListener cannot access its corresponding RuleSettings implementor. Shouldn't it?

So, I totally agree on where you came from with this design, it's a good solution for the current horrible state of the "onResponse/onError" logic, so es-specific.

But I can't help seeing here that **in the future** we could introduce a "ResponseContext" contract that would decouple the listener side like "RequestContext" did for the other part.

So yes, today this is good, but one day, once and if this is gonna be reasonable, we could re-consolidate the rule code and make it simple again.

Comment (sscarduzio) [inline src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java]: BTW, about SearchlogSyncRule.
It was introduced by @innotech-research because they needed it for their use case. It's not terribly useful for anyone else, but I left it in because it is a precursor of all the "onResponse/onError" logic. Which opens the road to document-level security and field-level security, which is a really cool feature instead. 👍

Comment (coutoPL) [inline src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java]: I don't agree with the first sentence. I would say: one-rule = one-class complexity ;) In object programming languages I tried to follow SOLID principles. Then maybe you have to create more classes to add new feature, but new dev can easier reason about what single class responsibility is and what it is expected to do. 

I tried to figure out what is a job of `onResponse` & `onFailure` and I noticed that it is loosely coupled with rule. So, it was natural to move it somewhere else. Maybe my abstraction is not so good - I don't understand very well logic in these methods.

Settings could be defined as it was before with new common, es independent, settings class - RawSettings. But IMO it is better to split parsing&validation from config interpretation. Factories which glue everything together are not so pretty - I know it. But it can be refactored eg. with annotations. 

Maybe, when ResponseContext will be introduced, rule behaviour should be redefined/extended. But still - responsibilites should be split into separate classes and rule would delegate work to them.

Comment (sscarduzio) [inline src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/SearchlogSyncRule.java]: Sure, fair enough. We can take it from there when we have ResponseContext, which will take some good effort in design. The stuff I've been doing with onResponse is badly missing some abstraction.

File: src/main/java/org/elasticsearch/plugin/readonlyrest/es/requestcontext/RCTransactionalIndices.java [hunk 5]
```diff
@@ -177,7 +178,7 @@ public void onCommit(Set<String> newIndices) {
         logger.debug("id: " + rc.getId() + " - Replacing indices. Old:" + getInitial() + " New:" + newIndices);
 
         if (newIndices.size() == 0) {
-          throw new ElasticsearchException(
+          throw es.rorException(
```
coutoPL: won't fix

coutoPL: won't fix

src/test/java/org/elasticsearch/plugin/readonlyrest/acl/definitions/ldaps/UnboundidAuthenticationLdapClientTests.java
```diff
@@ -30,12 +32,12 @@
...
-public class UnboundidLdapClientTests {
+public class UnboundidAuthenticationLdapClientTests {
 
   @ClassRule
   public static LdapContainer ldapContainer = LdapContainer.create("/test_example.ldif");
 
-  private UnboundidGroupsProviderLdapClient client = new UnboundidGroupsProviderLdapClient(
+  private UnboundidAuthenticationLdapClient client = new UnboundidAuthenticationLdapClient(
       new ConnectionConfig.Builder(ldapContainer.getLdapHost())
           .setPort(ldapContainer.getLdapPort())
```
coutoPL: fixed

src/test/java/org/elasticsearch/plugin/readonlyrest/acl/definitions/ldaps/UnboundidAuthenticationLdapClientTests.java
```diff
@@ -30,12 +32,12 @@
...
           .setPoolSize(10)
```
coutoPL: fixed

src/test/java/org/elasticsearch/plugin/readonlyrest/acl/definitions/ldaps/UnboundidGroupsProviderLdapClientTests.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class UnboundidGroupsProviderLdapClientTests {
+
+  @ClassRule
+  public static LdapContainer ldapContainer = LdapContainer.create("/test_example.ldif");
+
+  private UnboundidGroupsProviderLdapClient client = new UnboundidGroupsProviderLdapClient(
+      new ConnectionConfig.Builder(ldapContainer.getLdapHost())
+          .setPort(ldapContainer.getLdapPort())
+          .setPoolSize(10)
```
coutoPL: fixed

src/main/java/org/elasticsearch/plugin/readonlyrest/es/requestcontext/RequestContextImpl.java
```diff
@@ -57,106 +64,92 @@
...
 
-  public RequestContext(RestChannel channel, RestRequest request, String action,
-                        ActionRequest actionRequest, ClusterService clusterService,
-                        IndexNameExpressionResolver indexResolver, ThreadPool threadPool) {
-    super("rc");
-    this.channel = channel;
-    this.request = request;
-    this.action = action;
-    this.actionRequest = actionRequest;
-    this.clusterService = clusterService;
```
coutoPL: fixed

coutoPL: fixed

coutoPL:
@sscarduzio do you want me to change sth in this PR? Do you have any objections?
For me it's ready to merge.

sscarduzio:
Hey @coutoPL nice to have this big chunk merged, thanks a lot for making it happen! 
So, wrapping up on one of the biggest ROR crusades, that is, unify the ES version branches:

- [x] Decoupling ES specific from common code 
- [x] Decouple also the tests
- [ ] Modularize Gradle builds so the ES-specific is segregated into a separate gradle project
- [ ] Create a es52 gradle project
- [ ] Create a es51 gradle project
- [ ] Create a es50 gradle project

And  as discussed I can take the last 3 Right? Please amend anything I missed.

coutoPL:
ROR with separate gradle projects is almost ready. At the moment I have one module: `es53x`. I'm going to create also `es52x` module and join everything together with one build script. So, I'll leave you with job related to es51 and es50 modules.

sscarduzio:
That's awesome!
Question: will we be able to build all the versions using Gradle 3.5 instead of 2.13 (this was a change introduced in ES 5.4)?

coutoPL:
We must do that. No matter what. I saw your commits and I'm going to merge with it soon.

PR #218, opened 2017-05-14
jwt_auth rule: authenticate users through JSON Web Tokens

diegonc:
WIP! may need adjusting 😄 

The `jwt_auth` rule allows to grant access to clients that provide a valid JSON Web Token as part of the
request using the `Authorization` header.

There are two settings currently supported:
* _signature.key_: the secret used to check for the validity of the signature in the token
* _user_claim_: the claim in the token payload whose value will be used a replacement for `@user`

A sample configuration is below:

```yaml
    access_control_rules:
      - name: Valid JWT token is present
        type: allow
        jwt_auth:
          signature.key: 123456
          user_claim: user
```


sscarduzio:
Hey @diegonc thanks for putting this together, really useful! 
Travis CI is moaning about missing license headers. Can you copy paste it from other source files and stick it on top of each new file?

coutoPL:
you can use `./gradlew licenseFormat` to generate licences for new files

diegonc:
Thanks @coutoPL! I'll try with that command in case I forgot something when adding them manually.

sscarduzio:
@diegonc one thing about JWT secrets is that they are like plain text passwords. And you don't want those in your configuration files. A good practice is to extract the secret from environmental variables. Can we have a special value for signature key like "ENV"?

One more thing, you have "signature.key" the dot in yaml means a nested property of an object. Can we change it to "signature_key"?

diegonc:
Yes, environment variables make sense, indeed. 

Something like a magic prefix?

```yaml
    signature_key: "env:AN_ENV_VARIABLE"
```

or more config entries?

```yaml
    signature_key: "AN_ENV_VARIABLE"
    signature_key_source: "environment" 
```

or maybe some other approach.

> One more thing, you have "signature.key" the dot in yaml means a nested property of an object. Can we change it to "signature_key"?

Dot was intentional. The key could also be a path to a public key for RSA/ECDSA  based tokens or be base64 encoded data. I thought making `signature` an object could accommodate those use cases better in the future. But I guess a flat config will do, too.

sscarduzio:
If you saw the CERN presentation, you know I'd love the first option 😃

sscarduzio:
@diegonc tell me when we can merge this :)

diegonc:
Ok, I've got some integration tests running, so this is actually working 🎉 
Tomorrow I'll work on polishing it a bit:
- [ ]  `signature_key` instead of `signature.key`
- [ ]  Allow `env:VAR` to read key from environment. Use `text:secret` or drop plaintext support.
- [ ]  Make settings reader more robust. The sample at the top fails with ClassCastException because `123456` is parsed as a number and read as a string 😄 

----

These are the integration tests I've run. Just a few, but I hope they cover most relevant usecases.

→ reject-no-authorization-header
  GET https://127.0.0.1:9200/_cat/indices [401 Unauthorized, 165B, 1273ms]
  ✓  Status code is 401

→ reject-wrong-key
  GET https://127.0.0.1:9200/_cat/indices [401 Unauthorized, 165B, 405ms]
  ✓  Status code is 401

→ require-user-claim
  GET https://127.0.0.1:9200/_cat/indices [401 Unauthorized, 165B, 29ms]
  ✓  Status code is 401

→ accept-valid-token-with-user
  GET https://127.0.0.1:9200/_cat/indices [200 OK, 79B, 65ms]
  ✓  Status code is 200

→ reject-expired-token
  GET https://127.0.0.1:9200/_cat/indices [401 Unauthorized, 165B, 16ms]
  ✓  Status code is 401

sscarduzio:
BTW in theory that integration_tests folder is old and contained manual integration tests based on docker. 
We moved on into testcontainers. For example:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/master/src/test/java/org/elasticsearch/plugin/readonlyrest/es/IndicesReverseWildcardTests.java

coutoPL:
In ES modules branch I've got rid of this folder. Is it ok?

sscarduzio:
@coutoPL yes, makes sense! Clean it up 👍

diegonc:
I think this is ready. Tell me if you need a rebase, the history looks a bit messy 😅

sscarduzio:
Thanks! Hey, what's the deal with jackson dependencies? I just checked out your branch, it does build without them.

sscarduzio:
But.. The test runs without complaining 💅

sscarduzio:
I see.. Doesn't sound right. I opened an issue in their project.

https://github.com/jwtk/jjwt/issues/223

diegonc:
Maybe this is related. I cannot find much documentation about this decision on ES site 🤔 

https://github.com/elastic/elasticsearch/pull/14755
https://github.com/elastic/elasticsearch/commit/4b5f87cb7dcea80bca99f863a02e585b8c2c1e17

sscarduzio:
So, elasticsearch aside, JJWT only depends on jackson core, but their code is referencing to a class that is in jackson databind and the error arrives at runtime.

Maybe their tests work because of some other test-dependency pulling in jackson-databind transitively.
I still think it's their error.

But let's just make it work in ours for now.  Here it's time to merge stuff and release so let's keep this JJWT issue in the sidelines. I will still follow up on this, but let's not let it block us.

So, @diegonc if you roll back those dependencies I can merge it. 👍

diegonc:
Done

PR #220, opened 2017-05-16
Feature/es modules

coutoPL:
Readonly REST project was split into several gradle modules:
- ES dependent modules (now es52x, es53x)
- core (ES intependent)
- integration-tests 

New ES-module should be created when used Elasticsearch API will be changed and name of the module should reference first version with incompatible (used) API. gradle.properties file in ES-module should contain newest ES version (eg. es52x has set esVersion=5.2.2). 

Core module also have tests which also are ES independent.

Integration-tests module contains integration tests which are run for every ES-module. 

Running all tests: `./gradlew test`
Running integration tests: `./gradlew integration-tests:test`
Running core tests: `./gradlew core:test`
Building ROR version for ES 5.2.2: `./gradlew es52x:ror`
Building ROR version for ES 5.3.0: `./gradlew es53x:ror '-PesVersion=5.3.0'` (in gradle.properties 5.4.0 is set)

Caveats:
- integration tests are very slow ... JUnit Parameters are used to invoke tests for different ES version. But there is no way (as far as I know) to configure test case and then run all tests in it. So each tests should create docker image, run container. 
- `IndicesReverseWildcardTests` not working - don't know why (ignored at the moment)
- @sscarduzio, please search for //todos which I left to review/fix for you. Most of them are ES related, so it would be better that you will look at it.


coutoPL:
ok, done

sscarduzio:
Amazing job! I tried it and it actually builds packages for 5.3 and 5.2. But where is the 5.4 zip?

sscarduzio:
I will need to amend my s3 upload script and tagging mechanism to basically tag N times with N=#subprojects

coutoPL:
check `es53x/build/distributions/`. Look at `es53x` module gradle.properties - 5.4.0 version is set there. So by default, when you run `./gradlew ror` 2 versions should be build: 5.2.2 and 5.4.0. To build other version, eg. 5.3.0 you should also run `./gradlew es53x '-PesVersion=5.3.0'`. And it should be placed in `es53x/build/distributions/`.

sscarduzio:
is that because es53x actually contains valid code for 5.4 too?

coutoPL:
Yes, at the moment in 5.4.0 we don't use ES API which was changed. So there is no need to introduce module for es54x. And because of that I've created `'-PesVersion=5.3.0'` for build specific version compatible with given module :)

sscarduzio:
Great, less code for us! So would it then make more sense to name `es53x` as `esCurrent`? 
And when a breaking change comes, we do:

 ```bash 
cp -r esCurrent es$(FIRST_ES_SUPPORTED)_$(LAST_ES_SUPPORTED)
```

For example, if they release 5.5.0 and no change is required for us, then when es5.6.0 is released and breaks from 5.5.0, we can do:

```bash 
cp -r esCurrent es53x_55x
``` 
and start working on `esCurrent` to make it work with 5.6..

coutoPL:
It's up to you. I didn't do that because I think that changes module name affects git history. But if it's ok for you, we can follow this convention.

sscarduzio:
Oh that is a fair objection I didn't think of. Let's keep it this way. As long as we know the latest sub-project is the one to go for building the latest stuff, it will be more git-tidy.

sscarduzio:
>integration tests are very slow ... JUnit Parameters are used to invoke tests for different ES version. But there is no way (as far as I know) to configure test case and then run all tests in it. So each tests should create docker image, run container.

Actually there could be a way: making a tear down function that deletes all the indices, writes the .readonlyrest index and triggers the settings refresh. So we could reuse the same docker container.
It would be 10x faster, but it will introduce a tad of risk as it's not as super clean start as with a fresh container.
I've been reluctant to do this, but  on the other hand, it might make sense to merge simple tests together.

coutoPL:
TestNG has better parameterized test, but there is no official integration with it in testcontainers. But if we could integrate it with TestNG, there will be possible to run all tests in suite for given parameter (read: one container).

sscarduzio:
how much time can we gain from switching to TestNG?

sscarduzio:
I found the reason why `IndicesReverseWildcardTests` takes so long and fails: it's creating a new container for every test method!
Like, it creates #testMethods * #esProjects containers :D 
Does it have anything to do with the @Rule annotation?

coutoPL:
I don't know, because I've never used testNG.
JUnit calls constructor for each test and passes to it esVersion. So this is one place to create and configure container for given test. 
We'd like junit to call constructor once and then run all tests. But it doesn't work like that. 

We could do it by ourself: in raw junit test suite constructor, create containers for all versions and then in each test in loop (for each esVersion) call test logic. But this is ugly solution and IMO we should handle it at framework level. jUnit is limited in this aspect.

sscarduzio:
Here's a super dumb/explicit way to do it: how about we make the containers static fields and create them from the constructor only if they're null?

coutoPL:
bleh ... ugly, static, mutable, java hack :P 
But you're right it is fastest solution at the moment. Can change it in this PR.

sscarduzio:
:rofl: I agree, but it's a tiny effort, maximum result, self-contained, clear intentions readable in code. Keep the ugly stuff coming, I say.. haha 💃

sscarduzio:
I have the impression this trick can be generalised to quite a few other integration tests. By that speeding the heck up of the whole CI process. Am I rite? :)

coutoPL:
yes, I'll try to put it in BaseIntegrationTest class

sscarduzio:
Integration tests pass in my IDE. This is CI bullshit. I will merge this and print a dot every time a container is created. 
Once again, thanks a ton @coutoPL for this very much needed PR. Finally we can streamline our work.

coutoPL:
But I think that it must be some issue with MultiContainer, because on my machine one test also failed. I have to check it.

sscarduzio:
I ran all the tests together from the IDE and have intermittent failures (obviously re-testing the failed tests singularly they pass).

I've been running `docker ps` in the process and at times I saw 7 containers running at the same time. I've been thinking what if we run integration tests sequentially?

Can it be that we spawn too many processes and travis slows down and stuff goes in timeout?

PR #223, opened 2017-05-17
Oauth

tcharlot-datasweet:
Hello, 

We've developed a new authentication method to your plugin, based on OAuth. 
It's available on this repo: https://github.com/datasweet-fr/elasticsearch-readonlyrest-plugin/tree/oauth 
This authentication works with JWT (Java Web Token) as OAuth token. 
Here is what we've added to your plugin: 
    * New configuration parameters for the OAuth authentication 
    * OAuth authentication if an OAuth is enabled in config 
    * Token groups checking instead if OAuth is enabled 
    * Token verification as a new rule (only if OAuth is enabled): expiration date and integrity are checked for now 
    * Token UT added 

Also, to make this work, you'll need to install this file in your JAVA_HOME: jce_policy-8.zip Deploy the two jars in <java-home>/lib/security 
These files make the use of the jiron (use for deciphering the token) possible as it is using a certain key size instead of a regular one. (Those are custom jars given by Oracle) 
A small example to show how this works: 
An OAuth provider (Keycloak in our case) provides a JWT, encrypt it and set it as a cookie with the ES request ROR plugin receives the request, deciphers the token and add it to the request context 
Then the rules are checked (including the new token rule) 
Here's a configuration example: 
    oauth_enabled: true 
    cookieSecret: 3eeadfea0896c1a51b41d09aa1eab3d6c24d2154 #used to decipher the token retrieved in the cookie 
    cookieName: datasweet-oauth 
    tokenClientId: demo #name of the client where we'll get the group in the JWT under "resource_access", see <a href='https://jwt.io/' class='linkified' target='_blank'>https://jwt.io/</a> for more explanation 
    tokenSecret: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAkBpwYFlAqMcQhdbSOkkOHu3cU39EioKzWdmmkD8HZS0vfuiqij/PgqMJvF49EXwiclCLgGmf71pvZxazEHp2sqQmkUYeF9xAUPvRKurNE2krd69g+nGiOCvoNMZ10APeO3FU36EOtn2cxoYXUs/qkNBQZf1YZaI0BfWqjhwI9ndFwbx2D8o4YliGfJdagDJhueJ4X3L++pqfk7f8UYXehmE9T+2ymJPbSkx7Lw9YQFIoRYdWzAfFJebk1v0FUE1mmfTSDEye47UvYxec4StYzo1f1K3SgXwUTl1EuhClH1eX+XAuQYHWczR9sXH3kGAvKFsJ/azVBh0N1M1hs+N97wIDAQAB #public key used to verify the integrity of the token 

This small enhancement to your plugin is part a biggest one as we're planning other updates
This development is made to meet very specific needs (Keycloak as an oauth provider, hapijs cookie cipher). That's why if you don't want to merge it, this is not a big issue for us.
We just want to notify you of our changes in order to be in good standing with the choice of the GPL license.

Regards,


sscarduzio:
Hi @tcharlot-datasweet, good stuff. I recommend the same thing I recommended to @diegonc. Keeping the JWT key in the configuration file is dangerous, especially in ES (settings can be accessed via a REST call, if ROR is not well configured). So I'd add the possibility to get the key from the environmental variables.

PR #229, opened 2017-05-22
Bugfix/ldap search query

coutoPL:
LDAP search queries were not escaped, so eg. parenthesis in CN caused search fail.


sscarduzio:
Great, merging 👍

PR #233, opened 2017-05-29
changed: integration tests for different modules are run from bash sc…

coutoPL:
running integration tests for given module:
`./gradlew integration-tests:test '-PesModule=es52x' --debug`

I don't want to fight with gradle any more, so ... here you have, not perfect, but working solution.


coutoPL:
Yes, I have nothing to add.

sscarduzio:
I could not believe my eyes: it builds on MacOS now 🚀🦄

PR #242, opened 2017-06-13
Add action match needed for _cat/indices request

nitsanc:


sscarduzio:
what was this needed for, exactly?

PR #250, opened 2017-07-03
Add sha512 and unix rule

samy-orange:
A new rule has been added, the unix rule. 
It is based on /etc/shadow file syntax for example I configured my unix to hash my password with sha512 65535 rounds.
For account `test:test` the hash in /etc/shadow will be 
`test:$6$rounds=65535$d07dnv4N$QeErsDT9Mz.ZoEPXW3dwQGL7tzwRz.eOrTBepIwfGEwdUAYSy/NirGoOaNyPx8lqiR6DYRSsDzVvVbhP4Y9wf0`

To use it with the plugin you need to put this in your config file:
`auth_key_unix: test:$6$rounds=65535$d07dnv4N$QeErsDT9Mz.ZoEPXW3dwQGL7tzwRz.eOrTBepIwfGEwdUAYSy/NirGoOaNyPx8lqiR6DYRSsDzVvVbhP4Y9wf0`


sscarduzio:
So did I understand correctly this is basically a better `auth_key_sha256` i.e. with a crazier hashing? 
I thought it was actually trying to log into the local unix system. Would that be feasible?

samy-orange:
> So did I understand correctly this is basically a better auth_key_sha256 i.e. with a crazier hashing?

Not exactly, the `aut_key_ unix` rule use [apache crypt method](https://commons.apache.org/proper/commons-codec/apidocs/index.html), it's not only a better `auth_key_sha256`
With my example `test:$6$rounds=65535$d07dnv4N$QeErsDT9Mz.ZoEPXW3dwQGL7tzwRz.eOrTBepIwfGEwdUAYSy/NirGoOaNyPx8lqiR6DYRSsDzVvVbhP4Y9wf0`
`test` is the username and `$6$rounds=65535$d07dnv4N$QeErsDT9Mz.ZoEPXW3dwQGL7tzwRz.eOrTBepIwfGEwdUAYSy/NirGoOaNyPx8lqiR6DYRSsDzVvVbhP4Y9wf0` is the salt. 

The exact algorithm depends on the format of the salt string:
- SHA-512 salts start with $6$ and are up to 16 chars long.
- SHA-256 salts start with $5$ and are up to 16 chars long
- MD5 salts start with $1$ and are up to 8 chars long

The rounds=N option helps to improve key strengthening. The number of rounds has a larger impact on security than the selection of a hash function. For example, rounds=65535means that an attacker has to compute 65535hashes for each password he tests against the hash in your /etc/shadow. Therefore the attacker will be delayed by a factor of 65535.

We use this rule because it's more secure than a hashed password and we have just to put the content of /etc/shadow in the elastic config file

> I thought it was actually trying to log into the local unix system. Would that be feasible?

I don't know if you can log into the local system directly with java. 
You can parse the /etc/shadow file but it looks bad idea because you need to be root, also if you do that you need all account in the same machine.

sscarduzio:
Yep agreed, it's a (way, way) better version of the current hashed auth key. I meant it as opposed to reading system's /etc/passwd.

I like this thing, what's the best way for people to generate these super-hashes? Is there practically a command or a website that I can use to convert my password into that hash?

samy-orange:
- Configure your Unix system to generate password in your /etc/shadow, and just need to copy the line until second ":" separator

```
/etc/pam.d/common-password

password	[success=1 default=ignore]	pam_unix.so obscure sha512 rounds=65536
```
- With Debian you can use the command line, you need whois package (apt-get install whois): `mkpasswd -m sha-512 -R 65534`

- You can use python or perl, for example you can run this with the password in argument:

```
#!/usr/bin/python
import crypt
import random
import sys
import string



def sha512_crypt(password, salt=None, rounds=None):
    if salt is None:
        rand = random.SystemRandom()
        salt = ''.join([rand.choice(string.ascii_letters + string.digits)
                        for _ in range(8)])

    prefix = '$6$'
    if rounds is not None:
        rounds = max(1000, min(999999999, rounds or 5000))
        prefix += 'rounds={0}$'.format(rounds)
    return crypt.crypt(password, prefix + salt)


if __name__ == '__main__':
    if len(sys.argv) > 1:
        print sha512_crypt(sys.argv[1], rounds=65635)
    else:
        print "Argument is missing, <password>"
```
Finally you have to put your username at the begining of the hash with ":" separator, like :
`test:$6$rounds=65535$d07dnv4N$QeErsDT9Mz.ZoEPXW3dwQGL7tzwRz.eOrTBepIwfGEwdUAYSy/NirGoOaNyPx8lqiR6DYRSsDzVvVbhP4Y9wf0`

Vebryn:
Seems interesting! Hope this will be merged with master branch!

sscarduzio:
@samy-orange could you please sign the CLA? Otherwise I can't merge. See clahub details. Thank you!

samy-orange:
@sscarduzio Done !

sscarduzio:
Thanks. Could you please add some documentation on how to generate the hash somewhere in README.md?

samy-orange:
@sscarduzio yes i'll do that

samy-orange:
Which OS do you use ?
In my Debian 8 the python script work well

> $ python pass.py test
>$6$rounds=65635$QZ5EDRH6$xc46j8uR4M68EhcxWCapPehcNVc5TrFYyz/5UcUKlQ7raTULFNMbAOelQPhODtx/IZKzfBTGUMBN8UoF0d2SB0

sscarduzio:
Interesting, I'm on Mac OS (whatever the latest version is called)

_S

samy-orange:
crypt module is a Unix Specific Service.

Right at the top of the [docs ](http://docs.python.org/dev/library/crypt.html) for crypt:

```
    34.5. crypt — Function to check Unix passwords

    Platforms: Unix

```

sscarduzio:
Yeah I used docker to run mkpasswd. I need to figure out a way to let people create these password easily. Maybe a web form in the website?

samy-orange:
Client side you can try [node-crypt3](https://github.com/sendanor/node-crypt3)

You can use Java with [Apache Commons Codec](http://commons.apache.org/proper/commons-codec/archives/1.10/apidocs/org/apache/commons/codec/digest/Crypt.html#crypt(java.lang.String,%20java.lang.String)) or php with [crypt function](http://php.net/manual/fr/function.crypt.php)
```php
<?php
private function encrypt_sha512($password, $round = 'rounds=5000')
{
    $chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
    for ($i = 0; $i < 16; $i++) {
        $salt .= $chars[rand(0, strlen($chars) - 1)];
    }
    return (crypt($password, '$6$'. $round. '$' . $salt. '$'));
}
```

PR #271, opened 2017-09-13
Password ending in ':' fails

ruria:
When user has a password ending with ":" authentication fails.


sscarduzio:
Thank you @ruria! Awesome find, quite a subtle bug. 👍  If you add the license in the heading of the test it builds ok in Travis-ci. Can you please do that so I can merge?

PR #272, opened 2017-09-13
Password ending in ':' fails

ruria:
When user has a password ending with ":" authentication fails.When user has a password ending with ":" authentication fails.


sscarduzio:
@ruria can you add the GPL license heading on top of BasicAuthUtilsTests.java?

sscarduzio:
Thanks @ruria! Now buckle up, Travis takes A LOT of time to run all the dockerized tests. 🍿 😎

PR #293, opened 2017-11-09
DMH: custom audit log serializer

sscarduzio:
This PR adds the possibility to write your own custom audit log serializer class and configure it through the settings.

## Instructions
1. Create a new Java project in your IDE
2. Create a class like this:
```java
import tech.beshu.ror.AuditLogContext;
import tech.beshu.ror.requestcontext.AuditLogSerializer;

import java.util.HashMap;
import java.util.Map;

public class MyCustomSerializer implements AuditLogSerializer {
  public MyCustomSerializer(){
    System.out.println("MY CUSTOM SERIALIZER INSTANTIATED");
  }
  @Override
  public Map<String, ?> createLoggableEntry(AuditLogContext context) {
    System.out.println("making the loggable entry!");
    Map m =  new HashMap<>();
    m.put("myField", context.getAction());
    return m;
  }
}

```
3. Satisfy the two import above by copy-pasting the two classes from ROR core code base into your project (they have no other dependency).

4. Find the MyCustomSerializer.class somewhere in your build directory

5. ` jar cvf CUSTOMSERIALIZER.jar MyCustomSerializer.class`

6.  mv CUSTOMSERIALIZER.jar plugins/readonlyrest/

7. Your config/readonlyrest.yml should start like this
```yml
readonlyrest:
    audit_serializer: MyCustomSerializer
```

8. Start elasticsearch (with ROR installed) and grep for:
```
[2017-11-09T09:42:51,260][INFO ][t.b.r.r.SerializationTool] Using custom serializer: MyCustomSerializer
```


core/src/main/java/MyCustomSerializer.java
```diff
@@ -0,0 +1,35 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+
+//import tech.beshu.ror.AuditLogContext;
+//import tech.beshu.ror.requestcontext.AuditLogSerializer;
+//
+//import java.util.HashMap;
+//import java.util.Map;
+//
+//public class MyCustomSerializer implements AuditLogSerializer {
```
elaPa: It looks like a leftover - this class shouldn't be included.

core/src/main/java/tech/beshu/ror/requestcontext/SerializationTool.java
```diff
@@ -45,70 +50,94 @@
...
-    map.put("acl_history", req.getHistory().toString());
 
-    // The cumbersome, awkward security handling in Java. (TM) #securityAgainstProductivity
     final String[] res = new String[1];
 
     AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
       try {
-        res[0] = mapper.writeValueAsString(map);
+        if (customSerializer.isPresent()) {
+          res[0] = mapper.writeValueAsString(customSerializer.get().createLoggableEntry(logContext));
```
elaPa: By passing the logContext, you're limiting what we can add to our audit log entry by the fields that are included in logContext. Eg we would not be able to include the value of whitelisted headers (which is what we need to have).
createLoggableEntry should accept ResponseContext as a parameter, not AuditLogContext.

core/src/main/java/tech/beshu/ror/requestcontext/AuditLogSerializer.java
```diff
@@ -0,0 +1,27 @@
...
+
+package tech.beshu.ror.requestcontext;
+
+import tech.beshu.ror.AuditLogContext;
+
+import java.util.Map;
+
+public interface AuditLogSerializer {
+
+  Map<String, ?> createLoggableEntry(AuditLogContext context);
```
elaPa: As mentioned below, that method signature should be:

`Map<String, ?> createLoggableEntry(ResponseContext context);`

or maybe, even better, the interface should be generic:

`public interface AuditLogSerializer<T> {
    T createLoggableEntry(ResponseContext context);
}`

And then you can have a default implementation of AuditLogSerializer, that converts ResponseContext to AuditLogContext.

core/src/main/java/tech/beshu/ror/requestcontext/SerializationTool.java
```diff
@@ -45,70 +50,94 @@
     zuluFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
   }
 
-  public SerializationTool() {
+  private final LoggerShim logger;
+  private Optional<AuditLogSerializer> customSerializer;
```
elaPa: IMHO this should be: 
`private final AuditLogSerializer auditLogSerializer;`
and should either be set to the default serializer (that returns  AuditLogContext) or to the custom serializer (if specified)

core/src/main/java/MyCustomSerializer.java
sscarduzio: Yeah I left this here as an example of how to write a serializer until I make the docs.

core/src/main/java/tech/beshu/ror/requestcontext/SerializationTool.java
sscarduzio: Yes that would be more clean 👍

sscarduzio:
@elaPa thanks for the awesome review, check out the latest changes when you can :)

core/src/main/java/tech/beshu/ror/AuditLogContext.java
```diff
@@ -0,0 +1,290 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror;
+
+import java.util.Set;
+
+public class AuditLogContext {
```
elaPa: AuditLogContext is not used anywhere now, I guess it is not needed anymore?

core/src/main/java/tech/beshu/ror/requestcontext/DefaultAuditLogSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+import tech.beshu.ror.commons.ResponseContext;
+import tech.beshu.ror.commons.shims.request.RequestContextShim;
+
+import java.text.SimpleDateFormat;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.TimeZone;
+
+public class DefaultAuditLogSerializer implements AuditLogSerializer {
```
elaPa: NIT: This should be:
`AuditLogSerializer<Map<String, ?>>`

sscarduzio:
@elaPa what do you think of this? https://forum.readonlyrest.com/t/set-readonlyrest-audit-index/398

elaPa:
@sscarduzio thank you for checking - replied in that topic.

PR #298, opened 2017-12-07
add external auth validate option

presto9292:
If validate option is false, host name verification and certificate validation are turned off.
Default is true which preserves b/c


sscarduzio:
Thank you @presto9292!!

PR #330, opened 2018-04-02
Configurable headers and parameters for Group Service Api Calls.

rbkasat:
Updated the Group Service Configurations to support static headers and parameters to be configured.
Some API's need basic auth, some need Hmac token as request param, which can be statically configured in the configuration Yml. New updates won't affect the existing configurations and won't take effect unless configured.

### DIFF
--- a/core/src/test/java/tech/beshu/ror/acl/blocks/rules/impl/GroupsProviderAuthorizationAsyncRuleTests.java
+++ b/core/src/test/java/tech/beshu/ror/acl/blocks/rules/impl/GroupsProviderAuthorizationAsyncRuleTests.java
@@ -48,6 +48,9 @@ public class GroupsProviderAuthorizationAsyncRuleTests {
   public static WireMockContainer wireMockContainer = WireMockContainer.create("/groups_provider_authorization.json",
       "/group_provider_authorization_complex.json");
 
+  private final Function<ImmutableMap<String, String>, String> mapFormatter = map -> map.entrySet().stream().
+      map(y -> "      " + y.getKey() + ": " + y.getValue()).reduce("", (x, y) -> x + "\n" + y).concat("\n");
+
   @Test
   public void testUserAuthorizationSuccess() throws Exception {
     RuleExitResult result = createRuleRunMatch(Lists.newArrayList("group1", "group3"),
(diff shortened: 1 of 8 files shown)


sscarduzio:
Awesome! Thanks @rbkasat.
I have a really simple question: why is it called "default_query_parameters" and "default_headers"? I mean, why "default"?

core/src/test/java/tech/beshu/ror/acl/blocks/rules/impl/GroupsProviderAuthorizationAsyncRuleTests.java
```diff
@@ -59,7 +60,7 @@ public void testUserAuthorizationSuccess() throws Exception {
     result = createRuleRunMatch(Lists.newArrayList("group1", "group3"),
         "/complex/groups", "POST",
         ImmutableMap.of("Content-Type", "application/x-www-form-urlencoded",
-            "Authorization", "Basic base64_encoded_token"),
```
sscarduzio: Why not to use the standard auth header "Authorization"?

rbkasat:
Since these headers are static and will be added to each request by default. 

Just wanted to make expectation clear that the headers specified in this block will not be scraped from a request eg: if ES API call comes with a bunch of headers, people may want to use those headers instead of statically configured ones from yml file. 
same for request parameters.

This would solve the purpose mainly for authentication while calling APIs. Authentication details are generally static.

if a default doesn't sound very good, please suggest, I can get rid of it.

sscarduzio:
It's not that it doesn't sound good, it would be nice to have the headers and QP values being replaced by [@{dynamic_variables}](https://github.com/beshu-tech/readonlyrest-docs/blob/master/elasticsearch.md#dynamic-variables) like we have for the [kibana_index rule](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/master/core/src/main/java/tech/beshu/ror/acl/blocks/rules/impl/KibanaIndexSyncRule.java#L55), and many others:



This will enable the reuse of the values of the incoming HTTP request headers, or the resolved user's name. What do you think?

core/src/test/java/tech/beshu/ror/acl/blocks/rules/impl/GroupsProviderAuthorizationAsyncRuleTests.java
rbkasat: this is just a test to show even basic auth can be simply passed in one argument, no extra logic required. the header can be anything. it's just a name-value pair for users.

rbkasat:
yes, sounds like a good idea but can be done in iterations :) if you agree. the first iteration static params and second iteration may be dynamic params unless its a minor change?

PR #359, opened 2018-05-10
Add getContent() method to the RequestContextShim interface.

elaPa:
This is to enable including the body of the request in the custom
AuditLogSerializer.

This method was already provided in all concrete implementation, this is just exposing it via the interface.


elaPa:
Thank you for the super quick response (as always). I just signed the CLA (needed the approval from my employer, that's why it took a while ;) ).
Thank you!

PR #362, opened 2018-05-15
Result map was created but never used.

elaPa:
As the consequence, "match" and "block" were never included in the
audit log entry.


sscarduzio:
Weird one, must have been a result of a wild copy paste from old code. Thanks for spotting this 😮

sscarduzio:
Looks like serializationTest fails

Issue #390, opened 2018-09-26
jwt_auth rule throws ArrayIndexOutOfBoundsException


pdesoyres:
When jwt_auth is defined to use an external validator and when the jwt token passed is not a valid JWT, an ArrayIndexOutOfBoundsException is thrown in code 
```
String[] ar = token.get().split("\\.");
String tokenNoSig = ar[0] + "." + ar[1] + ".";
jws = parser.parseClaimsJwt(tokenNoSig).getBody();
```

We should check for the `ar` array length before trying to get `ar[1]`.

PR #391, opened 2018-09-26
fix issue [jwt_auth rule throws ArrayIndexOutOfBoundsException #390] :

pdesoyres:
When using external validator in jwt_auth rule,
 avoid ArrayIndexOutOfBoundsException to be thrown when JWT is not valid


sscarduzio:
Thank you @pdesoyres!

Issue #399, opened 2018-12-14
jwt_auth uses wrong cache key when trying to validate with an external validator


pdesoyres:
When jwt_auth has got an external validator with cache enabled, the key "x" is always used as the entry of the cache. So the cache will have only one entry and will respond the same for all tokens.

The responsible line of code is here : https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/526c5004b8f5eced8fe199bff9b5e72c989337b2/core/src/main/java/tech/beshu/ror/acl/blocks/rules/impl/JwtAuthSyncRule.java#L186

I suggest to use the token itself as the entry of the cache.

PR #400, opened 2018-12-14
jwt_auth uses wrong cache key when trying to validate with an external validator #399

pdesoyres:
pull request fixing issue #399


sscarduzio:
hello @pdesoyres, thanks for the pull request! Could you please make sure that the key of the token is hashed when used as a cache key? A token is like a password, we need to make sure it's hashed also in memory.

sscarduzio:
Thank you @pdesoyres !!!

PR #407, opened 2019-02-06
Allow custom jwt header prefix

lucasmogari:
I'm saving the token in cookies (https://docs.micronaut.io/latest/guide/index.html#cookieToken), so by default, the request is sent with a different header.

For example, instead of Authorization: Bearer <token>, allow this: Cookie: JWT=<token>

### DIFF
--- a/es63x/src/main/java/tech/beshu/ror/es/RequestInfo.java
+++ b/es63x/src/main/java/tech/beshu/ror/es/RequestInfo.java
@@ -62,7 +62,6 @@ import tech.beshu.ror.commons.utils.RCUtils;
 import tech.beshu.ror.commons.utils.ReflecUtils;
 
 import java.lang.reflect.Field;
-import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
@@ -445,17 +444,7 @@ public class RequestInfo implements RequestInfoShim {
 
   @Override
   public String extractRemoteAddress() {
-    InetSocketAddress remoteAddress = (InetSocketAddress) request.getRemoteAddress();
-    String remoteHost = null;
-    if (remoteAddress != null) {
-    	InetAddress address = remoteAddress.getAddress();
-    	if (address != null) {
-    		remoteHost = address.getHostAddress();
-    	}
-    }
-    if (remoteHost == null) {
-    	remoteHost = request.header("X-Forwarded-For");
-    }
+    String remoteHost = ((InetSocketAddress) request.getRemoteAddress()).getAddress().getHostAddress();
     // Make sure we recognize localhost even when IPV6 is involved
     if (RCUtils.isLocalHost(remoteHost)) {
       remoteHost = RCUtils.LOCALHOST;
--- a/es63x/src/main/java/tech/beshu/ror/es/RequestInfo.java
+++ b/es63x/src/main/java/tech/beshu/ror/es/RequestInfo.java
@@ -62,6 +62,7 @@ import tech.beshu.ror.commons.utils.RCUtils;
 import tech.beshu.ror.commons.utils.ReflecUtils;
 
 import java.lang.reflect.Field;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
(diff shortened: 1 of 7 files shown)


es63x/src/main/java/tech/beshu/ror/es/RequestInfo.java
```diff
@@ -444,7 +445,17 @@ public String extractLocalAddress() {
...
-    String remoteHost = ((InetSocketAddress) request.getRemoteAddress()).getAddress().getHostAddress();
+    InetSocketAddress remoteAddress = (InetSocketAddress) request.getRemoteAddress();
+    String remoteHost = null;
+    if (remoteAddress != null) {
+    	InetAddress address = remoteAddress.getAddress();
+    	if (address != null) {
+    		remoteHost = address.getHostAddress();
+    	}
+    }
+    if (remoteHost == null) {
```
sscarduzio: when is the remote address ever going to be null? Did this happen to you?

sscarduzio:
Nice one @lucasmogari! Pretty useful feature.

es63x/src/main/java/tech/beshu/ror/es/RequestInfo.java
lucasmogari: I got a NPE when using nginx in front of ES with the below configuration. I don't know why it happened.

```
server {
    listen       9201;
    server_name  localhost;

    location / {
        proxy_pass http://host.docker.internal:9200;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    }
}
```

sscarduzio: That's a problem: you should definitely always get at least the address of the proxy there. If not, there's a bug in how we extract the address from the request.

Also, falling back to X-Forwarded-For exposes people who don't use proxies to the risk an attacker injects an arbitrary address in that header and bypass all "hosts" rules.

One more thing is that when you do a change to a sub-project like es63x you should make all other sub-projects coherent. That is, find RequestInfo.java in other sub-projects and change them too.

sscarduzio: Could you please reproduce the no remote-address situation and try to see with some println or the debugger how comes it's null and if its' really null?

lucasmogari: Sure, I'll try to do that today.

lucasmogari: When requesting port 9200:
LocalAddress: /172.17.0.2:9200
RemoteAddress: /172.17.0.1:36310

When requesting from nginx:
LocalAddress: 0.0.0.0/0.0.0.0:9200
RemoteAddress: null

I'm running ES inside a docker container. So I don't know if this is problem.

I reverted the commit of RequestInfo and will create a new branch to investigate this better.

sscarduzio: thanks @lucasmogari! This is extra weird because the address should come from the TCP stack. 

Are both nginx and ES running in docker containers?

lucasmogari: Yes, it's really weird. I tried now with both ES and nginx in container and with local install and it's the same problem.

Now what is more weird is that if I debug and inspect the channel, it resolves the remote address. I'm clueless now.

screen shot 2019-02-07 at 12 07 14

sscarduzio: Oh, nice! So the information is there, but we can't get it out. Makes more sense. 

I was thinking maybe when you are paused in that break point, can you go with the "Evaluate Expression" feature in intellij Idea and try to call some methods until you are able to pull the address out in some alternative way? Would that work?

lucasmogari: After some research, I found that nginx with only 'proxy_pass' will proxy the request and close the connection. That's why the channel didn't get the remote address. In sun.nio.ch.SocketChannelImpl, the state was ST_KILLPENDING.

With the directives 'proxy_http_version' and 'proxy_set_header', the connection will keep alive until resolved.
```
    location / {
        proxy_pass http://127.0.0.1:9200;
        proxy_http_version 1.1;
        proxy_set_header Connection "";
    }
```

So no need to change RequestInfo!

Thank you for the assistance.

sscarduzio:
thank you @lucasmogari!

PR #410, opened 2019-02-08
Made HTTP Connection configurable

lillesand:
LDAP is already configurable, but this does not seem to be the case for other external autentication providers.

I considered adding a retry handler, but it doesn't look like that is supported in AsyncHttpClient, so it's probably better to not clutter things with some custom implementation of that (at least for now!).

Remaining work:

- [x] Docs
- [x] Consider adding retry support using [Apache DefaultHttpRequestRetryHandler](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/DefaultHttpRequestRetryHandler.html)

### DIFF
--- a/core/src/main/java/tech/beshu/ror/acl/definitions/DefinitionsFactory.java
+++ b/core/src/main/java/tech/beshu/ror/acl/definitions/DefinitionsFactory.java
@@ -76,7 +76,6 @@ public class DefinitionsFactory implements LdapClientFactory,
 
   public DefinitionsFactory(ESContext context, ACL acl) {
     this.acl = acl;
-    //this.httpClient = new ApacheHttpCoreClient(context);
     this.context = context;
     this.groupsProviderLdapClientsCache = CacheBuilder.newBuilder().build();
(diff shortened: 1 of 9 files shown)


sscarduzio:
Nice feature! Thanks @lillesand! Just don't forget the CLA, otherwise I can't merge 😉

lillesand:
Note that this will change the default behavior. Most notably, the default specified in [HttpConnectionSettings](core/src/main/java/tech/beshu/ror/settings/HttpConnectionSettings.java) will come in effect for all unconfigured external authentication providers. Especially for low traffic ES instances with slow authentication integrations that might lead to errors that aren't there today.

I'd say the default values are fairly conservative, but the will of course depend on context. I think it makes sense to do it this way as this in most contexts will be better than the defaults, but I'm definitely open for other views!

lillesand:
Okay, I think this is ready for review now 👍.

Let me know what your expectations on testing are – I haven't done any live testing of this, as I don't have a rig readily set up.

PR #417, opened 2019-03-27
fixed: kibana index and kibana access rules in config - processing or…

coutoPL:
…der; verbosity level issue


sscarduzio:
missing a license banner, will fix it myself after merge. Thanks.

PR #418, opened 2019-03-28
Bugfix/new code bugs

coutoPL:


sscarduzio:
should this fix also the equivalent of the .join() in the old core?

coutoPL:
> should this fix also the equivalent of the .join() in the old core?

not yet

PR #456, opened 2019-06-13
Feature/rordev 49

coutoPL:
- new algorithm of ROR starting
- more code in core 
- refactoring
- at the moment should work only for es70x


sscarduzio:
"The job exceeded the maximum log length, and has been terminated." 
We still have verbosity issues with testcontainers?

coutoPL:
should be fine now

PR #477, opened 2019-07-28
Feature/rordev 71

coutoPL:
Impersonation feature - admin user can make a call as a given user and if current configuration (authentication rules used) supports impersonation, it's going to be able to see the same result as if the user was doing the action. 

To be able to impersonate a user by the admin, the new section has to be added to ROR configuration, eg:

```
readonlyrest:
  [...]  
  impersonation:
    - impersonator: admin1  // the user who is going to impersonate the user
      auth_key: admin1:pass // admin authentication method (any auth method supported by ROR can be user)
      users: ["*"] // list of user IDs which can be impersonated ("*" means all users)
    - impersonator: admin2
      auth_key: admin2:pass
      users: ["dev2"]
```
To impersonate given user, `impersonate_as` header should be added to request, eg:

```curl -kv -u admin:pass -H "impersonate_as: dev2" -H "Content-Type: application/json" "https://localhost:90200/test1_index/_search?pretty"```

Due to impersonation feature, the forbidden response payload was changed, eg:
```
{
  "error": {
    "root_cause": [
      {
        "reason": "forbidden",
        "due_to": [
          "OPERATION_NOT_ALLOWED",
          "IMPERSONATION_NOT_SUPPORTED"
        ]
      }
    ],
    "reason": "forbidden",
    "due_to": [
      "OPERATION_NOT_ALLOWED",
      "IMPERSONATION_NOT_SUPPORTED"
    ]
  },
  "status": 401
}
```

`due_to` field contains a list of causes why request was forbidden. Possible values are listed below:
* FORBIDDEN_BY_BLOCK - block with forbidden policy was matched
* OPERATION_NOT_ALLOWED - no block was matched
* IMPERSONATION_NOT_ALLOWED - there was as least one block where impersonation failed, due to admin authentication or authorization failure
* IMPERSONATION_NOT_SUPPORTED - there was at least one block where authentication rule was not support impersonation

### CODE REVIEW

File: core/src/main/scala/tech/beshu/ror/acl/Acl.scala [hunk 1]
```diff
@@ -36,7 +38,15 @@ object AclHandlingResult {
   object Result {
     final case class Allow(blockContext: BlockContext, block: Block) extends Result
     final case class ForbiddenBy(blockContext: BlockContext, block: Block) extends Result
-    case object ForbiddenByUnmatched extends Result
+    final case class ForbiddenByMismatched(causes: NonEmptySet[Cause]) extends Result
+    case object ForbiddenByMismatched {
+      sealed trait Cause
+      object Cause {
+        case object ActionNotAllowed extends Cause
```
sscarduzio: Here the "Action" word is generic for the whole set of characteristics of the request? i.e. not the action as in the actions rule. Right?

coutoPL: yes. Maybe it should be called OperationNotAllowed

sscarduzio: Love it. Yes, let's change it and merge.

PR #491, opened 2019-08-26
Feature/rordev 85

coutoPL:
better current user metadata request handling


core/src/main/java/tech/beshu/ror/shims/request/RequestInfoShim.java
```diff
@@ -20,9 +20,11 @@
 import com.google.common.collect.Sets;
 import tech.beshu.ror.utils.MatcherWithWildcards;
 
+import javax.swing.text.html.Option;
```
sscarduzio: is this intentional?

coutoPL: hell no! will fix it

coutoPL: unused header. Removed

Issue #478, opened 2019-07-30
unable to authorize with unusual ldap schema


Vebryn:
Hi,

I'm unable to use ldap authorization because my ldap implements an unusual schema.

A ldap query on my group reveal that memberUid (uniqueMember) attribute contains only my login id (instead of full dn).

```
dn: cn=mygroup,ou=Group,dc=x,dc=x,dc=x
memberUid: jdurand
memberUid: sdupond
```

According to _UnboundidLdapService.scala_, _searchFilterFrom_ method hard code a full dn search :

```
  private def searchFilterFrom(mode: DefaultGroupSearch, user: LdapUser) = {
    s"(&${mode.groupSearchFilter}(${mode.uniqueMemberAttribute}=${Filter.encodeValue(user.dn.value.value)}))"
}
```

Is it possible to create a configuration parameter to customize this search filter ?

Best regards.

---

Ldapsearch on jdurand :

```
dn: uid=jdurand,ou=People,dc=x,dc=x,dc=x
accessgroup: mygroup
uid: jdurand
```

Elasticsearch logs :

```
[2019-07-30T12:14:48,122][DEBUG][t.b.r.a.b.d.l.LoggableLdapAuthenticationServiceDecorator] Trying to authenticate user [jdurand] with LDAP [ldap1]
[2019-07-30T12:14:48,245][DEBUG][t.b.r.a.b.d.l.LoggableLdapAuthenticationServiceDecorator] User [jdurand]  authenticated by LDAP [ldap1]
[2019-07-30T12:14:48,245][DEBUG][t.b.r.a.b.d.l.LoggableLdapAuthorizationServiceDecorator] Trying to fetch user [id=jdurand] groups from LDAP [ldap1]
[2019-07-30T12:14:48,266][DEBUG][t.b.r.a.b.d.l.i.UnboundidLdapAuthorizationService] LDAP search string: (&(cn=*)(memberUid=uid=jdurand,ou=People,dc=x,dc=x,dc=x)) | groupNameAttr: cn
[2019-07-30T12:14:48,283][DEBUG][t.b.r.a.b.d.l.LoggableLdapAuthorizationServiceDecorator] LDAP [ldap1] returned for user [jdurand] following groups: []
```

RoR configuration :

```
    - name: "mygroup access"
      ldap_auth:
        name: "ldap1"
        groups: ["mygroup"]
      indices: ["*"]
  ldaps:
    - name: "ldap1"
[...]

Issue #402, opened 2018-12-21
when objectClass=posixGroup, can't get user groups.


kingmorning:
Hi, i use this plugin to connect LDAP, 
In our LDAP, i can get user info by uid, but can not get user groups.(objectClass=posixGroup . not groupOfNames)
I checked the source code, and find. 
`   String searchString = String.format(
        "(&%s(%s=%s))",
        userGroupsSearchFilterConfig.getGroupSearchFilter(),
        userGroupsSearchFilterConfig.getUniqueMemberAttribute(),
        Filter.encodeValue(user.getDN())
    );`
user.getDN() has too many things, we just need a uid.
so i changed this code , just use user.getUid() in the searchString, and got the groups success.

I think maybe the searchString should be build by objectClass.

sscarduzio:
@coutoPL  any comments on this?

coutoPL:
We plan to reimpIement LDAP rules in a moment. I'll take this into consideration.

Vebryn:
Related to #300 and #478

Issue #300, opened 2017-12-17
uid in unique_member_atttribute


StupidScience:
Hello.
It is always `unique_member_atttribute=user_dn` in group search filter.
But in my case it should be `unique_member_atttribute=user_uid`.

I understand how to hardcode it but don't think this is the best solution.
Any chance to get an option in settings? Something like `user_attribute_in_group`.

slavavalkanova:
I am having the same issue -
even when I use the unique_member_atttribute: "member"  , in the debug logs I can see that it is returning
[DEBUG][t.b.r.a.d.l.u.UnboundidGroupsProviderLdapClient] LDAP search string: (&(cn=*)(uniqueMember=CN=***

Can you please get in touch with us? Thanks.

slavavalkanova:
Hi mate,
As I had the same issue, the attribute is working, but what we are using is mispelled:  triple "t"
Just write it properly and it will work:  unique_member_attribute

Vebryn:
> 
> 
> Hello.
> It is always `unique_member_atttribute=user_dn` in group search filter.
> But in my case it should be `unique_member_atttribute=user_uid`.
> 
> I understand how to hardcode it but don't think this is the best solution.
> Any chance to get an option in settings? Something like `user_attribute_in_group`.

Same for me #478

PR #492, opened 2019-08-29
Add a boolean group_attribute_is_dn to query user_dn or user_uid

Vebryn:
Hi,

Apache allows to authorize an user with its dn or uid (cf. http://httpd.apache.org/docs/2.4/mod/mod_authnz_ldap.html#authldapgroupattributeisdn). This change adds a boolean to reproduce Apache behavious with RoR. It solves #402, #300, and #478.

Best regards.


coutoPL:
LGTM. Thanks for your effort!

PR #501, opened 2019-09-14
Bugfix/rordev 114

coutoPL:


sscarduzio:
what was the problem? A bug in Snake Yaml?

coutoPL:
> what was the problem? A bug in Snake Yaml?

We were using in our core the newest circe-yaml which uses snakeyaml 1.24 underneath. When I was profiling ROR I've noticed that there are non garbage collected objects from snakeyaml. I've decided to downgrade snakeyaml to 1.17 (which is used by ES internally) and there was no leak any more. I've tried also next version, 1.18 and the leak comes back. So, I've decided to stay with 1.17. I've proved also that the leak is in snakeyaml, not in circe-yaml.

Same leak was also experienced by another user and reported here:
https://github.com/circe/circe-yaml/issues/73

sscarduzio:
Amazing job! Let's notify the reporters of the ror bug about this resolution

PR #503, opened 2019-09-21
[Rordev-101] Add new runtime variable for current group

pondzix:
- Added new runtime variable `@{acl:current_group}`
- Added namespace for user variable and support for both versions: `@{user}` and `@{acl:user}`
- Changed rules evaluation order - moved authorization rules before other rules requiring variable for current group
- Added `VariableContext` which is responsible for defining requirements for other rules in block when particular variable is used
- Added support for multivariables in uri regex rule


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -0,0 +1,111 @@
...
+  sealed trait VariableType
+
+  object VariableType {
+    trait User extends VariableType
+    trait CurrentGroup extends VariableType
+    trait Header extends VariableType
+    trait Jwt extends VariableType
+  }
+
+  trait UsingVariable {
```
coutoPL: IMO we should not extend rules. AFAIU we use it to provide proper validation of used variables in context of given block. This validation is correlated with configuration loading. All code we can find in rule should be related to processing the ES request. So, it doesn't fit there.

But we can solve this in pretty simple way - by defining proper typeclasses for given rules (at the level of rule instance creation). The validation will take advantage of code from the typeclasses, not rule classes directly.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -0,0 +1,111 @@
...
+ */
+package tech.beshu.ror.accesscontrol.blocks.variables.runtime
+
+import cats.data.NonEmptyList
+import tech.beshu.ror.accesscontrol.blocks.rules.Rule
+import tech.beshu.ror.accesscontrol.blocks.variables.runtime.MultiExtractable.SingleExtractableWrapper
+
+object VariableContext {
+
+  sealed trait VariableType
```
coutoPL: Do we have to introduce VariableType? Maybe we can use just Extractable to do the same? 
eg. SingleExtractable.UserIdVar & MultiExtractable.UserIdVar could be treated as VariableType.User.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -0,0 +1,111 @@
...
+      usedVariables.toList
+        .flatMap {
+          case RuntimeSingleResolvableVariable.ToBeResolved(extractables) => extractSingle(extractables)
+          case RuntimeMultiResolvableVariable.ToBeResolved(extractables) => extractMulti(extractables)
+          case _ => List.empty
+        }
+    }
+
+    private def extractSingle(extractables: NonEmptyList[SingleExtractable]) =
+      extractables.collect { case e: VariableContext.VariableType => e }
```
coutoPL: do we have to use collect? Do we want to filter sth? Maybe we can use just map - it's way safer when you want to add sth in future

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    private def extractMulti(extractables: NonEmptyList[MultiExtractable]) =
+      extractables.collect {
```
coutoPL: same here

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
```diff
@@ -136,6 +137,19 @@ object SingleExtractable {
...
+
+  case object CurrentGroupVar extends SingleExtractable with VariableType.CurrentGroup {
+
+    override def extractUsing(requestContext: RequestContext,
+                              blockContext: BlockContext): Either[ExtractError, String] = {
+      blockContext
+        .currentGroup
+        .map(_.value.value) match {
+        case Some(value) => Right(value)
+        case None => Left(ExtractError(s"Cannot extract user current group from block context"))
```
coutoPL: IMO the message should be changed to sth like this:
"There was no current group found for request: ${requestContext.id}"

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/RawRorConfigBasedCoreFactory.scala
```diff
@@ -180,7 +180,7 @@ class RawRorConfigBasedCoreFactory(implicit clock: Clock,
     rulesNelDecoder(definitions)
       .toSyncDecoder
       .emapE { rules =>
-        RulesValidator.validate(rules) match {
+        RulesValidator.validate(rules.sorted) match {
```
coutoPL: I wonder if we shouldn't change RulesValidator to BlocksValidator and pass there whole block. At the moment we have sorting of rules in two places. This is fragile part of the ROR code, so maybe we should be careful here. And it seems that we rather validate block, not rules per se.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/variables/RuntimeResolvableVariablesTests.scala
```diff
@@ -137,6 +146,37 @@ class RuntimeResolvableVariablesTests extends WordSpec with MockFactory {
...
+      "current group variable is used and initial group is present" in {
+        val requestContext = MockRequestContext(headers = Set(headerFrom("x-ror-current-group" -> "g1")))
+        val variable = forceCreateSingleVariable("@{acl:current_group}")
+          .resolve(
+            requestContext,
+            fromRequestContext(requestContext)
+          )
+        variable shouldBe Right("g1")
+      }
+      "current group multivariable is used and some groups has been added as available" in {
```
coutoPL: what about tests where currentGroup variable cannot be resolved?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
pondzix: I think introducing `VariableType` makes design and intention more explicit, because we define requirements for variables, not for extractables (there is for example `Const` extractable and it is not related to variables in any way). If I used extractables I would need to use `Extractable[_]` everywhere and when pattern matching think about single and multivariables like:
```scala 
case SingleExtractable.UserIdVar | MultiExtractable.UserIdVar =>
```
For example defining `Show` instances would be less readable. So `VariableType` is grouping extractables (single and multi) under set of available variables.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
pondzix: We want use only variables, so for example `Const` should not be collected.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
pondzix: explained in previous comment

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/RawRorConfigBasedCoreFactory.scala
pondzix: I created factory method where I'm validating passed rules. I think we don't want to create block and then proceed with validation, but instead validate all data before creation, so we can not create block in invalid state, right? :) I didn't make constructor for block private because it's used in tests...but maybe I should and then use factory method there as well?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -0,0 +1,128 @@
...
+      case rule: GroupsRule => rule.settings.groups.toList
+      case rule: HostsRule => rule.settings.allowedHosts.toNonEmptyList.toList
+      case rule: IndicesRule => rule.settings.allowedIndices.toNonEmptyList.toList
+      case rule: KibanaAccessRule => rule.settings.kibanaIndex :: Nil
+      case rule: KibanaIndexRule => rule.settings.kibanaIndex :: Nil
+      case rule: LocalHostsRule => rule.settings.allowedAddresses.toNonEmptyList.toList
+      case rule: UriRegexRule => rule.settings.uriPatterns.toNonEmptyList.toList
+      case rule: UsersRule => rule.settings.userIds.toNonEmptyList.toList
+      case rule: XForwardedForRule => rule.settings.allowedAddresses.toNonEmptyList.toList
+      case _ => List.empty
```
coutoPL: this approach is dangerous. Do you think next developer will remember to add rule which uses variables here? I don't think so. Compiler should help us. To do so, you have to defined typeclass instance for each rule type. But you cannot just replace `_` with rest of rules, because Rule is not sealed trait.

But you can provide typeclass for each rule, eg:
```
implicit val filterRuleUsingVariable: UsingVariable[FilterRule] = rule => rule.settings.filter :: Nil
```
Then you can count on compiler which will tell you that you that there is no such typeclass instance defined for newly created Rule class.

And it's better to not defined typeclass for abstract, base classes (like BaseSpecializedIndicesRule)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -0,0 +1,128 @@
...
+    sealed trait ComplianceResult
+    object ComplianceResult {
+      case object Compliant extends ComplianceResult
+      final case class NonCompliantWith(usageRequirement: UsageRequirement) extends ComplianceResult
+    }
+
+    def definedFor(variableType: VariableType): Option[UsageRequirement] = {
+      variableType match {
+        case v: VariableType.User => Some(UsageRequirement.OneOfRuleBeforeMustBeAuthenticationRule(v))
+        case v: VariableType.CurrentGroup => Some(UsageRequirement.OneOfRuleBeforeMustBeAuthorizationRule(v))
```
coutoPL: I've just realized that we shouldn't use this requirement. Current group also can be sent by kibana plugin.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/BlockValidator.scala
```diff
@@ -0,0 +1,81 @@
...
+      case Some(_) if rules.exists(_.isInstanceOf[AuthenticationRule]) => Validated.Valid(())
+      case Some(_) => Validated.Invalid(NonEmptyList.one(BlockValidationError.AuthorizationWithoutAuthentication))
+    }
+  }
+
+  private def validateRequirementsForRulesUsingVariables(allRules: NonEmptyList[Rule]): ValidatedNel[BlockValidationError, Unit] = {
+    allRules.toList
+      .map(rule => validateRequirementsForSingleRule(allRules)(rule)) match {
+      case Nil => Validated.Valid(())
+      case ::(head, tl) => NonEmptyList(head, tl).sequence_
```
coutoPL: auto generated cases for list are ugly ;)

coutoPL:
LGTM. Thanks :)

coutoPL:
@pondzix 
* please add at the beginning of PR description one line, which describes the changes according to @sscarduzio requirements
* prepare PR with modified doc (https://github.com/beshu-tech/readonlyrest-docs)

PR #504, opened 2019-09-21
Rordev 75

clutroth:


clutroth:
I pushed reflection approach. I tired to use `mustache-lang` module as compile time dependency, and use `ExtendedPluginClassLoader` in runtime, but I wasn't able to make it work. I was getting `NoClassDefFoundError`
```scala
object RequestInfo {
  private object SearchTemplateRequestIndices
    extends Logging {
    private val className = "SearchTemplateRequest"

    def unapply(arg: CompositeIndicesRequest): Option[Set[String]] = {
      if (arg.getClass.getCanonicalName.contains(className)) {
        val pluiginClassLoader = arg.getClass.getClassLoader
        val contextClassLoader = Thread.currentThread().getContextClassLoader
        logger.info(s"Plugin class loader: ${pluiginClassLoader}")
        logger.info("request class: org.elasticsearch.script.mustache.SearchTemplateRequest")
        logger.info(s"${pluiginClassLoader.loadClass("org.elasticsearch.script.mustache.SearchTemplateRequest")}")
//        contextClassLoader.loadClass()
        ScalaClassLoader(pluiginClassLoader).asContext{
        try {
          if (arg.isInstanceOf[org.elasticsearch.script.mustache.SearchTemplateRequest]) {
            val indices = arg.asInstanceOf[org.elasticsearch.script.mustache.SearchTemplateRequest]
              .getRequest.indices().toSet
            Some(indices)
          } else {
            logger.error(s"${arg.getClass.getCanonicalName} is not instance of ${classOf[org.elasticsearch.script.mustache.SearchTemplateRequest]}")
            Some(Set.empty)
          }
        } catch {
          case e: NoClassDefFoundError =>
            throw IndicesClassNotFoundOnClasspath(arg.getClass.getCanonicalName, e)
        }
        }
      } else None
    }
  }
  final case class IndicesClassNotFoundOnClasspath(className:String, error:NoClassDefFoundError)
    extends Error(s"$className recognized, but not found on classpath", error)
}

```

PR #509, opened 2019-09-25
Rordev 75

clutroth:


coutoPL:
@clutroth please check the travis

PR #510, opened 2019-09-26
changed: all rules should be evaluated during current_user_metadata r…

coutoPL:
bugfix: identity request will check all rules in each ACL block


sscarduzio:
we need a pre to test this at the customers'

core/src/main/scala/tech/beshu/ror/accesscontrol/acl/AccessControlList.scala
```diff
@@ -132,7 +132,7 @@ class AccessControlList(val blocks: NonEmptyList[Block])
 
   private def executeBlocksUserMetadataRulesOnly(block: Block, context: RequestContext) = {
```
clutroth: shouldn't method name be updated?

PR #513, opened 2019-10-06
[Rordev-54] Internode ssl improvements

pondzix:
**New** Internode ssl support for es5x, es60x, es61x and es62x
**New** Internode ssl integration test
**Enhancement** Splitting `verification` ssl property to `client_authentication` and `certificate_verification`


integration-tests-scala/src/test/scala/tech/beshu/ror/integration/ClusterStateWithInternodeSslTests.scala
```diff
@@ -0,0 +1,45 @@
...
+  )
+
+  private lazy val adminClusterStateManager = new ClusterStateManagerJ(container.nodesContainers.head.adminClient)
+
+  "Health check" should {
+    "be successful" when {
+      "internode ssl between nodes is enabled" in {
+        val response = adminClusterStateManager.healthCheck()
+
+        response.getResponseCode should be(200)
```
sscarduzio: should we also verify the count of cluster nodes = 3?

pondzix: I think it's not necessary.  During cluster initialization we check cluster health with configured retrying policy. If status is green (all nodes joined cluster)  we can continue with test execution, so here we are sure that cluster is in valid state and all (in this case 3) nodes are up.

core/src/main/scala/tech/beshu/ror/configuration/SslConfiguration.scala
```diff
@@ -83,7 +83,8 @@ final case class SslConfiguration(keystoreFile: JFile,
                                   keyAlias: Option[SslConfiguration.KeyAlias],
                                   allowedProtocols: Set[SslConfiguration.Protocol],
                                   allowedCiphers: Set[SslConfiguration.Cipher],
-                                  verifyClientAuth: Boolean)
+                                  clientAuthenticationEnabled: Boolean,
+                                  certificateVerificationEnabled: Boolean)
```
coutoPL: this class is used to describe external SSL and internode SSL. At the moment certificate verification can be disabled only for internode SSL. Maybe we should allow user to disable it also for external SSL (eg. self signed cert usage). 

@sscarduzio @pondzix WDYT?

es73x/src/main/scala/tech/beshu/ror/es/ssl/SSLNetty4InternodeServerTransport.scala
```diff
@@ -53,8 +53,9 @@ class SSLNetty4InternodeServerTransport(settings: Settings,
     override def initChannel(ch: Channel): Unit = {
       super.initChannel(ch)
       logger.info(">> internode SSL channel initializing")
+      logger.info(s"SSL!! $ssl")
```
coutoPL: to remove?

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/DockerfileBuilderOps.scala
```diff
@@ -0,0 +1,30 @@
...
+ */
+package tech.beshu.ror.utils.containers
+
+import org.testcontainers.images.builder.dockerfile.DockerfileBuilder
+
+object DockerfileBuilderOps {
+
+  implicit class OptionalCommandRunner(val builder: DockerfileBuilder) extends AnyVal {
+
+    def runWhen(condition: Boolean, command: String): DockerfileBuilder = {
```
coutoPL: maybe 2 parameters list will make the DSL nicer:

```
[...]
  .runWhen(isSthEnabled) {
    // comand
  }
```
?

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
```diff
@@ -42,15 +42,17 @@ object ReadonlyRestEsCluster {
                                   nodeDataInitializer: ElasticsearchNodeDataInitializer = NoOpElasticsearchNodeDataInitializer,
                                   clusterInitializer: ReadonlyRestEsClusterInitializer = NoOpReadonlyRestEsClusterInitializer,
                                   dependentServicesContainers: List[DependencyDef] = Nil,
-                                  configHotReloadingEnabled: Boolean = true): ReadonlyRestEsClusterContainer =
+                                  configHotReloadingEnabled: Boolean = true,
+                                  internodeSslEnabled: Boolean = false): ReadonlyRestEsClusterContainer =
```
coutoPL: maybe it's the time, when we should introduce some config class to gather all additional settings of ROR container?

integration-tests-scala/src/test/scala/tech/beshu/ror/integration/ClusterStateWithInternodeSslTests.scala
```diff
@@ -0,0 +1,45 @@
...
+ */
+package tech.beshu.ror.integration
+
+import com.dimafeng.testcontainers.ForAllTestContainer
+import org.scalatest.Matchers._
+import org.scalatest.WordSpec
+import tech.beshu.ror.utils.containers.{ReadonlyRestEsCluster, ReadonlyRestEsClusterContainer}
+import tech.beshu.ror.utils.elasticsearch.ClusterStateManagerJ
+
+class ClusterStateWithInternodeSslTests extends WordSpec with ForAllTestContainer {
```
coutoPL: we don't need commented, java counterpart of this class - please remove it (InternodeSSLTest) and its resources file

core/src/main/scala/tech/beshu/ror/configuration/SslConfiguration.scala
sscarduzio: Definitely we should set the HTTPS not to verify client cert by default. Otherwise admins should distribute certificates to all the clients before they can connect. Client cert verification is not expected by default in HTTPS, and this is also the current behaviour. So it should continue to be like this, or would break all customers' upgrades.

coutoPL: I'm not talking about defaults but possibility to use self signed certs. Seems that at the moment you cannot do that. Default should be the same as now

pondzix: I split `SslConfiguration` to `InternodeSslConfiguration` and `ExternalSslConfiguration` to make it more explicit

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/DockerfileBuilderOps.scala
pondzix: IMO current approach looks better: 
```scala
.run("...")
.runWhen(internodeSslEnabled, "...")
.run("...")
```
than: 
```scala
.run("...")
.runWhen(internodeSslEnabled) {
 "..."
}
.run("...")
```

PR #514, opened 2019-10-09
fixed: non-string ENV variables proper resolving

coutoPL:
Fix (ES) wrong non-string ENV variables resolving


coutoPL:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/issues/519

PR #515, opened 2019-10-10
es 7.4.0

clutroth:


es74x/src/main/scala/tech/beshu/ror/es/rradmin/RRAdminResponse.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.elasticsearch.common.xcontent.{ToXContent, ToXContentObject, XContentBuilder}
+import tech.beshu.ror.adminapi.AdminRestApi
+
+class RRAdminResponse(response: Either[Throwable, AdminRestApi.AdminResponse])
+  extends ActionResponse with ToXContentObject with Logging {
+
+  def this(response: AdminRestApi.AdminResponse) {
+    this(Right(response))
+  }
+  override def writeTo(out: StreamOutput): Unit = {???}
```
coutoPL: Seems that we need to provide some implementation in new version of ES for this method. IMO we should do this for previous version. So, maybe when we do that for es73x module, there won't be necessary to create new es74x? WDYT? Can it be done or there are any other incompatibilities?

clutroth: There are. Some constructors are getting `StreamInput`

PR #520, opened 2019-10-23
fixed: issue with backward compatibility of proxy_auth settings

coutoPL:
Fix (ES) issue with backward compatibility of proxy_auth settings


coutoPL:
fix: https://github.com/beshu-tech/readonlyrest-docs/issues/29

PR #522, opened 2019-10-24
es 7.4.0

clutroth:


es73x/src/main/scala/tech/beshu/ror/es/rradmin/rest/RestRRAdminAction.scala
```diff
@@ -27,8 +26,8 @@ import tech.beshu.ror.adminapi._
 import tech.beshu.ror.es.rradmin.{RRAdminAction, RRAdminRequest, RRAdminResponse}
 
 @Inject
-class RestRRAdminAction(settings: Settings, controller: RestController)
-  extends BaseRestHandler(settings) with RestHandler {
+class RestRRAdminAction(controller: RestController)
+  extends BaseRestHandler(null) with RestHandler {
```
clutroth: there's only difference between es73x, and es74x. `BaseRestHandler(null)`, and `BaseRestHandler()`

es73x/src/main/scala/tech/beshu/ror/es/providers/EsAuditSink.scala
```diff
@@ -40,7 +40,7 @@ class EsAuditSink(client: Client) extends AuditSink with Logging {
       .build
 
   override def submit(indexName: String, documentId: String, jsonRecord: String): Unit = {
-    bulkProcessor.add(new IndexRequest(indexName, "ror_audit_evt", documentId).source(jsonRecord, XContentType.JSON))
+    bulkProcessor.add(new IndexRequest(indexName).source(jsonRecord, XContentType.JSON))
```
coutoPL: I know it's deprecated, but IMO we should leave it as it was before. Types will be removed in ES 8.x, so we will have to prepare for it after all (https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html). But IMO we have to do it wisely, with proper tests. Let's abandon the idea to change it now.

es73x/src/main/scala/tech/beshu/ror/es/ReadonlyRestPlugin.scala
```diff
@@ -178,7 +178,7 @@ class ReadonlyRestPlugin(s: Settings, p: Path)
                                indexNameExpressionResolver: IndexNameExpressionResolver,
                                nodesInCluster: Supplier[DiscoveryNodes]): util.List[RestHandler] = {
     List[RestHandler](
-      new RestRRAdminAction(settings, restController)
+      new RestRRAdminAction(restController)
```
coutoPL: I don't agree with this change either. I see that implementation of BaseRestHandler in ES 7.3.2 has "todo" in base constructor, but:
* we have to be sure that all versions, starting from ES 7.0.0 to ES 7.3.2, don't use the settings
* passing null as settings strongly relays on knowing the implementation details, not on interface and documentation

Please, back to previous version.

core/src/main/scala/tech/beshu/ror/accesscontrol/ops.scala
```diff
@@ -47,13 +48,47 @@ import tech.beshu.ror.com.jayway.jsonpath.JsonPath
...
 import tech.beshu.ror.utils.FilterTransient
+import tech.beshu.ror.utils.uniquelist.UniqueList
 import upickle.default
 
 import scala.collection.SortedSet
 import scala.concurrent.duration.FiniteDuration
 import scala.language.{implicitConversions, postfixOps}
 import scala.util.Try
 
+object codecs {
```
coutoPL: is it used?

es74x/src/main/scala/tech/beshu/ror/es/providers/EsAuditSink.scala
```diff
@@ -0,0 +1,70 @@
...
+      .builder(client, new AuditSinkBulkProcessorListener)
+      .setBulkActions(AUDIT_SINK_MAX_ITEMS)
+      .setBulkSize(new ByteSizeValue(AUDIT_SINK_MAX_KB.toInt, ByteSizeUnit.KB))
+      .setFlushInterval(TimeValue.timeValueSeconds(AUDIT_SINK_MAX_SECONDS.toInt))
+      .setConcurrentRequests(1)
+      .setBackoffPolicy(BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(100), AUDIT_SINK_MAX_RETRIES))
+      .build
+
+  override def submit(indexName: String, documentId: String, jsonRecord: String): Unit = {
+    bulkProcessor.add(new IndexRequest(indexName).source(jsonRecord, XContentType.JSON))
```
coutoPL: same here (there is no tests for it at the moment, so we have to be careful)

es73x/src/main/scala/tech/beshu/ror/es/ReadonlyRestPlugin.scala
clutroth: You're right. I forgot to ensure versions before 7.3.2.

core/src/main/scala/tech/beshu/ror/accesscontrol/ops.scala
clutroth: No

coutoPL:
Thanks.

PR #542, opened 2019-12-27
Feature/rordev 177

coutoPL:
🚀New (ES) Example projects with custom audit log serializers


sscarduzio:
Hey @coutoPL, I've been trying the UX of these new example projects, and I notice we can do better:

1. The base case when you want to create a custom log serializer, is to make incremental changes to the default JSON schema. So the custom serializer should initially behave identically to the default one. I.e. extend the base serializer, and additionally give the user the opportunity to mutate the default JSON object. As far as I understood, the example is creating an entirely new, smaller JSON object from scratch instead.

2. Why we need that full package name (with all the subdirectories)? We have the opportunityto use the default package, and remove the user from knowing what a package is. The vast majority of our users are not JVM developers. It's already a very big stretch we are asking them to write code..

coutoPL:
> Hey @coutoPL, I've been trying the UX of these new example projects, and I notice we can do better:
> 
> 1. The base case when you want to create a custom log serializer, is to make incremental changes to the default JSON schema. So the custom serializer should initially behave identically to the default one. I.e. extend the base serializer, and additionally give the user the opportunity to mutate the default JSON object. As far as I understood, the example is creating an entirely new, smaller JSON object from scratch instead.
> 2. Why we need that full package name (with all the subdirectories)? We have the opportunityto use the default package, and remove the user from knowing what a package is. The vast majority of our users are not JVM developers. It's already a very big stretch we are asking them to write code..

good points - please see: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/543

PR #543, opened 2019-12-28
Feature/rordev 177

coutoPL:
custom audit log serializer example improvements


coutoPL:
@sscarduzio do you want me to add sth more?

sscarduzio:
In order to keep the number of files low, and all code in one place, and minimising the "choice" factor, I suggest keeping this example to a single class. And perhaps use comments to reveal the two alternative ways to implement the serializer. I.e. The creation of the new, entirely custom log event VS the alteration of the default,pre-populated log event.

So I'd say: let's have in one single class, some commented code and some commented text to reveal the former (simpler, but less common) option. And assume that everyone wants to implement slight modifications to the default schema.

coutoPL:
Sounds reasonable. Please, check out current version.

sscarduzio:
Thank you! Good to go for me :)

PR #549, opened 2020-01-13
[Rordev-181] Add attempted user to default audit log serializer

pondzix:
**Enhancement** Enrich logs and audit with attempted username

### DIFF
--- a/audit/src/main/scala/tech/beshu/ror/audit/AuditRequestContext.scala
+++ b/audit/src/main/scala/tech/beshu/ror/audit/AuditRequestContext.scala
@@ -37,5 +37,6 @@ trait AuditRequestContext {
   def loggedInUserName: Option[String]
   def impersonatedByUserName: Option[String]
   def involvesIndices: Boolean
+  def attemptedUser: Option[String]
 
 }
--- a/audit/src/main/scala/tech/beshu/ror/audit/AuditRequestContext.scala
+++ b/audit/src/main/scala/tech/beshu/ror/audit/AuditRequestContext.scala
@@ -38,5 +38,5 @@ trait AuditRequestContext {
   def impersonatedByUserName: Option[String]
   def involvesIndices: Boolean
   def attemptedUserName: Option[String]
-
+  def rawAuthHeader: Option[String]
 }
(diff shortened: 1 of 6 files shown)


audit/src/main/scala/tech/beshu/ror/audit/AuditRequestContext.scala
```diff
@@ -37,5 +37,6 @@ trait AuditRequestContext {
   def loggedInUserName: Option[String]
   def impersonatedByUserName: Option[String]
   def involvesIndices: Boolean
+  def attemptedUser: Option[String]
```
coutoPL: maybe attemptedUserName?

coutoPL: and please write a few tests eg. in EnabledAuditingToolsTests

coutoPL: I've just notice that we need to have one more fallback - to raw auth header

core/src/main/scala/tech/beshu/ror/accesscontrol/request/RequestContext.scala
```diff
@@ -75,11 +75,11 @@ object RequestContext extends Logging {
            history: Vector[Block.History])
           (implicit headerShow: Show[Header]): Show[RequestContext] =
     Show.show { r =>
-      def stringifyLoggedUser = {
+      def stringifyUser = {
         loggedUser match {
           case Some(DirectlyLoggedUser(user)) => s"${user.show}"
           case Some(ImpersonatedUser(user, impersonatedBy)) => s"${impersonatedBy.show} (as ${user.show})"
-          case None => "[user not logged]"
+          case None => r.attemptedUserName.map(name => s"attempted as '${name.value}'").getOrElse("[no info about user]")
```
coutoPL: I don't think this is a good idea. We don't want to place eg. basic auth header value in logs. Audit is a different thing

sscarduzio: Yeah I was thinking the same, in logs let's log the attempted user, but not the whole auth header in case base http auth parsing fails.

PR #557, opened 2020-01-27
[RORDEV-191] force_load_from_file settings decoding fix

coutoPL:
🐞Fix (ES) readonlyrest.force_load_from_file settings decoding issue


core/src/main/scala/tech/beshu/ror/configuration/EsConfig.scala
```diff
@@ -57,12 +58,17 @@ object EsConfig {
...
-    implicit val rorEsLevelSettingsDecoder: Decoder[RorEsLevelSettings] = Decoder.instance { c =>
-      for {
-        forceLoadFromFile <- c.downField("readonlyrest").downField("force_load_from_file").as[Option[Boolean]]
-      } yield RorEsLevelSettings(
-        forceLoadFromFile.getOrElse(false)
-      )
+    implicit val rorEsLevelSettingsDecoder: Decoder[RorEsLevelSettings] = {
+      Decoder.instance { c =>
+        val oneLine = c.downField("readonlyrest.force_load_from_file").as[Option[Boolean]]
+        val twoLines =  c.downField("readonlyrest").downField("force_load_from_file").as[Option[Boolean]]
```
sscarduzio: Did I understand correctly that nested yaml and inline yaml are not equivalent for our parser normally? This is unexpected. There can be a ton of other configuration keys where customers prefere one or the other style (ES and Kibana encourage this equivalence in examples).

PR #561, opened 2020-02-08
[RORDEV-195] cutoff ipv6 zone id

clutroth:
🧐Enhancement (ES) literal ipv6 with zone id is acceptable network address.


coutoPL:
let's bump the pre version

gradle.properties
```diff
@@ -1,3 +1,3 @@
 publishedPluginVersion=1.18.4
-pluginVersion=1.18.5-pre16
+pluginVersion=1.18.5-pre17
```
coutoPL: hmmm ... current version is 1.19.1-pre6 :D
btw please add license headers in new files

coutoPL: still missing license header in test file

PR #563, opened 2020-02-11
es 7.6.0 support

clutroth:
🧐Enhancement (ES) ES 7.6.0 support

### DIFF
--- a/es74x-cloud/gradle.properties
+++ b/es74x-cloud/gradle.properties
@@ -1,1 +1,1 @@
-esVersion=7.5.1
+esVersion=7.6.0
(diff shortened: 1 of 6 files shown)


es74x-cloud/gradle.properties
```diff
@@ -1 +1 @@
-esVersion=7.5.1
```
clutroth: @coutoPL

coutoPL: it's ok

sscarduzio:
Hey guys, yesterday night on my laptop I had 100% integration test pass forcing es 7.6.0 dependency on the pre-existing es74x subproject. 

So why did we need to create es76x? I'm sure maybe there are reasons I didn't see?

The philosophy behind this question being the following: 
(I'm not sure if we ever talked about this, so let's make sure we are on the same page)

It's imperative to keep the code base at the bare minimum. I think this is valid for ANY code base. More code ALWAYS results (or will result) in more problems. 

As developers, "more code" is our enemy. As developers in security, "more code" is the antichrist.

Therefore, unless there's a significant API change that breaks the current esXXx subproject, let's keep reusing it for the new ES versions. 
I'd even be ready to use some reflection hacks to keep us from creating another subproject.

coutoPL:
OMG, @sscarduzio please take a look at files changed.

sscarduzio:
Shit I have to go to bed earlier. Sorry lol.

Issue #565, opened 2020-02-21
Add availableGroups as dynamic variable


kiblik:
Hi,

I know `@{acl:user}`, `@{acl:current_group}`, `@{header:...}` or `@{jwt:...}` but is it possible to implement dynamic variable also for available groups (not only first match - `current_group`). It is useful if somebody is a member of more groups and needs access to all indices like `["@explode{acl:available_groups}-*"]`.

Thank you

coutoPL:
@kiblik thanks for suggestion. We've added it to our backlog.

kiblik:
Hi @coutoPL,

I came up with a solution for the previously mentioned problem. It was my first contact with Scala and also with your plugin so I hope everything is correct.
It would be very helpful if you could check my patch, review it and add to your plugin.

I'm just working with it and it would be very helpful if it will be available in the next release of the plugin.

[0001-Add-availableGroups-as-dynamic-variable.patch.tar.gz](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/files/4240290/0001-Add-availableGroups-as-dynamic-variable.patch.tar.gz)

Thank you.

coutoPL:
@kiblik it's great that you managed to add it yourself. Could you please create pull request to our repo? When you do this, there will be one important step of the pipeline - signing the contribution agreement. If you will have any problems with it, don't hesitate to ask.

coutoPL:
it's done, so I'll close the issue

PR #566, opened 2020-02-23
Add available groups as dynamic variable

kiblik:
#565 

🧐Enhancement (ES) new acl:available_groups variable

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
@@ -155,13 +155,7 @@ object SingleExtractable {
 
     override def extractUsing(requestContext: RequestContext,
                               blockContext: BlockContext): Either[ExtractError, String] = {
-      if(blockContext.availableGroups.size > 0)
-        Right(blockContext
-          .availableGroups
-          .map(_.value.value)
-          .mkString(","))
-      else
-        Left(ExtractError(s"There were no groups for request: ${requestContext.id.show}"))
+      Left(ExtractError(s"Available groups are usable only with @explode"))
     }
   }
 }
@@ -240,13 +234,17 @@ object MultiExtractable {
   }
 
   case object AvailableGroupsVar extends MultiExtractable with VariableType.AvailableGroups {
-    private val singleAvailableGroupsExtractable = SingleExtractable.AvailableGroupsVar
-
     override def extractUsing(requestContext: RequestContext,
                               blockContext: BlockContext): Either[ExtractError, NonEmptyList[String]] = {
-      singleAvailableGroupsExtractable
-        .extractUsing(requestContext, blockContext)
-        .map(NonEmptyList.one)
+      NonEmptyList.fromList(
+        blockContext
+          .availableGroups
+          .map(_.value.value)
+          .toList
+      ) match {
+        case Some(value) => Right(value)
+        case None => Left(ExtractError(s"There were no groups for request: ${requestContext.id.show}"))
+      }
     }
   }
 }
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
@@ -150,6 +150,17 @@ object SingleExtractable {
       }
     }
   }
+
+  case object AvailableGroupsVar extends SingleExtractable with VariableType.AvailableGroups {
+
+    override def extractUsing(requestContext: RequestContext,
+                              blockContext: BlockContext): Either[ExtractError, String] = {
+      if(blockContext.availableGroups.size > 0)
+        Right(blockContext.availableGroups.mkString(","))
+      else
+        Left(ExtractError(s"There was no groups for request: ${requestContext.id.show}"))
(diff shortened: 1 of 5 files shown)


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
```diff
@@ -224,4 +238,15 @@ object MultiExtractable {
         .map(NonEmptyList.one)
     }
   }
+
+  case object AvailableGroupsVar extends MultiExtractable with VariableType.AvailableGroups {
+    private val singleAvailableGroupsExtractable = SingleExtractable.AvailableGroupsVar
+
+    override def extractUsing(requestContext: RequestContext,
+                              blockContext: BlockContext): Either[ExtractError, NonEmptyList[String]] = {
+      singleAvailableGroupsExtractable
```
coutoPL: I think this implementation is not correct. Please check out our docs: https://github.com/beshu-tech/readonlyrest-docs/blob/master/elasticsearch.md#dynamic-variables-from-jwt-claims.

This is used for `exploding` a variable. Here, available groups are always list, so we don't have to parse string to sprit it (like we do eg. with `JwtPayloadVar`).

Could you please reimplement it?

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/variables/RuntimeResolvableVariablesTests.scala
```diff
@@ -187,6 +187,37 @@ class RuntimeResolvableVariablesTests extends WordSpec with MockFactory {
...
+            MockRequestContext.default,
+            fromRequestContext(MockRequestContext.default).withAddedAvailableGroups(UniqueNonEmptyList.of(groupFrom("g1"), groupFrom("g2")))
+          )
+        variable shouldBe Right("g1,g2")
+      }
+      "available groups multivariable is used and explode" in {
+        val variable = forceCreateMultiVariable("@explode{acl:available_groups}")
+          .resolve(
+            MockRequestContext.default,
+            fromRequestContext(MockRequestContext.default).withAddedAvailableGroups(UniqueNonEmptyList.of(groupFrom("g1,g2"), groupFrom("g3")))
```
coutoPL: "g1,g2" as a name of the group? this is tricky and maybe we should pass here "g1", "g2", "g3".

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/variables/RuntimeResolvableVariablesTests.scala
```diff
@@ -187,6 +187,37 @@ class RuntimeResolvableVariablesTests extends WordSpec with MockFactory {
...
+          )
+        variable shouldBe Right(NonEmptyList.of("g1,g2,g3"))
```
coutoPL: and here we should assert:
`variable shouldBe Right(NonEmptyList.of("g1", "g2", "g3"))`

coutoPL:
> #565

Thanks for the implementation. I think we could improve it a little bit. Could you take a look at my comments, please?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
kiblik: Yes, you right. I didn't know there is the implementation for lists. I suppose it has to be `csv`. I will check it.

coutoPL: you know, `MultiExtractable` trait was created to explode single value into list of values. Here you have a list, so in implementation of `MultiExtractable` all you need to do is just convert the list of groups to list of strings. In contrast to `SingleExtractable` where you had to merge the list to one string.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/variables/RuntimeResolvableVariablesTests.scala
kiblik: you right, I changed it.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/variables/RuntimeResolvableVariablesTests.scala
kiblik: you right, I changed it.
I was inspired by `acl:current_group` tests

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableVariable.scala
kiblik: Ok, I changed it (I moved evaluation to `MultiExtractable`) but I cannot remove `AvailableGroupsVar` from `SingleExtractable` because of it depends on https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/566/files#diff-7bfb77d84aae07205a914b05b3b3886aR174 and it depends on https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/566/files#diff-7bfb77d84aae07205a914b05b3b3886aR163. So I created exception there.

kiblik:
@coutoPL, based on your comments I updated my implementation. Please check it.

kiblik:
ping @coutoPL

PR #572, opened 2020-03-13
[RORDEV-227] multiple values for authorization header

coutoPL:
🐞Fix (ES) support for multiple header values
🧐Enhancement (ES/KBN) handling of encoded ROR headers in Authorization header values

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/request/EsRequestContext.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/request/EsRequestContext.scala
@@ -18,11 +18,11 @@ package tech.beshu.ror.accesscontrol.request
 
 import java.time.Instant
 
+import cats.data.NonEmptyList
 import com.softwaremill.sttp.Method
 import eu.timepit.refined.types.string.NonEmptyString
 import squants.information.{Bytes, Information}
 import tech.beshu.ror.accesscontrol.domain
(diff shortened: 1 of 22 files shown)


core/src/main/scala/tech/beshu/ror/accesscontrol/request/EsRequestContext.scala
```diff
@@ -99,28 +109,28 @@ class EsRequestContext private (rInfo: RequestInfoShim) extends RequestContext {
...
-  override val isCompositeRequest: Boolean =
+  override lazy val isCompositeRequest: Boolean =
     rInfo.extractIsCompositeRequest
 
-  override val isAllowedForDLS: Boolean =
+  override lazy val isAllowedForDLS: Boolean =
     rInfo.extractIsAllowedForDLS
 
-  override val hasRemoteClusters: Boolean =
+  override lazy val hasRemoteClusters: Boolean =
```
sscarduzio: I love all these lazy vals, it could make a nice performance boost, right?

coutoPL: potentially yes, but I'd not expect big gain

sscarduzio: Mainly I expect it would speed up when you only use very few rules in the whole ACL

PR #575, opened 2020-03-19
[RORDEV-207] [ES] env var working for SSL settings

clutroth:
🧐Enhancement (ES) Interpolating config with environment variables.

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
+++ b/core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
@@ -50,7 +50,7 @@ final class EsConfigFileLoader[CONFIG: Decoder]()(implicit envVarsProvider: EnvV
 
   private implicit val showResolvingError: Show[NonEmptyList[ResolvingError]] =
     Show.show { nel =>
-      nel.map(_.msg).mkString_(", ")
+      nel.map(_.msg).mkString_("\n\t* ")
     }
 }
 
--- a/core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
+++ b/core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
@@ -17,13 +17,19 @@
 package tech.beshu.ror.configuration
 
 import better.files.File
+import cats.Show
+import cats.data.NonEmptyList
+import cats.implicits._
 import io.circe.Decoder
+import tech.beshu.ror.accesscontrol.factory.JsonConfigStaticVariableResolver
+import tech.beshu.ror.accesscontrol.factory.JsonConfigStaticVariableResolver.ResolvingError
+import tech.beshu.ror.providers.EnvVarsProvider
 import tech.beshu.ror.utils.yaml
 
-trait LoadFromEsConfig {
+final class EsConfigFileLoader[CONFIG: Decoder]()(implicit envVarsProvider: EnvVarsProvider) {
 
-  protected def loadConfigFromFile[CONFIG: Decoder](file: File,
-                                                    configName: String): Either[MalformedSettings, CONFIG] = {
+  def loadConfigFromFile(file: File,
+                         configName: String): Either[MalformedSettings, CONFIG] = {
     file.fileReader { reader =>
       yaml
         .parser
@@ -31,12 +37,21 @@ trait LoadFromEsConfig {
         .left.map(e => MalformedSettings(s"Cannot parse file ${file.pathAsString} content. Cause: ${e.message}"))
         .right
         .flatMap { json =>
+          JsonConfigStaticVariableResolver.resolve(json)
+            .left.map(e => MalformedSettings(show"""Invalid $configName configuration. $e."""))
+        }
+        .flatMap { json =>
           implicitly[Decoder[CONFIG]]
             .decodeJson(json)
-            .left.map(e => MalformedSettings(s"Invalid $configName configuration"))
+            .left.map(_ => MalformedSettings(s"Invalid $configName configuration"))
         }
     }
   }
+
+  private implicit val showResolvingError: Show[NonEmptyList[ResolvingError]] =
+    Show.show { nel =>
+      nel.map(_.msg).mkString_(", ")
+    }
 }
 
 final case class MalformedSettings(message: String)
--- a/core/src/main/scala/tech/beshu/ror/configuration/SslConfiguration.scala
+++ b/core/src/main/scala/tech/beshu/ror/configuration/SslConfiguration.scala
@@ -24,18 +24,20 @@ import io.circe.{Decoder, DecodingFailure, HCursor}
 import monix.eval.Task
 import org.apache.logging.log4j.scala.Logging
 import tech.beshu.ror.configuration.SslConfiguration.{ExternalSslConfiguration, InternodeSslConfiguration, KeystoreFile, TruststoreFile}
+import tech.beshu.ror.providers.{EnvVarsProvider, OsEnvVarsProvider}
 
 import scala.language.implicitConversions
 
 final case class RorSsl(externalSsl: Option[ExternalSslConfiguration],
                         interNodeSsl: Option[InternodeSslConfiguration])
 
-object RorSsl extends LoadFromEsConfig with Logging {
+object RorSsl extends Logging {
 
   val noSsl = RorSsl(None, None)
 
   def load(esConfigFolderPath: Path): Task[Either[MalformedSettings, RorSsl]] = Task {
     implicit val sslDecoder: Decoder[RorSsl] = SslDecoders.rorSslDecoder(esConfigFolderPath)
+    implicit val envVarsProvider:EnvVarsProvider = OsEnvVarsProvider
     val esConfig = File(new JFile(esConfigFolderPath.toFile, "elasticsearch.yml").toPath)
     loadSslConfigFromFile(esConfig)
       .fold(
@@ -48,7 +50,8 @@ object RorSsl extends LoadFromEsConfig with Logging {
   }
 
   private def fallbackToRorConfig(esConfigFolderPath: Path)
-                                 (implicit rorSslDecoder: Decoder[RorSsl]) = {
+                                 (implicit rorSslDecoder: Decoder[RorSsl],
+                                  envVarsProvider: EnvVarsProvider) = {
     val rorConfig = FileConfigLoader.create(esConfigFolderPath).rawConfigFile
     logger.info(s"Cannot find SSL configuration is elasticsearch.yml, trying: ${rorConfig.pathAsString}")
     if (rorConfig.exists) {
@@ -59,8 +62,9 @@ object RorSsl extends LoadFromEsConfig with Logging {
   }
 
   private def loadSslConfigFromFile(config: File)
-                                   (implicit rorSslDecoder: Decoder[RorSsl]) = {
-    loadConfigFromFile[RorSsl](config, "ROR SSL")
+                                   (implicit rorSslDecoder: Decoder[RorSsl],
+                                    envVarsProvider: EnvVarsProvider) = {
+    new EsConfigFileLoader[RorSsl]().loadConfigFromFile(config, "ROR SSL")
   }
 }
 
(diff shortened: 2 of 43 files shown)


core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
```diff
@@ -17,26 +17,41 @@
...
             .decodeJson(json)
-            .left.map(e => MalformedSettings(s"Invalid $configName configuration"))
+            .left.map(_ => MalformedSettings(s"Invalid $configName configuration"))
         }
     }
   }
+
+  private implicit val showResolvingError: Show[NonEmptyList[ResolvingError]] =
+    Show.show { nel =>
+      nel.map(_.msg).mkString_(", ")
```
coutoPL: message in ResolvingError is human readable sentence, so concatenating such sentences with `,` could end up with weird thing. Use multiline string with eg. bullet points

core/src/main/scala/tech/beshu/ror/configuration/SslConfiguration.scala
```diff
@@ -24,18 +24,20 @@ import io.circe.{Decoder, DecodingFailure, HCursor}
...
                         interNodeSsl: Option[InternodeSslConfiguration])
 
-object RorSsl extends LoadFromEsConfig with Logging {
+object RorSsl extends Logging {
 
   val noSsl = RorSsl(None, None)
 
   def load(esConfigFolderPath: Path): Task[Either[MalformedSettings, RorSsl]] = Task {
     implicit val sslDecoder: Decoder[RorSsl] = SslDecoders.rorSslDecoder(esConfigFolderPath)
+    implicit val envVarsProvider:EnvVarsProvider = OsEnvVarsProvider
```
coutoPL: IMO we should pass this envprovider as implicit. The implementation of  EnvVarsProvider should be fixed as close to app boundary as possible.

core/src/test/scala/tech/beshu/ror/unit/configuration/EsConfigFileLoaderTest.scala
```diff
@@ -0,0 +1,58 @@
...
+      result.right.get shouldBe "John"
+    }
+    "fail for non existing vairable" in {
+      val result = loadFromTempFile[String](""""${WRONG_VARIABLE}"""")
+      result.left.get shouldBe a[MalformedSettings]
+      result.left.get.message should include("WRONG_VARIABLE")
+    }
+  }
+
+  private def loadFromTempFile[A: Decoder](content: String) = tempFile(content).map { file =>
```
coutoPL: formatting

core/src/test/scala/tech/beshu/ror/unit/configuration/EsConfigFileLoaderTest.scala
```diff
@@ -0,0 +1,58 @@
...
+  private implicit val envVarsProvider: EnvVarsProvider = name =>
+    name.value.value match {
+      case "USER_NAME" => Some("John")
+      case _ => None
+    }
+
+  "FileConfigLoader" should {
+    "decode file file" in {
+      val result = loadFromTempFile[String](""""encoded value"""")
+      result.right.get shouldBe "encoded value"
```
coutoPL: IMO it should be checked as follows:
`result shouldBe Right("encoded value")`

error message produced by scalatest will be better

core/src/test/scala/tech/beshu/ror/unit/configuration/EsConfigFileLoaderTest.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import better.files.File
+import io.circe.Decoder
+import org.scalatest.Matchers._
+import org.scalatest.WordSpec
+import tech.beshu.ror.configuration.{EsConfigFileLoader, MalformedSettings}
+import tech.beshu.ror.providers.EnvVarsProvider
+
+class EsConfigFileLoaderTest
+  extends WordSpec {
```
coutoPL: \n

core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
clutroth: Do you mean to use `"\n\n* "` instead of `", "`.

coutoPL: seems that `\n* ` will be enough

clutroth: I mean `"\n\t* "` :)

coutoPL: 👍

clutroth:
I want https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/576 to fail, to ensure test correctness, but travis has maintenance break now.

clutroth:
> I want #576 to fail, to ensure test correctness, but travis has maintenance break now.

failed

integration-tests-scala/src/test/scala/tech/beshu/ror/integration/plugin/ClusterStateWithInternodeSslTests.scala
```diff
@@ -18,22 +18,31 @@ package tech.beshu.ror.integration.plugin
...
 
+  override lazy val container: ReadonlyRestEsClusterContainer = {
+    ReadonlyRestEsCluster.createLocalClusterContainer(
+      name = "ROR1",
+      rorConfigFileName = "/cluster_state_internode_ssl/readonlyrest.yml",
+      clusterSettings = AdditionalClusterSettings(
+        numberOfInstances = 3,
+        internodeSslEnabled = true
+      )
+    ).mapContainer { c =>
```
coutoPL: I'd rather add new field to `AdditionalClusterSettings`, `envs: Map[String, String]` - we don't need to expose raw container, because the java api is not so clean and simply it looks poor.

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
```diff
@@ -92,14 +93,13 @@ final case class LocalClusterDef(name: String, rorConfigFileName: String, nodeDa
 final case class DependencyDef(name: String, containerCreator: Coeval[GenericContainer])
 
 class ReadonlyRestEsClusterContainer private[containers](rorClusterContainers: NonEmptyList[Task[ReadonlyRestEsContainer]],
-                                                         dependencies: List[DependencyDef],
-                                                         clusterInitializer: ReadonlyRestEsClusterInitializer)
+                                                         dependencies: List[DependencyDef])
   extends Container {
+//  def setEnvs(envs:Map[String,String])
```
coutoPL: commented code (to remove)

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
```diff
@@ -185,15 +184,13 @@ class ReadonlyRestEsRemoteClustersContainer private[containers](val localCluster
...
-
-trait ReadonlyRestEsClusterInitializer {
-  def initialize(adminClient: RestClient, container: ReadonlyRestEsClusterContainer): Unit
-}
-
-object NoOpReadonlyRestEsClusterInitializer extends ReadonlyRestEsClusterInitializer {
-  override def initialize(adminClient: RestClient, container: ReadonlyRestEsClusterContainer): Unit = ()
+final case class RorNodeSpecification(envs:Map[String,String])
+object RorNodeSpecification {
+  implicit val monoidRorNodeSpecification: Monoid[RorNodeSpecification] = {
```
coutoPL: sorry, it doesn't make sense for me. `Monoid` should be used when there is a structure which can be composed. I don't see where we'd like to use composition of `RorNodeSpecification`. So, we end up with usage monoid `empty`, additional dependency and far-fetched usage of Monoid concept. Please, get rid of it.

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
```diff
@@ -185,15 +184,13 @@ class ReadonlyRestEsRemoteClustersContainer private[containers](val localCluster
...
 
 }
-
-trait ReadonlyRestEsClusterInitializer {
-  def initialize(adminClient: RestClient, container: ReadonlyRestEsClusterContainer): Unit
-}
-
-object NoOpReadonlyRestEsClusterInitializer extends ReadonlyRestEsClusterInitializer {
-  override def initialize(adminClient: RestClient, container: ReadonlyRestEsClusterContainer): Unit = ()
+final case class RorNodeSpecification(envs:Map[String,String])
```
coutoPL: what about these ones?
```
xPackSupport: Boolean = false,	                                             
configHotReloadingEnabled: Boolean = true,
customRorIndexName: Option[String] = None,
internodeSslEnabled: Boolean = false
```
at least the first two can also be understood as `RorNodeSpecification` fields

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
clutroth: When using `Map` it's not monoid, but with `List` it will become monoid. It is composable structure. For now it supports only environment variables, but it can support more. Alternative was adding node envs list to AdditionalClusterSettings, not encapsulating this specific, low level setting.

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
clutroth: I don't see node specification here. I understand node specification, as low level container specification, not business configuration like those.

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
coutoPL: But I don't see usage of RorNodeSpecification as monoid. Now and in future. Please, remove this unnecessary complication.

tests-utils/src/main/scala/tech.beshu.ror.utils/containers/ReadonlyRestEsClusterContainer.scala
coutoPL: so the name is misleading and should reflect what it actually is

clutroth: I'm not sure if I got you. IMHO setting env variable in my test case shouldn't leak to ror image configuration, but this configuration should be extendable. Additionally extendable configuration should be separated form ROR higher level configs.

coutoPL: we could consider several level of settings:
* host/container settings
* jvm settings
* ES (+ RoR) node settings 

`RorNodeSpecification` could suggest that it's related strictly to ROR (no it isn't). It may also suggest that maybe it's a specification of ES node. That's why it's misleading for me.

Moreover, I noticed that in one (or maybe more - to verify) test ENV var is used. It's set like this: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/master/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/generic/EsImage.scala#L44 Maybe we should remove it and use new approach?

clutroth: Yes, this API can be used. There's so much layers for building. :/ I agree with this settings levels. I think problem was expressing it in `RorNodeSpecification`. I may propose `ContainerSpecification`.

coutoPL: Yes, we should (and we will) flat them, but after @pondzix big tests refactoring. 

Please, remove `.env("TEST_VAR", "dev")`. This is used only `DynamicVariablesTests`, so in java ones (@pondzix remember to add it in ContainerSpecification env in rewritten, scala equivalent of `DynamicVariablesTests`)

PR #589, opened 2020-05-01
bump circe

clutroth:


sscarduzio:
does this fix RODEV-244?

clutroth:
I don't think so. I have used newer circe in RORDEV196, so I decided to resolve bump independently. After all there was no issue, but our tests depends on specific behavior of circe 0.11. It  shouldn't have effect on plugin functionality.

coutoPL:
why separate PR for this?

clutroth:
Because circe upgrade failed tests, and I wanted to resolve it in isolation.

coutoPL:
ok, please close when it's not needed any more

PR #590, opened 2020-05-02
[Rordev-228] Filter rule modifications

pondzix:
**Enhancement** (ES) simplified filter rule, support for all kinds of queries


es74x/src/main/scala/tech/beshu/ror/es/request/context/types/SearchEsRequestContext.scala
```diff
@@ -40,14 +41,15 @@ class SearchEsRequestContext(actionRequest: SearchRequest,
...
+    SearchQueryDecorator.applyFilterToQuery(request, threadPool)
     request.indices(indices.toList.map(_.value.value): _*)
     Modified
   }
 
   // Cache disabling for this request is crucial for document level security to work.
   // Otherwise we'd get an answer from the cache some times and would not be filtered
   private def optionallyDisableCaching(): Unit = {
-    if(esContext.involveFilters) {
+    if (esContext.involveFilters) {
```
sscarduzio, line 52: The whole objetive here is to leave this cache enabled. In order to do so, we need to add the filter to the search request from the acl rule.

core/src/main/scala/tech/beshu/ror/accesscontrol/utils/FilterUtils.scala
```diff
@@ -0,0 +1,35 @@
...
+import tech.beshu.ror.Constants
+import tech.beshu.ror.utils.FilterTransient
+
+object FilterUtils {
+
+  def filterFromHeaderValue(value: String) = {
+    Option(FilterTransient.deserialize(value)) match {
+      case Some(ft) =>
+        val filter = ft.getFilter
+        if (!Strings.isNullOrEmpty(filter)) filter
```
sscarduzio, line 29: No serialisation/deserialisation should be necessary if we apply the filter to the search request in ACL space.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/FilterRule.scala
```diff
@@ -44,7 +43,15 @@ class FilterRule(val settings: Settings)
         case Left(_: Unresolvable) =>
           Rejected()
         case Right(filter) =>
-          Fulfilled(blockContext.withAddedContextHeader(Header(Name.transientFilter, filter)))
+          BlockContextUpdater[B] match {
```
coutoPL, line 46: IMO we should firstly check the block context type and then process only the ones which makes sense in context of the rule

coutoPL: we can also refactor `isAllowedForDLS` - it's rather related to search block context that request

coutoPL: please write a test where `isAllowedForDLS` returns false (current impl is a bit weird for me)

es51x/src/main/scala/tech/beshu/ror/es/request/SearchQueryDecorator.scala
```diff
@@ -0,0 +1,48 @@
...
+ */
+package tech.beshu.ror.es.request
+
+import org.apache.logging.log4j.scala.Logging
+import org.elasticsearch.action.search.SearchRequest
+import org.elasticsearch.index.query.{QueryBuilder, QueryBuilders}
+import tech.beshu.ror.Constants
+import tech.beshu.ror.accesscontrol.domain.Filter
+
+object SearchQueryDecorator extends Logging {
```
coutoPL, line 25: this name suggest that this is decorator pattern, but it's not, so it could be misleading. Mayber we can rename it (and refactor a bit) to SearchRequestOps class?

es51x/src/main/scala/tech/beshu/ror/es/request/SearchQueryDecorator.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def applyFilterToQuery(request: SearchRequest, filter: Option[Filter]) = {
```
coutoPL, line 27: even if method mutates the request, it's nice to return the mutated instance, to allow to use it like fluent interface

es51x/src/main/scala/tech/beshu/ror/es/request/context/types/MultiSearchEsRequestContext.scala
```diff
@@ -57,6 +59,7 @@ class MultiSearchEsRequestContext(actionRequest: MultiSearchRequest,
         .zip(modifiedPacksOfIndices)
         .foreach { case (request, pack) =>
           updateRequest(request, pack)
+          SearchQueryDecorator.applyFilterToQuery(request, blockContext.filter)
```
coutoPL, line 62: it should be moved to `updateRequest` method internals

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FilterRuleSuite.scala
```diff
@@ -83,8 +100,9 @@ object FilterRuleSuite {
 
   private def add3Docs(adminRestClient: RestClient, index: String, `type`: String): Unit = {
     val documentManager = new DocumentManagerJ(adminRestClient)
-    documentManager.insertDocAndWaitForRefresh(s"/$index/${`type`}/1", s"""{"db_name":"db_user1"}""")
-    documentManager.insertDocAndWaitForRefresh(s"/$index/${`type`}/2", s"""{"db_name":"db_user2"}""")
-    documentManager.insertDocAndWaitForRefresh(s"/$index/${`type`}/3", s"""{"db_name":"db_user3"}""")
+    documentManager.insertDocAndWaitForRefresh(s"/$index/${`type`}/1", s"""{"db_name":"db_user1", "code": 1}""")
```
coutoPL, line 103: please use DocumentManager instead, so you can get rid of esVersion check in method above

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FilterRuleSuite.scala
```diff
@@ -46,25 +43,45 @@ trait FilterRuleSuite
...
-
   "A filter rule" should {
     "show only doc according to defined filter" when {
-      "search api is used" in {
-        retry(times = 3) {
-          val result = searchManager.search("/test1_index/_search")
-          assertEquals(200, result.getResponseCode)
-          result.getSearchHits.size() should be(1)
-          result.getSearchHits.get(0).get("_source").asInstanceOf[JMap[String, String]].get("db_name") should be("db_user1")
+      "search api is used" when {
```
coutoPL, line 48: we need a test to show what happens when filter query is malformed

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/FilterRule.scala
```diff
@@ -44,14 +43,25 @@ class FilterRule(val settings: Settings)
         case Left(_: Unresolvable) =>
           Rejected()
         case Right(filter) =>
-          Fulfilled(blockContext.withAddedContextHeader(Header(Name.transientFilter, filter)))
+          blockContext match {
+            case bc: SearchRequestBlockContext => Fulfilled(addFilter(bc: SearchRequestBlockContext, filter).asInstanceOf[B])
```
coutoPL, line 47: this is why we should pattern match by `BlockContextUpdater[B]` instead (to avoid asInstanceOf and take advantage of exhausted pattern matching). But we can improve it in next (fields) change

PR #592, opened 2020-05-05
RORDEV-251 Adding unrestricted kibana access

sscarduzio:
Enhancement: kibana_access can be explicity set to unrestricted

### DIFF
--- a/core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/KibanaAccessRuleTests.scala
+++ b/core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/KibanaAccessRuleTests.scala
@@ -26,7 +26,7 @@ import tech.beshu.ror.accesscontrol.blocks.metadata.UserMetadata
 import tech.beshu.ror.accesscontrol.blocks.rules.KibanaAccessRule
 import tech.beshu.ror.accesscontrol.blocks.rules.Rule.RuleResult.{Fulfilled, Rejected}
 import tech.beshu.ror.accesscontrol.blocks.variables.runtime.RuntimeSingleResolvableVariable.AlreadyResolved
-import tech.beshu.ror.accesscontrol.domain.KibanaAccess.{RO, ROStrict, RW}
+import tech.beshu.ror.accesscontrol.domain.KibanaAccess.{RO, ROStrict, RW, Unrestricted}
 import tech.beshu.ror.accesscontrol.domain._
 import tech.beshu.ror.mocks.MockRequestContext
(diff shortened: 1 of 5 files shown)


core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/KibanaAccessRuleTests.scala
```diff
@@ -37,7 +37,25 @@ import scala.language.postfixOps
 
 class KibanaAccessRuleTests extends WordSpec with Inside with BlockContextAssertion {
 
+  val UNRESTRICTED_ACTIONS: Set[String] = Set(
```
coutoPL, line 40: In context of current implementation in KibanaAccessRule I don't get it - if:
1) implementation is correct - any action can be passed through the rule because there is no check one, so UNRESTRICTED_ACTIONS here makes no sense. And do we really need this mode? user just can remove the rule and behaviour of the block will be the same.
2) implementation is not correct - should we check if current action matches one of these actions if "unrestricted" mode is configured. Then these UNRESTRICTED_ACTIONS should be moved to `tech.beshu.ror.Constants`

Could you please check it or explain what we want to achieve?

sscarduzio: "why do we need tthis mode? can't they just remove the rule?" Objection accepted, in strict logical sense. 

But when it's time to document the feature, it's a bit harder for people to wrap their head around a set of possible values that is semantically incomplete. Or is completed by the omission of the rule itself.

Have you seen how common, after 3 years of supporting new customers, when people still try to make things work writingt things like:  `indices ["*"]` ? This is no different.

sscarduzio: Well, the UNRESTRICTED_ACTIONS in the unit test just contains a few random actions that are not supported by other actions sets. Does not intend to be a complete or representative set. I wasn't sure how to do a better job at it.

coutoPL: ok, let it be. So, let's modify the test. Because at the moment, looking at the test, you have an impression, that only the listed set of actions can be "unrestricted". Maybe we can sum all the sets of actions (the ones from Constant class), name the val "anyActionsSet". IMO this would avoid such misunderstanding like mine.

coutoPL: ok, please fill in description of the task in changelog style and we can merge

PR #605, opened 2020-06-18
[RORDEV-282] es 7.8.0 support

clutroth:
🚀New 7.8.0 support


coutoPL:
@clutroth  which part of ES API forced us to create new es module?

coutoPL:
ok, I've checked it. LGTM

PR #619, opened 2020-08-09
[Rordev-252] Variable support in fields rule

pondzix:
🚀New (ES) dynamic variable support in fields rule


sscarduzio:
can we fit this in 1.21.0?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/FieldsRule.scala
```diff
@@ -16,43 +16,46 @@
...
 
   override def check[B <: BlockContext : BlockContextUpdater](blockContext: B): Task[RuleResult[B]] = Task {
-    if (!blockContext.requestContext.isReadOnlyRequest) RuleResult.Rejected()
-    else RuleResult.Fulfilled(blockContext.withAddedContextHeader(transientFieldsHeader))
+    if (!blockContext.requestContext.isReadOnlyRequest)
+      RuleResult.Rejected()
+    else {
+      val maybeResolvedFields = resolveAll(settings.fields.toNonEmptyList, blockContext).toSet
+      maybeResolvedFields.toNonEmptySet match {
+        case Some(resolvedFields) if !DocumentField.areDifferentAccessModesUsedSimultaneously(resolvedFields.toList) =>
```
coutoPL, line 44: IMO it'd be better to recognise if two modes are mixed during parsing configuration at startup level, not at request handling level. `~@{user}` is negated field for every string resolved from user variables,  `@{user}` is regular field for every string resolved from user variables (also eg. `~john`)

coutoPL:
> can we fit this in 1.21.0?

It depends if we want to wait for this one. At the moment the above should be fix and one pipieline step fails.

PR #624, opened 2020-08-18
[RORDEV-315] bugfix: XFF in audit

coutoPL:
🐞Fix (ES) XFF entry in audit issue


audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializer.scala
```diff
@@ -68,7 +68,7 @@ class DefaultAuditLogSerializer extends AuditLogSerializer {
       .put("type", requestContext.`type`)
       .put("origin", requestContext.remoteAddress)
       .put("destination", requestContext.localAddress)
-      .put("xff", requestContext.headers.get("X-Forwarded-For").orNull)
+      .put("xff", requestContext.headers.find { case (key, _) => key.toLowerCase == "x-forwarded-for" }.map(_._2).orNull)
```
sscarduzio, line 71: why fixing the log serializer and not its request context? Now if someone has a custom serializer, they will still have case sensitive header map.  Which is not how HTTP should work.

coutoPL: this has nothing to http behaviour or standard. ROR always loads and stores headers in origin format. Then inside ROR code we use case insensitive `Eq` typeclass to compare header names. The bug was exactly in this line - we should also use it here. But we `AuditRequestContext` is just an interface with plain scala/java classes because audit serializers should be simple to implement by external users. That's why headers is just `Map[String, String]` as it was in old code. Backward compatibility is very important, so we should be careful and doesn't break it. I though about lowercasing all keys in this map but it does break the compatibility. I've introduced `requestHeaders: Headers` fields to allow user to get header value by key with case insensitive way.

PR #632, opened 2020-09-03
ES 7.9.1 support

clutroth:
🚀 added 7.9.0 support


muhviehstah:
maven did'nt find the classloader for 7.9.1

clutroth:
> maven did'nt find the classloader for 7.9.1

You're right. I fixed it.

PR #634, opened 2020-09-07
[RORDEV-314] Fixed config file permission issue with JDK 1.8 25.262-b10

pielas:
🐞Fix (ES) Config file permission issue with JDK 1.8.0 25.262-b10

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/configuration/EsConfig.scala
+++ b/core/src/main/scala/tech/beshu/ror/configuration/EsConfig.scala
@@ -16,17 +16,14 @@
  */
 package tech.beshu.ror.configuration
 
-
 import java.nio.file.Path
 
-import better.files.File
 import cats.data.EitherT
 import io.circe.Decoder
 import monix.eval.Task
 import tech.beshu.ror.configuration.EsConfig.LoadEsConfigError.{FileNotFound, MalformedContent}
 import tech.beshu.ror.configuration.EsConfig.RorEsLevelSettings
 import tech.beshu.ror.providers.EnvVarsProvider
-import tech.beshu.ror.utils.AccessControllerHelper.doPrivileged
 import tech.beshu.ror.utils.PrivilegedFile
 import tech.beshu.ror.utils.yaml.JsonFile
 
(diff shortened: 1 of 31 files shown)


es66x/build.gradle
```diff
@@ -53,7 +53,7 @@ compileScala {
 dependencies {
     compile project(path: ':core')
     compile             group: 'org.elasticsearch',                 name: 'elasticsearch' ,             version: esVersion
-//    compileClasspath    group: 'org.codelibs.elasticsearch.lib',    name: 'plugin-classloader',         version: esVersion // no 6.8.11&12 in mvn atm
```
coutoPL, line 56: I think you cannot uncomment it. That's why pipeline failed. There are no these two version published. This dependency is only needed to run the project with Idea.

core/src/main/scala/tech/beshu/ror/configuration/EsConfig.scala
```diff
@@ -38,7 +39,7 @@ object EsConfig {
           (implicit envVarsProvider:EnvVarsProvider): Task[Either[LoadEsConfigError, EsConfig]] = {
     val configFile = File(s"${esConfigFolderPath.toAbsolutePath}/elasticsearch.yml")
     (for {
-      _ <- EitherT.fromEither[Task](Either.cond(configFile.exists, (), FileNotFound(configFile)))
+      _ <- EitherT.fromEither[Task](Either.cond(doPrivileged(configFile.exists), (), FileNotFound(configFile)))
```
coutoPL, line 42: I'm little bit concerned that we should add `doPrivileged` in so many places. What about creating `File` adapter (I wanted to write decorator, but there is no sense to expose all methods from `better.file.File` class) and encapsulate `doPrivileged` in it?

eshome/plugins/readonlyrest/plugin-security.policy
```diff
@@ -9,4 +9,6 @@ grant {
   permission java.io.FilePermission "/proc/sys/net/core/somaxconn","read";
   permission java.lang.RuntimePermission "accessClassInPackage.sun.misc";
   permission java.lang.RuntimePermission "setContextClassLoader";
+  permission java.lang.RuntimePermission "createSecurityManager";
```
coutoPL, line 12: why this privilege is needed?

coutoPL (changes requested):

es66x/build.gradle
pielas: aah ok, I it didn't work in idea without that and I thought that they are available.

eshome/plugins/readonlyrest/plugin-security.policy
pielas: It turned out that these two permissions are needed when I removed using `sec-all.policy` which has `permission java.security.AllPermission;`. All of permissions from `plugin-security.policy` are not needed when `sec-all.policy` is used

coutoPL (approved):
LGTM

PR #639, opened 2020-09-17
[RORDEV-252] field rule improvements

pondzix:
🚀New (ES) Introduced configuration property defining FLS engine used by fields rule 
🧐Enhancement (ES) Fields rule performance improvement


integration-tests/src/test/scala/tech/beshu/ror/integration/plugin/FieldRuleMSearchApiPluginTests.scala
```diff
@@ -0,0 +1,22 @@
...
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.integration.plugin
+
+import tech.beshu.ror.integration.suites.fields.FieldRuleMSearchApiSuite
+import tech.beshu.ror.integration.utils.SingletonPluginTestSupport
+
+class FieldRuleMSearchApiPluginTests extends FieldRuleMSearchApiSuite with SingletonPluginTestSupport
```
coutoPL, line 22: I'd rather have one suite with all FLS tests. New test approach assumes that integration tests will reflect ES API (eg. snapshots api, index api) or concepts (like FLS, DLS).

integration-tests/src/test/scala/tech/beshu/ror/integration/proxy/FieldLevelSecurityProxyTests.scala
```diff
@@ -16,7 +16,6 @@
  */
 package tech.beshu.ror.integration.proxy
 
-import tech.beshu.ror.integration.suites.FieldLevelSecuritySuite
 
 // todo: enable when proxy supports fields rule
```
coutoPL, line 20: ATM proxy won't support Lucene level fields filtering so it'd be nice to have several modes for ROR FLS:
* old mode (FLS at Lucene level)
* new mode (QUERY modifications + fallback to old mode)
* proxy mode (QUERY modification + forbid)

Are we able to do it in a pretty simple way, or we should create a separate JIRA for proxy?

coutoPL (changes requested):

integration-tests/src/test/scala/tech/beshu/ror/integration/proxy/FieldLevelSecurityProxyTests.scala
coutoPL: let's create a jira.

coutoPL:
and one reminder:
```
 We should print a warning on startup if fields rule is used though, saying that all nodes should be installed for it to work well.
```

core/src/main/scala/tech/beshu/ror/accesscontrol/ops.scala
```diff
@@ -152,7 +153,6 @@ object show {
           showTraversable("indices", bc.indices) ::
           showOption("kibana_idx", bc.userMetadata.kibanaIndex) ::
           showTraversable("response_hdr", bc.responseHeaders) ::
-          showTraversable("context_hdr", bc.contextHeaders) ::
```
coutoPL, line 155: we don't want to show context headers any more (because there is no such things any more), but this log is helpful for support to figure out what was resolved in the matched block - it seems that we need to place here all information which will be helpful to figure out what was collected by the fields rule.

es78x/src/main/scala/tech/beshu/ror/es/request/context/types/SqlIndicesEsRequestContext.scala
```diff
@@ -47,21 +51,48 @@ class SqlIndicesEsRequestContext private(actionRequest: ActionRequest with Compo
   override protected def indicesFrom(request: ActionRequest with CompositeIndicesRequest): Set[IndexName] =
     sqlIndices.indices.flatMap(IndexName.fromString)
 
+  /** fixme: filter is not applied.
```
coutoPL, line 54: are you able to write a test which fails for this concrete example? we will @ignore it atm, but will be easier to fix it in future

es78x/src/main/scala/tech/beshu/ror/es/request/context/types/XpackAsyncSearchRequest.scala
```diff
@@ -62,14 +68,26 @@ class XpackAsyncSearchRequest private(actionRequest: ActionRequest,
       .getOrElse(throw new RequestSeemsToBeInvalid[ActionRequest]("Cannot extract SearchRequest from SubmitAsyncSearchRequest request"))
   }
 
-  //TODO cache is now disabled only when 'fields' rule is used.
-  //Remove after 'fields' rule improvements.
-  private def optionallyDisableCaching(request: SearchRequest): Unit = {
-    if (esContext.involvesFields) {
-      logger.debug("ACL involves fields, will disable request cache for SearchRequest")
-      request.requestCache(false)
+  private def filterFieldsFromResponse(fieldLevelSecurity: Option[FieldLevelSecurity])
```
coutoPL, line 71: we don't have to return Task

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/ops.scala
```diff
@@ -176,6 +177,16 @@ object show {
         showOption("user_origin", u.userOrigin) ::
         Nil flatten) mkString ";"
     }
+
+    implicit val flsStrategyShow: Show[FieldLevelSecurity] = Show.show[FieldLevelSecurity] { fls =>
+      fls.strategy match {
+        case Strategy.FlsAtLuceneLevelApproach => "fls_at_lucene_level"
```
coutoPL, line 183: why no "[strategy: ...]" here?

es55x/src/main/scala/tech/beshu/ror/es/request/SearchHitOps.scala
```diff
@@ -0,0 +1,69 @@
...
+                                                    fieldsRestrictions: FieldsRestrictions) = {
+      val (metadataFields, nonMetadataDocumentFields) = partitionFieldsByMetadata(fields)
+      val policy = new FieldsPolicy(fieldsRestrictions)
+      val filteredDocumentFields = nonMetadataDocumentFields.filter {
+        case (key, _) => policy.canKeep(key)
+      }
+      metadataFields ++ filteredDocumentFields
+    }
+
+
```
coutoPL, line 59: redundant \n

es55x/src/main/scala/tech/beshu/ror/es/request/SearchRequestOps.scala
```diff
@@ -48,4 +59,54 @@ object SearchRequestOps extends Logging {
...
+    }
+
+    private def modifyNotAllowedFieldsInQuery(notAllowedFields: NonEmptyList[UsedField.SpecificField]) = {
+      val currentQuery = request.source().query()
+      val newQuery = currentQuery.handleNotAllowedFields(notAllowedFields)
+      request.source().query(newQuery)
+      request
+    }
+
+
```
coutoPL, line 100: same here

coutoPL (changes requested):
We're almost there. 

What we need are:
* documentation (from user perspective)
* internal documentation (rough description of current approach, which helps support to solve user issues)
* description of the task

integration-tests/src/test/resources/field_level_security_mode/readonlyrest.yml
```diff
@@ -0,0 +1,32 @@
...
+      verbosity: error
+
+    - name: fls_mode_default
+      auth_key: "user1:pass"
+      fields: ["_source","allowedField"]
+
+    - name: fls_mode_hybrid_explicit
+      auth_key: "user2:pass"
+      fields: ["_source","allowedField"]
+      fields_mode: "hybrid"
```
coutoPL, line 22: we should avoid this approach in new rules or during extending existing rules. Please take a look at `proxy_auth` rule decoder and do it in the same way - by defining simple settings decoder and extended settings decoder.

Extended version can be defined like in the following example:
```
fields: 
  list: ["_source", "allowedField"]
  mode: "hybrid"
```

integration-tests/src/test/resources/field_level_security_mode/readonlyrest.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: fls_mode_legacy_explicit
+      auth_key: "user3:pass"
+      fields: ["_source","allowedField"]
+      fields_mode: "legacy"
+
+    - name: fls_mode_proxy_explicit
+      auth_key: "user4:pass"
+      fields: ["_source","allowedField"]
+      fields_mode: "proxy"
```
coutoPL, line 32: we should not allow user to set this one - this mode should be forced when ROR is run as a proxy

integration-tests/src/test/resources/field_level_security_mode/readonlyrest.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: fls_mode_hybrid_explicit
+      auth_key: "user2:pass"
+      fields: ["_source","allowedField"]
+      fields_mode: "hybrid"
+
+    - name: fls_mode_legacy_explicit
+      auth_key: "user3:pass"
+      fields: ["_source","allowedField"]
+      fields_mode: "legacy"
```
coutoPL, line 27: we should consider mode names changes. I don't know if "legacy" and "hybrid" are good names from perspective of client. From developer perspective they obviously make sense

coutoPL (changes requested):

integration-tests/src/test/resources/field_level_security_mode/readonlyrest.yml
sscarduzio: Please remember we are in he business of removing decision points from the user experience, not adding them. From the user point of view, what would functionally change between hybrid and legacy?

coutoPL: we were not managed to improve `fields` rule as we planned to do at the beginning. So, here we have a hybrid approach. The hybrid approach still requires to install ROR on all nodes. But we give the user an option to pick - the fast `fields` rule with ROR on one node (but for some requests ROR will return `forbidden`) or pretty fast `fields` rule with ROR on all nodes (all requests will be handled). The hybrid will be a default, but the user can pick the second one instead - depending on his requirements.

Obviously, the configuration above has no sense - we've discussed it and agreed that a global setting will be better. You will see it in the next commit

sscarduzio: OK for the default=hybrid, and people will just observe a speed up in a good most commonly used set of filtered requests. Instead I don't feel the case for "legacy" mode, as if somebody requires to set it, there must be a bug in the new hybrid solution. Instead for the "proxy" mode I'd say more pre-routing or pre-shard, as this enhancement is not only for proxy.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/GlobalStaticSettingsDecoder.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private def flsEngineFromString(str: String): Either[UnknownFlsEngine, FlsEngine] = str match {
+    case "lucene" => Right(GlobalSettings.FlsEngine.Lucene)
+    case "es_with_lucene" => Right(GlobalSettings.FlsEngine.ESWithLucene)
+    case "es" => Right(GlobalSettings.FlsEngine.ES)
+    case unknown => Left(UnknownFlsEngine(unknown))
+  }
+
+  final case class UnknownFlsEngine(value: String)
```
coutoPL, line 85: can be private

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -172,6 +173,9 @@ trait ReadonlyRest extends Logging {
               httpClientsFactory = httpClientsFactory,
               ldapConnectionPoolProvider
             )
+            if (engine.context.involvesFields) {
+              logger.warn("ACL contains fields rule. To make it work well, all nodes should have ROR plugin installed.")
```
coutoPL, line 177: WDYT about improving it? this warn makes no sense when fields rule is used in proxy mode or plugin mode with ES FLS engine

coutoPL (changes requested):

integration-tests/src/test/resources/field_level_security_mode/readonlyrest.yml
coutoPL: we have 3 options which will be configured globally, but only 2 of them will be mentioned in doc:
1. legacy mode (works as before) - when there will be a bug and the customer will complain about it, we can instruct him to enable this mode (it will be removed in future, when the solution warms up).
2. hybrid mode (default = es level fields handling + lucene level fields handling) - admin have to install ROR on each ES node + all requests will be handled
3. es level fields handling only (all requests which cannot be handled like this will be rejected) - admin can install ROR only on one node

So, by default you will have 2nd mode enabled - a fast and complete solution. If you don't want to install ROR on each node and you accept limitations of 3rd mode, you can set it up.

coutoPL (approved):
Great! Waiting for the description of the PR and we can merge. Thanks

PR #645, opened 2020-09-30
Removed unnecessary SecureStringHasher class

pielas:
🐞Fix (ES) issue when entropy of /dev/random could have been exhausted when using JwtToken rule


sscarduzio:
can you explain this?

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/JwtAuthRuleTests.scala
```diff
@@ -485,4 +524,19 @@ class JwtAuthRuleTests
       .returning(Task.now(authenticated))
     service
   }
+
+  private def cachedAuthService(authenticatedToken: String, unauthenticatedToken: String) = {
+    val service = mock[ExternalAuthenticationService]
+    (service.authenticate _)
+      .expects(where { credentials: Credentials => credentials.secret === PlainTextSecret(authenticatedToken.nonempty) })
```
coutoPL, line 531: I'd add here the constraint that it will be able to mock both tokens only once. Because if cache is broken, the test should fail

coutoPL (changes requested):

coutoPL:
and please fill the description of the PR (remember that it will be used here: https://readonlyrest.com/download/, so it should make sense from ROR's customer perspective)

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/JwtAuthRuleTests.scala
pielas: scalamock has such constraint by default. I even checked it out by breaking cache and test indeed failed. Although I added `.once()` to each mock to make it explicit.

PR #657, opened 2020-10-31
[RORDEV-364] [ES] ROR is not adding the "username" field in the case the user is authenticated via proxy_auth

clutroth:
🐞Fix (ES) "username" in audit entry when user is authenticated via proxy_auth


core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AuditingTool.scala
```diff
@@ -55,21 +55,21 @@ class AuditingTool(settings: Settings,
...
+          requestContext = toAuditRequestContext(allowedBy.requestContext, Some(allowedBy.blockContext), allowedBy.history),
+          verbosity = toAuditVerbosity(allowedBy.block.verbosity),
+          reason = allowedBy.block.show
         )
       case allow: ResponseContext.Allow[B] =>
         AuditResponseContext.Allowed(
-          toAuditRequestContext(allow.requestContext, None, allow.history),
-          toAuditVerbosity(Block.Verbosity.Info),
-          allow.block.show
+          requestContext = toAuditRequestContext(allow.requestContext, allow.history.collectFirst { case History(allow.block.name, _, blockContext) => blockContext }, allow.history),
```
coutoPL, line 64: yes, great catch. But I'm not sure if it's safe to use block context from the matched block. Because  current_user_metadata is processed by ACL a little bit differently - processing is not finished after first matched block, but find all blocks which can be matched. The weird logic of further processing in `AccessControlList#handleMetadataRequest` method. When request is allowed, there is `UserMetadata` object created which is passed to `Allowed` object. So, we should use it instead of BlockContext here.

So, there three methods of `AuditRequestContextBasedOnAclResult` will be a little bit different in this case:

```
  override val loggedInUserName: Option[String] = blockContext.flatMap(_.userMetadata.loggedUser.map(_.id.value.value))
  override val impersonatedByUserName: Option[String] =
    blockContext
      .flatMap(_.userMetadata.loggedUser)
      .flatMap {
        case DirectlyLoggedUser(_) => None
        case ImpersonatedUser(_, impersonatedBy) => Some(impersonatedBy.value.value)
      }
  override val involvesIndices: Boolean = blockContext match {
    case Some(_: GeneralIndexRequestBlockContext | _: TemplateRequestBlockContext | _: SnapshotRequestBlockContext) => true
    case _ => false
  }
```

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AuditingTool.scala
clutroth: I analyzed this flow, and I see, that difference between `block context's user metadata`, and `allowed's user metadata` is here tech.beshu.ror.accesscontrol.acl.AccessControlList#updateUserMetadataGroups. Difference is that, first user metadata is used in `RRMetadataResponse`, that is in rest response. Second user metadata is each, for block context. Serializing audit log event, we're in single rule context,  and it's history, not cumulative user metadata.

TLDR
There are tow user metadatas for two things in one place. That's misleading, but for me I'm using right one user metadata.

coutoPL: yes, correct, but ... here in audit log, we see:
* request data
* history of ACL request processing
* data collected in matched block (block context)

We have also two types of request in ROR:
* regular ES request (travels through the blocks and stops when all rules in block match)
* current user metadata request (travels through all blocks, collects matched blocks and then gathers all user metadata found in matched block in one final metadata [different blocks can provides different metadata parts])

`AuditRequestContextBasedOnAclResult` was designed prior to the introduction of the current user metadata request handling described above (previously it was treated like regular ES request). All the logic how to create final user metadata object from fragmented metadata objects was encapsulated in `AccessControlList#handleMetadataRequest`. That's why I'm sure that here we should use the user metadata taked from `Allowed` object, not collecting first one from history.

coutoPL: And please take a look at this:

```
override val involvesIndices: Boolean = blockContext match {
    case Some(_: GeneralIndexRequestBlockContext | _: TemplateRequestBlockContext | _: SnapshotRequestBlockContext) => true
    case _ => false
  }
```

It looks that it's a little bit outdates (and it's wrong atm). Maybe we should check here if given blockContext has `HasIndices` or `HasIndexPacks` type class instance.

clutroth: Done

core/src/main/scala/tech/beshu/ror/accesscontrol/acl/AccessControlList.scala
```diff
@@ -94,45 +95,41 @@ class AccessControlList(val blocks: NonEmptyList[Block])
...
-          .find { case (group, _) => group == pg }
-          .map { case (currentGroup, Matched(block, blockContext)) =>
-            val allGroups = UniqueList.fromList(allGroupsWithRelatedResults.map(_._1))
-            val userMetadata = createUserMetadata(blockContext, Some(currentGroup), allGroups)
-            (userMetadata, block)
-          }
+          .find {
+            case (`preferredGroup`, _) => true
+            case _ => false
+          }.map(someFirst)
```
coutoPL, line 106: formatting

core/src/main/scala/tech/beshu/ror/accesscontrol/acl/AccessControlList.scala
```diff
@@ -94,45 +95,41 @@ class AccessControlList(val blocks: NonEmptyList[Block])
...
+      val allGroups = UniqueList.fromList(allGroupsWithRelatedResults.map(_._1))
+      val userMetadata = updateUserMetadataGroups(blockContext, maybeGroup, allGroups)
+      (userMetadata, block)
     }
   }
 
-  private def createUserMetadata(blockContext: CurrentUserMetadataRequestBlockContext,
-                                 currentGroup: Option[Group],
-                                 availableGroups: UniqueList[Group]) = {
+  private def flatGroupWittMatchedCurrentMetadata(matchedResults: NonEmptyList[Matched[CurrentUserMetadataRequestBlockContext]]) = {
```
coutoPL, line 119: typo: with

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AccessControlLoggingDecorator.scala
```diff
@@ -77,6 +77,7 @@ class AccessControlLoggingDecorator(val underlying: AccessControl, auditingTool:
         case Success(resultWithHistory) =>
           resultWithHistory.result match {
             case UserMetadataRequestResult.Allow(userMetadata, block) =>
+              val blockContext = resultWithHistory.history.collectFirst(matchingBlockContext(block.name)).get
```
coutoPL, line 80: to remove

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AccessControlLoggingDecorator.scala
```diff
@@ -88,6 +89,10 @@ class AccessControlLoggingDecorator(val underlying: AccessControl, auditingTool:
       }
   }
 
+  private def matchingBlockContext[B <: BlockContext](blockName: Block.Name): PartialFunction[History[B], B] = {
```
coutoPL, line 92: to remove

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AuditRequestContextBasedOnAclResult.scala
```diff
@@ -21,21 +21,30 @@ import java.time.Instant
...
+                                                             userMetadata: Option[UserMetadata],
                                                              historyEntries: Vector[History[B]],
-                                                             loggingContext: LoggingContext)
+                                                             loggingContext: LoggingContext,
+                                                             override val involvesIndices: Boolean)
   extends AuditRequestContext {
 
   implicit val showHeader: Show[Header] = obfuscatedHeaderShow(loggingContext.obfuscatedHeaders)
+
+  private val maybeLoggedUser: Option[domain.LoggedUser] = userMetadata
```
coutoPL, line 45: why not just `loggedUser`?

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AuditRequestContextBasedOnAclResult.scala
```diff
@@ -65,18 +74,35 @@ class AuditRequestContextBasedOnAclResult[B <: BlockContext](requestContext: Req
...
-  override val involvesIndices: Boolean = blockContext match {
-    case Some(_: GeneralIndexRequestBlockContext | _: TemplateRequestBlockContext | _: SnapshotRequestBlockContext) => true
-    case _ => false
-  }
+
   override val attemptedUserName: Option[String] = requestContext.basicAuth.map(_.credentials.user.value.value)
   override val rawAuthHeader: Option[String] = requestContext.rawAuthHeader.map(_.value.value)
 }
+
+object AuditRequestContextBasedOnAclResult {
```
coutoPL, line 89: I'd rather:
* create extension method for a block context
* change the name, because `inspect indices` are too general
* move the class to BlockContext file

core/src/test/scala/tech/beshu/ror/unit/acl/logging/AuditRequestContextBasedOnAclResultTest.scala
```diff
@@ -0,0 +1,43 @@
...
+import tech.beshu.ror.accesscontrol.blocks.BlockContext.{HasIndexPacks, HasIndices}
+import tech.beshu.ror.accesscontrol.logging.AuditRequestContextBasedOnAclResult
+
+class AuditRequestContextBasedOnAclResultTest extends WordSpec with MockFactory {
+
+  import org.scalatest.Matchers._
+
+  "find HasIndexPacks" in {
+    implicitly[HasIndexPacks[BlockContext.FilterableMultiRequestBlockContext]]
+    val bc: BlockContext.FilterableMultiRequestBlockContext = BlockContext.FilterableMultiRequestBlockContext(null, null, null, null, null, null)
```
coutoPL, line 31: why not to use scalamock instead of creating object with null params?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/CurrentUserMetadataSuite.scala
```diff
@@ -20,20 +20,42 @@ import org.junit.Assert.assertEquals
...
     with BaseSingleNodeEsClusterTest {
   this: EsContainerCreator =>
 
   override implicit val rorConfigFileName = "/current_user_metadata/readonlyrest.yml"
+  private lazy val auditIndexManager = new AuditIndexManagerJ(basicAuthClient("admin", "container"), "audit_index")
 
   "An ACL" when {
     "handling current user metadata kibana plugin request" should {
       "allow to proceed" when {
+        "using proxy auth" excludeES(rorProxy) in {
```
coutoPL, line 40: I think that `QueryAuditLogSerializerSuite` is a better place for this test

coutoPL (changes requested):

core/src/test/scala/tech/beshu/ror/unit/acl/logging/AuditRequestContextBasedOnAclResultTest.scala
clutroth: I created those classes as `final case class`, so I can't mock them, and I don't want to change it.

core/src/main/scala/tech/beshu/ror/accesscontrol/acl/AccessControlList.scala
clutroth: ?

core/src/test/scala/tech/beshu/ror/unit/acl/logging/AuditRequestContextBasedOnAclResultTest.scala
coutoPL: I wrote about block context classes. But ok, we will change it in future

coutoPL (approved):
@clutroth please add the description to the PR and we can merge

PR #660, opened 2020-11-09
[RORDEV-325] internal audit events API + internal enhancements

coutoPL:


sscarduzio:
nice, let's write a bit of documentation about this, so we can start interacting with it from Kibana

PR #661, opened 2020-11-11
[Rordev-381] es 7.10.0 support

clutroth:
🚀New (ES) 7.10.0 support


coutoPL:
@clutroth I see no new pipeline task for ES710x

eshome/build.gradle
```diff
@@ -0,0 +1,86 @@
...
+apply plugin: 'java'
+
+configurations {
+    runtime
+    ror
+    netty
+}
+
+dependencies {
+    ror project(path: ':es79x')
```
coutoPL, line 42: I understand that we need this file to be able to run ROR with ES during development in IDEA? Could you please describe how to do it in our internal guide? It'd be nice to have sth like this: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/wiki/Development#run-es-as-a-java-application (I'll move the old guide some day)

clutroth: done

PR #665, opened 2020-12-04
[RORDEV-386] one more case where there is no user info in audit

clutroth:
🐞Fix (ES) "username" in audit entry when request is rejected


core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AuditingTool.scala
```diff
@@ -109,6 +110,15 @@ class AuditingTool(settings: Settings,
     }
   }
 
+  private def findLastBlockContext[B <: BlockContext](forbidden: ResponseContext.Forbidden[B]) =
+    forbidden.history
+      .flatMap(_.items)
+      .map(_.result)
+      .reverse
```
clutroth, line 117: I want to mention one thing. Here, I'm getting last `Fulfilled` rule. I see `blockContext` is updated in each `Fulfilled` result, so last `Fulfilled` should contain last, and I believe, and is, or should be proven in test?

coutoPL (approved):
LGTM! Thanks

PR #669, opened 2020-12-13
[RORDEV-76] srv ldap service discovery

pielas:
🚀New (ES) Support for LDAP server discovery using _ldaps._tcp SRV record


core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
```diff
@@ -0,0 +1,99 @@
...
+    with BaseYamlLoadedAccessControlTest
+    with BeforeAndAfterAll
+    with ForAllTestContainer
+    with Inside {
+
+  private val ldapContainer = new LdapContainer("LDAP1", "test_example.ldif")
+
+  private var dnsContainer: Option[DnsServerContainer] = None
+
+  override val container: LdapContainer = ldapContainer
```
coutoPL, line 43: maybe we can use MultipleContainers to gather ldap and dnsserver containers together?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
```diff
@@ -47,9 +47,22 @@ class LdapServicesSettingsTests(ldapConnectionPoolProvider: UnboundidLdapConnect
 
   private val containerLdap1 = new LdapContainer("LDAP1", "test_example.ldif")
   private val containerLdap2 = new LdapContainer("LDAP2", "test_example.ldif")
+  private var dnsContainer: Option[DnsServerContainer] = None
```
coutoPL, line 50: same here - why not to move it to MultipleContainters?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
```diff
@@ -199,6 +212,35 @@ class LdapServicesSettingsTests(ldapConnectionPoolProvider: UnboundidLdapConnect
           }
         )
       }
+      "server discovery is enabled" in {
+        assertDecodingSuccess(
+          yaml =
+            s"""
+               |  ldaps:
+               |  - name: ldap1
+               |    server_discovery:
```
coutoPL, line 221: what about the simple configuration test (booleanDiscoverySettingDecoder)?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/DnsServerContainer.scala
```diff
@@ -0,0 +1,49 @@
...
+          |$$ORIGIN example.org.
+          |@	3600	IN	SOA someorg.org.  someorg.com.  (2017042745 7200 3600 1209600 3600)
+          |_ldap._tcp.	 86400	IN	SRV	10	60     $srvServicePort	localhost.
+          |""".stripMargin)
+  ) {
+
+  def dnsPort: Int = {
+    // This is hack to obtain mapping to UDP port as testcontainers doesn't allow explicit mapping of UDP ports
+    // although it is mapping each port exposed by container(even UDP).
+    this.containerInfo.getNetworkSettings.getPorts.getBindings.asScala.get(new ExposedPort(53, InternetProtocol.UDP)).get.head.getHostPortSpec.toInt
```
coutoPL, line 42: as it's a hack I can image it may stop to work in future, so maybe instead of using `.get` we should use `.getOrElse(...)` and throw exception with meaningful message.

coutoPL (changes requested):
+ please update docs

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
pielas: I wanted to use MultipleContainers, but the problem is that I have to know mapped port of LdapContainer before I start DNS container. MultipleContainers is starting containers in independent way. This is why I'm starting DNS after containers are started

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
pielas: as above

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
coutoPL: OK, but I see we use it in two places so, it'd be nice to have container which behaves like you described. Please take a look at `EsClusterContainer` - this a multi ES containers cluster which uses external dependencies (like LDAP) which are also run in containers but have to be started before starting ES (+ROR obviously) nodes. So this is pretty what you need, but here DnsServerContainer has to wait for LDAP. Maybe we can generalise the EsClusterContainer solution to be ES independent or extract the code you can use in more generic case.

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
```diff
@@ -199,6 +212,64 @@ class LdapServicesSettingsTests(ldapConnectionPoolProvider: UnboundidLdapConnect
           }
         )
       }
+//  Commented out until feature allowing starting up ROR without LDAP connection will be developed
```
coutoPL, line 215: use `ignore` instead. It's easier to not forget about it because it's shown in test reports.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/DnsServerContainer.scala
```diff
@@ -0,0 +1,57 @@
...
+  def dnsPort: Int = {
+    // This is hack to obtain mapping of UDP port as testcontainers doesn't allow explicit mapping of UDP ports
+    // although it is mapping each port exposed by container(even UDP).
+    this.containerInfo
+      .getNetworkSettings
+      .getPorts
+      .getBindings
+      .asScala
+      .getOrElse(new ExposedPort(53, InternetProtocol.UDP), throw new Exception("Couldn't get value of docker mapped port 53/udp"))
+      .head
```
coutoPL, line 48: also here

coutoPL (changes requested):

coutoPL (approved):
LGTM. Thanks!

PR #671, opened 2020-12-19
Feature/[RORDEV-393] allow startup without ldap connection

pielas:
🚀 New (ES) New configuration option allowing to ignore LDAP connectivity problems

doc PR: https://github.com/beshu-tech/readonlyrest-docs/pull/76


coutoPL (changes requested):
Looks good, but please take a look at compilation issues

PR #680, opened 2021-01-09
[RORDEV-356] fields and filter rules should not match admin actions

pielas:


pielas:
Please check if such solution is enough and it won't break backward compatibility

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/FieldsRule.scala
```diff
@@ -45,7 +45,7 @@ class FieldsRule(val settings: Settings)
   override val name: Rule.Name = FieldsRule.name
 
   override def check[B <: BlockContext : BlockContextUpdater](blockContext: B): Task[RuleResult[B]] = Task {
-    if (blockContext.requestContext.isReadOnlyRequest) {
+    if (blockContext.requestContext.isReadOnlyRequest && !blockContext.requestContext.isRorAdminRequest) {
```
coutoPL, line 48: let's use pattern matching here - it's much more easier to reason about:

```
blockContext.requestContext match {
  case r if r. isRorAdminRequest => RuleResult.Fulfilled(blockContext)
  case r if r. isReadOnlyRequest => handleReadOnlyRequest(blockContext)
  case _ => RuleResult.Rejected()
}
```

As you see it's a little bit different, because we agreed that every ROR action should be matched. Then we should check if it's readonly.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/FilterRule.scala
```diff
@@ -37,7 +37,7 @@ class FilterRule(val settings: Settings)
 
   override def check[B <: BlockContext : BlockContextUpdater](blockContext: B): Task[RuleResult[B]] = Task {
     val requestContext = blockContext.requestContext
-    if (!requestContext.isAllowedForDLS) Rejected()
+    if (!requestContext.isAllowedForDLS || requestContext.isRorAdminRequest) Rejected()
```
coutoPL, line 40: here similar approach

core/src/main/scala/tech/beshu/ror/utils/RCUtils.java
```diff
@@ -56,10 +56,18 @@
       "indices:admin/resolve/index"
   ));
 
+  private static MatcherWithWildcards rorAdminRequestMatcher = new MatcherWithWildcards(Sets.newHashSet(
```
coutoPL, line 59: let's remove it and do it a little bit differently (in next comments I'll described what I mean)

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/FieldsRuleTests.scala
```diff
@@ -203,6 +203,18 @@ class FieldsRuleTests extends WordSpec with MockFactory with Inside {
         )
       }
     }
+    "request is ROR admin request" should {
+      "not match" in {
```
coutoPL, line 207: I though that the problem was the fact that ROR request was not matched by one block with filter rule (was matched by the next block)? Please double check it

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/FilterRuleTests.scala
```diff
@@ -118,6 +118,14 @@ class FilterRuleTests extends WordSpec with MockFactory {
         val requestContext = MockRequestContext.indices.copy(isAllowedForDLS = false)
         val blockContext = FilterableRequestBlockContext(requestContext, UserMetadata.empty, Set.empty, List.empty, Set.empty, Set.empty, None)
 
+        rule.check(blockContext).runSyncStep shouldBe Right(RuleResult.Rejected())
+      }
+      "request is ROR admin request" in {
```
coutoPL, line 123: same here

es55x/src/main/scala/tech/beshu/ror/es/request/context/BaseEsRequestContext.scala
```diff
@@ -145,6 +145,9 @@ abstract class BaseEsRequestContext[B <: BlockContext](esContext: EsContext,
     }
   }
 
+  override def isRorAdminRequest: Boolean = {
```
coutoPL, line 148: 1. let's remove it from here 
2. let's add `isRorAction` method to Action class
3. at the moment we have 4 ROR actions
4. let's use it like this in proper rules: `blockContext.requestContext.action.isRorAction`

es60x/src/main/scala/tech/beshu/ror/es/request/context/BaseEsRequestContext.scala
```diff
@@ -145,6 +145,10 @@ abstract class BaseEsRequestContext[B <: BlockContext](esContext: EsContext,
     }
   }
 
+  override def isRorAdminRequest: Boolean = {
```
coutoPL, line 148: to remove (here and in other files)

coutoPL (changes requested):

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/FieldsRuleTests.scala
pielas: resolved on slack

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/FilterRule.scala
```diff
@@ -36,26 +36,26 @@ class FilterRule(val settings: Settings)
...
-            case TemplateRequestBlockContextUpdater => Fulfilled(blockContext)
-            case GeneralIndexRequestBlockContextUpdater => Fulfilled(blockContext)
-            case AliasRequestBlockContextUpdater => Fulfilled(blockContext)
-            case MultiIndexRequestBlockContextUpdater => Fulfilled(blockContext)
-            case FilterableRequestBlockContextUpdater => addFilter(blockContext, filter)
-            case FilterableMultiRequestBlockContextUpdater => addFilter(blockContext, filter)
-          }
-      }
+    blockContext.requestContext match {
+      case r if !r.isAllowedForDLS || r.action.isRorAction => Rejected()
```
coutoPL, line 40: here we can spit it into two cases to avoid ||

coutoPL (changes requested):
pipeline failed

pielas:
> pipeline failed

Because `filter` rule rejects ror requests behaviour has been changed a little. I wonder if this is desirable or not. 
Failed test contains this in readonlyrest.yml:
```    
    - name: "User 1 - index1"
      filter: '{"query_string":{"query":"app:(cste)"}}'
      users: ["user1"]
      groups: [group2, group3]

    - name: "User 1 - index2"
      users: ["user1"]
      groups: [group2, group1]```
Previously `fetchMetadata` request was caught by first block, but after my change `filter` rule rejects this request and it falls into second block. I could just change this test, but I don't know ror too good to tell if that won't break backward compatibility for clients.

coutoPL:
> > pipeline failed
> 
> Because `filter` rule rejects ror requests behaviour has been changed a little. I wonder if this is desirable or not.
> Failed test contains this in readonlyrest.yml:
> 
> ```
>     - name: "User 1 - index1"
>       filter: '{"query_string":{"query":"app:(cste)"}}'
>       users: ["user1"]
>       groups: [group2, group3]
> 
>     - name: "User 1 - index2"
>       users: ["user1"]
>       groups: [group2, group1]
> ```
> 
> Previously `fetchMetadata` request was caught by first block, but after my change `filter` rule rejects this request and it falls into second block. I could just change this test, but I don't know ror too good to tell if that won't break backward compatibility for clients.

yes, the compatibility will be broken, but we should treat it as a bug. ROR internal requests should not be matched by the fields and filter rule. We have to obviously mention about it in our docs. Also changing the behaviour from MATCHED to MISMATCHED is rather secure - in worst case scenario user's request will be blocked. So, let's change the test.

coutoPL (approved):

coutoPL:
@pielas I'm merging it, but please fill in the description of the GH PR and create PR with doc changes

PR #681, opened 2021-01-09
[RORDEV-371] [LDAP] support transform username to uppercase/lowercase/capitalized/prefix/suffix

clutroth:
🧐Enhancement (ES) Usernames comparison can ignore cases


core/src/main/scala/tech/beshu/ror/utils/MatcherWithWildcards.java
```diff
@@ -14,141 +14,84 @@
...
+            if (nextIdx < 0) {
+                return false;
+            } else {
+                idx = nextIdx + patternTok.length();
+            }
+        }
 
-    for (String[] p : patternsList) {
-      // Ignore remote cluster related permissions if request didn't mean it
-      if (!remoteClusterRequested && Arrays.asList(p).contains(":")) {
```
clutroth, line 104: if haystack hasn't colon,and pattern has colon, miniglob won't match pattern, so `if(){continue}` expression doesn't change behaviour

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/Rule.scala
```diff
@@ -127,8 +124,11 @@ object Rule {
       }
     }
 
+    protected def caseMappingEquality: UserIdCaseMappingEquality
```
coutoPL, line 127: IMO at AuthenticationRule level we should have an info what about user ID case comparison strategy. Now, we have how the strategy is working. 

So here we should have instance of this trait:
```
 sealed trait UsernameCaseMapping
  object UsernameCaseMapping {
    case object CaseSensitive extends UsernameCaseMapping
    case object CaseInsensitive extends UsernameCaseMapping
  }
  ```

And when we create a matcher, the data above should be use to create proper function which knows how to apply the strategy during user id comparison.

File: core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/SessionMaxIdleRule.scala [hunk 3]
```diff
@@ -27,14 +27,17 @@ import tech.beshu.ror.accesscontrol.blocks.rules.Rule.{RegularRule, RuleResult}
...
 import tech.beshu.ror.providers.UuidProvider
 
 import scala.concurrent.duration.FiniteDuration
 
-class SessionMaxIdleRule(val settings: Settings)
-                        (implicit clock: Clock, uuidProvider: UuidProvider)
+final class SessionMaxIdleRule(val settings: Settings)
+                              (implicit clock: Clock,
+                               uuidProvider: UuidProvider,
+                               caseMappingEquality: UserIdCaseMappingEquality)
```
coutoPL, line 40: why do we need it here?

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
```diff
@@ -64,7 +65,7 @@ object domain {
   object User {
     final case class Id(value: NonEmptyString)
     object Id {
-      implicit val eqId: Eq[Id] = Eq.fromUniversalEquals
+      type UserIdCaseMappingEquality = CaseMappingEquality[User.Id]
```
coutoPL, line 68: shouldn't we provide an Eq[User.Id] instance based on `UsernameCaseMapping`?

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
```diff
@@ -161,6 +162,9 @@ object domain {
   }
 
   final case class Credentials(user: User.Id, secret: PlainTextSecret)
+  object Credentials {
+    implicit def eqCredentials(implicit caseMappingEquality: UserIdCaseMappingEquality): Eq[Credentials] = Eq.and(Eq.by(_.user), Eq.by(_.secret))
```
coutoPL, line 166: IMO we should implicitly require here `Eq[User.Id]` instance

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
```diff
@@ -288,6 +294,8 @@ object domain {
     }
   }
   object IndexName {
+    implicit val indexNameShow: Show[IndexName] = Show.show(_.value.value)
+    implicit val caseMappingEqualityIndexName: CaseMappingEquality[IndexName] = CaseMappingEquality.instance(identity)
```
coutoPL, line 298: I doubt we need it (for all non-userid types) - Eq should be enough

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/common.scala
```diff
@@ -121,7 +122,7 @@ object common extends Logging {
       .withError(ValueLevelCreationError(Message("Non empty list of groups are required")))
       .decoder
 
-  implicit val usersNesDecoder: Decoder[NonEmptySet[User.Id]] =
+  implicit def usersNesDecoder(implicit caseMappingEquality: UserIdCaseMappingEquality): Decoder[NonEmptySet[User.Id]] =
```
coutoPL, line 125: as I mentioned above - IMO this should be passed explicitly

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/SessionMaxIdleRuleDecoder.scala
```diff
@@ -21,13 +21,16 @@ import java.time.Clock
...
 import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.RulesLevelCreationError
 import tech.beshu.ror.accesscontrol.factory.decoders.common
 import tech.beshu.ror.accesscontrol.factory.decoders.rules.RuleBaseDecoder.RuleDecoderWithoutAssociatedFields
 import tech.beshu.ror.accesscontrol.utils.CirceOps._
 import tech.beshu.ror.providers.UuidProvider
 
-class SessionMaxIdleRuleDecoder(implicit clock: Clock, uuidProvider: UuidProvider)
+class SessionMaxIdleRuleDecoder(implicit clock: Clock,
+                                uuidProvider: UuidProvider,
+                                caseMappingEquality: UserIdCaseMappingEquality)
```
coutoPL, line 33: do we need it?

core/src/main/scala/tech/beshu/ror/accesscontrol/ops.scala
```diff
@@ -137,7 +136,6 @@ object show {
     implicit val headerNameShow: Show[Header.Name] = Show.show(_.value.value)
     implicit val kibanaAppShow: Show[KibanaApp] = Show.show(_.value.value)
     implicit val proxyAuthNameShow: Show[ProxyAuth.Name] = Show.show(_.value)
-    implicit val indexNameShow: Show[IndexName] = Show.show(_.value.value)
```
coutoPL, line 140: why? let's back to the previous version

core/src/main/scala/tech/beshu/ror/accesscontrol/ops.scala
```diff
@@ -146,8 +144,6 @@ object show {
     implicit val dnShow: Show[Dn] = Show.show(_.value.value)
     implicit val envNameShow: Show[EnvVarName] = Show.show(_.value.value)
     implicit val propNameShow: Show[PropName] = Show.show(_.value.value)
-    implicit val repositoryShow: Show[RepositoryName] = Show.show(_.value.value)
-    implicit val snapshotShow: Show[SnapshotName] = Show.show(_.value.value)
```
coutoPL, line 150: same here

core/src/main/scala/tech/beshu/ror/accesscontrol/request/RorSessionCookie.scala
```diff
@@ -49,8 +51,11 @@ object RorSessionCookie extends Logging {
     case object Invalid extends ExtractingError
   }
 
-  def extractFrom(context: RequestContext, user: LoggedUser)
-                 (implicit clock: Clock, uuidProvider: UuidProvider): Either[ExtractingError, RorSessionCookie] = {
+  def extractFrom(context: RequestContext,
+                  user: LoggedUser)
+                 (implicit clock: Clock,
+                  uuidProvider: UuidProvider,
+                  caseMappingEquality: UserIdCaseMappingEquality): Either[ExtractingError, RorSessionCookie] = {
```
coutoPL, line 58: do we need it here?

core/src/main/scala/tech/beshu/ror/utils/CaseMappingEquality.scala
```diff
@@ -0,0 +1,60 @@
...
+ */
+package tech.beshu.ror.utils
+
+import cats._
+import cats.implicits._
+import tech.beshu.ror.accesscontrol.domain.{Action, IndexName, RepositoryName, SnapshotName}
+
+import language.implicitConversions
+
+trait CaseMappingEquality[A] {
```
coutoPL, line 25: IMO we don't need it.

`sealed trait UsernameCaseMapping` and `def eqUserId(implicit caseCompatisonStrategy: UsernameCaseMapping): Eq[User.Id]` are enough

core/src/test/scala/tech/beshu/ror/integration/CaseInsensitiveGroupsWithProxyAuthAccessControlTests.scala
```diff
@@ -0,0 +1,79 @@
...
+import tech.beshu.ror.accesscontrol.domain.{Group, User}
+import tech.beshu.ror.mocks.MockRequestContext
+import tech.beshu.ror.utils.TestsUtils._
+import tech.beshu.ror.utils.uniquelist.UniqueList
+
+class CaseInsensitiveGroupsWithProxyAuthAccessControlTests extends AnyWordSpec with BaseYamlLoadedAccessControlTest with Inside {
+  override protected def configYaml: String =
+    """
+      |readonlyrest:
+      |  username_mapping: case_insensitive
```
coutoPL, line 35: why this is called "username mapping"? Do you think user will know what is going on? Why not more explicitly?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/SessionMaxIdleRule.scala
clutroth: because this rule checks user id equality. We can assume, that it's case insensitive, but can we guarantee, that user from block context, and cookie will have same cases?

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
clutroth: `UsernameCaseMapping` is `Order`, so it's `Eq`

core/src/main/scala/tech/beshu/ror/utils/CaseMappingEquality.scala
clutroth: `Eq` doesn't know about string patterns. To test pattern against `A` `mapCases(show(A))`, and pattern is mapped too `mapCases(pattern)`.

core/src/test/scala/tech/beshu/ror/integration/CaseInsensitiveGroupsWithProxyAuthAccessControlTests.scala
clutroth: I got it from jira. I changed it.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/Rule.scala
clutroth: Is `UsernameCaseMapping` `AuthenticationRule` parameter, or attribute? I guess you'd like to close `UsernameCaseMapping` in `AuthenticationRule`. `UsernameCaseMapping` isn't rule attribute, and shouldn't be interpreted bu rule. IMHO ROR should have one instance of ``UsernameCaseMapping`, and it shouldn't be interpreted by any of ror's components itself.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
clutroth: We need this for wildcard patterns

coutoPL:
@clutroth we need two more things:
* documentation
* description of the task

coutoPL (approved):
please fill in the description and create docs and we can merge

PR #713, opened 2021-04-08
[RORDEV-437] ldap circuit breaker pt1

pielas:


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/CircuitBreakerLdapServiceDecorator.scala
```diff
@@ -0,0 +1,94 @@
...
+import monix.eval.Task
+import tech.beshu.ror.accesscontrol.domain
+import tech.beshu.ror.accesscontrol.domain.User
+import monix.catnap._
+import tech.beshu.ror.utils.uniquelist.UniqueList
+
+import scala.concurrent.duration.FiniteDuration
+
+class CircuitBreakerLdapAuthenticationServiceDecorator(underlying: LdapAuthenticationService,
+                                                       maxFailures: Int,
```
coutoPL, line 30: can we gather all the CB settings into eg `CircuitBreakerConfig` class?

_(We can do it in a second part of the task)_

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/implementations/UnboundidLdapService.scala
```diff
@@ -336,6 +339,12 @@ object LdapConnectionConfig {
     final case class CustomUser(dn: Dn, password: PlainTextSecret) extends BindRequestUser
   }
 
+  sealed trait CircuitBreaker
```
coutoPL, line 342: It looks like we should defined it elsewhere. 

please take a look at:
* `ExternalAuthenticationService`
*  `ExternalAuthorizationService`

It'd be nice to have CB decorator available also for them (the same, like we do in case of caching).

_(We can do it in a second part of the task)_

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/implementations/UnboundidLdapService.scala
```diff
@@ -336,6 +339,12 @@ object LdapConnectionConfig {
...
+  object CircuitBreaker {
+    case object Disabled extends CircuitBreaker
+    final case class Enabled(maxFailures: Int, resetDuration: FiniteDuration Refined Positive) extends CircuitBreaker
```
coutoPL, line 345: maxFailures also can be `Int Refined Positive`

_(We can do it in a second part of the task)_

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/definitions/LdapServicesDecoder.scala
```diff
@@ -76,8 +76,8 @@ object LdapServicesDecoder {
   private def decodeLdapService(cursor: HCursor)
                                (implicit ldapConnectionPoolProvider: UnboundidLdapConnectionPoolProvider): Task[Either[DecodingFailure, LdapUserService]] = {
     for {
-      authenticationService <- (authenticationServiceDecoder: AsyncDecoder[LdapAuthenticationService])(cursor)
-      authortizationService <- (authorizationServiceDecoder: AsyncDecoder[LdapAuthorizationService])(cursor)
+      authenticationService <- (circuitBreakerAuthenticationServiceDecoder: AsyncDecoder[LdapAuthenticationService])(cursor)
+      authortizationService <- (circuitBreakerAuthorizationServiceDecoder: AsyncDecoder[LdapAuthorizationService])(cursor)
     } yield (authenticationService, authortizationService) match {
       case (Right(authn), Right(authz)) => Right {
         new ComposedLdapAuthService(authn.id, authn, authz)
```
coutoPL, line 83: I'm not sure this is a right level to apply the CB decorator. Please take a look at Caching decorator. 

In `decodeLdapService` we create an LDAP service (which can be authn, authz or authn&z/composed). Now, in case of authn&z service, we have two CBs applied and then composition. But IMO we should first compose authn and authz services into authn&z service and then apply (only one) CB at the top of the composition.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/definitions/LdapServicesDecoder.scala
```diff
@@ -222,6 +244,20 @@ object LdapServicesDecoder {
       }
   }
 
+  private implicit val enabledCircuitBreakerDecoder: Decoder[CircuitBreaker.Enabled] =
+    Decoder.forProduct2("max_retries", "reset_duration")(CircuitBreaker.Enabled.apply)
+
+  private lazy val circuitBreakerDecoder: Decoder[CircuitBreaker] =
+    Decoder
+      .instance { c =>
+        val circuitBreaker = c.downField("circuit_breaker")
```
coutoPL, line 253: * when `circuit_breaker` is not defined, we should enable it with default settings
* when `circuit_breaker` section is malformed, ROR should treat whole config as malformed

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
```diff
@@ -82,6 +82,39 @@ class LdapServicesSettingsTests(ldapConnectionPoolProvider: UnboundidLdapConnect
           }
         )
       }
+      "one LDAP service with circuit breaker is declared" in {
```
coutoPL, line 85: missing tests for:
* default CB in no explicit CB config declared
* malformed CB section

_(We can do it in a second part of the task)_

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
```diff
@@ -82,6 +82,39 @@ class LdapServicesSettingsTests(ldapConnectionPoolProvider: UnboundidLdapConnect
...
+               |    cache_ttl_in_sec: 60                                      # default 0 - cache disabled
+               |    circuit_breaker:
+               |      max_retries: 2
+               |      reset_duration: 2
+           """.stripMargin,
+          assertion = { definitions =>
+            definitions.items should have size 1
+            val ldapService = definitions.items.head
+            ldapService shouldBe a[CacheableLdapServiceDecorator]
+            //TODO: Check if returned CacheableLdapServiceDecorator has underlying CircuitBreaker decorator
```
coutoPL, line 113: todo left 

_(We can do it in a second part of the task)_

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/definitions/LdapServicesDecoder.scala
pielas: What should be default settings for CB?

coutoPL: let's say:
* maxFailures = 10
* resetDuration = 10 s

pielas: I gave it more though and I wonder how can we represent disabled CB state in config without requiring other fields present in that state. Will something like that be ok? 
```
circuit_breaker:
  enabled: false (optional)
  max_failures: 1 (required when enabled is true)
  reset_duration: 5s (required when enabled is true)
```

sscarduzio: Sorry if I jump into this a bit out of the blue, but I was thinking for time measures we could align to the existing naming policy? <what happens after the time>_in_<time unit of measurement> 

As seen in:
* connection_timeout_in_sec
* request_timeout_in_sec
* cache_ttl_in_sec

pielas: ok, I will make both variants `reset_duration` and `reset_duration_in_sec` allowed

coutoPL: @sscarduzio  `_in_sec` convention is deprecated. We maintain it for backward compatibility reasons. 

```
reset_duration: 60 sec
```

or 

```
reset_duration: 2 min
```

It's more elegant and flexible convention. We use it for any duration setting in our config. Maybe we should fix the docs (but there is already jira for this: RORDEV-255)

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/definitions/LdapServicesDecoder.scala
```diff
@@ -78,13 +79,19 @@ object LdapServicesDecoder {
     for {
       authenticationService <- (authenticationServiceDecoder: AsyncDecoder[LdapAuthenticationService])(cursor)
       authortizationService <- (authorizationServiceDecoder: AsyncDecoder[LdapAuthorizationService])(cursor)
-    } yield (authenticationService, authortizationService) match {
-      case (Right(authn), Right(authz)) => Right {
-        new ComposedLdapAuthService(authn.id, authn, authz)
+      circuitBreakerSettings <- AsyncDecoderCreator.from(circuitBreakerDecoder)(cursor)
+    } yield (authenticationService, authortizationService, circuitBreakerSettings) match {
+      case (Right(authn), Right(authz), Right(circuitBreakerConfig)) => Right {
+        new CircuitBreakerLdapServiceDecorator(
```
coutoPL, line 85: ok, but let's do this in separate method. It'd be nice to have similar decoder to `cachableLdapServiceDecoder`. But not implicit one. You can also remove implicit from `cachableLdapServiceDecoder` definition and apply decoders explicitly in `ldapServicesDefinitionsDecoder` method


_(We can do it in a second part of the task)_

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/definitions/LdapServicesDecoder.scala
```diff
@@ -222,6 +229,19 @@ object LdapServicesDecoder {
...
 
+
+  private lazy val circuitBreakerDecoder: Decoder[CircuitBreakerConfig] =
+    Decoder
+      .instance { c =>
+        val circuitBreaker = c.downField("circuit_breaker")
+        if (circuitBreaker.failed) {
+          Right(DEFAULT_CIRCUIT_BREAKER_CONFIG)
+        } else {
+          val decoder = Decoder.forProduct2("max_retries", "reset_duration")(CircuitBreakerConfig.apply)
```
coutoPL, line 240: do we have a test which tests what message will be seen by a user when he tries to set sth like this:

```
circuit_breaker:
  max_retries: 2
```

I wonder if this method of decoding will allow us to return meaningful error message? 
(Circe is pretty dump according to self explanatory decoding errors. We created own mechnism which makes it more user friendly, but sadly the cost is less pleasant decoding code.

_(We can do it in a second part of the task)_

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/definitions/UnboundidLdapAuthenticationServiceTests.scala
```diff
@@ -16,53 +16,92 @@
...
       .runSyncUnsafe()
   }
+
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    ldap1ContainerWithToxiproxy.enableNetwork()
+    ldap1ContainerWithToxiproxy.disableNetworkTimeout()
+  }
+
+  implicit class LdapAuthenticationServiceOps(authenticationService: LdapAuthenticationService) {
```
coutoPL, line 68: Let's move it at the end of the file (or at least after all test cases - because IMO they are more important than the assertion implmentation)

_(We can do it in a second part of the task)_

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/definitions/UnboundidLdapAuthenticationServiceTests.scala
```diff
@@ -124,6 +203,15 @@ class UnboundidLdapAuthenticationServiceTests
       .right.getOrElse(throw new IllegalStateException("LDAP connection problem"))
   }
 
+  private def createCircuitBreakerDecoratedSimpleAuthenticationService() = {
+    new CircuitBreakerLdapAuthenticationServiceDecorator(
+      createSimpleAuthenticationService(),
+      CircuitBreakerConfig(
+        maxFailures = Refined.unsafeApply(2),
```
coutoPL, line 210: this one should work when you import `auto` (sadly I don't know how we can achieve it for our custom PositiveFiniteDuration` class)

_(We can do it in a second part of the task)_

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/implementations/UnboundidLdapService.scala
coutoPL: (not resolving it, because in P2 we need to implement the CB for mentioned services)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/implementations/UnboundidLdapService.scala
```diff
@@ -291,6 +295,8 @@ final case class LdapConnectionConfig(connectionMethod: ConnectionMethod,
 
 object LdapConnectionConfig {
 
+  val DEFAULT_CIRCUIT_BREAKER_CONFIG = CircuitBreakerConfig(Refined.unsafeApply(10), Refined.unsafeApply(10 seconds))
```
coutoPL, line 298: camelCase is OK.

coutoPL (approved):
P1 can be merged. When artifacts are built, please sent the pre-build to the customer.

PR #725, opened 2021-05-29
[RORDEV-466] ResizeRequest handling

pielas:
🐞Fix (ES) Issue with handling ResizeRequest which made it unable to upgrade Kibana to version 7.12.0+

### DIFF
--- a/es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
+++ b/es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
@@ -1,0 +1,62 @@
+/*
+ *    This file is part of ReadonlyREST.
+ *
+ *    ReadonlyREST is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    ReadonlyREST is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.es.request.context.types
+
+import cats.data.NonEmptyList
+import cats.implicits._
+import org.elasticsearch.action.admin.indices.shrink.ShrinkRequest
+import org.elasticsearch.threadpool.ThreadPool
+import tech.beshu.ror.accesscontrol.AccessControlStaticContext
+import tech.beshu.ror.accesscontrol.domain.IndexName
+import tech.beshu.ror.es.RorClusterService
+import tech.beshu.ror.es.request.AclAwareRequestFilter.EsContext
+import tech.beshu.ror.es.request.context.ModificationResult
+import tech.beshu.ror.es.request.context.ModificationResult.{Modified, ShouldBeInterrupted}
+
+class ResizeEsRequestContext(actionRequest: ShrinkRequest,
+                             esContext: EsContext,
+                             aclContext: AccessControlStaticContext,
+                             clusterService: RorClusterService,
+                             override val threadPool: ThreadPool)
+  extends BaseIndicesEsRequestContext[ShrinkRequest](actionRequest, esContext, aclContext, clusterService, threadPool) {
+
+  override protected def indicesFrom(request: ShrinkRequest): Set[IndexName] = {
+    (request.getSourceIndex :: request.getShrinkIndexRequest.index() :: Nil)
+      .flatMap(IndexName.fromString)
+      .toSet
+  }
+
+  override protected def update(request: ShrinkRequest, filteredIndices: NonEmptyList[IndexName], allAllowedIndices: NonEmptyList[IndexName]): ModificationResult = {
+    val sourceIndex = IndexName.fromString(request.getSourceIndex)
+    val targetIndex = IndexName.fromString(request.getShrinkIndexRequest.index())
+
+    val isSourceIndexOnFilteredIndicesList = sourceIndex.exists(filteredIndices.toList.contains(_))
+    val isTargetIndexOnFilteredIndicesList = targetIndex.exists(filteredIndices.toList.contains(_))
+
+    if (isSourceIndexOnFilteredIndicesList && isTargetIndexOnFilteredIndicesList) {
+      Modified
+    } else {
+      if (!isSourceIndexOnFilteredIndicesList) {
+        logger.info(s"[${id.show}] Source index ShrinkRequest forbidden")
+      }
+      if (!isTargetIndexOnFilteredIndicesList) {
+        logger.info(s"[${id.show}] Target index ShrinkRequest forbidden")
+      }
+      ShouldBeInterrupted
+    }
+  }
+}
--- a/es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ShrinkEsRequestContext.scala
+++ b/es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ShrinkEsRequestContext.scala
@@ -47,7 +47,7 @@ class ShrinkEsRequestContext(actionRequest: ShrinkRequest,
     if (notAllowedIndices.isEmpty) {
       Modified
     } else {
-      throw new IllegalStateException(s"Shrink request is write request and such requests need all indices to be allowed. Not allowed indices=[${notAllowedIndices.map(_.value.value).mkString(",")}]")
+      throw new IllegalStateException(s"Resize request is write request and such requests need all indices to be allowed. Not allowed indices=[${notAllowedIndices.map(_.show).mkString(",")}]")
     }
   }
 }
--- a/es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ShrinkEsRequestContext.scala
+++ b/es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ShrinkEsRequestContext.scala
@@ -27,7 +27,7 @@ import tech.beshu.ror.es.request.AclAwareRequestFilter.EsContext
 import tech.beshu.ror.es.request.context.ModificationResult
 import tech.beshu.ror.es.request.context.ModificationResult.{Modified, ShouldBeInterrupted}
 
-class ResizeEsRequestContext(actionRequest: ShrinkRequest,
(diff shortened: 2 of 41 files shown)


es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.elasticsearch.action.admin.indices.shrink.ShrinkRequest
+import org.elasticsearch.threadpool.ThreadPool
+import tech.beshu.ror.accesscontrol.AccessControlStaticContext
+import tech.beshu.ror.accesscontrol.domain.IndexName
+import tech.beshu.ror.es.RorClusterService
+import tech.beshu.ror.es.request.AclAwareRequestFilter.EsContext
+import tech.beshu.ror.es.request.context.ModificationResult
+import tech.beshu.ror.es.request.context.ModificationResult.{Modified, ShouldBeInterrupted}
+
+class ResizeEsRequestContext(actionRequest: ShrinkRequest,
```
coutoPL, line 30: Let's maintain the convention of naming ROR ES request wrappers. Here `ShrinkEsRequestContext` (I understand that in newer versions of ES this is a ResizeRequest)

integration-tests/src/test/resources/resize/readonlyrest.yml
```diff
@@ -0,0 +1,18 @@
...
+  access_control_rules:
+
+    # ES containter initializer need this rule to configure ES instance after startup
+    - name: "CONTAINER ADMIN"
+      verbosity: error
+      type: allow
+      auth_key: admin:container
+
+    - name: "test1"
+      indices: ["test1_index", "test1_index_resized"]
```
coutoPL, line 17: it would be nice to test indices rule in this configuration:
`indices: ["test1_*"]`

es79x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
```diff
@@ -0,0 +1,62 @@
...
+
+class ResizeEsRequestContext(actionRequest: ResizeRequest,
+                             esContext: EsContext,
+                             aclContext: AccessControlStaticContext,
+                             clusterService: RorClusterService,
+                             override val threadPool: ThreadPool)
+  extends BaseIndicesEsRequestContext[ResizeRequest](actionRequest, esContext, aclContext, clusterService, threadPool) {
+
+  override protected def indicesFrom(request: ResizeRequest): Set[IndexName] = {
+    (request.getSourceIndex :: request.getTargetIndexRequest.index() :: Nil)
```
coutoPL, line 38: I see that target index request can have aliases. We should also add them here. ResizeRequest with target index with at least one forbidden alias should be rejected.

es79x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  override protected def update(request: ResizeRequest, filteredIndices: NonEmptyList[IndexName], allAllowedIndices: NonEmptyList[IndexName]): ModificationResult = {
+    val sourceIndex = IndexName.fromString(request.getSourceIndex)
+    val targetIndex = IndexName.fromString(request.getTargetIndexRequest.index())
+
+    val isSourceIndexOnFilteredIndicesList = sourceIndex.exists(filteredIndices.toList.contains(_))
+    val isTargetIndexOnFilteredIndicesList = targetIndex.exists(filteredIndices.toList.contains(_))
+
+    if (isSourceIndexOnFilteredIndicesList && isTargetIndexOnFilteredIndicesList) {
```
coutoPL, line 50: according to the comment above:
* it looks like this check is not necessary in general
* this is a write request (not read only), so `indices` rule semantics is as follows: match only when all requested index names are allowed
* this method won't be called if at least one index name, collected in `indicesFrom` method will be forbidden

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResizeSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+  private lazy val user1IndexManager = new IndexManager(basicAuthClient("dev1", "test"))
+
+  override def nodeDataInitializer = Some(ResizeSuite.nodeDataInitializer())
+
+  "A resize request" should {
+    "be able to proceed" when {
+      "user has permission to source index and dest index"  in {
+        val result = user1IndexManager.resize("test1_index", "test1_index_resized")
+
+        result.isSuccess should be(true)
```
coutoPL, line 46: IMO it's better to check a response code here and below

coutoPL (changes requested):

es79x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
pielas: hmm, so maybe we should throw invalid state exception from this method in case it eventually gets called?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResizeSuite.scala
pielas: I also changed it in `RemoteReindexSuite` and `ReindexSuite` because it was handled the same way

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RemoteReindexSuite.scala
```diff
@@ -74,36 +75,36 @@ trait RemoteReindexSuite
       "request specifies source and dest index that are allowed on both source and dest ES" in {
```
coutoPL, line 75: we'd rather merge suites then split them. That's why it'd be nice to merge these two reindex suites. Moreover, according to the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs.html), I see that ES classifies reindex operation as one of Document API operations. So maybe we can extend `DocumentApiSuite`?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResizeSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+        val result = user1IndexManager.resize("not_allowed_index", "not_allowed_index_resized")
+
+        result.responseCode should be(HttpStatus.SC_UNAUTHORIZED)
+      }
+      "user has permission to source index and but no permission to dest index"  in {
+        val result = user1IndexManager.resize("test1_index", "not_allowed_index_resized")
+
+        result.responseCode should be(HttpStatus.SC_UNAUTHORIZED)
+      }
+      "user has permission to dest index and but no permission to source index"  in {
```
coutoPL, line 76: what about aliases tests?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RemoteReindexSuite.scala
pielas: ok, but maybe let's do it in one of next PRs? I touched Reindex suites because I wanted to change response code matching according to what I've done in `ResizeSuite`

es55x/src/main/scala/tech/beshu/ror/es/request/context/types/ShrinkEsRequestContext.scala
```diff
@@ -0,0 +1,53 @@
...
+      .flatMap(IndexName.fromString)
+      .toSet
+  }
+
+  override protected def update(request: ShrinkRequest, filteredIndices: NonEmptyList[IndexName], allAllowedIndices: NonEmptyList[IndexName]): ModificationResult = {
+    val notAllowedIndices = indicesFrom(actionRequest) -- filteredIndices.toList.toSet
+    if (notAllowedIndices.isEmpty) {
+      Modified
+    } else {
+      throw new IllegalStateException(s"Shrink request is write request and such requests need all indices to be allowed. Not allowed indices=[${notAllowedIndices.map(_.value.value).mkString(",")}]")
```
coutoPL, line 50: let's use `.show` instead of `_.value.value`. Proper instance of Show type class can be found in `tech.beshu.ror.accesscontrol.show.logs._`

es63x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  override protected def update(request: ResizeRequest, filteredIndices: NonEmptyList[IndexName], allAllowedIndices: NonEmptyList[IndexName]): ModificationResult = {
+    val sourceIndex = IndexName.fromString(request.getSourceIndex)
+    val targetIndex = IndexName.fromString(request.getTargetIndexRequest.index())
+
+    val isSourceIndexOnFilteredIndicesList = sourceIndex.exists(filteredIndices.toList.contains(_))
+    val isTargetIndexOnFilteredIndicesList = targetIndex.exists(filteredIndices.toList.contains(_))
+
+    if (indicesFrom(actionRequest).subsetOf(filteredIndices.toList.toSet)) {
```
coutoPL, line 52: cannot we copy&paste implementation of the same class from different module?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RemoteReindexSuite.scala
coutoPL: ok

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResizeSuite.scala
```diff
@@ -0,0 +1,132 @@
...
+        "alias is used" in {
+          val result = user1IndexManager.resize("test1_index", "test1_index_resized", "test1_index_allowed_alias" :: Nil)
+
+          result.responseCode should be(200)
+        }
+      }
+    }
+    "not be able to proceed" when {
+      "user has permission to source index and dest index but no permission to alias" in {
+        val result = user1IndexManager.resize("test1_index", "test1_index_resized", "test1_index_not_allowed_alias" :: Nil)
```
coutoPL, line 62: let's use named parameters

coutoPL (changes requested):
Minor fixes needed. Let's merge with origin develop, because there was some refactoring according to indices names.

es63x/src/main/scala/tech/beshu/ror/es/request/context/types/ResizeEsRequestContext.scala
pielas: ah I worked on this file and that's why I missed it when I was copying and pasting proper implementation

PR #770, opened 2022-01-11
[RORDEV-580] API design

coutoPL:


coutoPL:
@Dzuming review please :)

Dzuming:
I assume that it's just for test settings and auth mock now, but something Is broken in the case of the settings section :P 
image

A few things about auth services mocks:
1. From a Business/Api point of view, do we need `NOT_CONFIGURED` mock status? Should we treat an empty user list as `NOT_CONFIGURED` mock status or just an empty list? On the client-side, we assume that an empty list means that we need to show a user 
image

2.  About your question in a Jira, yes we still want to display this list, and in addition, we also want to show a list of "local_users": We have two cases here
-  Every local user is defined correctly  
image
- Some of the local user usernames are not recognized 
image
and we show information about it, and the user is able to free type username
image

coutoPL:
> I assume that it's just for test settings and auth mock now, but something Is broken in the case of the settings section :P !

yeah, it can be misleading. Swagger UI duplicates the endpoints when more than one tag per endpoint is added. Will remove "Settings" and "Auth Mock" tags.

> 1. ~From a Business/Api point of view, do we need `NOT_CONFIGURED` mock status? Should we treat an empty user list as `NOT_CONFIGURED` mock status or just an empty list? On the client-side, we assume that an empty list means that we need to show a user~
>    EDIT: I checked Jira discussion about it, and let's leave it as it is, we will just interpret  `NOT_CONFIGURED` and empty users list, the same way -> by showing the `Configure` button

From the backend perspective, an empty list of used and NOT_CONFIGURED mock are two different things. When mock is not configured, we ROR with 403 response will return IMPERSONATION_NOT_SUPPORTED. It means that when the request was forbidden, at least one auth rule could not support impersonation. And then it means, that the admin (the impersonator) cannot be sure that ROR will behave exactly the same when the user (the one we are impersonating) calls it. 

> 2. About your question in a Jira, yes we still want to display this list, and in addition, we also want to show a list of "local_users": 

OK, will add the endpoint for fetching users today.

But do we need to return static users and groups? Because from ES ROR perspective and our ACL there is no strict connection between users and groups (at least not for every static rule). Maybe we don't need them at all. 

> 3. Seems that status property will have only one `PRESENT` value, from a client perspective, after opening the `Impersonate` tab we need to perform check status action ('not_configured", 'expired", 'present") and perform a specific action based on it. Maybe we need additional  `GET /_readonlyrest/admin/config/test/status` endpoint to fetch current test settings status ?

`GET /_readonlyrest/admin/config/test` returns 200 with JSON body with field status with two possible values: `PRESENT`  and `TEST_SETTINGS_NOT_CONFIGURED` (see examples and schema). Moreover, in the case of TEST_SETTINGS_NOT_CONFIGURED there should be nullable filed `recent_settings`, so you can use it too.

Dzuming:
> But do we need to return static users and groups? Because from ES ROR perspective and our ACL there is no strict 
> connection between users and groups (at least not for every static rule). Maybe we don't need them at all.

yes, we can just don't show the groups column in this case :)

> GET /_readonlyrest/admin/config/test returns 200 with JSON body with field status with two possible values: PRESENT 
> and TEST_SETTINGS_NOT_CONFIGURED (see examples and schema). Moreover, in the case of 
> TEST_SETTINGS_NOT_CONFIGURED there should be nullable filed recent_settings, so you can use it too.

Ohh, you are right we have 2 statuses already, maybe you can also provide the third status, `TEST_SETTINGS_EXPIRED` ?

coutoPL:
> Ohh, you are right we have 2 statuses already, maybe you can also provide the third status, `TEST_SETTINGS_EXPIRED` ?

I'm not sure if we should do that. Look at the machine state diagrams:
msd

Currently, we have 2 states. If a new, third stage, the starting stage (NOT_CONFIGURED) will be active only at the beginning of usage of the feature. It shows that sth is wrong with the state machine. 

Currently:
* if state == PRESENT => there are active Test Settings applied
* if state == NOT_CONFIGURED && recent_settings == null => there are no Test Settings applied and never were
* if state == NOT_CONFIGURED && recent_settings != null => there are no Test Settings applied (== the settings were invalidated or TTL was reached)

If we don't want to distinguish between these two cases:
* Test Settings were invalidated by calling `DELETE /_readonlyrest/admin/config/test`
* Test Settings were auto-invalidated because TTL reached
I'd not change anything here

coutoPL:
I've added a new endpoint definition: `GET /_readonlyrest/admin/config/test/localusers`

Dzuming:
>  If a new, third stage, the starting stage (NOT_CONFIGURED) will be active only at the beginning of usage of the feature. It 
> shows that sth is wrong with the state machine.

the question is, is something wrong or we just don't have to transition to this state implemented :P (because we don't need it)

Now I know that we can receive information about test settings expiration but

> if state == NOT_CONFIGURED && recent_settings != null => there are no Test Settings applied (== the settings were 
> invalidated or TTL was reached)

That's true, but from my perspective (the person who will use it), this condition is problematic
1. It's not 100% true, because test settings are present, but not valid/TTL reached, maybe I understand `NOT_CONFIGURED` wrong way 🤔 
2. We probably will need to write it down somewhere the information, how to achieve a specific state (now, we will try to achieve a specific state based on state machine :P )

coutoPL:
> the question is, is something wrong or we just don't have to transition to this state implemented :P (because we don't need it)
> 
> Now I know that we can receive information about test settings expiration but

We can do it like this:
Screenshot 2022-01-13 at 21 35 03

But I'm not sure how from a UI perspective you want to represent these two states: NOT_CONFIGURED and EXPIRED.

From the backend point of view, the states would mean:
* NOT CONFIGURED - no Test Settings were configured yet or previously configured Tests Settings were invalidated by calling proper endpoint
* EXPIRED - Test Settings were configured but their TTL was reached and they are not active anymore

> Looks good, could you also add example when username is not recognised (I assume that it will be undefined value on the users list), the case when we want too display possibility to free type, username in the dialog

Before I answer, let's recap:
* We can have a local user from rules like `auth_key`, `groups` (`users` section)
* We can have external users from rules like `ldap_auth`, `external_authentication` (configured using Auth Mocks)
* We can have the rules above, but without Auth Mocks configured (not supported by impersonation until the Auth Mock is not configured)
* We can have rules that don't support authentication like `auth_sha1` fully hashed (not supported at all)
* anonymous access (no authentication rule in the block)

And that's all. It looks like we don't have a case where we cannot "recognize" users. For the sake of the last point, we need the field to enter the username manually by the admin. I think we are able to detect that the last option is possible, so if we need to show the field conditionally, we could do that.

----

I've just realized that we missed one thing. Simone wrote:
> In order to send this information to the front end, the impersonatable users extraction API response could include an array of problematic block names (each ACL block has a unique name string) that are found to be problematic due to the incompatible authentication rules value

So I'm going to add this to the API design.

Dzuming:
What we want to achieve from a client perspective is:
- When test settings are not configured -> show initialize test settings modal (we allow users to initialize test settings based on current settings)
- When tests settings are expired or invalidated -> show expired test settings modal (we allow users to initialize test settings based on current settings or activate invalidated/expired current test settings)

based on it

> NOT CONFIGURED - no Test Settings were configured yet or previously configured Tests Settings were invalidated by calling proper endpoint
> EXPIRED - Test Settings were configured but their TTL was reached and they are not active anymore

This approach won't allow us to implement UI and we need to leave your initial approach :) 

> if state == PRESENT => there are active Test Settings applied

>  if state == NOT_CONFIGURED && recent_settings == null => there are no Test Settings applied and never were

> if state == NOT_CONFIGURED && recent_settings != null => there are no Test Settings applied (== the settings were 

> invalidated or TTL was reached)

coutoPL:
@Dzuming 
1. OK, it looks like this should be a final state machine model of Tests Settings
Screenshot 2022-01-16 at 20 50 54

2. I've added a flag that indicates that there were "unrecognized" users
3. Did you see the warnings design? Is it ok?

Dzuming:
1. Looks good, thanks!
2. Looks good
3. Yes, I had a chance to see it, also looks good :P

coutoPL:
OK, great. So, we have a design of the API. I'm going to merge this and we can adapt the existing endpoints and implement the new ones. 

Thanks!

PR #779, opened 2022-02-12
[RORDEV-589] Improved Test configuration API implementation

mateuszkp96:


mateuszkp96:
So far I have implemented integration with es714x and es79x modules. When everything is fine, it will integrate with other modules

core/src/main/scala/tech/beshu/ror/api/TestSettingsApi.scala
```diff
@@ -0,0 +1,260 @@
...
+        }
+      case _ =>
+        liftFailure("Malformed settings payload - only one settings value allowed")
+    }
+  }
+
+  private def ttlValue(json: io.circe.Json) = {
+    def liftFailure(failureMessage: String) = Left(TestSettingsResponse.UpdateTestSettings.FailedResponse(failureMessage))
+
+    json \\ "ttl" match {
```
coutoPL, line 101: the extraction of value under a specified key in this method is similar to the one above. And probably to the ConfigApi method. Maybe we can make util from these?

Or maybe we should refactor both API classes and use Circe decoders?

core/src/main/scala/tech/beshu/ror/api/TestSettingsApi.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  private def loadCurrentTestSettings()
+                                     (implicit requestId: RequestId): Task[TestSettingsResponse] = {
+    rorInstance
+      .currentTestSettings()
+      .map {
+        case TestSettings.NotConfigured =>
+          TestSettingsResponse.ProvideTestSettings.TestSettingsNotConfigured("ROR Test settings are not configured")
+        case TestSettings.Present(config, _, validTo) =>
+          val ttl = getTtl(validTo)
```
coutoPL, line 132: this is the TTL that was requested. We should not recalculate the current value of TTL. The `validTo` field will tell the caller that. Maybe the description in the API doc should be more precise (if so, let's fix it)

core/src/main/scala/tech/beshu/ror/api/TestSettingsApi.scala
```diff
@@ -0,0 +1,260 @@
...
+    object Type {
+      case object ProvideTestSettings extends Type
+      case object UpdateTestSettings extends Type
+      case object InvalidateTestSettings extends Type
+      case object ProvideLocalUsers extends Type
+    }
+  }
+
+  sealed trait TestSettingsResponse {
+    def status: String
```
coutoPL, line 200: let's not add this to the trait. Let's create a typeclass like this:
```
implicit StatusFromTestConfigResponse(val response: TestConfigResponse) extends AnyVal {
  def status: String = response match {...}
}
```

core/src/main/scala/tech/beshu/ror/api/TestSettingsApi.scala
```diff
@@ -0,0 +1,260 @@
...
+    object ProvideTestSettings {
+      final case class Warning(blockName: String,
+                               ruleName: String,
+                               message: String,
+                               hint: String)
+
+      final case class CurrentTestSettings(ttl: FiniteDuration,
+                                           validTo: Instant,
+                                           settings: RawRorConfig,
+                                           warnings: List[Warning]) extends TestSettingsResponse {
```
coutoPL, line 214: shouldn't we create here a new sealed trait for the endpoint Response?

core/src/main/scala/tech/beshu/ror/api/TestSettingsApi.scala
```diff
@@ -0,0 +1,260 @@
...
+        override val status: String = "TEST_SETTINGS_NOT_CONFIGURED"
+      }
+
+      final case class TestSettingsInvalidated(message: String, settings: RawRorConfig) extends TestSettingsResponse {
+        override val status: String = "TEST_SETTINGS_INVALIDATED"
+      }
+    }
+
+    object UpdateTestSettings {
+      final case class SuccessResponse(message: String) extends TestSettingsResponse {
```
coutoPL, line 228: both comments mentioned above also apply here and below

core/src/main/scala/tech/beshu/ror/boot/engines/BaseReloadableEngine.scala
```diff
@@ -102,6 +102,7 @@ private[engines] abstract class BaseReloadableEngine(val name: String,
       _ <- canBeReloaded(newConfig)
       newEngineWithConfig <- reloadWith(newConfig, newConfigEngineTtl)
       _ <- replaceCurrentEngine(newEngineWithConfig)
```
coutoPL, line 104: I noticed that it can return None in Right. It looks like we could throw an Invalid State exception when the result state is NotStartedYet and always return EngineWithConfig in Right

core/src/main/scala/tech/beshu/ror/boot/engines/BaseReloadableEngine.scala
```diff
@@ -189,9 +190,14 @@ private[engines] abstract class BaseReloadableEngine(val name: String,
 
   private def stop(engineWithConfig: EngineWithConfig)
                   (implicit requestId: RequestId): Unit = {
+    onStop()
     engineWithConfig.engine.shutdown()
     logger.info(s"[${requestId.show}] ROR $name engine (id=${engineWithConfig.config.hashString()}) stopped!")
   }
+
+  protected def onSuccessLoad(c: EngineWithConfig): Unit
```
coutoPL, line 198: I'm not sure if we have to define these two callback methods. Please, take a look at how we extract `engine` from  `currentEngine` State. It looks like we can extract raw config in the same way. Moreover, we have also info that the engine was not started yet or was stopped.

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestsettings/RRTestSettingsResponse.scala
```diff
@@ -0,0 +1,107 @@
...
+
+object RRTestSettingsResponse extends Logging {
+  def apply(response: Either[Throwable, TestSettingsApi.TestSettingsResponse]): RRTestSettingsResponse = {
+    response match {
+      case Left(ex) =>
+        logger.error("RRAdmin internal error", ex)
+        new RRTestSettingsResponse(TestSettingsApi.TestSettingsResponse.internalError)
+      case Right(value) =>
+        new RRTestSettingsResponse(value)
+
```
coutoPL, line 103: redundant \n

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestsettings/RRTestSettingsResponse.scala
```diff
@@ -0,0 +1,107 @@
...
+    response match {
+      case res: ProvideTestSettings.CurrentTestSettings => currentSettingsJson(builder, res)
+      case ProvideTestSettings.TestSettingsNotConfigured(message) => addResponseJson(builder, response.status, message)
+      case res: ProvideTestSettings.TestSettingsInvalidated => invalidatedSettingsJson(builder, res)
+      case UpdateTestSettings.SuccessResponse(message) => addResponseJson(builder, response.status, message)
+      case UpdateTestSettings.FailedResponse(message) => addResponseJson(builder, response.status, message)
+      case InvalidateTestSettings.SuccessResponse(message) => addResponseJson(builder, response.status, message)
+      case res: ProvideLocalUsers.SuccessResponse => provideLocalUsersJson(builder, res)
+      case ProvideLocalUsers.TestSettingsNotConfigured(message) => addResponseJson(builder, response.status, message)
+      case Failure(message) => addResponseJson(builder, response.status, message)
```
coutoPL, line 41: `ConfigApi.Failure` approach is a little legacy, I think. Here we should return 400 or 500 depending on the context:
1. malformed request
2. internal problem

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
```diff
@@ -236,14 +238,199 @@ trait BaseAdminApiSuite
...
+            testSettingsResponse.responseJson("status").str should be("TEST_SETTINGS_PRESENT")
+            testSettingsResponse.responseJson("settings").str should be(testSettings)
+            testSettingsResponse.responseJson("ttl").str.matches("""^(0|[1-9][0-9]*) seconds$""") should be(true)
+            isIsoDateTime(testSettingsResponse.responseJson("valid_to").str) should be(true)
+            testSettingsResponse.responseJson("warnings").arr should be(empty)
+          }
+
+          val dev1Ror1stInstanceSearchManager = new SearchManager(
+            clients.head.basicAuthClient("admin1", "pass"),
+            Map("IMPERSONATE_AS" -> "dev1")
```
coutoPL, line 277: BTW, let's change the header name to `X-ROR-Impersonating` (as agreed somewhere in ROR KBN reviews)

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/boot/engines/BaseReloadableEngine.scala
mateuszkp96: In `replaceCurrentEngine` method `getAndTransform` was called so result was the old engine. 
The old enginge is closed in transform part so we could ignore result and return just Unit type.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
```diff
@@ -236,14 +238,199 @@ trait BaseAdminApiSuite
         }
       }
     }
+    "provide a method for get current test settings" which {
```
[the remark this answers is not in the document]
mateuszkp96: I don't know which case you mean. In API we have 3: TEST_SETTINGS_PRESENT, TEST_SETTINGS_NOT_CONFIGURED, TEST_SETTINGS_INVALIDATED.
And each of them is tested:
image

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestsettings/RRTestSettingsResponse.scala
mateuszkp96: Removed

core/src/main/scala/tech/beshu/ror/api/ConfigApi.scala
```diff
@@ -133,27 +109,18 @@ class ConfigApi(rorInstance: RorInstance,
...
-            EitherT(RawRorConfig.fromString(configString).map(_.left.map(error => Failure(error.show))))
-          case None =>
-            liftFailure("Malformed settings payload - settings key value is not string")
-        }
-      case _ =>
-        liftFailure("Malformed settings payload - only one settings value allowed")
-    }
+  private def decodeUpdateRequest(payload: String): Either[Failure, UpdateConfigRequest] = {
+    io.circe.parser.decode[UpdateConfigRequest](payload)
+      .left.map(error => Failure(s"JSON body malformed: [${error.getMessage}]"))
```
mateuszkp96, line 119: I'm not sure if plain circe error is enough for this message. We can make pretty message with failed cursor (path).

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestconfig/RRTestConfigActionHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+      case Some(api) => doPrivileged {
+        implicit val requestId: RequestId = request.requestContextId
+        api
+          .call(request.getTestConfigRequest)
+          .runAsync { result =>
+            handle(result, listener)
+          }
+      }
+      case None =>
+        listener.onFailure(new Exception("TestConfig API is not available"))
```
mateuszkp96, line 45: When API is not available (because ROR is not started yet) could we return 503 http code? In this context we have only listener so maybe we should pass concrete exception and interpret this in other layer.

core/src/main/scala/tech/beshu/ror/boot/RorInstance.scala
```diff
@@ -193,12 +204,20 @@ object RorInstance {
     final case class EngineReloadError(underlying: IndexConfigReloadError) extends ScheduledReloadError
   }
 
+  sealed trait TestConfig
```
coutoPL, line 207: maybe we can move the definition to companion object of RorInstance class?

core/src/main/scala/tech/beshu/ror/boot/engines/ImpersonatorsReloadableEngine.scala
```diff
@@ -24,20 +24,40 @@ import tech.beshu.ror.RequestId
...
-import tech.beshu.ror.boot.RorInstance.RawConfigReloadError
+import tech.beshu.ror.boot.RorInstance.{RawConfigReloadError, TestConfig}
+import tech.beshu.ror.boot.engines.BaseReloadableEngine.EngineState
 import tech.beshu.ror.boot.engines.ConfigHash._
 import tech.beshu.ror.configuration.RawRorConfig
 
+import java.time.Clock
 import scala.concurrent.duration.FiniteDuration
 
 private[boot] class ImpersonatorsReloadableEngine(boot: ReadonlyRest,
```
coutoPL, line 35: let's rename it to TestConfigBasedReloadableEngine

core/src/main/scala/tech/beshu/ror/boot/engines/MainReloadableEngine.scala
```diff
@@ -33,11 +33,14 @@ import tech.beshu.ror.configuration.IndexConfigManager.SavingIndexConfigError.Ca
 import tech.beshu.ror.configuration.RawRorConfig
 import tech.beshu.ror.utils.ScalaOps.value
 
+import java.time.Clock
+
 private[boot] class MainReloadableEngine(boot: ReadonlyRest,
```
coutoPL, line 38: let's rename it to MainConfigBasedReloadableEngine

core/src/main/scala/tech/beshu/ror/boot/engines/ImpersonatorsReloadableEngine.scala
```diff
@@ -24,20 +24,40 @@ import tech.beshu.ror.RequestId
...
+                                                  clock: Clock)
   extends BaseReloadableEngine(
     "test", boot, None, reloadInProgress, rorConfigurationIndex
   ) {
 
+  def currentTestConfig()
+                       (implicit requestId: RequestId): Task[TestConfig] = {
+    Task.delay {
+      currentEngineState match {
+        case EngineState.NotStartedYet(None) =>
```
coutoPL, line 48: maybe we can write it as follows:
`case EngineState.NotStartedYet(None) | EngineState.Stopped =>`

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestconfig/RRTestConfigActionHandler.scala
coutoPL: yes, it looks ok

es79x/src/main/scala/tech/beshu/ror/es/actions/rrtestconfig/RRTestConfigActionHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+        listener.onFailure(new Exception("TestConfig API is not available"))
+    }
+  }
+
+  private def handle(result: Either[Throwable, TestConfigResponse],
+                     listener: ActionListener[RRTestConfigResponse]): Unit = result match {
+    case Right(response) =>
+      listener.onResponse(new RRTestConfigResponse(response))
+    case Left(ex) =>
+      logger.error("RRTestConfig internal error", ex)
```
coutoPL, line 54: we have in context requestId. Let's add it to the log like this:
`logger.error(s"[${requestId.show}] ...")`

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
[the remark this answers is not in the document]
coutoPL: oh, sorry. It's ok

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
```diff
@@ -26,8 +26,10 @@ import tech.beshu.ror.utils.elasticsearch.{DocumentManager, IndexManager, RorApi
 import tech.beshu.ror.utils.httpclient.RestClient
 import tech.beshu.ror.utils.misc.Resources.getResourceContent
 
+import java.time.format.DateTimeFormatter
 import scala.concurrent.duration._
 import scala.language.postfixOps
+import scala.util.Try
 
 trait BaseAdminApiSuite
```
coutoPL, line 34: are we able to add tests to see if we return 400 when the request body is malformed?

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestsettings/RRTestSettingsResponse.scala
coutoPL: but I'm not sure if we do return 400 for the malformed body (the exception case looks good)

core/src/main/scala/tech/beshu/ror/boot/RorInstance.scala
mateuszkp96: It's already in RorInstance companion object

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
mateuszkp96: image

es714x/src/main/scala/tech/beshu/ror/es/actions/rrtestsettings/RRTestSettingsResponse.scala
mateuszkp96: Status 400 is now returned

coutoPL (changes requested):
ok, everything is fine, but I want to ask you to refactor `ConfigApi` class to be more like `TestConfigApi` which has:
* better error handling (500 & 400)
* better error messages 
* better response models

PR #781, opened 2022-02-19
[RORDEV-515] fips compliant ssl

pielas:


core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+        keystore.load(keystoreFile, sslConfiguration.keystorePassword)
+        keystore
+      }
+      }
+  }
+
+  private def loadTruststore(sslConfiguration: SslConfiguration, fipsCompliant: Boolean) = {
+    sslConfiguration.truststoreFile
+      .map { truststoreFileName =>
+        Resource
```
coutoPL, line 73: similar code to the above. Can we extract it and reuse?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  private def prepareAlias(keystore: KeyStore, config: SslConfiguration) =
+    config.keyAlias match {
+      case None if keystore.aliases().hasMoreElements =>
+        val firstAlias = keystore.aliases().nextElement()
+        logger.info(s"ROR SSL: ssl.key_alias not configured, took first alias in keystore: $firstAlias")
+        firstAlias
+      case None =>
+        throw MalformedSslSettings("No alias found, therefore key found in keystore!")
```
coutoPL, line 98: hmm, are you sure the message is ok?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+            }
+          }
+      }.sequence
+  }
+
+  private def prepareAlias(keystore: KeyStore, config: SslConfiguration) =
+    config.keyAlias match {
+      case None if keystore.aliases().hasMoreElements =>
+        val firstAlias = keystore.aliases().nextElement()
+        logger.info(s"ROR SSL: ssl.key_alias not configured, took first alias in keystore: $firstAlias")
```
coutoPL, line 95: warn level?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+        firstAlias
+      case None =>
+        throw MalformedSslSettings("No alias found, therefore key found in keystore!")
+      case Some(keyAlias) =>
+        keyAlias.value
+    }
+
+  private def removeAllAliasesFromKeystoreBesidesOne(keystore: KeyStore, alias: String): Unit = {
```
coutoPL, line 103: shouldn't we warn the admin that this is being done?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+        _.map { truststore =>
+          val tmf = getTrustManagerFactoryInstance(fipsCompliant)
+          tmf.init(truststore)
+          tmf
+        }
+      }
+      .unsafeRunSync()) match {
+      case Success(kmf) => kmf.getOrElse(throw TrustManagerNotConfiguredException)
+      case Failure(exception) =>
+        throw UnableToInitializeTrustManagerFactoryUsingProvidedTruststore(exception.getMessage)
```
coutoPL, line 120: shouldn't we pass the whole cause exception, not just a message?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+        kmf
+      }
+  }
+
+  def prepareSSLContext(sslConfiguration: SslConfiguration, fipsCompliant: Boolean): SslContext = {
+    Try(SSLCertHelper
+      .getKeyManagerFactory(sslConfiguration, fipsCompliant)
+      .unsafeRunSync()) match {
+      case Success(keyManagerFactory) =>
+        val sslCtxBuilder = SslContextBuilder.forServer(keyManagerFactory)
```
coutoPL, line 139: what about moving it to the map method and calling unsafeRunSync at the end of monadic chain?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -0,0 +1,195 @@
...
+        }
+        sslCtxBuilder.build()
+      case Failure(exception: IOException) =>
+        throw UnableToLoadDataFromProvidedKeystoreException(sslConfiguration.keystoreFile, exception.getMessage)
+      case Failure(exception) =>
+        throw UnableToInitializeKeyManagerFactoryUsingProvidedKeystore(exception.getMessage)
+    }
+  }
+
+  def validateProtocolAndCiphers(sslContextBuilder: SslContextBuilder, config: SslConfiguration): Boolean =
```
coutoPL, line 157: maybe we can change the name to `areProtocolAndCiphersValid`?

es60x/src/main/scala/tech/beshu/ror/es/ReadonlyRestPlugin.scala
```diff
@@ -125,11 +130,15 @@ class ReadonlyRestPlugin(s: Settings, p: Path)
                                  xContentRegistry: NamedXContentRegistry,
                                  networkService: NetworkService,
                                  dispatcher: HttpServerTransport.Dispatcher): util.Map[String, Supplier[HttpServerTransport]] = {
+    val fipsCompliant = fipsConfig.fipsMode match {
```
coutoPL, line 133: maybe we can wrap the code into IsFibsCompliant typeclass and place next to FipsConfiguration class?

es61x/src/main/scala/tech/beshu/ror/es/ReadonlyRestPlugin.scala
```diff
@@ -116,6 +121,8 @@ class ReadonlyRestPlugin(s: Settings, p: Path)
 
   private var ilaf: IndexLevelActionFilter = _
 
+  SecurityProviderConfiguratorForFips.configureIfRequired(fipsConfig)
```
coutoPL, line 124: why we don't have to do this in the context of es60x?

es61x/src/main/scala/tech/beshu/ror/es/ReadonlyRestPlugin.scala
```diff
@@ -107,6 +108,10 @@ class ReadonlyRestPlugin(s: Settings, p: Path)
     .load(environment.configFile)
     .map(_.fold(e => throw new ElasticsearchException(e.message), identity))
     .runSyncUnsafe(timeout)(Scheduler.global, CanBlock.permit)
+  private val fipsConfig = FipsConfiguration
```
coutoPL, line 111: maybe we can create a helper to run it like this:
```
RorSsl
    .load(environment.configFile)
    .getOrThrowElasticsearchException
```

es61x/src/main/scala/tech/beshu/ror/es/ssl/SSLNetty4InternodeServerTransport.java
```diff
@@ -33,34 +32,33 @@
...
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
 public class SSLNetty4InternodeServerTransport extends Netty4Transport {
 
   private final Logger logger = LogManager.getLogger(this.getClass());
   private final InternodeSslConfiguration ssl;
+  private final Boolean fipsCompliant;
```
coutoPL, line 49: I wonder if we shouldn't add the flag to InternodeSslConfiguration? (the same applies to SslConfiguration). WDYT?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ClusterStateWithInternodeSslSuite.scala
```diff
@@ -66,7 +66,7 @@ trait ClusterStateWithInternodeSslSuite
           configHotReloadingEnabled = true,
           enableXPackSsl = true
         )
-      )
+    )
```
coutoPL, line 69: formatting

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FipsSslSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+  extends AnyWordSpec
+    with BaseEsClusterIntegrationTest
+    with ESVersionSupportForAnyWordSpecLike
+    with SingleClientSupport
+    with BeforeAndAfterAll {
+  this: EsContainerCreator =>
+
+  override implicit val rorConfigFileName = "/fips_ssl/readonlyrest.yml"
+
+
```
coutoPL, line 37: redundant `\n`s (here and below)

coutoPL (changes requested):
@pielas did you solve the problem of running FIBS SSL compliant code with java 11?

pielas:
> @pielas did you solve the problem of running FIBS SSL compliant code with java 11?

Kind of, If the library is included into the project it works with any java version

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
pielas: I've copy pasted this message from previous implementation. It was indeed wrong. Fixed it

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
pielas: Hmm I don't think it would be good idea to set this log message to warn because probably it's typical situation that alias is not configured

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
pielas: This method doesn't change any file on the disk. It's just modifying keystore in memory. I'm removing all unneeded keys to keep backward compatibility with previous implementation.

es61x/src/main/scala/tech/beshu/ror/es/ReadonlyRestPlugin.scala
pielas: We could, but currently we are throwing `ElasticsearchException` which is not accessible from the core and creating this helper in each module probably won't make sense. It would be easy to make such helper if we could throw some non elastic exception

es61x/src/main/scala/tech/beshu/ror/es/ssl/SSLNetty4InternodeServerTransport.java
pielas: I wanted to make `FipsCompliant` setting part of `SslConfiguration`. However there are plans to make ROR fully compatible with FIPS. Fully compliant FIPS would require SSL FIPS + some additional things. In that case I thought it would be probably better to have general configuration parameter for FIPS mode.

sscarduzio:
Guys this pr has been going on for months, what's the blocker here?

pielas:
@sscarduzio It was mainly delayed because I've encountered some issues with elastic limiting allowed privileges which came out during working on this PR. I found a way to overcome that and this PR is on it's really way to be closed. All tests should be passing now

coutoPL (approved):
Thanks! @pielas it looks like we need some documentation of these changes, right?

PR #784, opened 2022-02-24
[RORDEV-619] ES 8.0.0 support

sscarduzio:
🚀New (ES) New Support for 8.0.0, 8.0.1


sscarduzio:
OK it seems it's uploading finally :)

coutoPL:
> OK it seems it's uploading finally :)

yes, but I treat it as a workaround for now. Integration with ES 8.0.0 is the most important part of the task now. As a long-term solution, it have to be improved

PR #800, opened 2022-03-25
[RORDEV-644] LDAP groups_and

sscarduzio:
🚀New (ES) Added `groups_and` sub-rule for `ldap_auth`, and `ldap_authorization`.


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
```diff
@@ -75,6 +75,34 @@ object LdapAuthorizationRule {
...
 
   final case class Settings(ldap: LdapAuthorizationService,
-                            permittedGroups: UniqueNonEmptyList[Group],
+                            permittedGroups: GroupsLogic,
                             allLdapGroups: UniqueNonEmptyList[Group])
+
+  sealed trait GroupsLogic {
+    def groups: UniqueNonEmptyList[Group]
+
+    def availableGroupsFrom(userGroups: Set[Group], ruleGroups: Set[Group]): Option[UniqueNonEmptyList[Group]]
```
coutoPL, line 84: IMO it's better not to add methods to ADT that won't be overridden as val (see the example above). IMO it's better to create a typeclass (or extension method class). Eg.

```
implicit class GroupsLogicExecutor(val groupsLogic: GroupsLogic) extends AnyVal {
  def availableGroupsFrom(userGroups: Set[Group], ruleGroups: Set[Group]): Option[UniqueNonEmptyList[Group]] = {
    groupsLogic match {
      case And(_) => [do sth]
      case Or(_) => [do sth different]
    }
  }
}
```

then you will be able to use it in the same way:
```
val logic: GroupsLogic = ???
logic.availableGroupsFrom(...)
```

Btw. does the `availableGroupsFrom` needs two parameters? You have permitted groups in the GroupsLogic instance

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
```diff
@@ -32,40 +33,26 @@ import tech.beshu.ror.accesscontrol.domain.User.Id.UserIdCaseMappingEquality
...
-        mock
-          .users
-          .find(_.id === user)
-          .map(m => Groups.Present(UniqueList.of(m.groups.toSeq: _*)))
-          .getOrElse(Groups.Present(UniqueList.empty))
-      }
-      .getOrElse {
-        Groups.CannotCheck
-      }
+  override protected def mockedGroupsOf(user: User.Id, mocksProvider: MocksProvider)(implicit requestId: RequestId, eq: Eq[User.Id]): Groups = {
```
coutoPL, line 46: IMO the previous formatting makes the code a way easier to read

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
```diff
@@ -32,40 +33,26 @@ import tech.beshu.ror.accesscontrol.domain.User.Id.UserIdCaseMappingEquality
 import tech.beshu.ror.accesscontrol.domain.{Group, LoggedUser, User}
 import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
 
-class LdapAuthorizationRule(val settings: Settings,
-                            override val impersonation: Impersonation,
-                            override val caseMappingEquality: UserIdCaseMappingEquality)
-  extends BaseAuthorizationRule {
+class LdapAuthorizationRule(val settings: Settings, override val impersonation: Impersonation, override val caseMappingEquality: UserIdCaseMappingEquality) extends BaseAuthorizationRule {
```
coutoPL, line 36: the same here

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/base/BaseAuthorizationRule.scala
```diff
@@ -33,7 +33,7 @@ import tech.beshu.ror.accesscontrol.domain.{Group, LoggedUser, User}
 import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
 
 trait BaseAuthorizationRule extends AuthorizationRule with SimpleAuthorizationImpersonationSupport {
-
+  protected def calculateAllowedGroupsForUser(usersGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]]
```
coutoPL, line 36: in the `BaseGroupsRule` class we have a similar method:

```
protected def availableGroupsFrom(userGroups: Set[Group], ruleGroups: Set[Group]): Option[UniqueNonEmptyList[Group]]```

maybe it'd be nice to do it here in the same way?
1. it looks it's easier to reason about the method when it has two parameters
2. do you think the current name is better than the one from BaseGroupsRule? if so, maybe we should modify the BaseGroupsRule one? If not, let's make this one similar to the one from BaseGroupsRule
3. It'd ve nice to add `\n`s before and after the method declaration

File: core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala [hunk 5]
```diff
@@ -110,30 +112,40 @@ class LdapAuthorizationRuleDecoder(ldapDefinitions: Definitions[LdapService],
 }
 
 object LdapAuthorizationRuleDecoder {
+  def ERROR_MSG_NO_GROUPS_LIST(name: String) = s"Please specify one between 'groups' or 'groups_and' for LDAP authorization rule '${name}'"
```
coutoPL, line 115: 1. we don't use the Screaming Snake Case convention in our codebase (in scala in general too)
2. IMO, it's not a good practice to externalize messages like this only because we'd like to reuse them in tests. In tests, the strings/messages should be hardcoded

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
```diff
@@ -110,30 +112,40 @@ class LdapAuthorizationRuleDecoder(ldapDefinitions: Definitions[LdapService],
...
+
+  def ERROR_MSG_ONLY_ONE_GROUPS_LIST(name: String) = s"Please specify either 'groups' or 'groups_and' for LDAP authorization rule '${name}'"
 
   private def settingsDecoder(ldapDefinitions: Definitions[LdapService]): Decoder[LdapAuthorizationRule.Settings] =
     Decoder
       .instance { c =>
         for {
           name <- c.downField("name").as[LdapService.Name]
-          groups <- c.downField("groups").as[UniqueNonEmptyList[Group]]
+          groups_and <- c.downField("groups_and").as[Option[UniqueNonEmptyList[Group]]].map(_.map(GroupsLogic.And))
```
coutoPL, line 124: please use the `camelCase` convention instead for `val`s names

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
```diff
@@ -110,30 +112,40 @@ class LdapAuthorizationRuleDecoder(ldapDefinitions: Definitions[LdapService],
...
+          groups_or <- c.downField("groups").as[Option[UniqueNonEmptyList[Group]]].map(_.map(GroupsLogic.Or))
```
coutoPL, line 125: the same here

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
```diff
@@ -160,31 +172,34 @@ object LdapAuthRuleDecoder {
       .instance { c =>
         for {
           name <- c.downField("name").as[LdapService.Name]
-          groups <- c.downField("groups").as[UniqueNonEmptyList[Group]]
+          groups_and <- c.downField("groups_and").as[Option[UniqueNonEmptyList[Group]]].map(_.map(GroupsLogic.And))
```
coutoPL, line 175: all the same comments, I mentioned above, apply also here

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/base/BaseAuthorizationRule.scala
coutoPL: what about point 1?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
```diff
@@ -74,7 +74,28 @@ object LdapAuthorizationRule {
...
+
+    final case class And(override val groups: UniqueNonEmptyList[Group]) extends GroupsLogic
+  }
+
+  implicit class GroupsLogicExecutor(val groupsLogic: GroupsLogic) extends AnyVal {
+    def availableGroupsFrom(userGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]] = {
+      groupsLogic match {
+        case And(groups) =>
+          val intersection = userGroups intersect groups
+          if (intersection == groups.toSet) Some(groups) else None
```
coutoPL, line 94: the codacy warn is right - you compare two objects with different types. It's rather dangerous. You can `import cats.implicits._` and use `===` instead of `==` (compiler will fail here)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
```diff
@@ -74,7 +74,28 @@ object LdapAuthorizationRule {
     override val name = Rule.Name("ldap_authorization")
   }
 
-  final case class Settings(ldap: LdapAuthorizationService,
-                            permittedGroups: UniqueNonEmptyList[Group],
-                            allLdapGroups: UniqueNonEmptyList[Group])
+  final case class Settings(ldap: LdapAuthorizationService, permittedGroups: GroupsLogic, allLdapGroups: UniqueNonEmptyList[Group])
```
coutoPL, line 77: you add this to LdapAuthorizationRule. What about other authorization rules? I understand that this is not needed by the client now, but IMO we have to create a jira for `groups_and` support in other rules and add it to the backlog.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
coutoPL: `groups_and` => `groupsAnd`

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
```diff
@@ -160,50 +206,30 @@ object LdapAuthRuleDecoder {
...
+          groups_or <- c.downField("groups").as[Option[UniqueNonEmptyList[Group]]].map(_.map(GroupsLogic.Or))
```
coutoPL, line 210: `groups_or` => `groupsOr`

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/LdapAuthRuleSettingsTests.scala
```diff
@@ -18,13 +18,17 @@ package tech.beshu.ror.unit.acl.factory.decoders
...
 import tech.beshu.ror.accesscontrol.blocks.rules.LdapAuthRule
-import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.Reason.{MalformedValue, Message}
+import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.Reason.Message
 import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.RulesLevelCreationError
 import tech.beshu.ror.utils.SingletonLdapContainers
 
 class LdapAuthRuleSettingsTests
   extends BaseRuleSettingsDecoderTest[LdapAuthRule] {
 
+  def ERROR_MSG_NO_GROUPS_LIST(name: String) = s"Please specify one between 'groups' or 'groups_and' for LDAP authorization rule '${name}'"
```
coutoPL, line 28: please just hardcode the strings in the proper places. IMO the code brings nothing that can help

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/LdapAuthorizationRuleSettingsTests.scala
```diff
@@ -20,15 +20,21 @@ import eu.timepit.refined.auto._
...
-import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.Reason.{MalformedValue, Message}
+import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.Reason.Message
 import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.AclCreationError.RulesLevelCreationError
 import tech.beshu.ror.utils.SingletonLdapContainers
 import tech.beshu.ror.utils.uniquelist.UniqueNonEmptyList
 
 class LdapAuthorizationRuleSettingsTests
   extends BaseRuleSettingsDecoderTest[LdapAuthorizationRule] {
 
+  def ERROR_MSG_NO_GROUPS_LIST(name: String) = s"Please specify one between 'groups' or 'groups_and' for LDAP authorization rule '${name}'"
```
coutoPL, line 34: the same here

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/LdapAuthorizationRuleSettingsTests.scala
```diff
@@ -20,15 +20,21 @@ import eu.timepit.refined.auto._
...
+
+  def ERROR_MSG_ONLY_ONE_GROUPS_LIST(name: String) = s"Please specify either 'groups' or 'groups_and' for LDAP authorization rule '${name}'"
```
coutoPL, line 36: the same here

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
coutoPL: no commented, not fixed

coutoPL:
and please:
1. prepare PR with doc changes
2. bump pre-version
3. merge develop
4. add a changelog style description to this PR

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/LdapRulesDecoders.scala
sscarduzio: oh shoot. I forgot this!

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
sscarduzio: I tried and failed. The === won't pass the tests. Plus, `Intersection` and `groups.toSet` are the same type `Set[Group]`, so == should be ok, right?  🤔

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
sscarduzio: The only one implementor of `BaseAuthorizationRule` is `ExternalAuthorizationRule` (not entirely sure why! I.e. JWT and ror_kbn_auth could be good candidate to join the band for example).

Will add the task for ExternalAuthorizationRule for now.

sscarduzio:
DOCS https://github.com/beshu-tech/readonlyrest-docs/pull/111

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/LdapAuthorizationRule.scala
coutoPL: the answer is rather simple - `LdapAuthorizationRule` and `ExternalAuthorizationRule` are authorization rules only. `JwtAuthRule` and `RorKbnAuthRule` are authorization and authentication rules at the same time (aka "auth rules"). Usage of the common base is not so obvious.

But the JWT based rules have a "groups" field in the configuration. So, they should support "groups_and" too. Will you add the proper task for them too?

coutoPL (approved):
LGTM

PR #801, opened 2022-03-27
[RORDEV-606] Impersonation supported only by Test Engine

mateuszkp96:


es60x/src/main/scala/tech/beshu/ror/es/handler/AclAwareRequestFilter.scala
```diff
@@ -86,7 +87,16 @@ class AclAwareRequestFilter(clusterService: RorClusterService,
 
   def handle(engines: Engines,
              esContext: EsContext): Task[Unit] = {
-    val engine = esContext.pickEngineToHandle(engines)
+    esContext.pickEngineToHandleOrForbid(engines) match {
+      case Right(engine) =>
+        handleRequestWithEngine(engine, esContext)
+      case Left(forbiddenResponse) =>
+        Task.delay(esContext.listener.onFailure(forbiddenResponse))
```
coutoPL, line 94: In this class, we do not produce an ES response. Maybe it'd be better to produce it in IndexLevelActionFilter when the `handle` method is called? It could return `Task[Either[Error, Unit]]` and we could map the error (currently it'd be one instance) to `createTestSettingsNotConfiguredResponse` and call `onFailure`. WDYT?

es60x/src/main/scala/tech/beshu/ror/es/handler/response/ForbiddenResponse.scala
```diff
@@ -74,6 +75,10 @@ object ForbiddenResponse {
   def createRorNotEnabledResponse(): ForbiddenResponse =
     new ForbiddenResponse(None, RorNotEnabled :: Nil)
 
+  def createTestSettingsNotConfiguredResponse(): ForbiddenResponse = {
```
coutoPL, line 78: let's remove `{}` (other similar methods in this class don't use them)

coutoPL (changes requested):

coutoPL (approved):
LGTM. Thanks

PR #803, opened 2022-04-04
RORDEV-650 (WIP) Numeric string(s) array as indices rules argument won't pass decoder validation

sscarduzio:
Only tests added for now.

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
@@ -17,11 +17,11 @@
 package tech.beshu.ror.accesscontrol.factory
 
 import cats.data.NonEmptyList
-import eu.timepit.refined.types.string.NonEmptyString
-import io.circe.Json
 import cats.implicits._
-import tech.beshu.ror.accesscontrol.show.logs._
+import eu.timepit.refined.types.string.NonEmptyString
+import io.circe.{Encoder, Json}
 import tech.beshu.ror.accesscontrol.blocks.variables.startup.StartupResolvableVariableCreator.{createMultiVariableFrom, createSingleVariableFrom}
+import tech.beshu.ror.accesscontrol.show.logs._
 import tech.beshu.ror.providers.EnvVarsProvider
 
 object JsonConfigStaticVariableResolver {
@@ -43,7 +43,7 @@ object JsonConfigStaticVariableResolver {
         json.asString.flatMap(NonEmptyString.unapply) match {
           case Some(str) =>
             tryToResolveAllStaticMultipleVars(str, errors)
-              .map(resolvedStringToJson)
+              .map(s => resolvedStringToJson(s, json))
               .toList
           case None =>
             mapJson(json, errors) :: Nil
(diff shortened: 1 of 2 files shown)


sscarduzio:
I'm a bit stuck with this.  This is what I investigated so far:

- The Yaml parser seem to behave well, as you can see from the re-converted JSON.

image

- Following the debugger, the failure happens in the `IndicesRuleDecoder`
 
image

But I tragically get lost in the implicit typelevel stuff and cannot pinpoint what part of the decoding goes bad. Any advice?

coutoPL:
I see that the problem is somewhere in the `JsonConfigStaticVariableResolver` class. Check `RawRorConfigBasedCoreFactory#createCoreFromRorSection` method - in `rorSection` object we have a list of strings. In `resolvedRorSection` there is a list of numbers.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
```diff
@@ -54,17 +54,31 @@ object JsonConfigStaticVariableResolver {
...
   }
 
-  private def resolvedStringToJson(resolvedStr: String) = {
+  private def preserveNumbersAsStrings(original: Json, newValue: Json) = {
+    {
+      for {
+        _ <- original.asString
+        newNum <- newValue.asNumber
+      } yield Encoder.encodeString(newNum.toString)
+    }.getOrElse(newValue)
```
coutoPL, line 73: is it safe to "catch all exceptions" and return the `newValue` value? Maybe we should catch only a specific exception?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
```diff
@@ -54,17 +54,31 @@ object JsonConfigStaticVariableResolver {
...
+          resolvedStringToJson(resolved, json)
+        else json
       }
   }
 
-  private def resolvedStringToJson(resolvedStr: String) = {
+  private def preserveNumbersAsStrings(original: Json, newValue: Json) = {
+    {
+      for {
+        _ <- original.asString
```
coutoPL, line 70: I don't know if I understand what is going on here. We try to get a string from JSON and ignore the result and then try to get a number from the new value and encode it as a string or if all of these fail we return the newValue. It's rather hacky, so we should do one of the following:
1. write more declarative or at least descriptive function
2. explain in comment the intention of the hack

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
```diff
@@ -54,17 +54,31 @@ object JsonConfigStaticVariableResolver {
...
         }
-        resolvedStringToJson(resolved)
+        if (resolved =!= str)
+          resolvedStringToJson(resolved, json)
+        else json
       }
   }
 
-  private def resolvedStringToJson(resolvedStr: String) = {
+  private def preserveNumbersAsStrings(original: Json, newValue: Json) = {
```
coutoPL, line 67: the method is used only by the `resolvedStringToJson` method, so it should be defined after it (for a sake of readability)

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
sscarduzio: Wait, asString and asNumber return Option, so either they "manage" to parse as the respective type, or they return None. What is there for us to  catch here? 🤔

coutoPL: Ok, my bad

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/JsonConfigStaticVariableResolver.scala
sscarduzio: Nevermind, I played type tetris a bit too long when I did this and I was brain dead. I simplified the algorithm now and it should be evident what's happening.

coutoPL (approved):
LGTM

coutoPL:
@sscarduzio could you please add a changelog style description to the PR and paste the link to the PR in the corresponding jira?

PR #826, opened 2022-06-21
[RORDEV-594] PEM certificates support

pielas:
🚀New (ES) Added support for certificates in PEM format


sscarduzio:
@pielas what's the status of this?

pielas:
@sscarduzio it's generally ready. I just need to fix issues with tests on pipeline

coutoPL:
@pielas sth is wrong with es60x module:

```
2022-07-13T19:57:00.5198056Z [2022-07-13T19:31:19,069][ERROR][o.e.b.ElasticsearchUncaughtExceptionHandler] [ROR1_2] fatal error in thread [main], exiting
2022-07-13T19:57:00.5198318Z 
2022-07-13T19:57:00.5198603Z java.lang.NoClassDefFoundError: org/bouncycastle/openssl/PEMParser
2022-07-13T19:57:00.5198781Z 
2022-07-13T19:57:00.5199121Z 	at tech.beshu.ror.utils.SSLCertHelper$.$anonfun$loadPrivateKey$3(SSLCertHelper.scala:67) ~[?:?]
2022-07-13T19:57:00.5199395Z 
2022-07-13T19:57:00.5199745Z 	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87) ~[?:?]
```

We've excluded several intermediate ES modules from PR tests recently.
Please make sure that these modules pass the problematic test locally (on the develop all modules will be tested in the pipeline)

pielas:
@coutoPL ah, it turns out that BouncyCastle library, which I included to handle PEM files, causes JAR hell like other BouncyCastle libraries. Could we just exclude support for PEM files in es60?

coutoPL:
I think that there is no sense to fight with it (assuming that the jarhell is not easy to fix)

sscarduzio:
Parsing a pem file *requires* bouncy castle? 😳

pielas:
@sscarduzio No, it's not required. However parsing it without bouncy castle is more troublesome. Here's `loadPrivateKey` function without using BC https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/826/commits/efc9cde59904de6e021d70f300617244a45e747b#diff-1f6f852b3a8af9460e226a8940a31dd40d6b941674985610d0328e709292c3dd As you can see it requires to do few steps by hand and this code is able to load only RSA keys (because of `KeyFactory.getInstance("RSA")`). Handling other types is possible, but requires some more handling. Because such implementation is more error prone I decided to use BouncyCastle lib which takes care of everything

sscarduzio:
OK, I understand. It's ok if you want to skip 6.x

sscarduzio:
@pielas not much is left to do on this task, but it is running really late. Can we please wrap this up and close it in Jira ASAP?

pielas:
@sscarduzio I will close it as soon as possible. However I haven't felt well in recent days. I will try to wrap this up in the following days

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -51,6 +60,29 @@ object SSLCertHelper extends Logging {
     }
   }
 
+  private def loadPrivateKey(file: File): IO[PrivateKey] = {
```
coutoPL, line 63: can we move the private method below the method where it's used for the first time?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -51,6 +60,29 @@ object SSLCertHelper extends Logging {
...
+      .fromAutoCloseable(IO(new FileReader(file)))
+      .use { privateKeyFileReader => IO {
+        val pemParser = new PEMParser(privateKeyFileReader)
+        val privateKeyInfo = PrivateKeyInfo.getInstance(pemParser.readObject())
+        val converter = new JcaPEMKeyConverter()
+        converter.getPrivateKey(privateKeyInfo)
+      }}
+  }
+
+  private def loadCertificateChain(file: File): IO[Array[X509Certificate]] = {
```
coutoPL, line 74: the same here

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -51,6 +60,29 @@ object SSLCertHelper extends Logging {
...
+    Resource
```
coutoPL, line 75: `better.files` lib has sth similar:

```
import better.files._
  val file: File = ???
  file.fileInputStream { fileInputStream => ... }
```
maybe we should switch?

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
```diff
@@ -145,18 +181,34 @@ object SSLCertHelper extends Logging {
...
-  def getTrustManagerFactory(sslConfiguration: SslConfiguration, fipsCompliant: Boolean): TrustManagerFactory = {
-    loadTruststore(sslConfiguration, fipsCompliant)
+  def isPEMHandlingAvailable: Boolean = {
+    Try {
+      Class.forName("org.bouncycastle.openssl.PEMParser")
+    }
+      .isSuccess
+  }
+
+  def getTrustedCertificatesFromPemFile(fileBasedConfiguration: ClientCertificateConfiguration.FileBasedConfiguration): Array[X509Certificate] = {
```
coutoPL, line 191: let's move public methods at the beginning of the file

core/src/test/scala/tech/beshu/ror/unit/configuration/SslConfigurationTest.scala
```diff
@@ -20,7 +20,7 @@ import monix.execution.Scheduler.Implicits.global
 import org.scalatest.Inside
 import org.scalatest.matchers.should.Matchers._
 import org.scalatest.wordspec.AnyWordSpec
-import tech.beshu.ror.configuration.SslConfiguration._
+import tech.beshu.ror.configuration.SslConfiguration.{ClientCertificateConfiguration, _}
```
coutoPL, line 23: the previous version was better

integration-tests/src/test/resources/xpack_cluster_with_ror_nodes_and_internode_ssl/readonlyrest.yml
```diff
@@ -1,11 +1,9 @@
 readonlyrest:
```
coutoPL, line 1: do we need "testkeystore.bcfks" resource? I don't see its usage

integration-tests/src/test/resources/xpack_cluster_with_ror_nodes_and_internode_ssl/readonlyrest.yml
```diff
@@ -1,11 +1,9 @@
...
   ssl:
     enable: true
-    keystore_file: "elastic-certificates.p12"
-    keystore_pass: readonlyrest
-    truststore_file: "elastic-certificates.p12"
-    truststore_pass: readonlyrest
-    key_pass: readonlyrest
+    server_certificate_key_file: elastic-certificates-pkey.pem
```
coutoPL, line 4: maybe we should run XpackClusterWithRorNodesAndInternodeSslSuite twice? I mean, it'd be good to test it with p12 and pem. So, maybe we can have two configs (for modules > es60x obviously) and reuse the suite to create two plugin tests? WDYT?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/utils/SSLCertHelper.scala
coutoPL: ok

integration-tests/src/test/resources/xpack_cluster_with_ror_nodes_and_internode_ssl/readonlyrest.yml
pielas: sounds good. I wanted to test both functionalities by using pem files for server and keystore for internode, but this solution seems better

coutoPL (approved):
LGTM! Thanks

PR #827, opened 2022-06-24
[RORDEV-657] fixed ROR's cooperation with xpack nodes

coutoPL:
🐞Fix (ES) ROR node cooperation with X-pack nodes


sscarduzio:
Can you document/summarise what changed here? What are we able to do now that before this PR we could not?

coutoPL:
@sscarduzio There was a problem with creating index patterns in cluster which has eg. one ROR node and other x-pack nodes. Moreover, I figured out that previous solution (there was 2 attempts to fix it) didn't work when ROR node acted as coordination node (when it was a master or data node, the problem didn't exist). To fix it I had to sniff the communication between xpack nodes and fake the headers they are passing around.

sscarduzio:
Cool, thanks for the update 👍

PR #833, opened 2022-08-03
RORDEV 493 fail if upload (or git tag) fails

sscarduzio:


ci/ci-lib.sh
```diff
@@ -2,28 +2,52 @@
...
+    fi
+    echo "Deleting tag $GIT_TAG"
+    if [[ "$(uname -s)" == *"Linux"* ]]; then
+        git tag -d $GIT_TAG && git push origin :refs/tags/$GIT_TAG || (echo "Failed to delete tag $GIT_TAG" && exit 1)
+        return 0
     fi
     return 0
 }
 
+file . || (echo "FATAL: file command does not work or is not installed!" && exit 1)
```
coutoPL, line 50: do we need it?

ci/ci-lib.sh
```diff
@@ -2,28 +2,52 @@
...
-        git config --global user.name "Travis CI"
-        git tag $GIT_TAG -a -m "Generated tag from TravisCI build $TRAVIS_BUILD_NUMBER"
-        git push origin $GIT_TAG
+        git tag $GIT_TAG -a -m "Generated tag from TravisCI build $TRAVIS_BUILD_NUMBER" &&
+        git push origin $GIT_TAG || exit 1
+    fi
+    return 1
+}
+
+function __tag_exists {
```
coutoPL, line 27: do we need __ prefix?

ci/ci-lib.sh
```diff
@@ -2,28 +2,52 @@
...
     echo "Tagging as $GIT_TAG"
     if [[ "$(uname -s)" == *"Linux"* ]]; then
-        git remote set-url origin git@github.com:sscarduzio/elasticsearch-readonlyrest-plugin.git
-        git config --global push.default matching
-        git config --global user.email "builds@travis-ci.com"
-        git config --global user.name "Travis CI"
-        git tag $GIT_TAG -a -m "Generated tag from TravisCI build $TRAVIS_BUILD_NUMBER"
-        git push origin $GIT_TAG
+        git tag $GIT_TAG -a -m "Generated tag from TravisCI build $TRAVIS_BUILD_NUMBER" &&
+        git push origin $GIT_TAG || exit 1
```
coutoPL, line 22: no error log on fail?

coutoPL (changes requested):

ci/ci-lib.sh
sscarduzio: it's a way to say "private" in a language where there's no private methods, reminiscent from Python.

ci/ci-lib.sh
sscarduzio: This is checking a dependency for ci-uploader.sh, should belong to that file, indeed. 👍

coutoPL:
does it mean this is not ready yet?

sscarduzio:
What I did so far is functional to have this working (making the git command available), yet, something else is on the way: at the moment, Github does not get the right credentials for tagging

```
Host key verification failed.
fatal: Could not read from remote repository.
```

coutoPL:
can we mark the PR as a draft?

sscarduzio:
Yeah this thing didn't work. Will need to drastically simplify this whole dumpster fire and start over

PR #835, opened 2022-08-08
[RORDEV-588] Distributed and persistent Test configuration

mateuszkp96:
🧐Enhancement (ES)  added distributed and persistent Test configuration in the Impersonation Feature


core/src/main/scala/tech/beshu/ror/api/TestConfigApi.scala
```diff
@@ -126,19 +133,21 @@ class TestConfigApi(rorInstance: RorInstance)
...
     EitherT(
       rorInstance
         .forceReloadTestConfigEngine(config, ttl)
         .map {
           _.leftMap {
-            case RawConfigReloadError.ReloadingFailed(failure) =>
-              TestConfigResponse.UpdateTestConfig.FailedResponse(s"Cannot reload new settings: ${failure.message}")
-            case RawConfigReloadError.ConfigUpToDate(_) =>
+            case IndexConfigSavingError(error) =>
+              TestConfigResponse.UpdateTestConfig.FailedResponse(s"Cannot save new settings: ${error.show}")
```
coutoPL, line 144: the method is to force reload the config, but we inform the api user that settings cannot be saved. Sth is wrong here

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -89,25 +101,61 @@ class ReadonlyRest(coreFactory: CoreFactory,
...
+    EitherT(action.foldMap(compiler))
+      .leftMap(toStartingFailure)
+  }
+
+  private def toStartingFailure(error: LoadedTestRorConfig.Error) = {
+    error match {
+      case LoadedTestRorConfig.IndexParsingError(message) =>
+        StartingFailure(message)
+      case LoadedTestRorConfig.IndexUnknownStructure =>
+        StartingFailure(s"Test settings index is malformed")
```
coutoPL, line 115: I'm not sure if we should not allow starting ROR when test settings are invalid. 

The test settings are used by an admin. It doesn't apply to users, so it should not affect them. 
Maybe we can just ignore it, write a big error in the log and continue? Do you see any problems with this approach?

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -89,25 +101,61 @@ class ReadonlyRest(coreFactory: CoreFactory,
...
+
   private def startRor(esConfig: EsConfig,
-                       loadedConfig: LoadedRorConfig[RawRorConfig]) = {
+                       loadedConfig: LoadedRorConfig[RawRorConfig],
+                       loadedTestRorConfig: LoadedTestRorConfig[TestRorConfig]) = {
     for {
-      engine <- EitherT(loadRorCore(loadedConfig.value, esConfig.rorIndex.index))
-      rorInstance <- createRorInstance(esConfig.rorIndex.index, engine, loadedConfig)
+      mainEngine <- EitherT(loadRorCore(loadedConfig.value, esConfig.rorIndex.index))
+      testEngine <- loadedTestRorConfig.value match {
```
coutoPL, line 124: let;s move all of these to a new, private method

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -89,25 +101,61 @@ class ReadonlyRest(coreFactory: CoreFactory,
...
+        case TestRorConfig.NotSet =>
+          EitherT.right(Task.now(TestEngine.NotConfigured))
+        case config@TestRorConfig.Present(rawConfig, expiration) if !config.hasExpired(clock)=>
```
coutoPL, line 127: isExpired?

core/src/main/scala/tech/beshu/ror/boot/RorInstance.scala
```diff
@@ -112,72 +122,94 @@ class RorInstance private(boot: ReadonlyRest,
...
+  }
+
+  private def scheduleTestConfigReload(interval: FiniteDuration Refined Positive) = {
+    scheduleIndexConfigChecking(interval, "test", tryTestEngineReload)
+  }
+
+  private def scheduleIndexConfigChecking(interval: FiniteDuration Refined Positive,
+                                          name: String,
+                                          reloadTask: RequestId => Task[Either[ScheduledReloadError, Unit]]): Cancelable = {
+    logger.debug(s"[CLUSTERWIDE SETTINGS] Scheduling next in-index $name settings check within ${interval.value}")
```
coutoPL, line 155: let's stop here. 

Currently, we have two engines and two scheduled tasks.

Did you consider one scheduled task that tries to reload eg. a list of engines? What approach is better and why? WDYT?

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
mateuszkp96: I created an appropriate fallback config for each failure case.

core/src/main/scala/tech/beshu/ror/boot/RorInstance.scala
mateuszkp96: Finally, I used a list of engines

coutoPL (changes requested):
+ as we agreed, please create an epic branch (RORDEV-721) and set it as a target for this PR

core/src/main/scala/tech/beshu/ror/configuration/index/Config.scala
```diff
@@ -14,9 +14,15 @@
  *    You should have received a copy of the GNU General Public License
  *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
  */
-package tech.beshu.ror.integration.proxy
+package tech.beshu.ror.configuration.index
 
-import tech.beshu.ror.integration.suites.AdminApiAuthMockSuite
-import tech.beshu.ror.integration.utils.SingleNodeProxyTestSupport
+private[index] object Config {
+  object auditIndexConst {
```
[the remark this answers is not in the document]
mateuszkp96: I copied it from the old implementation. Now the name should be fine

docs/api/ror-internal-api-swagger.yaml
```diff
@@ -2,7 +2,7 @@ openapi: 3.0.3
 info:
   title: ROR Internal API
   description: The API is designed for ROR KBN plugin. It should be treated as internal. It means that backward compatibility won't be a first-class thing for us.
-  version: 1.0.0
+  version: 1.1.0
```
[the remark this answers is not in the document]
mateuszkp96: 1. I will create a task for changes merged on the epic branch. It Is possible that more API changes are required.
2. You have right that we break compatibility here, so `2.0.0` will be the appropriate API version. Previously possible was only the success code, so we cannot improve this endpoint without breaking the compatibility.

mateuszkp96:
I set the target branch to develop to pass CI checks. Before a merge, I will change the target branch to the epic branch.

coutoPL (approved):
LGTM! Thanks

PR #841, opened 2022-08-31
[RORDEV-717] LDAP high load handling improvements

coutoPL:
🧐Enhancement (KBN) handling high load when LDAP rules are used


sscarduzio:
What's a one-liner high level description of this change?

PR #843, opened 2022-09-03
[RORDEV-587] Distributed and persistent Auth Mocks configuration

mateuszkp96:
🧐Enhancement (ES) added distributed and persistent Auth Mocks configuration in the Impersonation Feature


core/src/main/scala/tech/beshu/ror/configuration/RorIndexNameConfiguration.scala
```diff
@@ -40,7 +40,10 @@ object RorIndexNameConfiguration extends Logging {
   }
 
   def load(esConfig: File): Task[Either[MalformedSettings, RorIndexNameConfiguration]] = Task {
-    new EsConfigFileLoader[RorIndexNameConfiguration]().loadConfigFromFile(esConfig, "Custom ROR index name")
+    new EsConfigFileLoader[RorIndexNameConfiguration]().loadConfigFromFile(
```
coutoPL, line 43: is it ok? Last time we talked about it I had doubts if this is ok. We use the loadConfigFromFile method, but the description says it's related to "ROR index".

core/src/main/scala/tech/beshu/ror/configuration/TestRorConfig.scala
```diff
@@ -28,7 +29,8 @@ object TestRorConfig {
   case object NotSet extends TestRorConfig
 
   final case class Present(rawConfig: RawRorConfig,
-                           expiration: Present.ExpirationConfig) extends TestRorConfig {
+                           expiration: Present.ExpirationConfig,
```
coutoPL, line 32: let's more `expiration` param to the end of params list

core/src/test/scala/tech/beshu/ror/unit/boot/RorIndexTest.scala
```diff
@@ -0,0 +1,385 @@
...
+  private val testRorConfigDocumentId = "2"
+
+  "A ReadonlyREST core" should {
+    "load ROR index name" when {
+      "no index is defined in config" should {
+        "start ROR with default index name" in {
+          val resourcesPath = "/boot_tests/index_config/no_index_defined/"
+          val indexJsonContentService = mock[IndexJsonContentService]
+
+          (indexJsonContentService.sourceOf _)
```
coutoPL, line 67: can we create a private helper method for mocking sourceOf?

core/src/test/scala/tech/beshu/ror/unit/boot/RorIndexTest.scala
```diff
@@ -0,0 +1,385 @@
...
+          mockCoreFactory(coreFactory, rorConfig)
+
+          (indexJsonContentService.saveContent _)
+            .expects(fullIndexName(defaultRorIndexName), mainRorConfigDocumentId, Map("settings" -> rorConfig.raw))
+            .once()
+            .returns(Task.now(Right(())))
+
+          val rorInstance = result.right.value
+
+          rorInstance
```
coutoPL, line 126: let's assign the result to a well-named variable (eg. forceReloadingResult or sth like that)

docs/api/ror-internal-api-swagger.yaml
```diff
@@ -612,6 +614,18 @@ components:
           type: string
           description: Human readable message
 
+    AuthMockUpdateFailedResponse:
```
coutoPL, line 617: how does it affect the API version?

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -136,21 +136,28 @@ class ReadonlyRest(coreFactory: CoreFactory,
...
+                  expiration = expirationConfig(expiration)
+                )
+              case Left(startingFailure) =>
+                logger.error(s"Unable to start test engine. Cause: ${startingFailure.message}. Test settings engine will be marked as invalidated.")
+                TestEngine.Invalidated(rawConfig, expirationConfig(expiration))
+            }
+        } yield testEngine
+      case TestRorConfig.Present(rawConfig, expiration, mocks) =>
+        Task
+          .delay(authServicesMocksProvider.update(mocks))
```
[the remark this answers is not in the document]
mateuszkp96: AuthMocksProvider` should load mocks stored in the index to be consistent with the other nodes. 
Imagine that you save mocks on one node, test settings expire, and then the second node starts. The `AuthMocksProvider` on the first node has mocks stored in memory, but the second one does not, so we have inconsistency here.
Additionally, If the user wants to refresh the test settings after expiration, he will have previously configured mocks. That is the same behavior as with mocks stored in memory.

core/src/main/scala/tech/beshu/ror/configuration/RorIndexNameConfiguration.scala
mateuszkp96: This class loads EsConfig JSON from the ES config file and tries to decode the given configuration using provided decoder.
I know the naming used here is misleading, so I refactored it a little bit.

docs/api/ror-internal-api-swagger.yaml
mateuszkp96: In the previous MR, I increased the major, so in this MR API version change is not required

core/src/main/scala/tech/beshu/ror/configuration/EsConfigFileLoader.scala
```diff
@@ -20,16 +20,25 @@ import better.files.File
...
+                              (implicit envVarsProvider: EnvVarsProvider) {
 
-  def loadConfigFromFile(file: File,
-                         configName: String): Either[MalformedSettings, CONFIG] = {
+  def loadConfig[CONFIG: Decoder](configName: String): Either[MalformedSettings, CONFIG] = {
+    loadEsConfigJson
+      .flatMap { json =>
+        implicitly[Decoder[CONFIG]]
+          .decodeJson(json)
+          .left.map(_ => MalformedSettings(s"Invalid $configName configuration"))
```
coutoPL, line 37: OK, it looks like I understand it now. We are loading yaml file and then trying to decode it to `CONFIG`. 
I'd ask you about some more improvements here:
1. Let's change MalformedSettings's message to: "Cannot load $configName from file ${file.pathAsString}" (and make sure that all `configName`s make sense in this message context
2. let's rename EsConfigFileLoader to YamlFileBasedConfigLoader (and remove all "es" from names)
3. cache result of `loadEsConfigJson` (lazy val?)

coutoPL (changes requested):

coutoPL (approved):
Thanks!

PR #854, opened 2022-10-06
[RORDEV-720] resolving index names with wildcards should take into consideration the current index state and request indices options

coutoPL:
🐞Fix (ES) resolving index names with wildcards should take into consideration the current index state and request indices options


sscarduzio:
Let's put a description on the PR title (ideally would be  nice some words on the branch name too)

coutoPL:
I just wanted to run tests for the single module. Not ready yet

PR #856, opened 2022-10-22
[RORDEV-707] Create data stream fix, data streams API tests

mateuszkp96:
🐞Fix (ES) Data streams creation issue fix

### DIFF
--- a/es710x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/ReflectionBasedIndicesEsRequestContext.scala
+++ b/es710x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/ReflectionBasedIndicesEsRequestContext.scala
@@ -57,7 +57,7 @@ class ReflectionBasedIndicesEsRequestContext private(actionRequest: ActionReques
     // Optimistic reflection attempt
     ReflecUtils.setIndices(
       actionRequest,
-      Sets.newHashSet("index", "indices", "setIndex", "setIndices"),
+      Sets.newHashSet("index", "indices", "setIndex", "setIndices", "name", "getName"),
       indices.toList.map(_.stringify).toSet.asJava
     )
   }
@@ -75,6 +75,8 @@ object ReflectionBasedIndicesEsRequestContext {
       .orElse(getIndicesUsingReflection(request, methodName = "getIndices"))
       .orElse(getIndicesUsingReflection(request, methodName = "index"))
       .orElse(getIndicesUsingReflection(request, methodName = "getIndex"))
+      .orElse(getIndicesUsingReflection(request, methodName = "name"))
+      .orElse(getIndicesUsingReflection(request, methodName = "getName"))
       .map(indices => indices.toList.toSet.flatMap(ClusterIndexName.fromString))
   }
 
--- a/es710x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/ReflectionBasedIndicesEsRequestContext.scala
+++ b/es710x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/ReflectionBasedIndicesEsRequestContext.scala
@@ -57,7 +57,7 @@ class ReflectionBasedIndicesEsRequestContext private(actionRequest: ActionReques
     // Optimistic reflection attempt
     ReflecUtils.setIndices(
       actionRequest,
-      Sets.newHashSet("index", "indices", "setIndex", "setIndices", "name", "getName"),
+      Sets.newHashSet("index", "indices", "setIndex", "setIndices"),
       indices.toList.map(_.stringify).toSet.asJava
     )
   }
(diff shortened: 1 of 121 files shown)


es710x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/ReflectionBasedIndicesEsRequestContext.scala
```diff
@@ -57,7 +57,7 @@ class ReflectionBasedIndicesEsRequestContext private(actionRequest: ActionReques
     // Optimistic reflection attempt
     ReflecUtils.setIndices(
       actionRequest,
-      Sets.newHashSet("index", "indices", "setIndex", "setIndices"),
+      Sets.newHashSet("index", "indices", "setIndex", "setIndices", "name", "getName"),
```
coutoPL, line 60: we should handle all requests from the `org.elasticsearch.action.datastreams` package explicitly. They should be handled before this generic index-based request handling. Let's treat the name of the stream as the index name at the moment. It should be safe enough but it'll unblock the data stream API (currently clients cannot use it because of the bug)

coutoPL (changes requested):

es711x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/datastreams/ReflectionBasedDataStreamsEsRequestContext.scala
```diff
@@ -0,0 +1,57 @@
...
+  def unapply(arg: ReflectionBasedActionRequest): Option[BaseIndicesEsRequestContext[ActionRequest]] = {
+    CreateDataStreamEsRequestContext.unapply(arg)
+      .orElse(DataStreamsStatsEsRequestContext.unapply(arg))
+      .orElse(DeleteDataStreamEsRequestContext.unapply(arg))
+      .orElse(GetDataStreamEsRequestContext.unapply(arg))
+      .orElse(MigrateToDataStreamEsRequestContext.unapply(arg))
+      .orElse(PromoteDataStreamEsRequestContext.unapply(arg))
+  }
+
+  private[datastreams] def tryMatchActionRequest(actionRequest: ActionRequest, expectedClassCanonicalName: String, methodName: String): MatchResult = {
```
coutoPL, line 39: `methodName` is too generic

coutoPL (changes requested):

integration-tests/src/test/scala/tech/beshu/ror/integration/plugin/DataStreamApiSuite.scala
```diff
@@ -0,0 +1,370 @@
...
+
+  private def createDocsInDataStream(streamName: String, count: Int): Unit = {
+    List.range(0, count).foreach { c =>
+      val doc = ujson.read(s"""{ "message":"test$c", "@timestamp": "${format(Instant.now())}"}""")
+      val response = dataStreamManager.addDocument(streamName, doc)
+      response.responseCode should be(201)
+    }
+  }
+
+  private def waitForReindexing(): Unit = {
```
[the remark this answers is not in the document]
mateuszkp96: Finally, I used `waitForRefresh` and I reduced a number of created documents. With `waitForRefresh`  the response is after 1 second

es710x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/datastreams/GetDataStreamEsRequestContext.scala
```diff
@@ -0,0 +1,53 @@
...
+
+private[datastreams] class GetDataStreamEsRequestContext(actionRequest: ActionRequest,
+                                                         indices: Set[ClusterIndexName],
+                                                         esContext: EsContext,
+                                                         aclContext: AccessControlStaticContext,
+                                                         clusterService: RorClusterService,
+                                                         override val threadPool: ThreadPool)
+  extends BaseReadDataStreamsEsRequestContext(actionRequest, indices, esContext, aclContext, clusterService, threadPool) {
+
+  override protected def indicesMethodName: String = "indices"
```
coutoPL, line 36: do we need it?

integration-tests/src/test/scala/tech/beshu/ror/integration/plugin/DataStreamApiSuite.scala
```diff
@@ -0,0 +1,364 @@
...
+  with BaseSingleNodeEsClusterTest
+  with ESVersionSupportForAnyFreeSpecLike
+  with BeforeAndAfterEach {
+  this: EsClusterProvider =>
+
+  override implicit val rorConfigFileName: String = "/data_stream_api/readonlyrest.yml"
+
+  private lazy val client = clients.head.adminClient
+  private lazy val adminDocumentManager = new DocumentManager(client, esVersionUsed)
+  private lazy val adminDataStreamManager = new DataStreamManager(client)
```
coutoPL, line 39: I see one problem with these tests. We test the API as an admin user which has no indices restriction. It would be better if we can use user1 | user2 | user3. Eg we should not be able to create a datastream that has ".ds-data-stream-dev-0001" index underneath if the user has no proper pattern allowed (eg. .ds-data-stream-dev*). I see you prepared proper ROR config, but these users were not used in this suite.

coutoPL (changes requested):

coutoPL (approved):
LGTM.

Please add changelog style description to the PR before merging it. 
Thanks

coutoPL (approved):

PR #877, opened 2023-01-22
[RORDEV-820] Data streams support in indices rule

mateuszkp96:
🐞Fix (ES) Data streams support in `indices` rule


integration-tests/src/test/resources/data_stream_api/readonlyrest.yml
```diff
@@ -59,12 +59,56 @@ readonlyrest:
...
\ No newline at end of file
+        - cluster:monitor/main
+
+    - name: "User7 access streams"
+      auth_key: "user7:pass"
+      indices:
+        - 'data-stream-dev'
+        - 'data-stream-test'
+      actions:
+        - indices:admin/data_stream/*
```
coutoPL, line 78: I'm not sure if we should add the actions in our tests. In general, using action is not recommended in ROR. Do we gain sth here? Maybe we can achieve it differently?

coutoPL (changes requested):

integration-tests/src/test/resources/data_stream_api/readonlyrest.yml
mateuszkp96: You are right - we don't need the actions rule here. The one rule per user is enough to be sure that the expected ACL block is tested

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/indicesrule/clusterindices/LocalIndicesManager.scala
```diff
@@ -16,46 +16,89 @@
...
-import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, IndexAttribute}
+import tech.beshu.ror.accesscontrol.domain.ClusterIndexName.Local
+import tech.beshu.ror.accesscontrol.domain.IndexAttribute
 import tech.beshu.ror.accesscontrol.matchers.IndicesMatcher
 import tech.beshu.ror.accesscontrol.request.RequestContext
 
 class LocalIndicesManager(requestContext: RequestContext,
-                          override val matcher: IndicesMatcher[ClusterIndexName.Local])
-  extends IndicesManager[ClusterIndexName.Local] {
+                          override val matcher: IndicesMatcher[Local])
```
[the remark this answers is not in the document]
mateuszkp96: In the `RemoteIndicesManager` the `Remote` was previously used, so I changed it here to `Local`. Now I changed it to LocalIndexName and Remote to RemoteIndexName in the `RemoteIndicesManager`.

coutoPL (approved):

PR #886, opened 2023-03-06
[RORDEV-637] Gradle 7.6.1

mateuszkp96:


ror-tools-core/build.gradle
```diff
@@ -114,25 +96,21 @@ compileScala {
     options.compilerArgs << "-Ywarn-macros:after"
 }
 
-tasks.withType(ScalaCompile) {
```
coutoPL, line 117: maybe this causes the patching step doesn't work in integration tests?

tests-utils/build.gradle
```diff
@@ -48,56 +36,38 @@ tasks.withType(ScalaCompile) {
...
-githook {
-    hooks {
-        "pre-commit" {
-            shell = "./gradlew license --rerun-tasks"
-        }
-    }
 }
 
-task wrapper(type: Wrapper) {
+tasks.register('wrapper', Wrapper) {
```
coutoPL, line 71: is it needed? I don't understand this task. Maybe we can try to remove it? (but please to it after solving all other issues)

coutoPL (changes requested):

sscarduzio:
guys any blocker on this?

coutoPL:
I asked Mateusz to switch to one of the task that our client needs at the moment (see the jira). 

But Mateusz dont hesitate to write a word about problems you have with this one ;)

mateuszkp96:
I resolved problems with all es modules except the proxy module. Gradle has changed dependency management. I need time to find the correct project structure for the proxy module. I think I'm close to a valid solution

coutoPL:
Maybe this is the right time to ask ourselves: do we need the ROR proxy any more?

sscarduzio:
Eventually no user adopted it. If it's an obstacle to moving forward let's drop it. We take it from git if we need it.

ror-tools-core/build.gradle
mateuszkp96: Should be fixed now

coutoPL:
yeah, so let's stop supporting the proxy mode. 
@mateuszkp96 could you please get rid of it? If it's better for you, you can do it in a separate PR

coutoPL (commented):
It looks good, but I want to play the project with the new gradle in Idea.

PR #889, opened 2023-03-18
[RORDEV-236] New audit sink type - log

mateuszkp96:
🚀New (ES) `audit` supports a new output type: `log`


core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -37,18 +39,83 @@ object AuditingSettingsDecoder extends Logging {
...
+      remoteAuditCluster <- decodeOptionalSetting[AuditCluster.RemoteAuditCluster](c)("cluster", fallbackKey = "audit_cluster")
+      maybeOutputs <- c.downField("audit").downField("outputs").as[Option[List[AuditSinkConfig]]]
+      sinkConfigs <- maybeOutputs match {
+        case Some(value) =>
+          for {
+            _ <- Either.cond(
+              List(auditIndexTemplate, customAuditSerializer, remoteAuditCluster).forall(_.isEmpty),
+              (),
+              DecodingFailure(
+                message = "Audit config contains 'outputs' section and deprecated audit properties." +
```
mateuszkp96, line 59: I am not sure if we should extend the `audit` section or consider the `v2` variant. The `audit` section contains many fallback keys and becomes harder to maintain. I will supply the rest of the unit tests when we make the decision

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/audit/EsIndexBasedAuditSink.scala
```diff
@@ -0,0 +1,50 @@
...
+ *
+ *    ReadonlyREST is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.accesscontrol.logging.audit
```
coutoPL, line 17: As we agreed it'd be better to distinguish between logging and audit. So, maybe it would be nice to refactor it and move `audit` out of the `logging` package?

core/src/main/scala/tech/beshu/ror/accesscontrol/logging/audit/AuditSink.scala
```diff
@@ -0,0 +1,40 @@
...
+    safeRunSerializer(auditEvent)
+      .map {
+        case Some(serializedEvent) => submit(auditEvent, serializedEvent)
+        case None => ()
+      }
+  }
+
+  def close(): Unit
+
+  protected def submit(event: AuditResponseContext, serializedEvent: JSONObject): Unit
```
coutoPL, line 35: and here too

core/src/test/resources/log4j2.properties
```diff
@@ -34,6 +34,23 @@ rootLogger.level = debug
 rootLogger.appenderRef.console.ref = console
 #rootLogger.appenderRef.rolling.ref = rolling
 
+appender.readonlyrest_audit_rolling.type = RollingFile
```
coutoPL, line 37: it's great. Don't forget to add it in the docs

core/src/test/scala/tech/beshu/ror/integration/AccessControlAuditTests.scala
```diff
@@ -16,20 +16,22 @@
...
 import tech.beshu.ror.utils.TestsUtils.header
 
+import java.time.{Clock, Instant, ZoneId}
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future, Promise}
-import monix.execution.Scheduler.Implicits.global
-
 import scala.language.postfixOps
 
 class AccessControlAuditTests extends AnyWordSpec with BaseYamlLoadedAccessControlTest {
```
coutoPL, line 37: should we rename this test?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditingSettingsTests.scala
```diff
@@ -19,27 +19,27 @@ package tech.beshu.ror.unit.acl.factory
...
 import tech.beshu.ror.providers._
 import tech.beshu.ror.utils.TestsUtils._
-import java.time.{Clock, ZoneId, ZonedDateTime}
-
-import io.lemonlabs.uri.Uri
 
+import java.time.{Clock, ZoneId, ZonedDateTime}
 import scala.reflect.ClassTag
 
 class AuditingSettingsTests extends AnyWordSpec with Inside {
```
coutoPL, line 45: should we rename this test name?

core/src/test/scala/tech/beshu/ror/unit/acl/logging/AuditingToolTests.scala
```diff
@@ -16,143 +16,164 @@
...
-        )
-
-        val requestContext = MockRequestContext.indices.copy(timestamp = someday.toInstant, id = RequestContext.Id("mock-1"))
-        val responseContext = Errored(requestContext, new Exception("error"))
-
-        auditingTool.audit(responseContext).runSyncUnsafe()
+    "log sink is used" should {
+      "saved audit log to file defined in log4j config" in {
+        for {
+          file <- File("./core/build/tmp/readonlyrest_audit.log").toTemporary
```
coutoPL, line 140: we have a hardcoded core module name here. Maybe we should take advantage of better.files' temporary files?

gradle.properties
```diff
@@ -1,3 +1,3 @@
 publishedPluginVersion=1.47.0
-pluginVersion=1.48.0-pre3
+pluginVersion=1.48.0-pre4
```
coutoPL, line 2: pre5?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
coutoPL: ok, we can provide v2 variant, but under a few assumptions:
* backward compatibility for in-index audit
* the new type of sink can be supported in v2 only (if users want to use it, they should adapt the v2)
* from the perspective of code we don't need to persist the information that we handle two versions of configuration 
* we should fix all audit examples in our docs to the format

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -34,21 +39,91 @@ import scala.util.{Failure, Success, Try}
...
-          auditIndexTemplate <- decodeOptionalSetting[RorAuditIndexTemplate](c)("index_template", fallbackKey = "audit_index_template")
-          customAuditSerializer <- decodeOptionalSetting[AuditLogSerializer](c)("serializer", fallbackKey = "audit_serializer")
-          remoteAuditCluster <- decodeOptionalSetting[AuditCluster.RemoteAuditCluster](c)("cluster", fallbackKey = "audit_cluster")
-        } yield AuditingTool.Settings(
-          auditIndexTemplate.getOrElse(RorAuditIndexTemplate.default),
-          customAuditSerializer.getOrElse(new DefaultAuditLogSerializer),
-          remoteAuditCluster.getOrElse(AuditCluster.LocalAuditCluster)
-        )
+  lazy val instance: Decoder[Option[AuditingTool.Settings]] = {
+    for {
```
coutoPL, line 43: why not `auditSettingsDecoder.orElse(DeprecatedAuditSeettingsDecoder.instance)`?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -34,21 +39,91 @@ import scala.util.{Failure, Success, Try}
...
+      auditSettings <- auditSettingsDecoder
+      deprecatedAuditSettings <- DeprecatedAuditSeettingsDecoder.instance
+    } yield auditSettings.orElse(deprecatedAuditSettings)
+  }
+
+  private val auditSettingsDecoder: Decoder[Option[AuditingTool.Settings]] = Decoder.instance { c =>
+    for {
+      isEnabled <- c.downField("audit").downField("enabled").as[Option[Boolean]]
```
coutoPL, line 51: it'd be good if we could be able to disable one sink only? And maybe instead of Option, we will introduce Enabled(...) | Disabled sealed trait hierarchy?

coutoPL (changes requested):

coutoPL (approved):
LGTM. Thanks

PR #890, opened 2023-03-20
[RORDEV-637] 'proxy' module removal

mateuszkp96:


coutoPL (commented):
WDYT about removing `tech.beshu.ror.integration.plugin` package and making all trait suites from `tech.beshu.ror.integration.suites` classes?

PR #893, opened 2023-03-27
[RORDEV-871] Simple format for audit outputs config

mateuszkp96:


core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -182,6 +210,12 @@ object AuditingSettingsDecoder extends Logging {
       .overrideDefaultErrorWith(error)
   }
 
+  private def unsupportedOutputTypeError(unsupportedType: String) = {
+    AuditingSettingsCreationError(Message(
+      s"Unsupported 'type' of audit output: $unsupportedType. Supported types: [index, log]"
```
coutoPL, line 215: let's pass the supported types list in the param. If some strings are repeated, it's better to have them closed each other (now it's high probability that someone in the future forgets to add the new supported type here)

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -59,14 +59,26 @@ object AuditingSettingsDecoder extends Logging {
   }
 
   private def decodeAuditSettings = {
+    decodeAuditSettingsWith(auditSinkConfigExtendedDecoder)
```
coutoPL, line 62: why not `decodeAuditSettingsWith(auditSinkConfigSimpleDecoder).orElse(decodeAuditSettingsWith(auditSinkConfigExtendedDecoder))`?

shouldn't we decode the simple case first?

coutoPL (changes requested):

coutoPL (approved):
LGTM. Thanks

PR #896, opened 2023-03-30
[RORDEV-636] Scala 2.13.10

mateuszkp96:


core/build.gradle
```diff
@@ -170,7 +163,7 @@ tasks.withType(ScalaCompile).configureEach {
             }
 
     // Add whatever flags you typically add
-    parameters += ["-Ypartial-unification", "-Ywarn-macros:after", "-feature"] //, "-Xfatal-warnings"]
+    parameters += ["-Ywarn-macros:after", "-feature", "-Ymacro-annotations"]
```
coutoPL, line 166: can we add fatal-warnings now? If it's not so simple, what do we need now to have it?

coutoPL (changes requested):

core/build.gradle
mateuszkp96: I added it with `-deprecation:false` because over 1000 files are affected. I will remove this flag in the next MRs

coutoPL (approved):
LGTM

PR #898, opened 2023-04-02
[RORDEV-847] CVE fixes & dependencies upgrade

mateuszkp96:
🚨Security Fix (ES) CVE-2022-45688

### DIFF
--- a/audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
+++ b/audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
@@ -18,7 +18,6 @@ package tech.beshu.ror.audit.adapters;
 
 import org.json.JSONObject;
 
-@Deprecated
 class JsonObjectJavaAdapter {
 
     static JSONObject create(Object obj) {
--- a/audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
+++ b/audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
@@ -14,17 +14,14 @@
  *    You should have received a copy of the GNU General Public License
  *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
  */
+package tech.beshu.ror.audit.adapters;
 
-plugins {
-    id "groovy-gradle-plugin"
-}
+import org.json.JSONObject;
 
-repositories {
-    gradlePluginPortal()
-}
+@Deprecated
+class JsonObjectJavaAdapter {
 
-dependencies {
-    implementation "com.star-zero.gradle:githook:1.2.1"
-    implementation "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:0.16.1"
-    implementation "org.owasp:dependency-check-gradle:8.2.1"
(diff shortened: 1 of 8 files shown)


audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.adapters;
+
+import org.json.JSONObject;
+
+@Deprecated
+class JsonObjectJavaAdapter {
```
mateuszkp96, line 22: New constructors have appeared in the `json` lib and this caused the `ambiguous constructors` issue in our scala code. 
This adapter is needed to use the desired constructor explicitly.

supressions_cve.xml
```diff
@@ -42,55 +42,46 @@
...
    ]]></notes>
-        <sha1>2cd0a87ff7df953f810c344bdf2fe3340b954c69</sha1>
-        <cve>CVE-2022-1471</cve>
+        <sha1>60458f877d055d0c9114d9e1a2efb737b4bc282c</sha1>
+        <cve>CVE-2020-8908</cve>
     </suppress>
     <suppress>
         <notes><![CDATA[
-   file name: async-http-client-netty-utils-2.10.5.jar
+   file name: bc-noncert-1.0.2.4.jar
```
mateuszkp96, line 68: I added suppressions for `bc-noncert`, but I don't know how to fix it (no upgraded lib).

audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.adapters;
+
+import org.json.JSONObject;
+
+@Deprecated
```
coutoPL, line 21: deprecated? what to use instead? could you please add the message

supressions_cve.xml
coutoPL: isn't it a false positive? 
We use `bcpkix-fips`, `bctls-fips` and transitively `bc-fips` (which is excluded and replaced by `bc-noncert` because the latest fix to `bc-fibs` is not published in the mvn repo by bouncycastle).

According to the CVE description, only `bc-fibs` has `[CVE-2020-26939](https://github.com/advisories/GHSA-72m5-fvvv-55m6)` but only for < 1.0.2. 

So, we are good, aren't we?

coutoPL (changes requested):

coutoPL:
... and please add ROR changelog style description

audit/src/main/java/tech/beshu/ror/audit/adapters/JsonObjectJavaAdapter.java
mateuszkp96: Fixed

coutoPL (approved):
LGTM

PR #907, opened 2023-05-09
[ROREV-899] wildcards support in external to local groups mapping

coutoPL:
🧐Enhancement (ES) external group name patterns support in the external to local groups mapping


mateuszkp96 (changes requested):
Some unit tests failed

PR #908, opened 2023-05-17
[RORDEV-901] fixed: resolving ENVs to YAML number

coutoPL:
🐞Fix (ES) resolving ENVs to YAML number in ROR settings


mateuszkp96 (approved):
LGTM

PR #910, opened 2023-05-26
[RORDEV-716] Fix for invalid action name warnings in case of ROR actions

mateuszkp96:


core/src/test/scala/tech/beshu/ror/unit/acl/domain/ActionTest.scala
```diff
@@ -0,0 +1,49 @@
...
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.unit.acl.domain
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+import tech.beshu.ror.accesscontrol.domain.Action
+
+class ActionTest extends AnyWordSpec with Matchers {
```
coutoPL, line 23: OK, but IMO we need one more test. Let's provide a core integration test (the one that extends BaseYamlLoadedAccessControlTest) and check if `Action` rule with the old ROR action name will be matched now

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
```diff
@@ -461,28 +463,47 @@ object domain {
...
+    private val rorActionByOutdatedName: Map[String, Action] = Map(
+      "cluster:ror/user_metadata/get" -> rorUserMetadataAction,
+      "cluster:ror/config/manage" -> rorConfigAction,
+      "cluster:ror/testconfig/manage" -> rorTestConfigAction,
+      "cluster:ror/authmock/manage" -> rorAuthMockAction,
+      "cluster:ror/audit_event/put" -> rorAuditEventAction,
+      "cluster:ror/config/refreshsettings" -> rorOldConfigAction
+    )
+
+    def apply(value: String): Action = {
```
coutoPL, line 494: I'm not sure if this is the right approach. We will see when you write the test I mentioned. 

---
I was rather thinking that we should do it like this:
```
sealed trait Action
object Action {
  final case class EsAction(name) extends Action
  sealed trait RorAction extends Action
  object RorAction {
    case object RorUserMetadataAction extends RorAction
    ...
  }
  
  def from(value: String): Action = {
    // two strings can lead to create RorUserMetadataAction
  }
}

File: core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala [hunk 3]
```diff
@@ -461,28 +463,47 @@ object domain {
...
-    val rorTestConfigAction = Action("cluster:ror/testconfig/manage")
-    val rorAuthMockAction = Action("cluster:ror/authmock/manage")
-    val rorAuditEventAction = Action("cluster:ror/audit_event/put")
-    val rorOldConfigAction = Action("cluster:ror/config/refreshsettings")
+    val rorUserMetadataAction: Action = rorAction("user_metadata/get")
+    val rorConfigAction: Action = rorAction("config/manage")
+    val rorTestConfigAction: Action = rorAction("testconfig/manage")
+    val rorAuthMockAction: Action = rorAction("authmock/manage")
+    val rorAuditEventAction: Action = rorAction("audit_event/put")
+    val rorOldConfigAction: Action = rorAction("config/refreshsettings")
```
coutoPL, line 472: in the sealed trait hierarchy, this will be the same that the `rorConfigAction` is.

coutoPL (changes requested):

gradle.properties
```diff
@@ -1,3 +1,3 @@
 publishedPluginVersion=1.48.0
-pluginVersion=1.49.0-pre5
```
coutoPL, line 2: let's change it to 1.50.0-pre1 too

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
mateuszkp96: We need to maintain both versions because we have two actions in elasticsearch modules: `tech.beshu.ror.es.actions.rrconfig.RRConfigActionType` `tech.beshu.ror.es.actions.rradmin.RRAdminActionType`

core/src/test/scala/tech/beshu/ror/integration/ActionYamlLoadedAccessControlTest.scala
```diff
@@ -0,0 +1,101 @@
...
+      |    actions: ["cluster:r*"] # old format
+    """.stripMargin
+
+  "An ACL" when {
+    "actions rule is defined" should {
+      "allow to proceed" when {
+        "it is an metadata request and the action is on the configured list with new name" in {
+          val request = MockRequestContext.metadata.copy(action = Action.RorAction.RorUserMetadataAction)
+          val result = acl.handleRegularRequest(request).runSyncUnsafe()
+          result.history should have size 1
```
coutoPL, line 62: maybe instead of this assertion, let's assert matched block name (it's more readable when test fails)

coutoPL (changes requested):

coutoPL:
please add changelog style description

core/src/test/scala/tech/beshu/ror/integration/ActionYamlLoadedAccessControlTest.scala
coutoPL: Allow is defined like that `final case class Allow[B <: BlockContext](blockContext: B, block: Block)` ;) we don't have to have `lastMatchedBlockName` method

PR #929, opened 2023-08-09
[RORDEV-969] tests refactoring

coutoPL:
* now, when the status code assertion is not matched, in the error there will be a body of the response

### DIFF
--- a/integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResponseFieldRuleSuite.scala
+++ b/integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResponseFieldRuleSuite.scala
@@ -16,20 +16,20 @@
  */
 package tech.beshu.ror.integration.suites
 
-import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 import tech.beshu.ror.integration.suites.base.support.BaseSingleNodeEsClusterTest
 import tech.beshu.ror.integration.utils.{ESVersionSupportForAnyWordSpecLike, SingletonPluginTestSupport}
 import tech.beshu.ror.utils.containers.ElasticsearchNodeDataInitializer
 import tech.beshu.ror.utils.elasticsearch._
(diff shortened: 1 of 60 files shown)


integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResponseFieldRuleSuite.scala
```diff
@@ -72,7 +72,7 @@ class ResponseFieldRuleSuite
       val result = searchManager.search(
         ujson.read("""{"query": {"terms":{"user_id": ["alice", "bob"]}}}""")
       )
-      result.responseCode should be(200)
+      result should have statusCode 401
```
mateuszkp96, line 75: Status has changed from 200 to 401. Is this OK?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RorDisabledSuite.scala
```diff
@@ -58,15 +58,15 @@ class RorDisabledSuite
...
+      result should have statusCode 200
     }
     "return information that ROR is disabled" when {
       "ROR API endpoint is being called" in {
         val user1MetadataManager = new RorApiManager(basicAuthClient("user1", "pass"), esVersionUsed)
 
         val result = user1MetadataManager.fetchMetadata()
 
-        result.responseCode should be(403)
+        result should have statusCode 402
```
mateuszkp96, line 69: Status has changed from 403 to 402. Is this OK?

mateuszkp96 (changes requested):

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/ResponseFieldRuleSuite.scala
coutoPL: thanks for catching this. It was a problem with the custom matcher

mateuszkp96 (approved):
LGTM

PR #930, opened 2023-08-12
[RORDEV-937] stack-management screen fix in case of `xpack.security.enabled: true`

coutoPL:
🐞Fix (ES) stack-management screen fix in case of `xpack.security.enabled: true`

### DIFF
--- a/es63x/src/main/scala/tech/beshu/ror/es/utils/XPackSecurityAuthenticationHeader.scala
+++ b/es63x/src/main/scala/tech/beshu/ror/es/utils/XPackSecurityAuthenticationHeader.scala
@@ -53,7 +53,7 @@ object XPackSecurityAuthenticationHeader {
       output.writeString(userName)
     } else {
       output.writeString(userName)
-      output.writeStringArray(Array("kibana_admin"))
+      output.writeStringArray(Array("kibana_admin", "superuser"))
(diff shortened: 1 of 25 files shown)


es63x/src/main/scala/tech/beshu/ror/es/utils/XPackSecurityAuthenticationHeader.scala
```diff
@@ -53,7 +53,7 @@ object XPackSecurityAuthenticationHeader {
       output.writeString(userName)
     } else {
       output.writeString(userName)
-      output.writeStringArray(Array("kibana_admin"))
```
Dzuming, line 56: I think we need only a superuser role because this role contains all kibana_admin permissions (it's something like unrestricted kibana_access)

Dzuming (approved):

es63x/src/main/scala/tech/beshu/ror/es/utils/XPackSecurityAuthenticationHeader.scala
coutoPL: ok, I've tested in our sandbox and using ECK PoC and it seems you are right

Dzuming (approved):

PR #931, opened 2023-08-15
[RORDEV-539] additional validation in `kibana` rule

coutoPL:
🧐Enhancement (ES) additional `hide_apps` validation in the `kibana` rule


core/src/test/scala/tech/beshu/ror/Example.scala
```diff
@@ -0,0 +1,28 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror
+
+// todo: remove
```
mateuszkp96, line 19: Todo left

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/misc.scala
```diff
@@ -44,21 +46,48 @@ object AuditCluster {
...
+
+final case class JsRegex private(value: NonEmptyString)
+object JsRegex extends Logging {
+  private val extractRawRegex = """\/(.*)\/""".r
+
+  def compile(str: NonEmptyString)
+             (implicit jsCompiler: JsCompiler): Either[CompilationResult, JsRegex] =
+    str.value match {
+      case extractRawRegex(regex) =>
+        jsCompiler.compile(s"new RegExp('$regex')") match {
```
coutoPL, line 76: @Dzuming this is how I validate the JS regex. WDYT?

mateuszkp96 (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/misc.scala
Dzuming: The first thing worth mentioning is that on Kibana plugin side we use this library https://www.npmjs.com/package/xregexp for regular expressions. In theory, it shouldn't be a problem since it uses  `ES6 regular expression syntax`, and we inform users to use https://docs.readonlyrest.com/kibana#hiding-kibana-apps this syntax.

The next thing is, that we have "special" syntax for regexes, see [docs](https://docs.readonlyrest.com/kibana#hiding-kibana-apps)

⚠️IMPORTANT Pipe operator needs to be escaped correctly when it's declared in the regular expression \\|. The regular expression must be declared between double quote "/<regular-expression/"

It means that there should be a logic to detect if hide_apps is a regexp or just a regular string. 


About the code:

If I understand correctly you want to call the native js regexp method here and check if compile or not, IMO every string will be compiled via `new RegExp()`

image

I think you should use the same approach as in the case of the Kibana plugin

```
const isRegexExpression = expression.startsWith('/') && expression.endsWith('/');
```

and the rest is treated as a string expression.

About js code injection, I don't think there should be a problem with it, at least I can't see a way to do it and evaluate.

es63x/src/main/scala/tech/beshu/ror/es/utils/EsPatchVerifier.scala
```diff
@@ -25,28 +26,44 @@ import scala.util.Try
 
 object EsPatchVerifier extends Logging {
 
+  private val determinedPatchState: Atomic[Option[Boolean]] = Atomic(Option.empty[Boolean])
+
+  def isPatched: Boolean = determinedPatchState.get() match {
```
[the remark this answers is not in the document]
coutoPL: yeah, it's not used because at the end I didn't have to use it. But I'd not remove this method because it could be useful in the future.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/misc.scala
coutoPL: > The next thing is, that we have "special" syntax for regexes, see [docs](https://docs.readonlyrest.com/kibana#hiding-kibana-apps)

Yeah, I saw that. When we compile the regex we detect the case when it's not a regex but only a string. See:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/931/files#diff-921b0c3473600d8156fc15d42fbcf62d4c5679dcb54625b09ca94d75b66d78f4R70

|> ⚠️IMPORTANT Pipe operator needs to be escaped correctly when it's declared in the regular expression \|. The regular expression must be declared between double quote "/<regular-expression/"

Here is an example of how it's escaped:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/931/files#diff-b446842a0ca9e4e6f881f0041dc55b3785ee3a8b37bdd74e5e4a5b80242b0dfaR27

|> About js code injection, I don't think there should be a problem with it, at least I can't see a way to do it and evaluate.

I've experimented with it a little bit and I found a way to inject some code (see `JsRegexTests`). Now I'm trying to validate the input. Please take a look (see `JsRegex#validateInput` method)

Dzuming: > Here is an example of how it's escaped:

`hide_apps: ["user2_app1", "user2_app2", "/^Analytics\\|(?!(Maps)$).*$/"]`

What I meant by my comment  is, that for the JS `RegExp` method, all of these hide_apps values are valid regex expressions, Maybe we shouldn't care about it, but if yes, we should detect which of them are regex, before a validation

> I've experimented with it a little bit and I found a way to inject some code (see JsRegexTests). Now I'm trying to validate the input. Please take a look (see JsRegex#validateInput method)

You are right, we are asking for trouble here. Did you consider the conversion of the regex syntax to the Java syntax and validate it in Java?

mateuszkp96 (approved):
LGTM

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/misc.scala
coutoPL: > What I meant by my comment is, that for the JS RegExp method, all of these hide_apps values are valid regex expressions,

Yes, I'm aware. We assume that if some string is wrapped into `/`s, it means it's a regex. I took that into consideration. It's also tested in `KibanaHideAppsRuleSettingsTests` (please take a look).

> Did you consider the conversion of the regex syntax to the Java syntax and validate it in Java?
Yes, I did. But haven't found any reliable conversion method, so I decided to use the JS engine implementation in Java.

Dzuming: Ok, LGTM

Dzuming (approved):

PR #933, opened 2023-08-23
[RORDEV-931] fix: getting data streams API

coutoPL:
🐞Fix (ES) [getting data streams when not full names of backing indices are declared in the `indices` rule](https://forum.readonlyrest.com/t/forbidden-for-creating-component-templates/2372/7)

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
@@ -23,6 +23,7 @@ import cats.kernel.Monoid
 import com.comcast.ip4s.{Cidr, Hostname, IpAddress}
 import eu.timepit.refined.api.Refined
 import eu.timepit.refined.auto._
+//import eu.timepit.refined.types.all.NonEmptyString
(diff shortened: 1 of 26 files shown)


core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
```diff
@@ -23,6 +23,7 @@ import cats.kernel.Monoid
 import com.comcast.ip4s.{Cidr, Hostname, IpAddress}
 import eu.timepit.refined.api.Refined
 import eu.timepit.refined.auto._
+//import eu.timepit.refined.types.all.NonEmptyString
```
mateuszkp96, line 26: Commented code

mateuszkp96 (approved):

core/src/main/scala/tech/beshu/ror/accesscontrol/domain.scala
coutoPL: I'm preparing some refactor related to this task. I promise to fix it in the next PR

PR #942, opened 2023-09-21
[RORDEV-993] _search & `fields` rule issue

coutoPL:
🚨Security Fix (ES) [`fields` rule didn't work well in the case of ES 7.10.0 and later and more than 10 documents in the response](https://forum.readonlyrest.com/t/field-rule-not-working-when-exceeding-a-certain-no-of-docs/2415)


mateuszkp96 (approved):
LGTM

PR #949, opened 2023-10-21
[RORDEV-975] the `indices` rule improvements

coutoPL:
🧐Enhancement (ES) when the `kibana` rule is used within the same block with the `indices` rule, there is no need to explicitly allow kibana-related indices


mateuszkp96 (approved):
LGTM

mateuszkp96 (approved):

mateuszkp96 (approved):

PR #963, opened 2023-12-06
[RORDEV-1050] Scroll API bugfix

coutoPL:
🚨Security Fix (ES) Scroll API: protected data could leak when the `fields` rule was used with `fls_engine` set to `es` or `es_with_lucene` 

------------

Hint for the reviewer:
The fixed class is `SearchRequestOps`. The rest of the changes are just a cleanup or refactoring.


mateuszkp96 (approved):
LGTM

PR #969, opened 2023-12-17
[RORDEV-813] Structured groups in JWT authorization

mateuszkp96:


core/src/main/scala/tech/beshu/ror/accesscontrol/utils/ClaimsOps.scala
```diff
@@ -54,52 +77,83 @@ class ClaimsOps(val claims: Claims) extends Logging {
...
+  def groupsClaim(groupIdsClaimName: Jwt.ClaimName,
+                  groupNamesClaimName: Option[Jwt.ClaimName]): ClaimSearchResult[UniqueList[Group]] = {
+
+    (for {
+      groupIds <- readGroupIds(groupIdsClaimName)
+      groupNames <- readGroupNames(groupNamesClaimName)
+      searchResult = groupIds match {
+        case Found(ids) => ClaimSearchResult.Found {
+          groupNames match {
+            case Found(names) if names.size == ids.size =>
```
coutoPL, line 89: I wonder if we shouldn't treat the case where `names.size != ids.size` as sth that we should warn a user? For sure this is not what is expected.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/JwtAuthRuleTests.scala
```diff
@@ -65,7 +65,7 @@ class JwtAuthRuleTests
             AuthorizationTokenDef(Header.Name.authorization, "Bearer "),
             SignatureCheckMethod.Hmac(secret.getEncoded),
             userClaim = None,
-            groupsClaim = None
+            groupsConfig = None
```
coutoPL, line 68: do we cover all cases we have converted in the LDAP case?
for sure it would be nice to check the case where `ids.size != names.size`

integration-tests/src/test/resources/jwt_auth/readonlyrest.yml
```diff
@@ -37,7 +37,8 @@ readonlyrest:
     - name: jwt3
       signature_key: "1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890"
       user_claim: user
-      roles_claim: roles
+      group_ids_claim: role_ids
```
coutoPL, line 40: more realistic example would be claim like this:
```
"roles": [
  { "id":"r1", "name": "role 1"},
  { "id":"r2", "name": "role 2"}
]
```

are we able to modify the test suite?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/utils/ClaimsOps.scala
mateuszkp96: I added debug logs describing what's happening.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/JwtAuthRuleTests.scala
mateuszkp96: I added a few tests with group names and improved some test assertions.

coutoPL (approved):
LGTM

PR #976, opened 2024-01-17
[RORDEV-1080] EsAuditSinkService should not close NodeClient, but the bulk processor

coutoPL:
[technical change]


mateuszkp96 (approved):
LGTM

PR #985, opened 2024-02-14
[RORDEV-814] Structured groups in local groups authorization rules

mateuszkp96:


core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/rules/auth/GroupsRuleSettingsTests.scala
```diff
@@ -413,6 +451,73 @@ sealed abstract class GroupsRuleSettingsTests[R <: BaseGroupsRule : ClassTag](ru
...
+                 |readonlyrest:
+                 |
+                 |  access_control_rules:
+                 |
+                 |  - name: test_block1
+                 |    ${ruleName.name.value}: [group1, group3]
+                 |
+                 |  users:
+                 |  - username: cartman
+                 |    groups:
```
coutoPL, line 479: when we consider the advanced group mapping:

the old/basic syntax is:
`- <local_group_id> : [ <external_group_id_pattern> ]`

we should support the old/basic syntax, but the new/extended syntax should be:
````
- local_group: 
    id: <local_group_id> // required
    name: <local_group_name> // required
   external_group_ids: [ <external_group_id_pattern> ] 
```

So two comments here:
1. there is no sense to support the extended/new syntax and optional name - if the user needs to use the name (so they want to distinguish between name and id), they should use the new/extended syntax
2. IMO it's better to add the `local_group` key to make it clear

core/src/main/scala/tech/beshu/ror/accesscontrol/utils/CirceOps.scala
```diff
@@ -304,6 +325,12 @@ object CirceOps {
       downFields(name).asWithError[Option[NonEmptyString]](s"Field $name cannot be empty")
     }
 
+    def downFieldAs[T: Decoder](name: String): Decoder.Result[T] = {
+      value.downField(name).as[T].adaptError {
+        case error: DecodingFailure => error.modifyError(errorMessage => s"Error for field '$name': $errorMessage")
```
coutoPL, line 330: why do we need to do it?

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
```diff
@@ -171,7 +171,10 @@ class CurrentUserMetadataAccessControlTests
       |    auth_key: "user2:pass"
       |
       |  - username: user4
-      |    groups: ["group5", "group6"]
+      |    groups:
+      |      - id: group5
+      |        name: "Group 5"
+      |      - id : group6
```
coutoPL, line 177: maybe in this case we should also require to use the full syntax. WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/utils/CirceOps.scala
mateuszkp96: We did not have tests for failure cases. When I added such ones, the error messages were like "Non-empty list of user ID patterns is required". Without the context, it's hard to locate the malformed configuration part. I know it's not an elegant error handling, but it was the simplest way to present some context in the error messages without modifying the domain.

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
```diff
@@ -171,7 +171,13 @@ class CurrentUserMetadataAccessControlTests
       |    auth_key: "user2:pass"
       |
       |  - username: user4
-      |    groups: ["group5", "group6"]
+      |    groups:
+      |      - local_group:
```
coutoPL, line 175: oh, little misunderstanding here. When we use `GroupMappings.Simple` there is no need to specify that it's a local group. But the `GroupMappings.Advanced` is used together with defining different group name, it's better "wrap" the local group attributes under the `local_group` key.

coutoPL: So in this case the config should look like this:
```
- username: user4
  groups:
    - id: group5
      name: group 5
```

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
coutoPL: oh, I see, I suggested that. I don't know why. I was wrong.

coutoPL (approved):
LGTM

PR #991, opened 2024-03-06
[RORDEV-1092] Internode SSL `certificate_verification: true` was causing problems with nodes discovery

coutoPL:
🐞Fix (ES) [Internode SSL `certificate_verification: true` was causing problems with nodes discovery](https://forum.readonlyrest.com/t/upgrade-elasticsearch-8-2-to-8-x-leads-to-ssl-problems/2480)


mateuszkp96 (approved):
LGTM

PR #992, opened 2024-03-13
[RORDEV-1152] Compile modules test-utils and integration-tests with Scala 3.3.3

mgoworko:
Modules test-utils and integrations-tests are now compiled with Scala 3 (v3.3.3 LTS). Library upgrades and minor code adjustments.


integration-tests/src/test/scala/tech/beshu/ror/integration/utils/ESVersionSupport.scala
```diff
@@ -34,8 +34,7 @@ trait ESVersionSupportForAnyWordSpecLike extends ESVersionSupport {
     string.taggedAs(firstTestTag, otherTestTags: _*)
 }
 
-trait ESVersionSupportForAnyFreeSpecLike extends ESVersionSupport {
-  this: AnyFreeSpecLike =>
+trait ESVersionSupportForAnyFreeSpecLike extends ESVersionSupport with AnyFreeSpecLike {
```
coutoPL, line 37: why do we have to do it like that?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsClusterContainer.scala
```diff
@@ -162,7 +163,11 @@ object EsClusterSettings {
   }
 
   final case class NodeType(securityType: SecurityType,
-                            numberOfInstances: Int Refined Positive = 1)
+                            numberOfInstances: Int Refined Positive = positiveInt(1))
+
+  inline def positiveInt(inline i: Int): Refined[Int, Positive] = {
```
coutoPL, line 168: there is no auto conversion?

tests-utils/src/main/scala/tech/beshu/ror/utils/misc/EsStartupChecker.scala
```diff
@@ -16,82 +16,85 @@
...
+                }
+              case Mode.Accessible =>
+                logger.info(s"[$name] ES is ready")
+                Task.unit
+            }
+          case otherStatus =>
+            logger.info(s"[$name] ES not ready yet, received HTTP $otherStatus")
+            Task.raiseError(ClusterNotReady)
+        }
+        response.close()
```
coutoPL, line 84: shouldn't we introduce cats Resource somewhere here?

coutoPL (changes requested):
please merge the current develop branch

tests-utils/src/main/scala/tech/beshu/ror/utils/misc/EsStartupChecker.scala
mgoworko: done, response wrapped in Resource

integration-tests/src/test/scala/tech/beshu/ror/integration/utils/ESVersionSupport.scala
mgoworko: Scala 3 compiler does not allow access to protected class FreeSpecStringWrapper otherwise. It may or may not have something to do with https://docs.scala-lang.org/scala3/reference/dropped-features/this-qualifier.html (it is not the same, but perhaps some side effect of more strict checks)

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsClusterContainer.scala
coutoPL: ok, it's not a big deal

coutoPL (approved):
Great! LGTM
Let me prepare the epic branch for the scala 3 migration and we are ready to merge

PR #994, opened 2024-03-20
[RORDEV-1128] additional validations: `kibana` rule should not be used with some other rules in the same block

coutoPL:
🧐Enhancement (ES) additional [validations](https://docs.readonlyrest.com/elasticsearch#configuring-an-acl-with-filter-fields-rules-when-using-kibana): `kibana` rule should not be used with some other rules in the same block


mateuszkp96 (approved):
LGTM

PR #1005, opened 2024-04-18
[RORDEV-1121] LDAP connector features: groups server-side filtering + skip user search option when user attribute is CN

coutoPL:
🧐Enhancement (ES) LDAP Connector feature: groups server-side filtering
🧐Enhancement (ES) LDAP Connector feature: skip user search option when user attribute is CN
🧐Enhancement (ES) rich, context-based debug logging in the LDAP connector and LDAP-related rules

---

**Groups server-side filtering feature**
Allows to not get all groups for a given user from the LDAP server, but allows to filter these groups on the LDAP server side which is going to return only the subset of all the groups

---

**Skip user search option when user attribute is CN**
When the user attribute is CN we can configure the LDAP connector to skip the user searching in LDAP. LDAP user name DN will be created from the username passed to ROR and the `search_user_base_DN` (like this: `DN: cn=${username},${search_user_base_DN}`)

---

**Rich, context-based debug logging in the LDAP connector and LDAP-related rules**

This can be enabled in the `log4j2.properties` file by adding these two lines:
```
logger.ror_ldap.name=tech.beshu.ror.accesscontrol.blocks.definitions.ldap.implementations
logger.ror_ldap.level=debug
```


coutoPL:
review in: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1002

mateuszkp96 (approved):
LGTM

PR #1006, opened 2024-04-23
[ROEDEV-1166] fixed: Stats API when `xpack.security.enabled: true` is set

coutoPL:
🐞Fix (ES) proper `/_stats` handling when `xpack.security.enabled: true` is set


mateuszkp96 (approved):
LGTM

PR #1007, opened 2024-04-25
[RORDEV-1174] fix: permissions for modified jar should be preserved

coutoPL:
🐞Fix (ES) patched files permission issue


mateuszkp96 (approved):
LGTM

PR #1018, opened 2024-05-17
[RORDEV-1152] ES testcontainer fixes

mgoworko:
Fixes:
- ES container startup timeout was reduced to 1 minute, instead of intended 2 minutes, now restored to the 5 minutes default from before Scala 3 migration
- testcontainers-scala downgraded back to 0.40.17 - I could not consistently reproduce it, but there may be a bug in 0.41.x, that manifests when using Dockerfile and copying files (testcontainers-java/issues/1348, testcontainers-java/issues/7239)


coutoPL (approved):
LGTM

PR #1019, opened 2024-05-20
[RORDEV-1193] taking into consideration OSS version of ES when the ROR settings are being configured

coutoPL:
🐞Fix (ES) no possibility to use ROR SSL with Elasticsearch OSS version


mateuszkp96 (approved):
LGTM

mk-software-pl:
@mateuszkp96 it was a problem with non-docker-based installation, so I had to refactor the code a little bit. Could you please take a look?

coutoPL:
@mateuszkp96 it was a problem with non-docker-based installation, so I had to refactor the code a little bit. Could you please take a look?

es717x/src/main/scala/tech/beshu/ror/es/ReadonlyRestEsConfig.scala
```diff
@@ -21,20 +21,18 @@ import monix.eval.Task
 import tech.beshu.ror.configuration.{EnvironmentConfig, FipsConfiguration, MalformedSettings, RorBootConfiguration, RorSsl}
 import tech.beshu.ror.utils.ScalaOps._
 
-import java.nio.file.Path
-
 final case class ReadonlyRestEsConfig(bootConfig: RorBootConfiguration,
```
mateuszkp96, line 24: Let's move this class to the core module (I don't see any differences between current modules)

coutoPL: good catch 👍

mateuszkp96 (approved):
LGTM

PR #1020, opened 2024-05-29
[RORDEV-1202] fixed: typo in LDAP server side filtering setting name

coutoPL:
"sever_side_groups_filtering" -> "server_side_groups_filtering"

The fix is backward compatible. We still support the setting name with the typo.


mateuszkp96 (approved):
LGTM

PR #1022, opened 2024-06-07
[RORDEV-811] Structured groups support in LDAP authorization

mateuszkp96:
🚀New (ES) Structured groups support - the ability to configure group names for the Kibana tenancy selector


core/src/test/resources/test_example.ldif
```diff
@@ -96,20 +96,26 @@ dn: cn=group1,ou=Groups,dc=example,dc=com
 objectClass: top
 objectClass: groupOfUniqueNames
 cn: group1
+o: Group 1
```
coutoPL, line 99: why `o`? maybe we could use a more descriptive key?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/BaseDecoderTest.scala
```diff
@@ -30,6 +31,16 @@ import tech.beshu.ror.utils.TestsUtils.unsafeNes
 abstract class BaseDecoderTest[F[_] : Comonad, A, B](decoder: ADecoder[F, A])
   extends AnyWordSpec with Inside {
 
+  def assertDecodingSuccess(yamls: NonEmptyList[String],
+                            assertion: A => Unit): Unit = {
+    yamls
+      .toList
+      .foreach { yaml =>
+        println(s"Testing yaml: $yaml")
```
coutoPL, line 39: println left

integration-tests/src/test/resources/ldap_integration/readonlyrest.yml
```diff
@@ -123,6 +123,8 @@ readonlyrest:
       search_groups_base_DN: "ou=Groups,dc=example,dc=com"
       user_id_attribute: "uid"                                  # default "uid"
       unique_member_attribute: "uniqueMember"                   # default "uniqueMember"
+      group_id_attribute: "cn"                                  # default "cn"
```
coutoPL, line 126: Can we use the new format in the integration tests? IMO, the legacy format should be used only in the unit tests

coutoPL (changes requested):

core/src/test/resources/test_example.ldif
mateuszkp96: We already use `ou` and I didn't want to confuse this attribute. As you can see, `ou` `o`, and `description` are valid attributes, and `o` seemed best for our case. Do you have any other option?

image
image
image

coutoPL: oh, OK. Didn't know that. So, let's leave it like this :)

coutoPL (approved):
LGTM

PR #1035, opened 2024-08-03
[RORDEV-1223] Fix for datastream backing indices access in the indices rule

mateuszkp96:
🐞Fix (ES) [data stream backing indices access issue with the indices rule](https://forum.readonlyrest.com/t/requested-index-doesnt-exist/2573)

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/RemoteIndicesManager.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/RemoteIndicesManager.scala
@@ -76,7 +76,7 @@ class RemoteIndicesManager(requestContext: RequestContext,
       }
   }
 
-  override def indicesPerDataStreamMap: Task[Map[RemoteIndexName, Set[RemoteIndexName]]] = {
+  override def backingIndicesPerDataStreamMap: Task[Map[RemoteIndexName, Set[RemoteIndexName]]] = {
     remoteDataStreams(requestContext.indexAttributes)
       .map {
         _.foldLeft(Map.empty[RemoteIndexName, Set[RemoteIndexName]]) {
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/RemoteIndicesManager.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/RemoteIndicesManager.scala
@@ -60,7 +60,6 @@ class RemoteIndicesManager(requestContext: RequestContext,
     remoteDataStreams(requestContext.indexAttributes).map(_.map(_.dataStream))
   }
 
-
   override def allDataStreamAliases: Task[Set[RemoteIndexName]] = {
     requestContext
       .allRemoteIndicesAndAliases.map(_.flatMap(_.aliases))
@@ -82,7 +81,8 @@ class RemoteIndicesManager(requestContext: RequestContext,
       .map {
         _.foldLeft(Map.empty[RemoteIndexName, Set[RemoteIndexName]]) {
           case (acc, fullRemoteDataStream) =>
-            val backingIndicesPerDataStream = Map(fullRemoteDataStream.dataStream -> fullRemoteDataStream.aliases)
+            val backingIndicesPerDataStream =
+              Map(fullRemoteDataStream.dataStream -> fullRemoteDataStream.backingIndices.map(index => RemoteIndexName(index, fullRemoteDataStream.clusterName)))
             mapMonoid.combine(acc, backingIndicesPerDataStream)
         }
       }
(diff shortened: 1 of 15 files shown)


integration-tests/src/test/scala/tech/beshu/ror/integration/suites/CrossClusterCallsSuite.scala
```diff
@@ -94,22 +97,56 @@ class CrossClusterCallsSuite
...
         "he queries remote indices only" in {
-          val result = user1SearchManager.search("etl2:test1_index")
+          val result = user1SearchManager.search("private2:test1_index")
           result should have statusCode 200
           result.searchHits.arr.size should be(2)
         }
+        "he queries remote data streams only by data stream name" excludeES (allEs7xBelowEs77x) in {
+          val result = user1SearchManager.search("private2:test1_ds")
+          result should have statusCode 200
+          result.searchHits.arr.size should be(3)
```
coutoPL, line 120: TBH this is not a good approach (I'm talking based on my experience with ROR tests that passes locally but fail in the pipeline). 

When there are more than 3 hists, the test will fail here. For sure we will be interested in what the search result returned. But there is no such log so there should be a problem with the analysis. 

There are at least two solutions here:
1. The same what we verify here, will also be verified by the last assertion. So, this one can be skipped.
2. nicely wrapped scalatest's [Checkpoint](https://www.scalatest.org/scaladoc/3.1.2/org/scalatest/Checkpoints$Checkpoint.html)

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/CrossClusterCallsSuite.scala
```diff
@@ -94,22 +97,56 @@ class CrossClusterCallsSuite
...
+          val searchResults = result.searchHits.map(_("_source").obj("message").str)
+          searchResults.sorted should be(List("test0", "test1", "test2"))
+        }
+        "he queries remote data streams only by data stream name with wildcard" excludeES (allEs7xBelowEs77x) in {
+          val result = user1SearchManager.search("private2:test1_d*")
+          result should have statusCode 200
+          result.searchHits.arr.size should be(3)
+          val searchResults = result.searchHits.map(_("_source").obj("message").str)
```
coutoPL, line 128: what's inside the message?  I cannot find where eg. test0 is defined

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/SearchApiSuite.scala
```diff
@@ -204,13 +271,69 @@ object SearchApiSuite {
...
-                                        documentManager: DocumentManager) = {
+  private def createDataStream(adminTemplateManager: IndexTemplateManager,
+                               adminDataStreamManager: DataStreamManager,
+                               dataStreamName: String,
+                               indexTemplateName: String): Unit = {
+    adminTemplateManager.createTemplate(indexTemplateName, indexTemplate(dataStreamName)).force()
+    adminDataStreamManager.createDataStream(dataStreamName)
+  }
+
+  private def indexTemplate(dataStreamName: String) = ujson.read(
```
coutoPL, line 298: It seems we have the same (or similar) in the DataStreamApiSuite. I understand that creating a data stream is not one operation (from the ES API perspective). But for the sake of our tests, it'd be nice to have an util which encapsulates all the methods under one: `def createDataStream(name: String)`. WDYT about creating such manager?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/SearchApiSuite.scala
```diff
@@ -204,13 +271,69 @@ object SearchApiSuite {
...
+       |  }
+       |}
+       |""".stripMargin
+  )
+
+  private def createDocsInDataStream(adminDocumentManager: DocumentManager, indexManager: IndexManager, streamName: String, count: Int): Unit = {
+    def format(instant: Instant) = instant.toString
+
+    def createDoc(c: Int): Unit = {
+      val doc = ujson.read(s"""{ "message":"test$c", "@timestamp": "${format(Instant.now())}"}""")
```
coutoPL, line 325: ah, here. Maybe instead of "test" it'd be better to have the "message" string in here?

coutoPL (commented):
- some tests failed (probably because we have to add some "excludes" here and there)
- let's add a markdown link to the Gustavo's thread from forum to the fix description

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/RemoteIndicesManager.scala
```diff
@@ -82,7 +81,8 @@ class RemoteIndicesManager(requestContext: RequestContext,
       .map {
         _.foldLeft(Map.empty[RemoteIndexName, Set[RemoteIndexName]]) {
           case (acc, fullRemoteDataStream) =>
-            val backingIndicesPerDataStream = Map(fullRemoteDataStream.dataStream -> fullRemoteDataStream.aliases)
```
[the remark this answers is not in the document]
mateuszkp96: It is symmetrical now - this method should return backing indices, and by mistake, it was returning aliases

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/CrossClusterCallsSuite.scala
mateuszkp96: Now messages are hardcoded in the suite code

coutoPL (approved):
It's ok, but it seems that we have some troubles found by tests in the pipeline. 
Please test locally `es79x` and `es711x` to be sure that they work too (we will find out when it's merged to develop branch).

AFAIS data stream APi was introduced in es 7.9 so it'd be nice to run locally `es78x` too.

mateuszkp96:
> It's ok, but it seems that we have some troubles found by tests in the pipeline. Please test locally `es79x` and `es711x` to be sure that they work too (we will find out when it's merged to develop branch).
> 
> AFAIS data stream APi was introduced in es 7.9 so it'd be nice to run locally `es78x` too.

It works as expected for modules es79x and es711x (tests passing) and es78x (skips tests)

PR #1038, opened 2024-08-11
[RORDEV-1019] prompt_for_basic_auth disabled by default

mateuszkp96:
⚠️Warning (ES) The prompt for basic authorization is disabled by default. To keep the previous behavior, set `readonlyrest.prompt_for_basic_auth` to `true` in the ROR configuration


core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/rules/elasticsearch/FieldsRuleSettingsTest.scala
```diff
@@ -288,7 +288,7 @@ class FieldsRuleSettingsTest extends BaseRuleSettingsDecoderTest[FieldsRule] {
               |""".stripMargin,
           assertion = errors => {
             errors should have size 1
-            errors.head should be(GeneralReadonlyrestSettingsError(Message("Unknown fls engine: 'something'. Supported: 'es_with_lucene', 'es'.")))
+            errors.head should be(GeneralReadonlyrestSettingsError(Message("Unknown fls engine: 'something'. Supported: 'lucene', 'es', 'es_with_lucene'(default).")))
```
coutoPL, line 291: AFAIK we don't officially say that `lucene` mode is supported. We left it as a "last resort" option, but it should never be needed to be used. Let's add proper comment somewhere and let's revert the previous message

integration-tests/src/test/resources/admin_api/readonlyrest_first_update.yml
```diff
@@ -1,5 +1,7 @@
 readonlyrest:
 
+  prompt_for_basic_auth: true
```
coutoPL, line 3: I understand that we should assert 403 instead of 401 currently? IMO we should do that. 
Obviously, we should have a test when we check that when `prompt_for_basic_auth: true` is set and the request is forbidden, the HTTP 401 is returned instead of 403. But in the rest of tests, we should use 403.

coutoPL (changes requested):
and please modify the docs too

coutoPL (approved):
LGTM

PR #1040, opened 2024-09-02
[RORDEV-1266] configurable max size of ROR yaml settings

coutoPL:
🚀New (ES) configurable ROR YAML settings max size


mateuszkp96 (approved):
LGTM

PR #1043, opened 2024-09-20
[RODEV-1280] create/restore snapshot: excluded indices cannot be defined before the included indices in the indices list

coutoPL:
🐞Fix (ES) create/restore snapshot with indices excludes


mateuszkp96 (approved):
LGTM

PR #1044, opened 2024-09-26
[RORDEV-1267] "response_if_req_forbidden" at the block level

mateuszkp96:
🚀New (ES) There is an option to define a custom response for users in ACL block with the 'forbid' policy

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/response/ForbiddenResponseContext.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/response/ForbiddenResponseContext.scala
@@ -36,8 +36,8 @@ final class ForbiddenResponseContext(aclStaticContext: Option[AccessControlStati
       .getOrElse(GlobalSettings.defaultForbiddenRequestMessage)
   }
 
-  def causes: List[String] = {
-    forbiddenCauses.toList.map(_.show)
+  def causes: NonEmptyList[String] = {
+    forbiddenCauses.map(_.show)
   }
 
   def doesRequirePassword: Boolean = {
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/response/ForbiddenResponseContext.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/response/ForbiddenResponseContext.scala
@@ -1,0 +1,128 @@
+/*
+ *    This file is part of ReadonlyREST.
+ *
+ *    ReadonlyREST is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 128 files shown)


core/src/main/scala/tech/beshu/ror/accesscontrol/response/ForbiddenResponseContext.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  import ForbiddenResponseContext.forbiddenCauseShow
+
+  def responseMessage: String = {
+    forbiddenRequestMessageFromBlock
+      .orElse(customForbiddenRequestMessage)
+      .getOrElse(GlobalSettings.defaultForbiddenRequestMessage)
+  }
+
+  def causes: List[String] = {
```
coutoPL, line 39: NonEmptyList?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/CoreFactoryTests.scala
```diff
@@ -98,6 +98,84 @@ class CoreFactoryTests extends AnyWordSpec with Inside with MockFactory {
...
+            |
+            |  - name: test_block2
+            |    type:
+            |      policy: forbid
+            |    auth_key: test:test
+            |
+            |  - name: test_block3
+            |    type:
+            |      policy: forbid
+            |      response_if_request_forbidden: "you are unauthorized to access this resource"
```
coutoPL, line 153: maybe `response_message`?
(because the policy has already told us that the matched request is forbidden, WDYT?)

core/src/test/scala/tech/beshu/ror/unit/acl/logging/AuditingToolTests.scala
```diff
@@ -99,7 +99,7 @@ class AuditingToolTests extends AnyWordSpec with MockFactory with BeforeAndAfter
               requestContext,
               new Block(
                 Block.Name("mock-block"),
-                Block.Policy.Forbid,
+                Block.Policy.Forbid(None),
```
coutoPL, line 102: WDYT about having None as a default value?

gradle.properties
```diff
@@ -1,5 +1,5 @@
 publishedPluginVersion=1.60.0
-pluginVersion=1.61.0-pre1
+pluginVersion=1.61.0-pre2
```
coutoPL, line 2: we have pre2, let's change it to pre3

coutoPL (changes requested):
It's great that you moved the common code from ES modules to core 👍

coutoPL (approved):
LGTM

PR #1048, opened 2024-10-26
[RORDEV-1259] Validation for duplicate usernames in user's definitions

mateuszkp96:
🧐Enhancement (ES) Validation added for multiple username entries  in the `users` section 

## Summary by CodeRabbit

- **New Features**
	- Introduced user validation functionality to ensure unique usernames in user definitions.
	- Added validation logic to the user definitions decoding process for improved error handling.

- **Bug Fixes**
	- Enhanced error handling for duplicate usernames and malformed user definitions in test scenarios.

- **Chores**
	- Updated plugin version in configuration files.


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/user/BasicAuthenticationRulesSimilarity.scala
```diff
@@ -0,0 +1,88 @@
...
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.BasicAuthenticationRule
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.*
+import tech.beshu.ror.utils.Similarity
+import tech.beshu.ror.utils.Similarity.SimilarityMeasure
+
+private[user] object BasicAuthenticationRulesSimilarity {
+
+  private type SupportedRule = AuthKeyRule | AuthKeyUnixRule | AuthKeyHashingRule
+
+  def similarity(x: BasicAuthenticationRule[?], y: BasicAuthenticationRule[?]): SimilarityMeasure = {
```
coutoPL, line 30: This implementation looks very nice, but I'm not sure if this is what the user needs. 
IMO we should compare the `username` fields. In code, this is represented by `UserIdPatterns` case class. 

We already have the `withUserIdParamsCheck`. It checks if the user defined in the auth rule matches the pattern defined in the `usernme` field. The user noticed that we have no validation of `usenname` patterns uniqueness. 
In the client's example the validation won't help, because username in the auth rule is hashed. 
But it seems there should be another validation, which can be defined like this:

- if there are at least two the same usename pattens (without wildcard) it should failed to validate

And that's it. Is it? Or I'm missing sth?

TBH, the "username" seems to be a redundant setting. In case of auth rules with local users it seems that it is not necessary. In case of rules like LDAP auth we can use it to cut out some users. In case of local users it makes no sense.

coutoPL (approved):
LGTM

PR #1052, opened 2024-11-08
[RORDEV-1274][TECH] covariant set + logs refactoring

coutoPL:
In this PR:
1. in the ROR codebase we use ConvariantSet instead of std Set (it's our wrapper around the std set)
2. in ROR logs, we don't use `toString`. From now on, we will use Show instances instead
3. refactoring related to these two and some class renaming and packages renaming and scala3 imports syntax

What is worth paying attention to in this PR?
1. `CovariantSet` case class
2. `implicits` object
3. `syntax` object
4. `tech.beshu.ror.es.handler.request` package of one of the refactored modules


mateuszkp96 (approved):
LGTM

PR #1058, opened 2024-11-18
[ES] KBN 8.16.x adaptations

coutoPL:


core/src/main/scala/tech/beshu/ror/accesscontrol/domain/indices.scala
```diff
@@ -88,7 +88,9 @@ object KibanaIndexName {
     """^_security_solution$""".r, // eg. .kibana_security_solution
     """^_security_solution_\d+\.\d+\.\d+$""".r, // eg. .kibana_security_solution_8.8.0
     """^_task_manager$""".r, // eg. .kibana_task_manager
-    """^_task_manager_\d+\.\d+\.\d+$""".r, // eg. .kibana_task_manager_8.8.0
+    """^_task_manager_\d+\.\d+\.\d+$""".r, // eg. .kibana_task_manager_8.8.0,
+    """^_usage_counters$""".r, // eg. .kibana_usage_counters
+    """^_usage_counters_\d+\.\d+\.\d+$""".r, // eg. .kibana_usage_counters_8.16.0
```
Dzuming, line 93: I see that it is possible to call all .kibana-related indices with a long suffix like 8.16.0_001, but we don't support it. Is there any reason not to do it?
image

coutoPL: IMO we can leave it as it is. There is big deal IMO. AFAIR we have in the backlog some jira with support datastreams in kibana index/indinces

Dzuming: I'm not sure if I understand you response, so I will reformulate my question.

1. In the case of all .kibana-based indices, there are three options to call them, 
1) without suffix, for example `.kibana`
2) with a short suffix, for example `.kibana_8.16.0`
3) With a long suffix, for example `.kibana_8.16.0_001`

In the current state, in some cases, we handle a long suffix (for example `"""^_alerting_cases_\d+\.\d+\.\d+_\d+$""".r, // eg. .kibana_alerting_cases_8.11.3_001`), and in other we not (for example usage_counters index). So in the current situation when I try to get for example alerting_cases, it's ok

image


however, in the case of ingest, there is a 403 error

image

Shouldn't we be consistent in both cases?

PR #1072, opened 2025-01-25
[RORDEV-1268] Added group rules NOT_ANY_OF and NOT_ALL_OF

mgoworko:


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -231,7 +232,8 @@ abstract class BaseGroupsRule(val settings: Settings)
     val externalAvailableGroups = sourceBlockContext.userMetadata.availableGroups
     for {
       externalGroupsMappedToLocalGroups <- mapExternalGroupsToLocalGroups(groupMappings, externalAvailableGroups)
-      availableLocalGroups <- calculateAllowedGroupsForUser(potentiallyAvailableGroups, PermittedGroupIds(externalGroupsMappedToLocalGroups))
+      potentiallyPermitted = GroupIds(externalGroupsMappedToLocalGroups)
+      availableLocalGroups <- GroupsLogic.Or(potentiallyPermitted).availableGroupsFrom(potentiallyAvailableGroups)
```
mgoworko, line 236: This change restores an original implementation. 
- original implementation was an instersect on sets
- later it was modfied to use `calculateAllowedGroupsForUser` ot the specific rule (in my opinion it was not necessary; it was part of a change that introduced AND rule, but IMO it did not have any effect on the final result, tests detected no changes)
- restored in this PR as GroupsLogic.Or (because simple intersect, as in original impl will no longer work, much changed)
- this line needs second opinion

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -42,7 +42,9 @@ abstract class BaseGroupsRule(val settings: Settings)
     with Logging {
 
   protected def calculateAllowedGroupsForUser(userGroups: UniqueNonEmptyList[Group],
-                                              permittedGroupIds: PermittedGroupIds): Option[UniqueNonEmptyList[Group]]
+                                              groupIds: GroupIds): Option[UniqueNonEmptyList[Group]]
```
mgoworko, line 45: coutoPL comment migrated from old PR:

_**at the moment it's hard to guess what the groupIds is. Previously, here we had user groups and permitted groups. It was rather easy to distinguish between these two. IMO we should find a better name for groupIds param (and the type too).**_

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/BlockContext.scala
```diff
@@ -36,15 +36,20 @@ sealed trait BlockContext {
 
   def responseTransformations: List[ResponseTransformation]
 
-  def isCurrentGroupEligible(permittedGroupIds: PermittedGroupIds): Boolean = {
-    userMetadata.currentGroupId match {
-      case Some(preferredGroupId) =>
-        requestContext.uriPath.isCurrentUserMetadataPath || permittedGroupIds.matches(preferredGroupId)
-      case None =>
+  def isCurrentGroupEligible(groupsPotentiallyPermittedByRule: GroupsPotentiallyPermittedByRule): Boolean = {
```
mgoworko, line 39: coutoPL comment migrated from old PR

I see some problem here. ROR request could have a header with the current group (the current group is related to the picked value in the tenancy selector value at the ROR KBN side).

Eg. GET /_search/index1 -H "x-ror-current-group: group1"

During block evaluation (the one with an authorization rule in it), ROR determines the allowed groups list (obviously, allowed groups can have not only full names groups, but also groups with wildcards).
The isCurrentGroupEligible method simply checks if the current group matches the allowed groups.

Currently, we deal here with potentially permitted groups. But we should deal with the already resolved and truly permitted groups. Maybe it's about the naming but seeing "All" allowed groups in this place, I'm pretty sure, sth is wrong here.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -58,8 +60,8 @@ abstract class BaseGroupsRule(val settings: Settings)
         resolveGroupIds(blockContext) match {
           case None =>
             Task.now(Rejected())
-          case Some(resolvedPermittedGroupIds) if blockContext.isCurrentGroupEligible(resolvedPermittedGroupIds) =>
-            continueCheckingWithUserDefinitions(blockContext, resolvedPermittedGroupIds)
+          case Some(groupIds) if blockContext.isCurrentGroupEligible(groupsPotentiallyPermittedByRule(groupIds)) =>
```
mgoworko, line 63: coutoPL comment migrated from old PR:

**_The previous name was a little bit better. In this class, groups can have several contexts: groups from the settings, groups from the rule, groups from the external system, and resolved groups (and maybe even more). Maybe, we should have more types to distinguish them. Anyway, when we call it groupIds we miss the information about the mentioned context.

The comment applies to other places like this in this file._**

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/ExternalAuthorizationRule.scala
```diff
@@ -78,7 +80,7 @@ object ExternalAuthorizationRule {
   }
 
   final case class Settings(service: ExternalAuthorizationService,
-                            permittedGroupsLogic: GroupsLogic,
+                            groupsLogic: GroupsLogic,
```
mgoworko, line 83: coutoPL comment migrated from old PR:

**_the same comment about naming and worsened context_**

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -138,32 +147,59 @@ object PermittedGroupIds {
...
+
+  sealed trait NegativeGroupsLogic extends GroupsLogic {
+    val forbiddenGroupIds: GroupIds
+  }
+
+  final case class NotAnyOf(override val forbiddenGroupIds: GroupIds) extends NegativeGroupsLogic
+
+  final case class NotAllOf(override val forbiddenGroupIds: GroupIds) extends NegativeGroupsLogic
+
+  sealed trait CombinedGroupsLogic extends GroupsLogic {
```
mgoworko, line 183: coutoPL comment migrated from old PR:

**_WDYT about defining it like this:

final case class CombinedGroupsLogic(permittedGroupsLogic: PositiveGroupsLogic, forbiddenGroupsLogic: NegativeGroupsLogic)
Can it be done like that?_**

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
mgoworko: Answered in other comment in BaseGroupsRule, quite huge refactor done, and there is now `groupsLogic` instead of `groupIds`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/ExternalAuthorizationRule.scala
mgoworko: In this place, I think that the name "groupsLogic" is correct. ExternalAuthorizationRule supports all available logics, including negative and combined. So I think the name had to be changed to `groupsLogic` because of added types of GroupsLogic

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
mgoworko: Done, it simplified the logic and allowed to combine nagative rules also with AND

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/BlockContext.scala
mgoworko: Splitted into 2 methods: 
- `def isCurrentGroupEligible` used to check if the groups is eligible based on list of permitted (as before this change)
- `def isCurrentGroupPotentiallyEligible` - used to reject non-eligible groups as soon as possible and continue checking only for potentially eligible. 

Those 2 different usages already existed before my changes, but there were no negative rules, so the same method served them.

IMO there was no error in logic before this new refactor, but it was indeed unclear and prone to future errors. Unless I don;t know something and there is still some error :D

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
mgoworko: I refactored it a bit
- introduced type `ResolvableGroupsLogic`, with implementations corresponding to `GroupsLogic`
- in BaseGroupsRule.Settings, now there is `resolvableGroupsLogic: ResolvableGroupsLogic` instead of `groupIds: ResolvableGroupIds`
- for example, calculateAllowedGroupsForUser and groupsPotentiallyPermittedByRule are now handled in the base class, because in settings we have more info about the logic, instead of raw list of group ids
- that way, much more logic is generic and defined in the base class `BaseGroupsRule`, instead of implementation for specific rules
- and there is less confusion about the names and purposes

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -69,18 +66,22 @@ abstract class BaseGroupsRule(val settings: Settings)
   override protected def authorize[B <: BlockContext : BlockContextUpdater](blockContext: B): Task[RuleResult[B]] =
     Task.now(RuleResult.Fulfilled(blockContext))
 
+  private def calculateAllowedGroupsForUser(userGroups: UniqueNonEmptyList[Group],
```
coutoPL, line 69: the first usage of this method is in L111. Can we move it after its first usage?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -280,13 +281,20 @@ abstract class BaseGroupsRule(val settings: Settings)
...
+  private def resolveGroupsLogic[B <: BlockContext](blockContext: B) = {
+    settings.resolvableGroupsLogic.resolve(blockContext)
   }
 }
 
 object BaseGroupsRule {
 
-  final case class Settings(permittedGroupIds: ResolvablePermittedGroupIds,
-                            usersDefinitions: NonEmptyList[UserDef])
+  final case class Settings[LOGIC <: ResolvableGroupsLogic](resolvableGroupsLogic: LOGIC,
```
coutoPL, line 291: Why do you think we should stop calling it the `permittedGroup` (I'm talking about the param name)?

I see the following reasons to don't do that:
1. there is no need to add info about the resolvability because the intention is expressed with type
2. in this file, when we talk about groups, we can have in mind:
  * permitted/allowed groups - the ones configured in the settings
  * available groups - the ones that the rule calculates to be available in the tenancy selector on the ROR KBN side
  * local groups - the ones defined in ROR settings
  * external groups - the ones that come from the external services (like LDAP)
So, saying, "here is a group logic", the next, natural question might be: "Which ones?" ;)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/ExternalAuthorizationRule.scala
coutoPL: I'm not sure if I agree. I mean, your explanation makes sense, but in this setting field can be described as: "This is the logic that is supposed to calculate the available groups".

So, our external connector says: `user1 has groups A, B, C`. The logic says: `any but D`. It allows us to calculate available groups (or permitted groups)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/ExternalAuthorizationRule.scala
```diff
@@ -35,19 +36,19 @@ class ExternalAuthorizationRule(val settings: ExternalAuthorizationRule.Settings
...
+  private val userMatcher = PatternsMatcher.create[User.Id](settings.users.toSet)
 
   override protected def loggedUserPreconditionCheck(user: LoggedUser): Either[Unit, Unit] = {
     Either.cond(userMatcher.`match`(user.id), (), ())
   }
 
   override protected def userGroups[B <: BlockContext](blockContext: B,
                                                        user: LoggedUser,
-                                                       permittedGroupIds: Set[GroupIdLike])
+                                                       groupsPotentiallyPermittedByRule: GroupsPotentiallyPermittedByRule)
```
coutoPL, line 51: I see that the `GroupsPotentiallyPermittedByRule' escapes the `BaseGroupsRule' context. That's suspicious. Together with the IllegalStateException exception thrown in the `LdapAuthorizationRule` class, makes me wonder if there is a problem here.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -197,9 +224,87 @@ object GroupsLogic {
...
+
+  implicit class CombinedGroupsLogicExecutor(val groupsLogic: GroupsLogic.CombinedGroupsLogic) extends AnyVal {
+    def availableGroupsFrom(userGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]] = {
+      groupsLogic.positiveGroupsLogic.availableGroupsFrom(userGroups)
+        .flatMap(groupsLogic.negativeGroupsLogic.availableGroupsFrom)
+    }
+  }
+}
+
+sealed trait ResolvableGroupsLogic {
```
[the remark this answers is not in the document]
mgoworko: I created GroupsLogicResolver interface, with implementations for all variants of logics. Current implementation is similar to the previous one, but I think that a bit better.

But I don't think that it can be defined exactly or almost exactly as GroupsLogicExecutor. The problem is, that we need to have the information about the specific type of GroupsLogic that will be resolved. The GroupsLogic is not an argument, but expected result. In GroupsLogicExecutor the situation is opposite - the GroupsLogic is an argument

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -27,23 +27,20 @@ import tech.beshu.ror.accesscontrol.blocks.rules.Rule.AuthenticationRule.Eligibl
...
 import tech.beshu.ror.accesscontrol.blocks.{BlockContext, BlockContextUpdater}
-import tech.beshu.ror.accesscontrol.domain.{Group, PermittedGroupIds, ResolvablePermittedGroupIds, User}
+import tech.beshu.ror.accesscontrol.domain.*
 import tech.beshu.ror.accesscontrol.matchers.GenericPatternMatcher
-import tech.beshu.ror.accesscontrol.orders.*
 import tech.beshu.ror.implicits.*
 import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
 
-abstract class BaseGroupsRule(val settings: Settings)
+abstract class BaseGroupsRule[LOGIC <: ResolvableGroupsLogic](val settings: Settings[LOGIC])
```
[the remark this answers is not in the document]
mgoworko: Generic type removed

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/BlockContext.scala
```diff
@@ -36,15 +36,24 @@ sealed trait BlockContext {
...
+  def isCurrentGroupEligible(permittedGroupIds: GroupIds): Boolean = {
     userMetadata.currentGroupId match {
       case Some(preferredGroupId) =>
         requestContext.uriPath.isCurrentUserMetadataPath || permittedGroupIds.matches(preferredGroupId)
       case None =>
         true
     }
   }
+
+  def isCurrentGroupPotentiallyEligible(groupsPotentiallyPermittedByRule: GroupsPotentiallyPermittedByRule): Boolean = {
```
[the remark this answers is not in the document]
mgoworko: `GroupsPotentiallyPermittedByRule` removed, described in comment in `ExternalAuthorizationRule`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
mgoworko: Method removed. It had single, simple usage

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/ExternalAuthorizationRule.scala
mgoworko: I now removed completely the `GroupsPotentiallyPermittedByRule`. All necessary information is already present inside `GroupsLogic`. There was no exception thrown in previous iteration, I changed it after dicsussions. Now changed back to implementation without exceptions (correct situations guaranteed at the type level of rule config)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -55,11 +52,11 @@ abstract class BaseGroupsRule(val settings: Settings)
     Task
       .unit
       .flatMap { _ =>
-        resolveGroupIds(blockContext) match {
+        resolveGroupsLogic(blockContext) match {
           case None =>
             Task.now(Rejected())
-          case Some(resolvedPermittedGroupIds) if blockContext.isCurrentGroupEligible(resolvedPermittedGroupIds) =>
-            continueCheckingWithUserDefinitions(blockContext, resolvedPermittedGroupIds)
+          case Some(groupsLogic) if blockContext.isCurrentGroupPotentiallyEligible(groupsLogic) =>
```
coutoPL, line 58: The `isCurrentGroupPotentiallyEligible` method has worried me since I first saw it in the code.
The BlockContext's `isCurrentGroupEligible` method is used to check if the current group matches one of the permitted groups (the available groups found in an authorization rule). And there is one exception - any current group is permitted for the CurrentUserMetadata request. (I've changed the implementation of the method a little bit to make it more visible).

In authorization rules, we used this method at the very beginning of the rule handling code, but it's for the sake of optimization. Let's consider the code:

```scala
override protected def authenticate[B <: BlockContext : BlockContextUpdater](blockContext: B): Task[RuleResult[B]] = {
    Task
      .unit
      .flatMap { _ =>
        resolveGroupIds(blockContext) match {
          case None =>
            Task.now(Rejected())
          case Some(resolvedPermittedGroupIds) if blockContext.isCurrentGroupEligible(resolvedPermittedGroupIds) =>
            continueCheckingWithUserDefinitions(blockContext, resolvedPermittedGroupIds)
          case Some(_) =>
            Task.now(Rejected())
        }
      }
  } 
```

We resolve the group logic and get the resolved permitted groups (from the logic). There is a complicated rule code ahead of us, but we know what groups are permitted (I'm talking about the old code, without the new "not" group logic), so we can check if the current group is eligible now. We don't have to calculate the available groups (eg. by fetching them from LDAP and executing the logic on them before we check). In the end, the available groups won't contain anything more than we already have in the permitted groups list. 

But the situation is different since we introduced the "not" group logic. As you noticed (that's why you named it "potentially eligible") we cannot check the current groups eligibility now because we could have "not" group logic. 

But we missed the checking at the end (when the available groups are calculated) - I added a test that shows this problem. Please take a look.

I'd propose to get rid of the optimization at the moment. Let's not use the `isCurrentGroupPotentiallyEligible` method and move the current group eligibility checking to the end of the rule business code. Then we will decide if it's worth optimizing or if it's better to have a less optimal, but more readable and less fragile (or error-prone) implementation,

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
mgoworko: I'm not sure I get it. In the test that You added the rule (NOT_ALL_OF) can be written as:
```
NOT ( g1 AND g2)
```

User has only one group: g1, so the final result is 

```
NOT(g1 AND g2) = NOT(TRUE AND FALSE) = NOT(FALSE) = TRUE
```

So the rule passes for user with single group - g1. This new test expects the rule to fail, but IMO it correctly passes.


In the GroupsNotAnyOfRuleTests (for NOT_ANY_OF rule) there already exists a test, that is exactly the same as the one You added for NOT_ALL_OF ("A GroupsNotAnyOfRule" should "not match" when "user has one forbidden group"). And in that test the NOT_ANY_OF rule correctly rejects user. But NOT_ALL_OF in my understanding works correctly.

coutoPL: The group logic is ok. I just wanted to show that it doesnt take into consideration the current group.

mgoworko: You are right, there was missing check against the `availableGroups` based on UserDef. It was missing only for `Mode.WithoutGroupsMapping`. 

For the other mode, `Mode.WithGroupsMapping`, this check seems to be present in the line that we discussed before  `availableLocalGroups <- GroupsLogic.Or(potentiallyPermitted).availableGroupsFrom(potentiallyAvailableGroups)` (intersect of available and external mapped to local)

Additionally, the tests are performed using `/_readonlyrest/metadata/current_user/`, I changed it to `/_readonlyrest/admin/audit/event/` for negative logic tests.

https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1072/commits/097c000334daf18cc7dedceb2b1728b129cd2ea6

mgoworko: Removing the optimization (isCurrentGroupPotentiallyEligible) won't fix the problem without some kind of fix of the underlying issue. So maybe let's not remove the optimization right now?

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -197,9 +225,104 @@ object GroupsLogic {
...
+  }
+
+  implicit class CombinedGroupsLogicExecutor(val groupsLogic: GroupsLogic.CombinedGroupsLogic) extends AnyVal {
+    def availableGroupsFrom(userGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]] = {
+      groupsLogic.positiveGroupsLogic.availableGroupsFrom(userGroups)
+        .flatMap(groupsLogic.negativeGroupsLogic.availableGroupsFrom)
+    }
+  }
+
+  sealed trait GroupsLogicResolver {
```
coutoPL, line 258: I have a feeling we don't need this thing. At least not in so specialized form.

We have `RuntimeSingleResolvableVariable[T]` & `RuntimeMultiResolvableVariable[T]`.

The first one creates an object of T from String (eg. it could be "group_@{user}" - it takes the user name from the block context and creates eg. "group_user1" group name).

The second one creates a nonempty list of T objects from String (eg. it could be `group_@explode{env:test-multi1}` - it takes the test-multi1 env and takes it value (it's assumed it's a comma-separated string IIRC) and creates eg. "group_value1", "group_value2" groups.

Now, we have GroupsLogic which can be seen as a container for groups (putting aside the logic it carries but concentrating only on the data type). From String, we have to create one GroupLogic, which holds several groups. 

I thought about Functor for the RuntimeMultiResolvableVariable, but our case is a little bit different. The `RuntimeMultiResolvableVariable` is defined like that:

```scala
private[runtime] trait RuntimeResolvableVariable[VALUE] {

  def resolve(blockContext: BlockContext): Either[Unresolvable, VALUE]
}

sealed trait RuntimeMultiResolvableVariable[T] extends RuntimeResolvableVariable[NonEmptyList[T]]
```

So, in this case, the String will be resolved to NonEmptyList[T]. We need it to be resolved to sth, which is a container for T (aka package). 

Could you please take a look at it and try to do sth I described? 
It will be generic solution that we could use in this task and in similar places in the future.

mgoworko: The reason why it was so complicated, was the crude support for CombinedGroupsLogic. I refactored it and made it much more simple. In the next task/PR it will be possible to add decoding and handling of combined logic.

Description of how I see it:

1. Before this PR, the data and logic of the BaseGroupsRule was spread between 3 places
- a raw list of resolvable group ids was decoded in `BaseGroupsRuleDecoder`, but without information what this list means
- it was passed in base rule settings, and only resolved during the generic `BaseGroupsRule` handling
- but the final logic of OR/AND was applied in the specific implementation GroupsOrRule/GroupsAndRule

2. In this PR:
- the information about the rule - both data and logic - is created early, when decoding, and then passed to the BaseGroupsRule
- the logic and data is contained in single entity - the `GroupsLogicResolver`, which is resolved during the execution of the rule

3. So:
- the GroupsLogicResolver is a wrapper for `RuntimeResolvableVariable` that contains data, and for the logic (information of how to transform the resolved data into the executable GroupsLogic) 
- I tried various approaches, but I don't see a way of representing info about both data and behavior on the level of `RuntimeResolvableVariable`
- You said to put aside logic and focus on data - but AFAIK the resolvable variables need to be wrapped into other utils/classes in order to add meaning to the underlying data. Before this change that meaning was added in the AND/OR rules, but it is not possible now because of Combined logic.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -197,9 +225,104 @@ object GroupsLogic {
...
+
+    final case class ForNotAnyOfGroupsLogic(permittedResolvableGroupIds: ResolvableGroupIds) extends NegativeGroupsLogicResolver {
+      override def resolve[B <: BlockContext](blockContext: B): Option[NotAnyOf] =
+        permittedResolvableGroupIds.resolveGroupIds(blockContext).map(NotAnyOf.apply)
+
+      override def usedVariables: NonEmptyList[RuntimeMultiResolvableVariable[GroupIdLike]] =
+        permittedResolvableGroupIds.groupIds.toNonEmptyList
+    }
+
+    final case class ForCombinedGroupsLogic[
```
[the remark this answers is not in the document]
mgoworko: The entire `GroupsLogicResolver` refactored into

```
  sealed trait GroupsLogicResolver {
    def resolve[B <: BlockContext](blockContext: B): Option[GroupsLogic]

    def usedVariables: NonEmptyList[RuntimeMultiResolvableVariable[GroupIdLike]]
  }

  object GroupsLogicResolver {
    def apply(groupIds: ResolvableGroupIds,
              creator: GroupIds => GroupsLogic): GroupsLogicResolver = new GroupsLogicResolver {
      override def resolve[B <: BlockContext](blockContext: B): Option[GroupsLogic] =
        groupIds.resolveGroupIds(blockContext).map(creator)

      override def usedVariables: NonEmptyList[RuntimeMultiResolvableVariable[GroupIdLike]] =
        groupIds.groupIds.toNonEmptyList
    }
  }
```

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -275,18 +271,24 @@ abstract class BaseGroupsRule(val settings: Settings)
...
 }
 
 object BaseGroupsRule {
 
-  final case class Settings(permittedGroupIds: ResolvablePermittedGroupIds,
+  final case class Settings(permittedGroupsLogicResolver: GroupsLogicResolver,
                             usersDefinitions: NonEmptyList[UserDef])
+
+  object Settings {
+    def apply(groupIds: ResolvableGroupIds,
```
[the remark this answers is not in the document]
mgoworko: Was used in tests. I adjusted tests and removed it.

coutoPL (changes requested):
1. could you please adapt tests if the proposition is ok for you?
2. please take a look at failing tests (before my changes there were some failed tests)
3. please bump pre version
4. did you remove the test I've added before? don't we need it?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -70,17 +67,17 @@ abstract class BaseGroupsRule(val settings: Settings)
     Task.now(RuleResult.Fulfilled(blockContext))
 
   private def continueCheckingWithUserDefinitions[B <: BlockContext : BlockContextUpdater](blockContext: B,
-                                                                                           resolvedPermittedGroupIds: PermittedGroupIds): Task[RuleResult[B]] = {
+                                                                                           groupsLogic: GroupsLogic): Task[RuleResult[B]] = {
```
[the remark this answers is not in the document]
mgoworko: Done

Answering the points 1-4 from the comment:
1. Tests adapted
2. All tests seem to pass now after changes
3. Pre version bumped
4. I did not remove the test - I renamed it and moved it to BaseGroupsRuleTests. It is now executed for all types of rules. I also added second test for situation, where groups mapping is configured. 

Also discussed outisde PR - test for no UserDefs returned from, LDAP or external service. We have such test - "user doesn't have any permitted group" in `LdapAuthRuleTests`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -275,18 +271,16 @@ abstract class BaseGroupsRule(val settings: Settings)
               .filter(m => m.externalGroupIdPatternsMatcher.`match`(externalGroup.id))
               .map(_.local)
           }
-          .map(_.id)
```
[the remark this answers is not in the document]
mgoworko: Ok, I reverted it and restored comparing by groupId alone. 

I wanted to revert to the oldest original implementation of intersect 
`externalGroupsMappedToLocalGroups.toSet.intersect(potentiallyAvailableGroups.toSet)`

Now it has to use GroupsLogic.Or (because it already implements logic of matching groups and groupIds)
`GroupsLogic.Or(potentiallyPermitted).availableGroupsFrom(potentiallyAvailableGroups)`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseAuthorizationRule.scala
```diff
@@ -98,9 +97,9 @@ private[auth] trait BaseAuthorizationRule
 
   private def authorizeLoggedUser[B <: BlockContext : BlockContextUpdater](blockContext: B,
                                                                            user: LoggedUser,
-                                                                           userGroupsProvider: (B, LoggedUser, Set[GroupIdLike]) => Task[UniqueList[Group]]): Task[RuleResult[B]] = {
-    if (blockContext.isCurrentGroupEligible(groupsPermittedByRule)) {
-      userGroupsProvider(blockContext, user, groupsPermittedByRule.groupIds.toSet)
+                                                                           userGroupsProvider: (B, LoggedUser) => Task[UniqueList[Group]]): Task[RuleResult[B]] = {
+    if (blockContext.isCurrentGroupPotentiallyEligible(groupsLogic)) {
```
coutoPL, line 101: I wonder if in this place we don't have the same issue we had in the case of the `BaseGroupsRule` class. We checked if the current group is potentially eligible but not if it's really eligible in the case of the "not" group logic.

Could you please write a test to prove the issue (or prove there is no such issue) too?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableGroupsLogic.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+package tech.beshu.ror.accesscontrol.blocks.variables.runtime
+
+import cats.data.NonEmptyList
+import tech.beshu.ror.accesscontrol.blocks.BlockContext
+import tech.beshu.ror.accesscontrol.domain.{GroupIdLike, GroupIds, GroupsLogic}
+import tech.beshu.ror.accesscontrol.utils.RuntimeMultiResolvableVariableOps.resolveAll
+import tech.beshu.ror.utils.uniquelist.UniqueNonEmptyList
+
+case class RuntimeResolvableGroupsLogic[GL <: GroupsLogic](groupIds: UniqueNonEmptyList[RuntimeMultiResolvableVariable[GroupIdLike]],
```
coutoPL, line 25: it's not a data class because of the function `GroupIds => GL` as a param

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsAndRule.scala
```diff
@@ -18,20 +18,14 @@ package tech.beshu.ror.accesscontrol.blocks.rules.auth
 
 import tech.beshu.ror.accesscontrol.blocks.rules.Rule
 import tech.beshu.ror.accesscontrol.blocks.rules.Rule.{Name, RuleName}
-import tech.beshu.ror.accesscontrol.domain.{CaseSensitivity, Group, GroupsLogic, PermittedGroupIds}
-import tech.beshu.ror.utils.uniquelist.UniqueNonEmptyList
+import tech.beshu.ror.accesscontrol.domain.{CaseSensitivity, GroupsLogic}
 
-final class GroupsAndRule(override val settings: BaseGroupsRule.Settings,
+final class GroupsAndRule(override val settings: BaseGroupsRule.Settings[GroupsLogic.And],
```
coutoPL, line 23: Please change the name to `Rule.Name("groups_all_of")` (it will be our default starting now). The "group_and" we will treat as a legacy name. 

Let's do the same for the GroupsOrRule

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/LdapRulesDecoders.scala
```diff
@@ -109,22 +109,40 @@ class LdapAuthorizationRuleDecoder(ldapDefinitions: Definitions[LdapService],
...
-        case (_, _, None, None) =>
+        case (_, _, None, None, None, None) =>
           Left(RulesLevelCreationError(Message(errorMsgNoGroupsList(LdapAuthorizationRule.Name))))
-        case (_, _, Some(_), Some(_)) =>
-          Left(RulesLevelCreationError(Message(errorMsgOnlyOneGroupsList(LdapAuthorizationRule.Name))))
-        case (name, ttl, Some(groupsOr), None) =>
+        case (name, ttl, Some(groupsOr), None, None, None) =>
           createLdapAuthorizationRule(name, ttl, groupsOr, ldapDefinitions)
-        case (name, ttl, None, Some(groupsAnd)) =>
+        case (name, ttl, None, Some(groupsAnd), None, None) =>
```
coutoPL, line 126: oh, it's getting complicated. Are we able to somehow extract the groups logic parsing to a separate, private method?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/LdapRulesDecoders.scala
```diff
@@ -159,22 +177,40 @@ class LdapAuthRuleDecoder(ldapDefinitions: Definitions[LdapService],
...
-        case (_, _, None, None) =>
+        case (_, _, None, None, None, None) =>
           Left(RulesLevelCreationError(Message(errorMsgNoGroupsList(LdapAuthRule.Name))))
-        case (_, _, Some(_), Some(_)) =>
-          Left(RulesLevelCreationError(Message(errorMsgOnlyOneGroupsList(LdapAuthRule.Name))))
-        case (name, ttl, Some(groupsOr), None) =>
+        case (name, ttl, Some(groupsOr), None, None, None) =>
           createLdapAuthRule(name, ttl, groupsOr, ldapDefinitions, impersonatorsDef, mocksProvider, userIdCaseSensitivity)
-        case (name, ttl, None, Some(groupsAnd)) =>
+        case (name, ttl, None, Some(groupsAnd), None, None) =>
```
coutoPL, line 194: the same here

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/LdapRulesDecoders.scala
```diff
@@ -221,6 +255,10 @@ private object LdapRulesDecodersHelper {
     s"${ruleName.show} rule requires to define 'groups_or'/'groups' or 'groups_and' arrays (but not both)"
   }
 
+  private[rules] def negativeGroupsRuleNotAllowedForLdapWithGroupsFiltering = {
+    s"It is not allowed to use groups_not_any_of and groups_not_all_of rule, when LDAP server-side groups filtering is enabled. Consider using a combined rule, which applies two rules at the same time: groups_or/groups_and together with groups_not_any_of/groups_not_all_of."
```
coutoPL, line 259: Do we support the combined rule in this change? 
If not, let's change the message for now.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/GroupsAndRuleTests.scala
```diff
@@ -17,29 +17,33 @@
...
 
-class GroupsAndRuleTests extends BaseGroupsRuleTests {
+class GroupsAndRuleTests extends BaseGroupsPositiveRuleTests[GroupsLogic.And] {
 
-  override def createRule(settings: GroupsRulesSettings, caseSensitivity: CaseSensitivity): BaseGroupsRule = {
+  override def createRule(settings: GroupsRulesSettings[GroupsLogic.And], caseSensitivity: CaseSensitivity): BaseGroupsRule[GroupsLogic.And] = {
     new GroupsAndRule(settings, caseSensitivity)
   }
 
+  override def resolvableGroupsLogic(groupIds: UniqueNonEmptyList[RuntimeMultiResolvableVariable[GroupIdLike]]): RuntimeResolvableGroupsLogic[GroupsLogic.And] = {
```
coutoPL, line 38: This method is almost the same in each of the implementation classes. It seems we need only the creator of the logic from the resolvable group list

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/BaseGroupsRule.scala
```diff
@@ -231,7 +227,8 @@ abstract class BaseGroupsRule(val settings: Settings)
     val externalAvailableGroups = sourceBlockContext.userMetadata.availableGroups
     for {
       externalGroupsMappedToLocalGroups <- mapExternalGroupsToLocalGroups(groupMappings, externalAvailableGroups)
-      availableLocalGroups <- calculateAllowedGroupsForUser(potentiallyAvailableGroups, PermittedGroupIds(externalGroupsMappedToLocalGroups))
+      potentiallyPermitted = GroupIds(externalGroupsMappedToLocalGroups)
+      availableLocalGroups <- GroupsLogic.Or(potentiallyPermitted).availableGroupsFrom(potentiallyAvailableGroups)
```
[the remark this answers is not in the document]
mgoworko: done, `availableLocalGroupsFromExternalGroupsMappedToLocalGroups`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/LdapAuthorizationRule.scala
```diff
@@ -37,20 +37,33 @@ class LdapAuthorizationRule(val settings: Settings,
...
+        ldap match {
+          case ldap: LdapAuthorizationService.WithoutGroupsFiltering =>
+            ldap.groupsOf(user.id)
+          case ldap: LdapAuthorizationService.WithGroupsFiltering =>
+            ldap.groupsOf(user.id, groupsLogic.positiveGroupsLogic.permittedGroupIds.groupIds.toSet)
+        }
     }
   }
 
   override protected def calculateAllowedGroupsForUser(usersGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]] =
```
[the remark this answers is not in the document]
mgoworko: Done. No longer exposed, there was the same implementation for both External and Ldap rules

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/LdapAuthorizationRule.scala
```diff
@@ -76,6 +89,26 @@ object LdapAuthorizationRule {
...
+    final case class NegativeGroupsLogicSettings(ldap: LdapAuthorizationService.WithoutGroupsFiltering,
+                                                 groupsLogic: GroupsLogic.NegativeGroupsLogic) extends Settings
+
+    final case class PositiveGroupsLogicSettings(ldap: LdapAuthorizationService,
+                                                 groupsLogic: GroupsLogic.PositiveGroupsLogic) extends Settings
+
+    final case class CombinedGroupsLogicSettings(ldap: LdapAuthorizationService,
+                                                 groupsLogic: GroupsLogic.CombinedGroupsLogic) extends Settings
+
+    def apply(ldap: LdapAuthorizationService,
```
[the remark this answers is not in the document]
mgoworko: Removed. Similar create methods added in test files where they were used.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseAuthorizationRule.scala
mgoworko: Yes, there was an issue with `preferredGroupId`, again not detected by tests. I added 8 tests in `LdapAuthorizationRuleTests`, two of them specifically for this case.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/RuntimeResolvableGroupsLogic.scala
mgoworko: I made it a case class in order to provide equals method. I noticed it too and reverted comparing instances of this class in tests, but forgot to revert making it  `case`. Reverted.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsAndRule.scala
mgoworko: I will be a part of task RORDEV-1388. It needs to be changed in multiple places (everywhere, where the current deprecated name of OR rule is used).

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/LdapRulesDecoders.scala
mgoworko: Done. It was possible to extract common decoder for GroupsLogic for both Authorization and Auth decoders

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/LdapRulesDecoders.scala
mgoworko: The combined rule is supported for LDAP. Just without the new config syntax (making a config section for both lists), which will be added in RORDEV-1388.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/GroupsAndRuleTests.scala
mgoworko: Done, also for `GroupsRuleSettingsTests`

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -115,19 +120,20 @@ object GroupIdLike {
   }
 }
 
-final case class PermittedGroupIds(groupIds: UniqueNonEmptyList[_ <: GroupIdLike]) {
-  private[PermittedGroupIds] lazy val matcher = PatternsMatcher.create[GroupIdLike](groupIds)
+final case class GroupIds(groupIds: UniqueNonEmptyList[_ <: GroupIdLike]) {
```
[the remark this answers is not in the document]
mgoworko: ok, that's better than having `groupIds.groupIds`

coutoPL (approved):
LGTM

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -138,32 +144,42 @@ object PermittedGroupIds {
...
+  sealed trait NegativeGroupsLogic extends GroupsLogic {
+    val forbiddenGroupIds: GroupIds
+  }
+
+  final case class NotAnyOf(override val forbiddenGroupIds: GroupIds) extends NegativeGroupsLogic
+
+  final case class NotAllOf(override val forbiddenGroupIds: GroupIds) extends NegativeGroupsLogic
+
+  final case class CombinedGroupsLogic(positiveGroupsLogic: PositiveGroupsLogic,
+                                       negativeGroupsLogic: NegativeGroupsLogic) extends GroupsLogic
```
mateuszkp96, line 173: Let's add some tests for CombinedGroupsLogic usage within AuthorizationRules. Currently, only the settings parsing is tested.

mateuszkp96 (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
mgoworko: Added tests. One invalid situation was detected, I modified implementation for CombinedGroupsLogic 

Also, I extracted common decoder for GroupsLogic for all Auth rules.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -197,9 +213,39 @@ object GroupsLogic {
...
+  }
+
+  implicit class GroupsLogicNotAnyOfExecutor(val groupsLogic: GroupsLogic.NotAnyOf) extends AnyVal {
+    def availableGroupsFrom(userGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]] = {
+      val forbiddenGroupDetected = groupsLogic.forbiddenGroupIds.ids.toList.widen[GroupIdLike].foldLeft(false) {
+        case (false, forbiddenGroup) =>
+          userGroups.toList.exists(userGroup => forbiddenGroup.matches(userGroup))
+        case (true, _) =>
+          true
+      }
```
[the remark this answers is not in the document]
mgoworko: done

mateuszkp96 (changes requested):

mateuszkp96 (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/GroupsLogicDecoder.scala
```diff
@@ -0,0 +1,112 @@
...
+    val (cursor, key) = c.downFieldsWithKey(field, fields: _*)
+    cursor.as[Option[GL]].map(_.map((_, key)))
+  }
+
+  private[rules] def errorMsgNoGroupsList[R <: Rule](ruleName: RuleName[R]) = {
+    s"${ruleName.show} rule requires to define 'groups_or'/'groups'/'groups_and'/'groups_not_all_of' arrays"
+  }
+
+  private[rules] def errorMsgOnlyOneGroupsList[R <: Rule](ruleName: RuleName[R]) = {
+    s"${ruleName.show} rule requires to define 'groups_or'/'groups'/'groups_and'/'groups_not_all_of' arrays (but not all)"
```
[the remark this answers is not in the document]
mgoworko: Removed legacy `groups` from message, added missing name of the new rule

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/userAndGroups.scala
```diff
@@ -197,9 +213,40 @@ object GroupsLogic {
       UniqueNonEmptyList.from(someMatchedUserGroups)
     }
   }
+
+  implicit class GroupsLogicNotAllOfExecutor(val groupsLogic: GroupsLogic.NotAllOf) extends AnyVal {
+    def availableGroupsFrom(userGroups: UniqueNonEmptyList[Group]): Option[UniqueNonEmptyList[Group]] = {
+      val matchedNotPermittedGroups = groupsLogic.forbiddenGroupIds.filterOnlyPermitted(userGroups)
+      val allForbiddenGroupsDetected = matchedNotPermittedGroups.size == groupsLogic.forbiddenGroupIds.ids.size
+      if (allForbiddenGroupsDetected) None
```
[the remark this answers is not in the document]
mgoworko: Great catch. I added 2 tests that detected this issue and changed implementation.

mateuszkp96 (approved):
LGTM

PR #1076, opened 2025-02-12
[RORDEV-1263] Data stream support in ROR audit

mateuszkp96:


core/src/main/scala/tech/beshu/ror/accesscontrol/audit/DataStreamAuditSinkCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  private def setupDataStream(service: DataStreamService, settings: DataStreamSettings): Task[Unit] = {
+    for {
+      _ <- Task.delay(logger.info(s"Trying to setup ROR audit sink with data stream ${settings.dataStreamName.show}.."))
+      _ <- service.createIndexLifecyclePolicy(settings.lifecyclePolicy, settings.lifecyclePolicyJson(service.capabilities.ilmMaxPrimaryShardSize))
+      _ <- service.createComponentTemplateForMappings(settings.componentTemplateMappingsId, settings.mappings, settings.mappingsMetadata)
+      _ <- service.createComponentTemplateForIndex(settings.componentTemplateSettingsId, settings.lifecyclePolicy, settings.indexSettingsMetadata)
+      _ <- service.createIndexTemplate(settings.indexTemplate, settings.dataStreamName, NonEmptyList.of(settings.componentTemplateMappingsId, settings.componentTemplateSettingsId), settings.indexTemplateMetadata)
+      _ <- service.createDataStream(settings.dataStreamName)
+      _ <- Task.delay(logger.info(s"ROR audit data stream ${settings.dataStreamName.show} created."))
```
mateuszkp96, line 56: Here, we should handle ACKs from service responses. We will probably need to query the resources to ensure they are available. Some of the queries would require new methods based on reflection.
Now, if the request fails on any step, it cannot be resumed (we need something like creating resources if not exist for every one of the given resources)

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/DataStreamAuditSinkCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+      )
+    }
+
+    val indexTemplateMetadata: Map[String, String] = {
+      Map(
+        "description" -> "Index template for ReadonlyREST audit data stream"
+      )
+    }
+
+    def lifecyclePolicyJson(maxPrimaryShardSizeInRollover: Boolean): Value = {
```
coutoPL, line 100: I'm not sure if we should do it like this. We define it at a core level, but it may change in the future. Maintaining compatibility at the core level will be difficult. At the same time, we have modules, where the compatibility is easier to maintain.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/DataStreamAuditSinkCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+
+    def indexTemplate: TemplateName = templateNameFrom(s"${dataStreamName.value.value}-template")
+
+    private def templateNameFrom(value: String) = {
+      TemplateName
+        .fromString(value)
+        .getOrElse(throw new IllegalStateException("Template name should be non-empty"))
+    }
+
+    val mappings: Value = serializeMappings(documentMappings)
```
coutoPL, line 80: AFAIR in case of the ROR audit index we use dynamic mapping, don't we? 
Shouldn't we do it in the data stream case in the same way?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/DataStreamAuditSinkCreator.scala
mateuszkp96: Let's assume that creating an index lifecycle request returned the following response:
```
{
  "acknowledged": false
}
```
This means that the master node did not confirm the request. As a result, the index lifecycle policy was not successfully registered in the cluster.
In the next request, we attempt to create a component template that references this lifecycle policy. However, since the master node is unaware of the index lifecycle policy, the request fails.

If you retry to save ROR settings again, you send a creation request for the lifecycle policy and you can get a conflict. The policy name was created in a previous attempt. This leaves us stuck with an index lifecycle policy but without a data stream.

To avoid this, we should proceed with further requests only after the lifecycle policy is confirmed by the master node (we should check the acknowledge response). Or if we resume the creation process, we should skip the request if the lifecycle policy already exists.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/EsDataStreamBasedAuditSink.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  private def mappingsForSerializer(serializer: AuditLogSerializer): Map[String, FieldType] = {
+    serializer match {
+      case _: QueryAuditLogSerializer => QueryAuditLogSerializer.defaultIndexedMappings
+      case _: DefaultAuditLogSerializer => DefaultAuditLogSerializer.defaultIndexedMappings
+      case _ =>
+      // ES data streams require the timestamp field.
+      // It would be hard to pass the additional field mappings through the ROR yaml config, so only the timestamp field is here
+      Map("@timestamp" -> FieldType.Date)
```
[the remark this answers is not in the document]
mateuszkp96: I left only dynamic mappings, like in the index-based approach. These mappings are available on the backing indices level. If the customer would like to use them on the data stream level (to query by the data stream name instead of backing indices), it seems the mappings need to be static.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/indices.scala
```diff
@@ -538,3 +539,60 @@ object RorAuditIndexTemplate {
...
+  def apply(name: NonEmptyString): Either[CreationError, RorAuditDataStream] = from(name)
+
+  def from(name: NonEmptyString): Either[CreationError, RorAuditDataStream] = {
+    validateFormat(name)
+      .map(DataStreamName.Full.fromNes)
+      .map(RorAuditDataStream.apply)
+      .leftWiden[CreationError]
+  }
+
+  private def validateFormat(value: NonEmptyString): Either[CreationError.FormatError, NonEmptyString] = {
```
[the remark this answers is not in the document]
mateuszkp96: I agree, but we are creating a lifecycle policy, component, and index templates with IDs that do not have such format constraints. Without checking the format we can left with some trash because the data stream creation step is the last one.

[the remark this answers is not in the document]
coutoPL: ok, good point

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/EsDataStreamBasedAuditSink.scala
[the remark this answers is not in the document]
coutoPL: does it mean the user won't be able to do sth like this:
```
GET readonlyrest_audit/_search
{
  "query": {
    "range": {
      "@timestamp": {
        "gte": "now-1d/d",
        "lt": "now/d"
      }
    }
  }
}
```

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -263,21 +266,28 @@ class ReadonlyRest(coreFactory: CoreFactory,
 }
 
 object ReadonlyRest {
-  type AuditSinkCreator = AuditCluster => AuditSinkService
+  trait AuditSinkCreator {
```
coutoPL, line 269: I wonder if this trait should be defined in the `tech.beshu.ror.accesscontrol.audit` package. WDYT?

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -198,44 +198,47 @@ class ReadonlyRest(coreFactory: CoreFactory,
...
+        )
+      }
   }
 
   private def createAuditingTool(core: Core)
-                                (implicit loggingContext: LoggingContext): Option[AuditingTool] = {
+                                (implicit loggingContext: LoggingContext): Task[Option[AuditingTool]] = {
     core.rorConfig.auditingSettings
-      .flatMap(settings => AuditingTool.create(settings, auditSinkCreator)(environmentConfig.clock, loggingContext))
+      .map(settings => AuditingTool.create(settings, auditSinkCreator.index, auditSinkCreator.dataStream)(using environmentConfig.clock, loggingContext))
```
coutoPL, line 239: maybe we could pass here `AuditSinkCreator` instead of extracting data stream and index creators at this level. WDYT?

core/src/main/scala/tech/beshu/ror/utils/DurationOps.scala
```diff
@@ -28,13 +28,28 @@ object DurationOps {
   implicit class RefinedDurationOps(val duration: Duration) extends AnyVal {
     def toRefinedPositive: Either[String, PositiveFiniteDuration] = duration match {
       case v: FiniteDuration if v.toMillis > 0 =>
+        v.toCoarsest
```
coutoPL, line 31: to remove?

core/src/main/scala/tech/beshu/ror/utils/DurationOps.scala
```diff
@@ -28,13 +28,28 @@ object DurationOps {
...
         Left(s"Cannot map '${duration.toString}' to finite duration.")
     }
 
     def toRefinedPositiveUnsafe: PositiveFiniteDuration =
       toRefinedPositive.fold(err => throw new IllegalArgumentException(err), identity)
+
+    def inShortFormat: String = {
+      val d = duration.toCoarsest
+      val unit = d.unit match {
+        case java.util.concurrent.TimeUnit.DAYS => "d"
```
coutoPL, line 43: can we import java.util.concurrent.TimeUnit as JTimeUnit?

core/src/test/scala/tech/beshu/ror/unit/acl/logging/AuditingToolTests.scala
```diff
@@ -61,39 +61,43 @@ class AuditingToolTests extends AnyWordSpec with MockFactory with BeforeAndAfter
           "request was allowed and verbosity level was ERROR" in {
             val auditingTool = AuditingTool.create(
               settings = auditSettings(new DefaultAuditLogSerializer),
-              auditSinkServiceCreator = _ => mock[AuditSinkService]
-            ).get
+              indexAuditSinkCreator = _ => mock[IndexBasedAuditSinkService],
```
coutoPL, line 64: shouldn't we add datastream-related tests in this suite?

core/src/test/scala/tech/beshu/ror/unit/boot/ReadonlyRestStartingTests.scala
```diff
@@ -1354,16 +1352,16 @@ class ReadonlyRestStartingTests
         mapWithIntervalFrom(refreshInterval) ++
           mapWithMaxYamlSize(maxYamlSize) ++
           Map(
-            "com.readonlyrest.settings.loading.delay" -> "0"
+            "com.readonlyrest.settings.loading.delay" -> "1"
```
coutoPL, line 1355: is it intentional?

core/src/test/scala/tech/beshu/ror/unit/boot/RorIndexTest.scala
```diff
@@ -216,27 +216,20 @@ class RorIndexTest extends AnyWordSpec
...
-
+                               configPath: String) = {
     implicit val environmentConfig: EnvironmentConfig = new EnvironmentConfig(
       propertiesProvider = TestsPropertiesProvider.usingMap(
-        mapWithIntervalFrom(refreshInterval) ++
-          Map(
-            "com.readonlyrest.settings.loading.delay" -> "0"
-          )
+        Map(
+          "com.readonlyrest.settings.loading.delay" -> "1"
```
coutoPL, line 223: the same here

es67x/src/main/scala/tech/beshu/ror/es/IndexLevelActionFilter.scala
```diff
@@ -92,11 +92,16 @@ class IndexLevelActionFilter(nodeName: String,
     startRorInstance()
   }
 
-  private def auditSinkCreator: AuditSinkCreator = {
-    case AuditCluster.LocalAuditCluster =>
-      new EsAuditSinkService(client)
-    case remote: AuditCluster.RemoteAuditCluster =>
-      HighLevelClientAuditSinkService.create(remote)
+  private def auditSinkCreator: AuditSinkCreator = new AuditSinkCreator {
```
coutoPL, line 95: Maybe instead of throwing the illegal state ex, we can make this state impossible?
Eg. by having two AuditSinkCreator traits? One supporting only index sink. And the second one supporting both?

es717x/src/main/scala/tech/beshu/ror/es/services/EsAuditSinkService.scala
```diff
@@ -16,22 +16,25 @@
...
 import tech.beshu.ror.constants.{AUDIT_SINK_MAX_ITEMS, AUDIT_SINK_MAX_KB, AUDIT_SINK_MAX_RETRIES, AUDIT_SINK_MAX_SECONDS}
-import tech.beshu.ror.es.AuditSinkService
+import tech.beshu.ror.es.{DataStreamBasedAuditSinkService, IndexBasedAuditSinkService}
+import tech.beshu.ror.es.utils.XContentJsonParserFactory
 
-@Inject
-class EsAuditSinkService(client: Client,
-                         threadPool: ThreadPool)
-  extends AuditSinkService
+final class EsAuditSinkService(client: NodeClient, threadPool: ThreadPool, jsonParserFactory: XContentJsonParserFactory)
```
coutoPL, line 35: I wonder if we shouldn't rename it to `NodeClientBasedAuditSinkService`. WDYT?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/DataStreamAuditSinkCreator.scala
coutoPL: yes, IMO it's very important that all of these resource creation methods are idempotent. If the resource (eg. ilm policy exists, we should not create one).

core/src/test/scala/tech/beshu/ror/unit/boot/ReadonlyRestStartingTests.scala
mateuszkp96: It's intentional. The FiniteDuration type has been changed to the PositiveFiniteDuration. The value of 0 became invalid, so the default was used (5 seconds) in the tests. With this change, I reduced all unit test execution time by 5 minutes.

core/src/test/scala/tech/beshu/ror/unit/boot/RorIndexTest.scala
mateuszkp96: The same here

core/src/test/scala/tech/beshu/ror/unit/boot/ReadonlyRestStartingTests.scala
coutoPL: but with this little change, we change the interface too. Because some may have this configured to 0. Is it worth it?

mateuszkp96: The type was changed a few months ago, I only changed the value used in tests to shorten test execution.

integration-tests/src/test/resources/log4j2_es_7.10_and_newer.properties
```diff
@@ -85,4 +85,7 @@ appender.header_warning.type = HeaderWarningAppender
 appender.header_warning.name = header_warning
 
 logger.ror_ldap.name=tech.beshu.ror.accesscontrol.blocks.definitions.ldap.implementations
-logger.ror_ldap.level=debug
\ No newline at end of file
+logger.ror_ldap.level=debug
+
+logger.ror.name=tech.beshu.ror
+logger.ror.level=debug
```
coutoPL, line 91: should we keep it?

gradle.properties
```diff
@@ -1,5 +1,6 @@
 publishedPluginVersion=1.62.0
-pluginVersion=1.63.0-pre5
+pluginVersion=1.63.0-pre6
 pluginName=readonlyrest
 
 org.gradle.jvmargs=-Xmx6144m
+org.gradle.parallel=false
```
coutoPL, line 6: why do we need it?

coutoPL (approved):
Only small comments.
In general it LGTM 👍

PR #1080, opened 2025-02-20
[RORDEV-1388] New sytax of groups rules, refactor and common decoder

mgoworko:
🚀New (ES) [New, unified and simplified syntax of groups rules]

## Summary by CodeRabbit

- **New Features**
  - Enhanced access control rule definitions offer more intuitive and consistent configuration of group-based permissions.
  - Introduced new group logic classes: `AllOfGroupsRule` and `AnyOfGroupsRule` for improved rule definitions.
- **Refactor**
  - Unified the decoding and matching logic for group rules to improve reliability and clarity.
  - Streamlined error reporting and diagnostics related to group configurations.
  - Consolidated multiple implicit definitions into a single generic method for handling group rules.
- **Chores**
  - Upgraded the plugin version to 1.63.0-pre7.


mgoworko:
https://github.com/coderabbitai review

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/BlockValidator.scala
```diff
@@ -54,8 +54,7 @@ object BlockValidator {
       .map(_.rule)
       .collect { case a: AuthenticationRule => a }
       .filter {
-        case _: GroupsOrRule => false
-        case _: GroupsAndRule => false
+        case _: GroupsRule[_] => false
```
mgoworko, line 57: Here and in `UserDefinitionsDecoder` - the original implementation was created when there was only an "or" groups rule. I think that it is correct to match all groups rules here.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/GroupsLogicRepresentationDecoder.scala
```diff
@@ -0,0 +1,182 @@
...
+import tech.beshu.ror.accesscontrol.factory.RawRorConfigBasedCoreFactory.CoreCreationError.RulesLevelCreationError
+import tech.beshu.ror.accesscontrol.factory.decoders.rules.auth.GroupsLogicRepresentationDecoder.{GroupsLogicDecodingResult, declaredName}
+import tech.beshu.ror.accesscontrol.utils.CirceOps.*
+import tech.beshu.ror.accesscontrol.utils.{SyncDecoder, SyncDecoderCreator}
+import tech.beshu.ror.implicits.*
+
+import scala.compiletime.ops.string.Matches
+import scala.compiletime.{constValue, error}
+
+private[auth] class GroupsLogicRepresentationDecoder[
```
mgoworko, line 33: Single, unified decoder implementation for all types of rules - `GroupsRule`, LDAP, JWT, external. Unified config structure and error handling. For `GroupsRule` the `RuntimeResolvableGroupsLogic` implementation is used, but for LDAP, JWT and External - the `GroupsLogic`. This decoder can work on both.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsAndRule.scala
```diff
@@ -1,35 +0,0 @@
...
- *    You should have received a copy of the GNU General Public License
- *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
- */
-package tech.beshu.ror.accesscontrol.blocks.rules.auth
-
-import tech.beshu.ror.accesscontrol.blocks.rules.Rule
-import tech.beshu.ror.accesscontrol.blocks.rules.Rule.{Name, RuleName}
-import tech.beshu.ror.accesscontrol.domain.{CaseSensitivity, GroupsLogic}
-
-final class GroupsAndRule(override val settings: BaseGroupsRule.Settings[GroupsLogic.And],
```
mk-software-pl, line 23: Yes, it could be done like that. But we'd rather have a class per rule even if the class is trivial. I see the value in this approach.

mk-software-pl (changes requested):
and please remember about docs changes ;)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
```diff
@@ -35,7 +35,9 @@ import tech.beshu.ror.accesscontrol.matchers.GenericPatternMatcher
 import tech.beshu.ror.implicits.*
 import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
 
-abstract class BaseGroupsRule[GL <: GroupsLogic](val settings: Settings[GL])
+class GroupsRule[GL <: GroupsLogic](override val name: Rule.Name,
```
coutoPL, line 38: Yes, it could be done like that. But we'd rather have a class per rule even if the class is trivial. I see the value in this approach.

coutoPL (changes requested):
and please remember about docs changes ;)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
mgoworko: Ok, will restore classes per rule.

https://github.com/beshu-tech/readonlyrest-docs/pull/246 <= docs, I can't assign reviewer there

mgoworko: Done, restored classes per rule.

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/rules/http/HeadersAllOfRuleSettingsTests.scala
```diff
@@ -26,7 +26,7 @@ import tech.beshu.ror.syntax.*
 import tech.beshu.ror.unit.acl.factory.decoders.rules.BaseRuleSettingsDecoderTest
 import tech.beshu.ror.utils.TestsUtils.*
 
-class HeadersAndRuleSettingsTests extends BaseRuleSettingsDecoderTest[HeadersAndRule] {
+class HeadersAllOfRuleSettingsTests extends BaseRuleSettingsDecoderTest[HeadersAndRule] {
```
coutoPL, line 29: auto-rename mistake?

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/http/HeadersAllOfRuleTests.scala
```diff
@@ -32,7 +32,7 @@ import tech.beshu.ror.accesscontrol.request.RequestContext
 import tech.beshu.ror.syntax.*
 import tech.beshu.ror.utils.TestsUtils.*
 
-class HeadersAndRuleTests extends AnyWordSpec with MockFactory {
+class HeadersAllOfRuleTests extends AnyWordSpec with MockFactory {
```
coutoPL, line 35: auto-rename mistake?

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/http/HeaderAnyOfRuleTests.scala
```diff
@@ -32,7 +32,7 @@ import tech.beshu.ror.accesscontrol.request.RequestContext
 import tech.beshu.ror.syntax.*
 import tech.beshu.ror.utils.TestsUtils.*
 
-class HeaderOrRuleTests extends AnyWordSpec with MockFactory {
+class HeaderAnyOfRuleTests extends AnyWordSpec with MockFactory {
```
coutoPL, line 35: auto-rename mistake?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
```diff
@@ -35,7 +35,29 @@ import tech.beshu.ror.accesscontrol.matchers.GenericPatternMatcher
 import tech.beshu.ror.implicits.*
 import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
 
-abstract class BaseGroupsRule[GL <: GroupsLogic](val settings: Settings[GL])
+class AnyOfGroupsRule(override val name: Rule.Name, override val settings: Settings[GroupsLogic.AnyOf])
```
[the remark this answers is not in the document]
mgoworko: Restored all rule names in this file, and then used in the decoder.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/ExternalAuthorizationRuleTests.scala
```diff
@@ -65,7 +65,7 @@ class ExternalAuthorizationRuleTests
               assertMatchRule(
```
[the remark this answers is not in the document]
mgoworko: I will have to check if it would be possible. I agree it would be a great idea if viable. I made changes or otherwise responded for other review comments, and will analyze this further later.

On the other hand - there are already unit tests for all logic's and after my changes the decoder is the same for all rules. Maybe it is enough?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/GroupsLogicRepresentationDecoder.scala
coutoPL: Yeah, it's a good approach to extract the decoding of the groups logic to a separate decoder. I just didn't catch the "representation" idea. 

Looking from the perspective of authorization rules, each uses the groups logic. In [this comment](https://github.com/beshu-tech/readonlyrest-docs/pull/246#discussion_r1980163314) I showed two examples. At the moment (it may change in a moment ;)) both look the same. That's why I'm not sure if we need all of these generalisations:
```
private[auth] class GroupsLogicRepresentationDecoder[
  RULE_REPRESENTATION,
  POSITIVE_RULE_REPRESENTATION <: RULE_REPRESENTATION,
  NEGATIVE_RULE_REPRESENTATION <: RULE_REPRESENTATION,
  ALL_OF_REPRESENTATION <: POSITIVE_RULE_REPRESENTATION : Decoder,
  ANY_OF_REPRESENTATION <: POSITIVE_RULE_REPRESENTATION : Decoder,
  NOT_ALL_OF_REPRESENTATION <: NEGATIVE_RULE_REPRESENTATION : Decoder,
  NOT_ANY_OF_REPRESENTATION <: NEGATIVE_RULE_REPRESENTATION : Decoder,
](combinedRuleCreator: (POSITIVE_RULE_REPRESENTATION, NEGATIVE_RULE_REPRESENTATION) => RULE_REPRESENTATION) {
```

Why? Because:
1. there is no domain equivalent of the representation - it's a technical thing
2. we have only two implementations (one anonymous) - it's hard to reason about the necessity of the technical thing when the cases are not so clear
3. when the decoder looks the same in all contexts, do we really need the generalisation? (the decoder of the GroupsLogic can be separated from the decoder of the groups rule /the latter one will just use the first one/)

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/rules/auth/GroupsRuleSettingsTests.scala
```diff
@@ -19,59 +19,48 @@ package tech.beshu.ror.unit.acl.factory.decoders.rules.auth
...
-
-class GroupsNotAnyOfRuleSettingsTests extends GroupsRuleSettingsTests[GroupsLogic.NotAnyOf, GroupsNotAnyOfRule](GroupsNotAnyOfRule.Name) {
-  override protected def groupsLogicCreator: GroupIds => GroupsLogic.NotAnyOf = GroupsLogic.NotAnyOf.apply
-}
-
-sealed abstract class GroupsRuleSettingsTests[GL <: GroupsLogic, R <: BaseGroupsRule[GL] : ClassTag](ruleName: RuleName[R])
-  extends BaseRuleSettingsDecoderTest[R]
-    with Inside {
+  private val testParams = Table[String, String, GroupIds => GroupsLogic](
+    ("deprecated_subrule_name", "subrule_name_in_user_belongs_to_groups_section", "creator"),
```
coutoPL, line 53: sth is strange here. In the first column we have the `deprecated_subrule_name`. At the same time, we have `groups_any_of` in this column...

In context of the local groups rule we have:
1. the simplified syntax eg .`groups_any_of: [...]`
2. the extended syntax (with the ability to compose group logics): 
    ```yaml
    user_belongs_to_groups:
      any_of: [...]
      not_all_of: [...]
    ```
3. deprecated names (only for the simple syntax?). eg. `groups_or: []` or `groups: []`

I think in this table we missed some of these. Or flatten them in a way I don't understand.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/ExternalAuthorizationRuleTests.scala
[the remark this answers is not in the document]
coutoPL: you are right. We should have:
1. groups logic unit test
2. groups logic decoder unit tests
3. [authorization rule] unit tests (with arbitrarily picked group logic, but without focusing on the type of groups logic)
4. [authorization rule] rule decoder unit tests (it's the same case as the 3rd one ... but in the context of decoders - we have tested the groups logic decoders, so we don't have to deal with it in these tests)

Makes sense? Are we able to do it in this PR, or is it better to create a jira for the refactoring?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
```diff
@@ -35,7 +35,86 @@ import tech.beshu.ror.accesscontrol.matchers.GenericPatternMatcher
...
+class AllOfGroupsRule(override val name: Rule.Name, override val settings: Settings[GroupsLogic.AllOf])
+                     (override implicit val userIdCaseSensitivity: CaseSensitivity)
+  extends GroupsRule[GroupsLogic.AllOf](name, settings)
+
+object AllOfGroupsRule {
+  case object NameInsideGroupsSubruleSection extends RuleName[AllOfGroupsRule] {
+    override val name: Name = Rule.Name("all_of")
+  }
+
+  case object NameV2 extends RuleName[AllOfGroupsRule] {
```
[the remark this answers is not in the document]
mgoworko: Refactored, each rule now has `ExtendedSyntaxName`, `SimpleSyntaxName`, and for older rules `DeprecatedSimpleSyntaxNameVX`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
```diff
@@ -35,7 +35,86 @@ import tech.beshu.ror.accesscontrol.matchers.GenericPatternMatcher
...
+    override val name: Name = Rule.Name("groups_all_of")
+  }
+
+  case object NameV1 extends RuleName[AllOfGroupsRule] {
```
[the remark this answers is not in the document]
mgoworko: Refactored, each rule now has `ExtendedSyntaxName`, `SimpleSyntaxName`, and for older rules `DeprecatedSimpleSyntaxNameVX`

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/ExternalAuthorizationRuleTests.scala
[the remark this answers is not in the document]
mgoworko: 1. We have those, `GroupsLogicExecutorTests`
2. A bit more than simple unit testing, but we have those too: `GroupsRuleSettingsTests``
3. We have those only for `LdapAuthenticationRuleSettingsTests` for now.
4. I think that the decoder is sufficiently tested in `GroupsRuleSettingsTests` - all types of logics are decoded and tested there.

It would be possible to add more authorization rule tests - perhaps common for all rules. But extracting some generic tests for all auth rules is IMO outside the scope of this change.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
```diff
@@ -35,7 +35,86 @@ import tech.beshu.ror.accesscontrol.matchers.GenericPatternMatcher
...
+    override val name: Name = Rule.Name("roles")
+  }
+}
+
+class AllOfGroupsRule(override val name: Rule.Name, override val settings: Settings[GroupsLogic.AllOf])
+                     (override implicit val userIdCaseSensitivity: CaseSensitivity)
+  extends GroupsRule[GroupsLogic.AllOf](name, settings)
+
+object AllOfGroupsRule {
+  case object NameInsideGroupsSubruleSection extends RuleName[AllOfGroupsRule] {
```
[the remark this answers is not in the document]
mgoworko: Refactored, each rule now has `ExtendedSyntaxName`, `SimpleSyntaxName`, and for older rules `DeprecatedSimpleSyntaxNameVX`

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/rules/auth/GroupsRuleSettingsTests.scala
mgoworko: Separated 2 tables for tests of simple syntax and for tests of extended syntax.

```  
private val simpleSyntaxTestParams = Table[String, GroupIds => GroupsLogic](
    ("simple_syntax_name", "creator"),
    ("roles", GroupsLogic.AnyOf.apply),
    ("groups", GroupsLogic.AnyOf.apply),
    ("groups_or", GroupsLogic.AnyOf.apply),
    ("groups_any_of", GroupsLogic.AnyOf.apply),
    ("groups_and", GroupsLogic.AllOf.apply),
    ("groups_all_of", GroupsLogic.AllOf.apply),
    ("groups_not_all_of", GroupsLogic.NotAllOf.apply),
    ("groups_not_any_of", GroupsLogic.NotAnyOf.apply)
  )

  private val extendedSyntaxTestParams = Table[String, GroupIds => GroupsLogic](
    ("extended_syntax_name", "creator"),
    ("any_of", GroupsLogic.AnyOf.apply),
    ("all_of", GroupsLogic.AllOf.apply),
    ("not_all_of", GroupsLogic.NotAllOf.apply),
    ("not_any_of", GroupsLogic.NotAnyOf.apply)
  )
  ``

File: core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala [hunk 14]
```diff
@@ -287,7 +366,33 @@ abstract class BaseGroupsRule[GL <: GroupsLogic](val settings: Settings[GL])
   }
 }
 
-object BaseGroupsRule {
-  final case class Settings[GL <: GroupsLogic](permittedGroupsLogic: RuntimeResolvableGroupsLogic[GL],
-                                               usersDefinitions: NonEmptyList[UserDef])
+object GroupsRule {
+
+  case object GroupsSubruleSectionName extends RuleName[GroupsRule[GroupsLogic]] {
+    override val name: Name = Rule.Name("user_belongs_to_groups")
```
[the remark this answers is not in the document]
coutoPL: ok, I thought about it, and I think (at least at the moment) should do it like this:
1. all_of - `groups_all_of` and `groups.all_of` 
...
5. combined:
```
groups:
  all_of/any_of: []
  not_all_of/not_any_of: []
```

Let's not change too much in this PR.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/AllOfGroupsRule.scala
```diff
@@ -0,0 +1,47 @@
...
+
+class AllOfGroupsRule(override val name: Rule.Name, override val settings: Settings[GroupsLogic.AllOf])
+                     (override implicit val userIdCaseSensitivity: CaseSensitivity)
+  extends BaseGroupsRule[GroupsLogic.AllOf](name, settings)
+
+object AllOfGroupsRule {
+  implicit val allOfTypeInfo: Creator[GroupsLogic.AllOf] = new AllOfGroupsRule(_, _)(_)
+
+  case object ExtendedSyntaxName extends RuleName[AllOfGroupsRule] {
+    override val name: Name = Rule.Name("all_of")
```
coutoPL, line 33: I forgot about one more thing. Group name is used in audit, so it should be "stable identifier". We cannot change it depending on the configuration of the rule. I totally forgot about it.

So, I want to ask you (again, sorry) about changing it a little bit. For the rule name, let's pick the one that describes it clearly.  In this case, it would probably be "groups_all_of". Other names (all_of, groups_and, roles_and) won't be alternative names, but we will treat them as configuration keys only. The rule name should always be the same (and not passed with the constructor, but determined at the code level - as we do in the rest of ROR rules).

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/CombinedGroupsRule.scala
```diff
@@ -17,19 +17,14 @@
...
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.BaseGroupsRule
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.BaseGroupsRule.{Creator, Settings}
+import tech.beshu.ror.accesscontrol.domain.*
 
-final class GroupsAndRule(override val settings: BaseGroupsRule.Settings[GroupsLogic.And],
-                          override val userIdCaseSensitivity: CaseSensitivity)
-  extends BaseGroupsRule(settings) {
+class CombinedGroupsRule(override val name: Rule.Name, override val settings: Settings[GroupsLogic.Combined])
+                        (override implicit val userIdCaseSensitivity: CaseSensitivity)
+  extends BaseGroupsRule[GroupsLogic.Combined](name, settings)
```
coutoPL, line 26: this rule should also have a fixed name. I suggest `groups_combined` event if this name won't be used as a yaml key. But it will be used in audit and logs. 

Alternative proposition would be `groups_{all_of/any_of}_{not_all_of/not_any_of}`. 

WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseGroupsRule.scala
```diff
@@ -288,6 +290,26 @@ abstract class BaseGroupsRule[GL <: GroupsLogic](val settings: Settings[GL])
 }
 
 object BaseGroupsRule {
-  final case class Settings[GL <: GroupsLogic](permittedGroupsLogic: RuntimeResolvableGroupsLogic[GL],
-                                               usersDefinitions: NonEmptyList[UserDef])
+
+  case object BaseGroupsRuleExtendedSyntaxName extends RuleName[BaseGroupsRule[GroupsLogic]] {
```
coutoPL, line 294: In reference to my previous comment, I'd suggest to get rid of this.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseGroupsRule.scala
```diff
@@ -288,6 +290,26 @@ abstract class BaseGroupsRule[GL <: GroupsLogic](val settings: Settings[GL])
...
+
+  def name(ruleName: String): RuleName[BaseGroupsRule[GroupsLogic]] = new RuleName[BaseGroupsRule[GroupsLogic]] {
+    override val name: Rule.Name = Rule.Name(ruleName)
+  }
+
+  final case class Settings[+GL <: GroupsLogic](permittedGroupsLogic: RuntimeResolvableGroupsLogic[GL],
+                                                usersDefinitions: NonEmptyList[UserDef])
+
+  trait Creator[GL <: GroupsLogic] {
+    def create(name: Rule.Name,
```
coutoPL, line 306: ... and creator won't need the name param

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/GroupsAnyOfRuleTests.scala
```diff
@@ -20,21 +20,23 @@ import cats.data.NonEmptyList
...
-  override def createRule(settings: GroupsRuleSettings[GroupsLogic.Or], caseSensitivity: CaseSensitivity): BaseGroupsRule[GroupsLogic.Or] = {
-    new GroupsOrRule(settings, caseSensitivity)
+  override def createRule(ruleName: Rule.Name, settings: GroupsRuleSettings[GroupsLogic.AnyOf], caseSensitivity: CaseSensitivity): BaseGroupsRule[GroupsLogic.AnyOf] = {
+    BaseGroupsRule.Creator[GroupsLogic.AnyOf].create(ruleName, settings, caseSensitivity)
   }
 
-  protected def groupsLogicCreator: GroupIds => GroupsLogic.Or = GroupsLogic.Or.apply
+  protected def groupsLogicCreator: GroupIds => GroupsLogic.AnyOf = GroupsLogic.AnyOf.apply
 
   "A GroupsRule" should {
```
coutoPL, line 41: let's change it too

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/GroupsRule.scala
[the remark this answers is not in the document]
mgoworko: Done, with `groups` instead of `user_belongs_to_groups`.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/AnyOfGroupsRule.scala
```diff
@@ -0,0 +1,52 @@
...
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.BaseGroupsRule.{Creator, Settings}
+import tech.beshu.ror.accesscontrol.domain.*
+
+class AnyOfGroupsRule(override val name: Rule.Name, override val settings: Settings[GroupsLogic.AnyOf])
+                     (override implicit val userIdCaseSensitivity: CaseSensitivity)
+  extends BaseGroupsRule[GroupsLogic.AnyOf](name, settings)
+
+
+object AnyOfGroupsRule {
+  implicit val anyOfTypeInfo: Creator[GroupsLogic.AnyOf] = new AnyOfGroupsRule(_, _)(_)
```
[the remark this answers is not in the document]
mgoworko: typo/remnant from old implementation, fixed

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/AllOfGroupsRule.scala
mgoworko: I think it is now done as specified in this comment, please check.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/CombinedGroupsRule.scala
mgoworko: Done, with `groups_combined`. Using the second proposition would require refactor and doing some workaround for type erasure.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseGroupsRule.scala
```diff
@@ -288,6 +290,21 @@ abstract class BaseGroupsRule[GL <: GroupsLogic](val settings: Settings[GL])
 }
 
 object BaseGroupsRule {
-  final case class Settings[GL <: GroupsLogic](permittedGroupsLogic: RuntimeResolvableGroupsLogic[GL],
-                                               usersDefinitions: NonEmptyList[UserDef])
+
+  def name(ruleName: String): RuleName[BaseGroupsRule[GroupsLogic]] = new RuleName[BaseGroupsRule[GroupsLogic]] {
```
coutoPL, line 294: do we need it? I see one usage and it seems we can get rid of it

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/groups/GroupsLogicRepresentationDecoder.scala
```diff
@@ -0,0 +1,176 @@
...
+      case _ => cursor.as[Option[REPRESENTATION]].map(_.map((_, key)))
+    }
+  }
+
+  private[rules] def errorMsgNoGroupsList[R <: Rule](ruleName: RuleName[R]) = {
+    s"${ruleName.show} rule requires to define 'groups_or'/'groups_and'/'groups_not_any_of'/'groups_not_all_of' arrays"
+  }
+
+  private[rules] def errorMsgOnlyOneGroupsList[R <: Rule](ruleName: RuleName[R]) = {
+    s"${ruleName.show} rule requires to define 'groups_or'/'groups_and'/'groups_not_any_of'/'groups_not_all_of' arrays (but not all)"
```
coutoPL, line 158: should we fix the message (deprecated rule names used)?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/groups/GroupsLogicRepresentationDecoder.scala
```diff
@@ -0,0 +1,176 @@
...
+    val fields = ruleNames.map(_.name.value)
+    val (cursor, key) = c.downFieldsWithKey(field, fields: _*)
+    cursor match {
+      case _: FailedCursor => Right(None)
+      case _ => cursor.as[Option[REPRESENTATION]].map(_.map((_, key)))
+    }
+  }
+
+  private[rules] def errorMsgNoGroupsList[R <: Rule](ruleName: RuleName[R]) = {
+    s"${ruleName.show} rule requires to define 'groups_or'/'groups_and'/'groups_not_any_of'/'groups_not_all_of' arrays"
```
coutoPL, line 154: the same here

coutoPL (changes requested):
last minor comments

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseGroupsRule.scala
mgoworko: I removed it from here. But we need to have the RuleName instance for the `RuleDecoder` It is used in the base decoder implementation as the path. So it is move in a slightly changed form to `object ruleDecoders`

gradle.properties
```diff
@@ -1,5 +1,5 @@
 publishedPluginVersion=1.62.0
-pluginVersion=1.63.0-pre5
+pluginVersion=1.63.0-pre6
```
coutoPL, line 2: let's bump to pre7

mgoworko: merged with latest develop and done

PR #1083, opened 2025-02-28
[RORDEV-1317] Ask for confirmation before ror-tools app performs patching - flag or interactive mode

mgoworko:
⚠️Warning (ES) Acknowledgement needs to be accepted before Elasticsearch patching process. For scripts, you can [set a flag](https://docs.readonlyrest.com/elasticsearch#id-3.-patch-elasticsearch) to automate the process.

## Summary by CodeRabbit

- **New Features**
  - Introduced an explicit confirmation step for Elasticsearch patching, ensuring users explicitly acknowledge the implications before proceeding.
  - Added configurable options for container startup, allowing custom entry points for more flexible deployments.

- **Enhancements**
  - Improved error handling and logging to provide clearer, more informative feedback during operations.
  - Expanded and refined test configurations, enhancing reliability and overall stability.

- **Chores**
  - Updated build configurations and dependencies to support streamlined testing and improved resource management.


ror-tools/src/test/resources/test-es-not-patched/lib/elasticsearch-8.15.0.jar
mgoworko, line 1: Can we (from license point of view) embed those jars for tests?

coutoPL: Is this [jar](https://mvnrepository.com/artifact/org.elasticsearch/elasticsearch)? If yes, we don't have to do it.
If not, it's better not to do that. 

But we can test with docker, do we?

docker-envs/base-ror-docker/es/install-ror-es-using-file.sh
```diff
@@ -13,6 +13,6 @@ echo "Installing ES ROR from file..."
 /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch file:///tmp/ror.zip
 if verlte "6.5.0" "$ES_VERSION"; then
   echo "Patching ES ROR..."
-  /usr/share/elasticsearch/jdk/bin/java -jar /usr/share/elasticsearch/plugins/readonlyrest/ror-tools.jar patch
+  /usr/share/elasticsearch/jdk/bin/java -jar /usr/share/elasticsearch/plugins/readonlyrest/ror-tools.jar patch --I-understand-implications-of-ES-patching yes
```
coutoPL, line 16: In KBN ROR plugin it's [I_UNDERSTAND_AND_ACCEPT_KBN_PATCHING ](https://docs.readonlyrest.com/kibana#patching-kibana-acknowledgment-in-a-silent-mode). Let's make it on the ROR ES side symmetric (and case insensitive)

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
 
   // todo:
   // 1. option: return success when already patched/unpatched
   // 2. restore backup when fails to patch
   def main(args: Array[String]): Unit = {
-    OParser
-      .parse(parser, args, Config(Command.Verify(None)))
-      .foreach { config =>
+    handle(args) match {
+      case Result.CommandNotParsed => ()
```
coutoPL, line 35: do nothing when the command is not parsed?

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
+      case Result.Success => ()
+      case Result.Failure(exitCode) => sys.exit(exitCode)
+    }
+  }
+
+  def handle(args: Array[String]): Result = {
```
coutoPL, line 41: maybe we can move it to some trait? the handle method could be a protected one in this trait

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
+    OParser.parse(parser, args, Config(Command.Verify(None), UserUnderstandsImplicationsOfESPatching.AnswerNotGiven)) match {
+      case None =>
+        Result.CommandNotParsed
+      case Some(config) =>
         Try {
           config.command match {
             case Command.Patch(customEsPath) =>
-              val esDirectory = esDirectoryFrom(customEsPath)
-              new PatchAction(EsPatch.create(esDirectory)).execute()
+              config.userUnderstandsImplicationsOfESPatching match {
```
coutoPL, line 49: let's move this to private method

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
-      .foreach { config =>
+    handle(args) match {
+      case Result.CommandNotParsed => ()
+      case Result.Success => ()
+      case Result.Failure(exitCode) => sys.exit(exitCode)
+    }
+  }
+
+  def handle(args: Array[String]): Result = {
+    OParser.parse(parser, args, Config(Command.Verify(None), UserUnderstandsImplicationsOfESPatching.AnswerNotGiven)) match {
```
coutoPL, line 42: I don't know why it's called config :( Maybe we should change it? Is it rather `Arguments`? Or sth like that?

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
+      case None =>
+        Result.CommandNotParsed
+      case Some(config) =>
         Try {
           config.command match {
             case Command.Patch(customEsPath) =>
-              val esDirectory = esDirectoryFrom(customEsPath)
-              new PatchAction(EsPatch.create(esDirectory)).execute()
+              config.userUnderstandsImplicationsOfESPatching match {
+                case UserUnderstandsImplicationsOfESPatching.Yes =>
```
coutoPL, line 50: let's import `import UserUnderstandsImplicationsOfESPatching.*`

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
+                  performPatching(customEsPath)
+                case UserUnderstandsImplicationsOfESPatching.No =>
+                  patchingAbortedBecauseUserDidNotAcceptConsequences()
+                case UserUnderstandsImplicationsOfESPatching.AnswerNotGiven =>
+                  if (askForConfirmation() == "yes") performPatching(customEsPath)
+                  else patchingAbortedBecauseUserDidNotAcceptConsequences()
+              }
```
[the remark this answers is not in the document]
coutoPL: or return a sealed trait

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
+
+  private def patchingAbortedBecauseUserDidNotAcceptConsequences(): Result = {
+    println("You have to confirm, that You understand the implications of ES patching in order to perform it.\nYou can read about patching in our documentation: https://docs.readonlyrest.com/elasticsearch#id-3.-patch-elasticsearch.")
+    Result.Failure(1)
+  }
+
+  private def askForConfirmation(): String = {
+    println("Elasticsearch needs to be patched to work with ReadonlyREST. You can read about patching in our documentation: https://docs.readonlyrest.com/elasticsearch#id-3.-patch-elasticsearch.")
+    print("Do you understand the implications of ES patching? (yes/no): ")
+    StdIn.readLine()
```
coutoPL, line 88: ensure case insenitivness

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,43 +16,82 @@
...
+  }
+
+  private def performPatching(customESPath: Option[Path]): Result = {
```
coutoPL, line 91: shouldn't we add similar methods from unpatch and verify? to be consistent and for the sake of redability

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -68,6 +107,20 @@ object RorToolsApp {
...
     note(""),
+    opt[String]("I-understand-implications-of-ES-patching").optional()
+      .valueName("<yes/no>")
+      .validate {
+        case "yes" => success
+        case "no" => success
+        case other => failure(s"ERROR: Invalid value [$other]. Only values 'yes' and 'no' can be provided as an answer.")
+      }
+      .action { (answer, config) =>
+        answer match {
```
coutoPL, line 118: answer.toLowerCase?

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppTest.scala
```diff
@@ -0,0 +1,172 @@
...
+import org.scalatest.matchers.should.Matchers.{equal, should}
+import tech.beshu.ror.tools.RorToolsApp.Result
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, IOException, PrintStream}
+import java.nio.file.Files.*
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.{Files, Path, Paths}
+import java.util.UUID
+
+class RorToolsAppTest extends AnyFunSuite with BeforeAndAfter {
```
coutoPL, line 31: it'd be good to run the test with the different ES version (at least with the ones listed below:

Screenshot 2025-03-01 at 19 41 52 

The addition of xpack jars is not an option for sure. But I imagine we can pull a proper Elasticsearch image, copy the ES folder and run the tests. WDYT?

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppTest.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  test("Patching is successful for ES installation that was not patched (with consent given in arg)") {
+    withTestDirectory("test-es-not-patched") { testDirectory =>
+      val (result, output) = captureResultAndOutput {
+        RorToolsApp.handle(Array("patch", "--I-understand-implications-of-ES-patching", "yes", "--es-path", testDirectory))
+      }
+      result should equal(Result.Success)
+      output should include(
```
coutoPL, line 39: it'd be good to add some additional verification step. Eg. we could calculate the hash of some ES dirs before and after patching and checking if the hash is ok. 

It'd be good to have "There and Back Again" test too.

coutoPL (changes requested):

ror-tools/src/test/resources/test-es-not-patched/lib/elasticsearch-8.15.0.jar
coutoPL: Hmm, I'm not sure if we should do it like this. The `integration-tests` module should rather be called the `e2e-test` module (if this is misleading, we should change it in the near future). 

Now, we want to test our patcher. From the patcher point of view, these would be e2e tests too - it makes sense. But from the ROR plugin perspective, patching is only one of the steps. In the `integration-tests` module we test the patching step already.

That's why I think the integration tests of the patcher should be placed in the `ror-tools` module (this module produces the patcher jar). We should add the dependency (test-level) to the `tests-utils` module and take advantage of the `Elasticsearch` class.

There is no problem with using test-containers-based tests in other modules than the `integration-tests` one. E.g. in `LdapConnectivityCheckYamlLoadedAccessControlTests` we did it.

docker-envs/base-ror-docker/es/install-ror-es-using-file.sh
coutoPL: yeah, I agree it looks better but IMO our plugins management should be similar. It would be little surprising for a user.  I'm OK with the approach you proposed as long as you add this in the ROR KBN plugin too (preferably in the backward compatible way ;))

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,284 @@
...
+class RorToolsAppSuite extends AnyWordSpec with ESVersionSupportForAnyWordSpecLike with BeforeAndAfterAll {
+
+  implicit val scheduler: Scheduler = Scheduler.computation(10)
+
+  private def localPath = Path.of(getClass.getResource(s"/temp_directory_for_ror_tools_app_suite").getPath)
+
+  private def esLocalPath = Path.of(getClass.getResource(s"/temp_directory_for_ror_tools_app_suite").getPath + "/es")
+
+  // Before performing tests in this suite:
+  // - the ES container is started (using security variant RorWithXpackSecurity)
```
coutoPL, line 52: IMO we should not use the `RorWithXpackSecurity` container. IMO we need the `Elasticsearch` class only. 
Why? The `RorWithXpackSecurity` container has ES with ROR installed. Moreover, AFAIR, when it starts, it does the patching step.

What do we want to test? 
- patching/unpatching/verify of ror-tools end-to-end

What do we need to test it?
- the `ror-tools` jar
- folder with ES binaries

What limitations do we have?
- we cannot include the ES binaries in our repo (in general it would be a bad idea)

Taking into consideration all of the above, we can do it like this:
1. before tests we run the container and copy the folder to the local temp folder (and we can stop the container - we won't need it anymore)
2. we copy the ES binaries to some test-specific folder and run the test operation: `/my-test-specific-folder/elasticsearch/jdk/bin/java -jar ror-tools.jar patch --I-understand-and-accept-ES-patching yes`
3. we check if the ES sources were patched correctly

As you can see, using this method we can test the ror-tools in the e2e fashion. The ES container is only a source of the test environment only.

coutoPL (changes requested):

ror-tools/src/test/resources/test-es-not-patched/lib/elasticsearch-8.15.0.jar
mgoworko: Test moved to ror-tools, as discussed. It works. It required moving some test resources from integration-tests to tests-utils and the same Gradle trick was needed - assembling plugin in Gradle task as dependency, instead of in runtime as before.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RorToolsAppSuite.scala
mgoworko: As discussed, test moved to ror-tools. The new tests invokes the patcher app as a scala method invocation - and not the ror-tools.jar execution. The execution from jar is tested in all old integration tests - it is executed when tests ES container is created.

(we cannot run the ror-tools.jar here in those new tests - it would be impossible to assert detailed result and text output, and the exit code would stop the JVM)

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,231 @@
...
+
+class RorToolsAppSuite extends AnyWordSpec with BeforeAndAfterAll {
+
+  implicit val scheduler: Scheduler = Scheduler.computation(10)
+
+  private val tempDirectory = BetterFile.newTemporaryDirectory()
+  private val localPath = tempDirectory.path
+  private val esLocalPath = (tempDirectory / "es").path
+
+  // Before performing tests in this suite:
```
coutoPL, line 45: if we add the `prepareElasticsearchAndRorBinaries` the comment should be defined close to it, not to the `beforeAll` method.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,231 @@
...
+  // - on each test this file is uncompressed and the fresh copy of the ES directory is used
+  override protected def beforeAll(): Unit = {
+    withTestEsContainer { esContainer =>
+      esContainer.execInContainer("tar", "-cvf", "/tmp/elasticsearch.tar", "-C", "/usr/share/elasticsearch", "modules", "bin", "lib", "plugins", "tmp")
+      esContainer.copyFileFromContainer("/tmp/elasticsearch.tar", s"$localPath/elasticsearch.tar")
+    }
+    super.beforeAll()
+  }
+
+  override protected def afterAll(): Unit = {
```
coutoPL, line 59: I would move the afterAll after the tests. It's not needed to understand the logic of the tests

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,231 @@
...
+    super.afterAll()
+    tempDirectory.clear()
+  }
+
+  "ROR tools app" should {
+    "Patching is successful for ES installation that was not patched (with consent given in arg)" in withFreshEsDirectory { () =>
```
coutoPL, line 65: TBH, I wonder if we should use `withFreshEsDirectory`. Now, you have to remember to add it. At the same time, you have access to `esLocalPath` (if you didn't have such access, and the `withFreshEsDirectory' would give it to you, you wouldn't have a chance to make a mistake). 

Maybe we should consider using `beforeEach` instead?

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,231 @@
...
+    try {
+      manager.start()
+        .map(_ => withStartedEs(manager.esContainer))
+        .runSyncUnsafe(5 minutes)
+    } finally {
+      manager.stop().runSyncUnsafe()
+    }
+  }
+
+  private def unTar(tarPath: Path, outputPath: Path): Unit = {
```
coutoPL, line 198: let's move the method to some utils? Please review existing utils objects/classes

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  private def withFreshEsDirectory(perform: () => Unit): Unit = {
+    try {
+      DirectoryUtils.clean(esLocalPath)
+      File(s"$esLocalPath").mkdirs
+      unTar(Path.of(s"$localPath/elasticsearch.tar"), Path.of(s"$esLocalPath"))
+      File(s"$esLocalPath/plugins/readonlyrest/").mkdirs
+      Files.copy(
```
coutoPL, line 217: I'm not sure why we do it like that? Cannot we just use the whole `/usr/share/elasticsearch` folder we downloaded?

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/CapturingOutputAndMockingInput.scala
```diff
@@ -0,0 +1,38 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.tools.utils
+
+import monix.execution.atomic.AtomicAny
+import tech.beshu.ror.tools.core.utils.InOut
+
+class CapturingOutputAndMockingInput(mockedInput: Option[String] = None) extends InOut {
+
+  val outputBuffer: AtomicAny[String] = AtomicAny[String]("")
```
coutoPL, line 24: private?

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/DirectoryUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.{Files, Path}
+import java.security.MessageDigest
+import scala.jdk.CollectionConverters.*
+
+object DirectoryUtils {
+
+  def clean(dir: Path): Unit = {
+    if (dir.toFile.exists()) {
+      Files.walkFileTree(dir, new java.nio.file.SimpleFileVisitor[java.nio.file.Path]() {
```
coutoPL, line 29: better-files have the `delete` (with true flag as a parameter) to recursively delete folder files. Can we use it instead?

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/DirectoryUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+          java.nio.file.FileVisitResult.CONTINUE
+        }
+      })
+    }
+  }
+
+  def calculateHash(dir: Path): String = {
+    val digest = MessageDigest.getInstance("SHA-256")
+
+    val fileList = Files.walk(dir)
```
coutoPL, line 46: maybe we can use better-files also here?

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/TestEsContainerManager.scala
```diff
@@ -0,0 +1,63 @@
...
+import tech.beshu.ror.utils.containers.*
+import tech.beshu.ror.utils.containers.EsContainerCreator.EsNodeSettings
+import tech.beshu.ror.utils.containers.images.ReadonlyRestWithEnabledXpackSecurityPlugin
+import tech.beshu.ror.utils.containers.images.domain.Enabled
+
+import java.io.Console as _
+import scala.concurrent.duration.*
+import scala.language.postfixOps
+
+class TestEsContainerManager extends EsContainerCreator {
```
coutoPL, line 31: I'm not sure about the name. I don't like the Manager suffix for sure. Looking at the implementation (and its usage in tests), I'd say it's an example ES container with ROR installed. So maybe `ExampleEsWithRorContainer`? WDYT?

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/TestEsContainerManager.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  private val uniqueClusterId: AtomicInt = AtomicInt(1)
+
+  lazy val esContainer: EsContainer = createEsContainer
+
+  def start(): Task[Unit] = Task.delay(esContainer.start())
```
coutoPL, line 37: I wonder if we have to manage the lifecycle of the container on our own. I understand that we don't need the container when we copy the tar from it, but does it hurt us that it's working while the tests are running? Maybe we can leave the lifecycle management for the test containers and scala-test? WDYT?

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/TestEsContainerManager.scala
```diff
@@ -0,0 +1,63 @@
...
+  private def createEsContainer: EsContainer = {
+    val clusterName = s"ROR_${uniqueClusterId.getAndIncrement()}"
+    val nodeName = s"${clusterName}_1"
+    create(
+      nodeSettings = EsNodeSettings(
+        nodeName = nodeName,
+        clusterName = clusterName,
+        securityType = SecurityType.RorWithXpackSecurity(
+          ReadonlyRestWithEnabledXpackSecurityPlugin.Config.Attributes.default.copy(
+            rorConfigReloading = Enabled.Yes(1 hour),
```
coutoPL, line 50: Let's set No here. Currently, one may ask: "why 1 hour?" ;)

tests-utils/src/main/resources/ldap.ldif
```diff
@@ -1,230 +0,0 @@
-#################################
```
coutoPL, line 1: I guess it wasn't used?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsClusterContainer.scala
```diff
@@ -179,7 +179,7 @@ object EsVersion {
 
 sealed trait SecurityType
 object SecurityType {
-  final case class RorWithXpackSecurity(attributes: ReadonlyRestWithEnabledXpackSecurityPlugin.Config.Attributes) extends SecurityType
+  final case class RorWithXpackSecurity(attributes: ReadonlyRestWithEnabledXpackSecurityPlugin.Config.Attributes, performInstallation: Boolean = true) extends SecurityType
```
coutoPL, line 182: IMO we should have this flag for `RorSecurity` too. It's (almost) always better to be symmetrical.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainerWithRorSecurity.scala
```diff
@@ -64,7 +64,7 @@ object EsContainerWithRorSecurity extends StrictLogging {
                                            rorConfig: ReadonlyRestPlugin.Config) = {
     DockerImageCreator.create(
       Elasticsearch.create(esVersion, esConfig)
-        .install(new ReadonlyRestPlugin(esVersion, rorConfig))
+        .install(new ReadonlyRestPlugin(esVersion, rorConfig, performInstallation = true))
```
coutoPL, line 67: It doesn't look good here. We "install" ReadonlyRestPlugin here but we can set `performInstallation = false`

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -76,6 +76,7 @@ class Elasticsearch(esVersion: String,
         file = log4jFileFromResources
       )
       .user("root")
+      .runWhen(Version.greaterOrEqualThan(esVersion, 9, 0, 0), "microdnf install -y tar")
```
coutoPL, line 79: let's add a note that we need it in `ror-tools` tests

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
```diff
@@ -60,20 +61,29 @@ object ReadonlyRestPlugin {
...
 }
 class ReadonlyRestPlugin(esVersion: String,
-                         config: Config)
+                         config: Config,
+                         performInstallation: Boolean)
   extends Elasticsearch.Plugin {
 
   override def updateEsImage(image: DockerImageDescription): DockerImageDescription = {
-    image
+    val withoutInstallation = image
```
coutoPL, line 69: Oh, I get it now. I thought we were disabling the patching, not the ROR installation itself. 
It would be good if we could allow ES to install ROR and disable patching. I guess the problem is that the ES doesn't start (because ROR makes it explode, right?). 

When you run ES docker from console you can override its entrypoint (`docker run -ti docker.elastic.co/elasticsearch/elasticsearch:9.0.0-beta1 bash`). Are we able to sth similar with test-containers?

coutoPL (changes requested):

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
mgoworko: Done, with some tweaks (`/bin/sh -c "while true; do sleep 30; done`)
- in order to keep the container alive
- also I had to modify checking container readiness - it depended on ES healthcheck endpoint

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsClusterContainer.scala
mgoworko: I limited the usages and visibility of `performPatching` to only 2 places: 
- EsContainerWithRorAndXpackSecurity (where it is created with default settings - with patching enabled) 
- TestEsContainerManager for ror-tools tests - where it is created with custom settings

So it is no longer visible here and generally in utils used for standard integration tests.

tests-utils/src/main/resources/ldap.ldif
mgoworko: Yes, `integrations-tests` had its own version, same as `core` tests.
This PR changed the way resources are handled - we now reuse resources from tests-utils for integration-tests and ror-utils tests. This file was unused anyway and started causing conflicts.

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/TestEsContainerManager.scala
mgoworko: Refactored, this class now exposes only single public method
```
def withTestEsContainer(withStartedEs: EsContainer => Unit): Unit
```

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/TestEsContainerManager.scala
mgoworko: Renamed and added comment about what is happening in this container and how/why it is different.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
mgoworko: I remove it. It was needed, because before this iteration of review ROR plugin was not installed. Now it is installed and those files are present

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
mgoworko: Moved to FileUtils in this module.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
mgoworko: Done, moved to beforeEach and afterEach.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  private val tempDirectory = BetterFile.newTemporaryDirectory()
+  private val localPath = tempDirectory.path
+  private val esLocalPath = (tempDirectory / "es").path
+
+  // Before performing tests in this suite:
+  // - the ES container is started (using security variant RorWithXpackSecurity)
+  // - but the ROR plugin is not installed on container creation/start
+  // - the `/usr/share/elasticsearch` is compressed and downloaded from the container
+  // - the tar file is stored in the /temp_directory_for_ror_tools_app_suite directory in resources
```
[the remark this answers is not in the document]
mgoworko: After all other refactors after this iteration of review I don't think it is necessary:
- the tar is downloaded once from ES container (in fact, that was the recommended way anywhere I looked - it is very complicated and discouraged to download multiple files using testcontainers - it is not supported at all)
- the fresh environment still needs to be created for every test - so we would still have to copy the directory of pre-untarred files anyway
- there are issues with file permissions (I encountered them earlier - tar simply restores the directory structure, but copying files caused permission issues in some directories)

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -68,6 +158,20 @@ object RorToolsApp {
...
+        case "no" => success
+        case other => failure(s"ERROR: Invalid value [$other]. Only values 'yes' and 'no' can be provided as an answer.")
+      }
+      .action { (answer, config) =>
+        answer.toLowerCase match {
+          case "yes" => config.copy(userUnderstandsImplicationsOfESPatching = UserUnderstandsAndAcceptsESPatching.Yes)
+          case "no" => config.copy(userUnderstandsImplicationsOfESPatching = UserUnderstandsAndAcceptsESPatching.No)
+        }
+      }
+      .text("Optional, when provided with value 'yes', it confirms that the user understands and accepts the implications of ES patching. The patching can therefore be performed. When not provided, user will be asked for confirmation in interactive mode."),
```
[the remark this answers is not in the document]
mgoworko: I will push docs PR tomorrow.

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -112,13 +216,31 @@ object RorToolsApp {
...
 
   private sealed trait Command
+
   private object Command {
     final case class Patch(customEsPath: Option[os.Path]) extends Command
     final case class Unpatch(customEsPath: Option[os.Path]) extends Command
     final case class Verify(customEsPath: Option[os.Path]) extends Command
   }
 
+  private sealed trait UserUnderstandsAndAcceptsESPatching
```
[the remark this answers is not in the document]
mgoworko: private object PatchingConsent {
    case object Accepted extends PatchingConsent
    case object Rejected extends PatchingConsent
    case object AnswerNotGiven extends PatchingConsent
  }

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,50 +16,140 @@
...
-            println(s"UNEXPECTED ERROR: ${ex.printStackTrace()}")
-            sys.exit(1)
+    RorToolsAppHandler.handle(args)(ConsoleInOut) match {
+      case Result.Success =>
+        ()
+      case Result.Failure(exitCode) =>
+        // The details of why the command was not parsed should have been
+        // already printed by the scopt library parser
+        sys.exit(exitCode)
+      case Result.CommandNotParsed =>
```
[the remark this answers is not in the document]
mgoworko: It was not enabled, added

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -16,50 +16,140 @@
...
+        sys.exit(1)
```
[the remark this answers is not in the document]
mgoworko: I moved the comment to this case. It was a mistake that it was 2 lines higher. Result.Failure is now case object

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/InOut.scala
```diff
@@ -0,0 +1,45 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.tools.core.utils
+
+import scala.io.StdIn
+
+trait InOut {
```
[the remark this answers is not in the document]
mgoworko: I'm not sure about changing it to logback. It is the interface of the console app. From what I understand it should use the console input/output. Logback is not designed for that purpose IMO. I think that for now it is ok to have a single interface.   

I added it for 2 reasons:
- It was unclear how scopt handled the output and the Console and Predef interfaces were mixed in the code. That way, with single interface it can be changed/fixed across the entire tool quickly 
- Second reason is not very good, because test code is leaking to tests. But there was no better alternative - it was impossible to capture the stdin/out output of the ror-tools  in tests - Gradle blocked any kind on capturing

ror-tools/src/test/scala/tech/beshu/ror/tools/utils/DirectoryUtils.scala
mgoworko: I tried it, but it was not shorter or more concise - iterating over files looks more or less the same.

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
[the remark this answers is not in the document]
coutoPL: I meant the link we have in other our messages returned by `ror-tools` we should add also in this message

coutoPL (approved):
LGTM.

[One missing comment left](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1083#discussion_r2012956714). Please take a look

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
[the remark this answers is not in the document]
mgoworko: Sorry, I missed the comment. Done, added the same message and link to documentation as in other places + modified test output assertion to include it.

PR #1094, opened 2025-03-31
[Tech] Download library sources in IntelliJ IDE

mgoworko:
## Summary by CodeRabbit

- **Chores**
	- Enhanced IDE integration to automatically download source code, ensuring a smoother debugging and development experience.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,13 @@ plugins {
     id 'com.bmuschko.docker-remote-api'             version '9.4.0'
     id 'io.github.gradle-nexus.publish-plugin'      version '2.0.0-rc-1'
     id 'org.gradle.crypto.checksum'                 version '1.4.0'
+    id "idea"
+}
+
+idea {
+    module {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -21,6 +21,13 @@ plugins {
     id 'com.bmuschko.docker-remote-api'             version '9.4.0'
     id 'io.github.gradle-nexus.publish-plugin'      version '2.0.0-rc-1'
     id 'org.gradle.crypto.checksum'                 version '1.4.0'
+    id "idea"
+}
+
+idea {
+    module {
+        downloadSources = true
```
mgoworko, line 29: Just a little change to make working with Idea less painful. The plugin interfaces with Idea and causes the Gradle sync inside Idea to download library sources. 

I think that it will not affect jars and builds outside Idea (I checked that locally), but probably someone else should check on their side, on different configuration, too.

coutoPL (approved):

PR #1095, opened 2025-04-01
[RORDEV-1435] Improved patching algorithm

mgoworko:
🧐Enhancement (ES) Patcher - more robust Elasticsearch patcher + better logs messages

## Summary by CodeRabbit

- **Chores**
  - Updated the build pipeline to use Ubuntu 22.04 for improved compatibility and performance.
  - Renamed environment variables related to Elasticsearch patching consent for clearer user understanding.

- **New Features**
  - Introduced a unified Elasticsearch patch executor with enhanced patch status reporting and error handling.
  - Added detailed error messages and structured error types to improve user feedback during patching.
  - Improved patching consent flow with explicit handling of no input and silent mode guidance.

- **Bug Fixes**
  - Corrected command argument formatting for Elasticsearch patching commands.

- **Refactor**
  - Simplified patching logic by renaming methods and using boolean patch status checks.
  - Removed deprecated patch action classes and logging decorators to streamline codebase.
  - Enhanced test coverage for patching, unpatching, and verification commands, including interactive consent scenarios.
  - Updated input reading methods to safely handle absent input by returning optional values.
  - Modified patching methods to return detailed metadata about patched files for better tracking.
  - Added patch metadata management to track file hashes and detect unauthorized modifications.
  - Improved internal patch verification with more specific status cases and error handling.
  - Streamlined Java environment selection and patching options in installation scripts based on version checks.
  - Added file hash calculation and patch metadata encoding/decoding for robust patch state tracking.
  - Refactored patch classes to unify patching operations and metadata reporting.
  - Introduced JAR manifest modification to record patch version and detect patched files.
  - Enhanced plugin directory metadata handling with structured patch information.
  - Updated test utilities and input mocking to support optional input scenarios.


ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/RorToolsError.scala
```diff
@@ -0,0 +1,34 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.tools.core.utils
+
+sealed class RorToolsError(val message: String)
+
+object RorToolsError {
```
mgoworko, line 21: Those issues are now handled as errors - with error message and exit code 1, but without throwing exception and showing stacktrace. Exceptions are now reserved only for technical/unexpected situations, and not standard behavior, taken into account in logic.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -81,7 +84,29 @@ class RorToolsAppSuite extends AnyWordSpec with BeforeAndAfterAll with BeforeAnd
           |""".stripMargin
       )
     }
-    "Patching does not start when user declines to accept implications of patching (in arg)" in {
+    "Patching successful first time, on second try not started because already patched" in {
```
mgoworko, line 87: I think that the tests now cover all cases from diagrams https://github.com/sscarduzio/readonlyrest_kbn/pull/581

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -122,29 +121,46 @@ trait RorTools {
     private def askUserAboutPatchingConsent(): PatchingConsent = {
       inOut.println("Elasticsearch needs to be patched to work with ReadonlyREST. You can read about patching in our documentation: https://docs.readonlyrest.com/elasticsearch#id-3.-patch-elasticsearch.")
       inOut.print("Do you understand the implications of ES patching? (yes/no): ")
-      inOut.readLine().toLowerCase match
-        case "yes" => PatchingConsent.Accepted
-        case _ => PatchingConsent.Rejected
+      inOut.readLine() match {
```
mgoworko, line 124: That is the logic, that handles error with input being null (console input not possible)

docker-image/ror-entrypoint.sh
```diff
@@ -4,7 +4,15 @@ INVOKE_ROR_TOOLS="/usr/share/elasticsearch/jdk/bin/java -jar /usr/share/elastics
 
 if $INVOKE_ROR_TOOLS verify | grep -q "Elasticsearch is NOT patched"; then
 
-  if [ -n "$I_UNDERSTAND_IMPLICATION_OF_ES_PATCHING" ] && [[ "${I_UNDERSTAND_IMPLICATION_OF_ES_PATCHING,,}" == *"yes"* ]]; then
+  if [ -n "$I_UNDERSTAND_AND_ACCEPT_ES_PATCHING" ]; then
```
mgoworko, line 7: Missing in previous PR, it was done for ES 6x entrypoint

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,120 @@
...
+  private def backup(): Unit = {
+    inOut.println("Creating backup ...")
+    Try(esPatch.performBackup()) match {
+      case Success(_) =>
+        rorPluginDirectory.updatePatchedByRorVersion()
+      case Failure(ex) =>
+        rorPluginDirectory.clearBackupFolder()
+        throw ex
+    }
+  }
```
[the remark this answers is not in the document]
mgoworko: Backup fails in case of technical errors - some problems with files/directories. IMO it is ok to keep throwing exceptions in those cases.

docker-envs/base-ror-docker/es/install-ror-es-using-file.sh
```diff
@@ -13,6 +13,6 @@ echo "Installing ES ROR from file..."
 /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch file:///tmp/ror.zip
 if verlte "6.5.0" "$ES_VERSION"; then
   echo "Patching ES ROR..."
-  /usr/share/elasticsearch/jdk/bin/java -jar /usr/share/elasticsearch/plugins/readonlyrest/ror-tools.jar patch --I_UNDERSTAND_AND_ACCEPT_ES_PATCHING yes
+  /usr/share/elasticsearch/jdk/bin/java -jar /usr/share/elasticsearch/plugins/readonlyrest/ror-tools.jar patch --I_UNDERSTAND_AND_ACCEPT_ES_PATCHING=yes
```
coutoPL, line 16: hmmm, I've just noticed this subtle difference. Space vs `=`. 
Why did it change?

coutoPL, line 16: I see the scripts can use old versions of ROR (like it's done in the ROR Sandbox)
If so, we have to be backward compatible here. Please try to run the stack for ROR 1.62.0 ;)

mgoworko: I changed it to be consistent with KBN tools - they use the syntax with `=` everywhere. Both syntaxes are supported and I added tests - both variants are tested.

mgoworko: I added an additional check (if) for ROR version. The new flag is added for versions >=1.64 (taking into account -pre versions). Tested with 1.62 and 1.64 and with ES 8.13. Was indeed failing before this change, now the container is starting.

mgoworko: And today added additional adjustments for testing with ROR zip file or from sources - it required passing the version env too, same change as in ror-sandbox.

mgoworko: I applied the changes from https://github.com/beshu-tech/ror-sandbox/pull/67#discussion_r2030191735 (the docker example in this repo is similar to ror-sandbox, so the comment from there apllied also here)

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
coutoPL: did you check the [ROR KBN patcher tests](https://github.com/sscarduzio/readonlyrest_kbn/blob/develop/test/KibanaPatcher.test.ts)?

coutoPL (changes requested):

docker-envs/base-ror-docker/es/install-ror-es-using-api.sh
```diff
@@ -16,8 +18,28 @@ fi
 
 echo "Installing ES ROR $ROR_VERSION..."
 /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch "https://api.beshu.tech/download/es?esVersion=$ES_VERSION&pluginVersion=$ROR_VERSION&email=ror-sandbox%40readonlyrest.com"
-if verlte "6.5.0" "$ES_VERSION"; then
-  echo "Patching ES ROR $ROR_VERSION..."
-  /usr/share/elasticsearch/jdk/bin/java -jar /usr/share/elasticsearch/plugins/readonlyrest/ror-tools.jar patch --I_UNDERSTAND_AND_ACCEPT_ES_PATCHING yes
+
+if [[ -z "$ROR_VERSION" ]]; then
```
coutoPL, line 22: shouldn't it be `if [ ! -v ROR_VERSION ] || [ -z "$ROR_VERSION" ]; then`?
`

coutoPL: BTW, we have the same check above

docker-envs/base-ror-docker/es/install-ror-es-using-api.sh
```diff
@@ -16,8 +18,28 @@ fi
...
+  echo "No $ROR_VERSION variable is set"
+  exit 2
```
coutoPL, line 24: `exit 3`

docker-envs/base-ror-docker/es/install-ror-es-using-api.sh
```diff
@@ -16,8 +18,28 @@ fi
...
+fi
+
+# Set Java path based on ES version
+if verlte "7.0.0" "$ES_VERSION"; then
+  JAVA_BIN_PATH="/usr/share/elasticsearch/jdk/bin/java"
+elif verlte "6.7.0" "$ES_VERSION"; then
+  JAVA_BIN_PATH="$JAVA_HOME/bin/java"
+else
+  echo "Unsupported ES version: $ES_VERSION"
+  exit 1
```
coutoPL, line 34: `exit 4`

docker-envs/base-ror-docker/es/install-ror-es-using-file.sh
```diff
@@ -11,8 +13,29 @@ fi
...
+fi
+
+# Set Java path based on ES version
+if verlte "7.0.0" "$ES_VERSION"; then
+  JAVA_BIN_PATH="/usr/share/elasticsearch/jdk/bin/java"
+elif verlte "6.7.0" "$ES_VERSION"; then
+  JAVA_BIN_PATH="$JAVA_HOME/bin/java"
+else
+  echo "Unsupported ES version: $ES_VERSION"
+  exit 1
```
coutoPL, line 30: exit 3

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,137 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+final class EsPatchExecutor(rorPluginDirectory: RorPluginDirectory,
+                            esPatch: EsPatch)
+                           (implicit inOut: InOut) {
+
+  def patch(): Either[RorToolsError, Unit] = {
+    checkWithPatchedByFileAndEsPatch() match {
+      case EsPatchStatus.PatchedWithCurrentRorVersion(rorVersion) =>
```
coutoPL, line 32: Let's do `import EsPatchStatus.*`.IMO The errors are clear without adding this context explicitly.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,137 @@
...
+        Left(EsAlreadyPatchedError(rorVersion))
+      case EsPatchStatus.PatchedWithOtherRorVersion(expectedRorVersion, patchedByRorVersion) =>
+        Left(EsPatchedWithDifferentVersionError(expectedRorVersion, patchedByRorVersion))
+      case EsPatchStatus.NotPatched =>
+        backup()
```
coutoPL, line 37: Maybe we can introduce `doPatch` private method, as we did with `doRestore`? WDYT?

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  }
+
+  def restore(): Either[RorToolsError, Unit] = {
+    checkWithPatchedByFileAndEsPatch() match {
+      case EsPatchStatus.PatchedWithCurrentRorVersion(_) =>
+        doRestore()
+      case EsPatchStatus.PatchedWithOtherRorVersion(expectedRorVersion, patchedByRorVersion) =>
+        Left(EsPatchedWithDifferentVersionError(expectedRorVersion, patchedByRorVersion))
+      case EsPatchStatus.NotPatched if esPatch.isPatchApplied =>
```
coutoPL, line 52: I think this smells. Does it mean we should have another status? Is it tested?
Maybe we have to describe this use case better.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,137 @@
...
+        // The patched_by file may be missing for some reason (so the status is EsPatchStatus.NotPatched),
+        // but the EsPlugin implementation recognizes as patched, so we perform restoring
+        doRestore()
+      case EsPatchStatus.NotPatched =>
+        Left(EsNotPatchedError)
+    }
+  }
+
+  def verify(): Unit = {
```
coutoPL, line 61: AFAIR, in the KBN changes, I mentioned that we should return status code 0 when verify tells us the ES is patched. And otherwise return -1. It would be helpful when the patcher is used in scripts.

please taka a look at ROR KBN patcher tests.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,137 @@
...
+    val resultOfCheckingPatchedByFile =
+      rorPluginDirectory.readPatchedByRorVersion() match {
+        case None =>
+          EsPatchStatus.NotPatched
+        case Some(patchedByRorVersion) if patchedByRorVersion == currentRorVersion =>
+          EsPatchStatus.PatchedWithCurrentRorVersion(currentRorVersion)
+        case Some(patchedByRorVersion) =>
+          EsPatchStatus.PatchedWithOtherRorVersion(currentRorVersion, patchedByRorVersion)
+      }
+    resultOfCheckingPatchedByFile match {
```
coutoPL, line 107: Can we do it in one pattern match?

coutoPL (changes requested):

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,168 @@
...
+  private def validatePatchedFiles(): Option[Either[List[os.Path], Unit]] = {
+    rorPluginDirectory
+      .readPatchMetadataFile()
+      .map(_.map { filePatchMetadata =>
+        val currentHash = FileUtils.calculateFileHash(filePatchMetadata.path.wrapped)
+        if (filePatchMetadata.hash == currentHash) Right(()) else Left(filePatchMetadata.path)
+      }.partitionMap(identity) match {
+        case (paths, _) if paths.nonEmpty => Left(paths)
+        case (_, _ :: _) => Right(())
+      })
```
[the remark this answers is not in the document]
coutoPL: yeah, it seems we missed `case (_, Nil) => ???`

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,168 @@
...
+      case PatchedWithCurrentRorVersion(_) =>
+        inOut.println("Elasticsearch is currently patched, restoring ...")
+        doRestore()
+      case PatchedWithOtherRorVersion(expectedRorVersion, patchedByRorVersion) =>
+        Left(EsPatchedWithDifferentVersionError(expectedRorVersion, patchedByRorVersion))
+      case NotPatched =>
+        Left(EsNotPatchedError)
+      case IllegalFileModificationsDetectedInPatchedFiles(files) =>
+        Left(IllegalFileModificationsDetectedInPatchedFilesError(files))
+      // We have to handle this case explicitly, because older versions of patcher did not save the patched_by file
```
coutoPL, line 56: Do we? Our patching procedure says: "unpatch with old ROR", "install new ROR", "patch with new ROR". 
So, it seems, there is no sense in considering the case when there was no file, because it was patched with old ROR version patcher. But obviously, we should consider what to do when the file is removed (e.g., together with the backup folder). How do we figure out if the ES is patched? I wonder if instead of the file with hashes, we should modify the  MANIFEST.MF in modified jar adding info about ROR patched version. 
If we do that, we could iterate through the files we need to patch and verify that they're patched or not. It's harder to break than just removing the folder.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,168 @@
...
+
+import tech.beshu.ror.tools.core.patches.base.EsPatchExecutor.EsPatchStatus
+import tech.beshu.ror.tools.core.patches.base.EsPatchExecutor.EsPatchStatus.*
+import tech.beshu.ror.tools.core.patches.internal.RorPluginDirectory
+import tech.beshu.ror.tools.core.utils.RorToolsError.*
+import tech.beshu.ror.tools.core.utils.{EsDirectory, FileUtils, InOut, RorToolsError}
+
+import scala.util.{Failure, Success, Try}
+
+final class EsPatchExecutor(rorPluginDirectory: RorPluginDirectory,
```
coutoPL, line 27: Let's consider the ES upgrade.

Let's say: 
1. user has ES 8.0.0, installs ROR 1.63.0 and patches it
2. user upgrades the ES to 9.0.0 without unpatching and ROR
3. user starts the new ES

Is ROR able to detect that the installation is broken?

mgoworko: That situation can be detected, if the jar files in /lib directory have changed during the update (and according to some sources on the Internet, the /lib directory is replaced during upgrade; but I guess it depends on how the ES is installed and how it is upgraded)

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/filePatchers/CopyTransportNetty4JarToPluginPatchCreator.scala
```diff
@@ -0,0 +1,54 @@
...
+  override def create(rorPluginDirectory: RorPluginDirectory,
+                      esVersion: SemVer): CopyTransportNetty4JarToPluginPatch =
+    new CopyTransportNetty4JarToPluginPatch(rorPluginDirectory)
+}
+
+private[patches] class CopyTransportNetty4JarToPluginPatch(rorPluginDirectory: RorPluginDirectory)
+  extends FilePatch {
+
+  override def backup(): Unit = ()
+
```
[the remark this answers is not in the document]
mgoworko: The jar file is copied from ES directory and is still present there. There is no need to backup

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -72,6 +74,21 @@ private [patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
     os.write(patchedByFilePath, readCurrentRorVersion())
   }
 
+  def readPatchMetadataFile(): Option[List[FilePatchMetadata]] = {
```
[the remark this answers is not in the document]
coutoPL: Oh, I see. Maybe it'd be better to have sth like this:
```scala
final case class EsPatchMetadata(patchedFilesMetadata: List[FilePatchMetadata]) 
```

and this method could be named like this:

```scala
def readEsPatchMetadata(): Option[ESPatchMetadata]
```

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -72,6 +74,21 @@ private [patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
...
+    Option.when(os.exists(patchMetadataFilePath)) {
+      (os.read(patchMetadataFilePath): String)
+        .split("\n")
+        .map(_.split("=", 2)) // split only on the first '='
```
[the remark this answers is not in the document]
coutoPL: ok, great.

WDYT about defining sth like this:
```scala
trait Encoder[T] {
  def encode(value: T): String
}

trait Decoder[T] {
  def decode(value: String): Either[Throwable, T]
}

trait Codec[T] extends Encoder[T] with Decoder[T]
```
 and provide proper implementation of `Coder[EsPatchMetadata]` and then write a bunch of unit tests for it?

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -17,14 +17,16 @@
...
 import tech.beshu.ror.tools.core.utils.EsDirectory
 import tech.beshu.ror.tools.core.utils.EsUtil.{findTransportNetty4JarIn, readonlyrestPluginPath}
 
-private [patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
+private[patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
 
   private val rorPath: Path = readonlyrestPluginPath(esDirectory.path)
   private val backupFolderPath: Path = rorPath / "patch_backup"
   private val patchedByFilePath: Path = backupFolderPath / "patched_by"
+  private val patchMetadataFilePath: Path = backupFolderPath / "patch_metadata"
```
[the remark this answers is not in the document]
coutoPL: TBH, it's not a big deal, but:
- We don't need backward compatibility between ROR versions because every upgrade of ROR (or ES) should be followed by the steps: `unpatch ROR, uninstall ROR`. After these two steps, the ROR backup folder is removed, so no patched_by file is there anymore.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
coutoPL: 1. So the use case we consider here is as follows:
- the is a backup folder
- there is no `patched_by` file
IMO, this use case is less likely than the use case where the whole backup folder was removed. 
I imagine that the user forgets to unpatch ROR, but they uninstall it. Could you please check what happens with the backup folder in this use case? I'm afraid it may be removed during the uninstall process. But even if it persists, more likely user removes the folder than the single file in it. 

2. Yeah, for sure it's simpler. But I wonder if the solution won't be more resilient to the problem I mentioned in the 1st point. It's very important to detect the corrupted state. At the end of the day, we don't have to say exactly what went wrong (it'd be nice, but it's not the most important thing), but we have to figure out the corrupted state and inform the user about it. We don't want to allow patching already-patched ES binaries.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
coutoPL: Ok, great. Are we able to simulate this in our tests?

coutoPL (commented):
a few more questions

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
mgoworko: I added explicit handling of ES version in the plugin. The metadata file now looks something like that 

```
{
  "rorVersion" : "1.64.0-pre6",
  "esVersion" : "9.0.0",
  "patchedFilesMetadata" : [
    {
      "path" : "/tmp/3483890482512061106/es/lib/elasticsearch-9.0.0.jar",
      "hash" : "2d115f55b9befc1381284756292901af10ec7b28b90eecd5350ea08d613ba92a"
    }
  ]
}
```

And the error message for this case is ```The patch was performed on Elasticsearch version 1.2.3, but currently installed version is 9.0.0. It should be unpatched using ROR for the ES version used for patching and patched again. For patching instructions see our docs: https://docs.readonlyrest.com/elasticsearch#id-3.-patch-elasticsearch```

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/filePatchers/JarManifestModifier.scala
```diff
@@ -0,0 +1,90 @@
...
+          jarOutput.putNextEntry(newEntry)
+          jarOutput.write(newManifestContent.getBytes("UTF-8"))
+          jarOutput.closeEntry()
+        } else {
+          // Copy file other than manifest without changes to the temp jar
+          val newEntry = new JarEntry(name)
+          jarOutput.putNextEntry(newEntry)
+          zipInput.transferTo(jarOutput)
+          jarOutput.closeEntry()
+        }
```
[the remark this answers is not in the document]
mgoworko: I did not push part of the change by mistake, will fix tomorrow

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,203 @@
...
+  object EsPatchStatus {
+    final case class PatchedWithCurrentRorVersion(currentRorVersion: String) extends EsPatchStatus
+
+    final case class PatchPerformedOnOtherEsVersion(currentEsVersion: String, patchPerformedOnEsVersion: String) extends EsPatchStatus
+
+    final case class PatchedWithOtherRorVersion(currentRorVersion: String, patchedByRorVersion: String) extends EsPatchStatus
+
+    final case class CorruptedPatchWithoutValidMetadata(backupFolderIsPresent: Boolean, patchedJarFiles: List[PatchedJarFile]) extends EsPatchStatus
+
+    final case class IllegalFileModificationsDetectedInPatchedFiles(files: List[os.Path]) extends EsPatchStatus
```
coutoPL, line 193: CorruptedPatchWithoutValidMetadata and IllegalFileModificationsDetectedInPatchedFiles are similar cases, don't they?

They refer to a corrupted ES installation, but the reasons are different. Maybe we should have another subhierarchy, or at least the names should be similar.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -53,23 +57,33 @@ private [patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
...
   }
 
-  def updatePatchedByRorVersion(): Unit = {
-    os.remove(patchedByFilePath, checkExists = false)
-    os.write(patchedByFilePath, readCurrentRorVersion())
+  def readEsPatchMetadata(): Option[EsPatchMetadata] = {
+    if (os.exists(patchMetadataFilePath)) {
+      EsPatchMetadataCodec.decode(readFile(patchMetadataFilePath)) match {
+        case Right(metadata) => Some(metadata)
+        case Left(error) => throw new IllegalStateException(s"Cannot decode ROR patch metadata file: $error")
```
coutoPL, line 72: Let's add the content of the file to the exception. It will help us with analysing the issues (clients will likely send only the stacktrace)

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -84,5 +98,12 @@ private [patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
       .getOrElse(throw new IllegalStateException(s"Cannot read ROR version from ${pluginPropertiesFilePath}"))
   }
 
-  def isRorPluginPath(path: Path): Boolean = path.startsWith(rorPath)
+  private def readFile(path: Path): String = os.read(path)
```
coutoPL, line 101: do we need it? ;)

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/filePatchers/JarManifestModifier.scala
```diff
@@ -0,0 +1,65 @@
...
+object JarManifestModifier {
+
+  private val patchedByRorVersionPropertyName = "Patched-By-Ror-Version"
+
+  def addPatchedByRorVersionProperty(file: File, rorVersion: String): Unit = {
+    Using(new JarFile(file.toJava)) { jarFile =>
+      val tempJarFile = File(s"temp-${UUID.randomUUID()}.jar")
+      val manifest = jarFile.getManifest
+      manifest.getMainAttributes.putValue(patchedByRorVersionPropertyName, rorVersion)
+      Using(new JarOutputStream(tempJarFile.newOutputStream.buffered, manifest)) { jarOutput =>
```
coutoPL, line 36: maybe we can extract this to a well-named private method to increase readability? WDYT?

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -132,6 +208,178 @@ class RorToolsAppSuite extends AnyWordSpec with BeforeAndAfterAll with BeforeAnd
           |  -h, --help               prints this usage text""".stripMargin
       )
     }
+    "Patching not started because there is a metadata file indicating that the ES is already patched" in {
+      File(backupDirectory.path).createDirectory()
+      patchMetadataFile.createFile()
+      patchMetadataFile.write(
```
coutoPL, line 214: IMO, we have too many details at this level. Reading the test, the reader wants to know what is the given state. They are not interested on how the state was achieved. So, maybe we should have a private method like this:
```
givenPatchMetadata(rorVersion = "0.0.1", esVersionUsed, patchedFilesMetadata = List.empty)
```

All the details should be in the method

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -132,6 +208,178 @@ class RorToolsAppSuite extends AnyWordSpec with BeforeAndAfterAll with BeforeAnd
...
+      }
+      verifyResult should equal(Result.Success)
+      verifyOutput should include(
+        """Checking if Elasticsearch is patched ...
+          |Elasticsearch is patched! ReadonlyREST can be used"""
+          .stripMargin
+      )
+    }
+
+    "The patch is not detected when metadata file is missing and `verify` command is executed" excludeES(allEs6x, allEs7xBelowEs74x) in {
```
coutoPL, line 321: Could you please remind me why we have to exclude these modules here?

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -132,6 +201,128 @@ class RorToolsAppSuite extends AnyWordSpec with BeforeAndAfterAll with BeforeAnd
...
+      patchedByFile.delete()
+      patchedByFile.exists() should be(false)
+
+      val (unpatchResult, unpatchOutput) = captureResultAndOutput {
+        RorToolsTestApp.run(Array("unpatch", "--es-path", esLocalPath.toString))(_)
+      }
+      unpatchResult should equal(Result.Failure)
+      unpatchOutput should include(
+        """Checking if Elasticsearch is patched ...
+          |ERROR: Elasticsearch is likely patched by an older version of ROR, but there is no valid patch metadata present."""
```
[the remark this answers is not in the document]
coutoPL: What do we do with this message? Should we tell the user that it's "likely patched by an older version of ROR"? 
Or maybe we are not sure what happens exactly, and the one thing we are sure, is that the ES installation is corrupted. And maybe the only advice we can give is that the ES should be reinstalled. WDYT?>

coutoPL (changes requested):
Great! It seems all the important things we already have here. Only small minor things left

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
```diff
@@ -0,0 +1,203 @@
...
+        throw ex
+    }
+  }
+
+  private def checkWithPatchedByFileAndEsPatch(): EsPatchStatus = {
+    inOut.println("Checking if Elasticsearch is patched ...")
+    val currentRorVersion = rorPluginDirectory.readCurrentRorVersion()
+    val currentEsVersion = readEsVersion(rorPluginDirectory.esDirectory)
+    rorPluginDirectory.readEsPatchMetadata() match {
+      case Some(metadata) if metadata.rorVersion == currentRorVersion && metadata.esVersion.major == currentEsVersion.major =>
```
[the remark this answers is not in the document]
coutoPL: yeah, until now ;) see [this](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/99eba67fed84ff0b4fb5db17fd007e008c51625b/ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/modifiers/bytecodeJars/ModifyEntitlementInitializationClass.scala#L61)

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
mgoworko: ```As a result, the Elasticsearch is in a corrupted state. ES must be reinstalled.```

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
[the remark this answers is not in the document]
mgoworko: Modified, new full output:
```
Checking if Elasticsearch is patched ...
ERROR: Elasticsearch is either patched by an older version of ROR or corrupted.
 - if ES has been patched using some older ROR version, then try unpatching using that older ROR version
 - otherwise the ES installation is corrupted and ES must be reinstalled
Problems:
 - backup catalog is present, but there is no metadata file
 - file x-pack-core-9.0.0.jar was patched by ROR 1.63.0
 - file x-pack-ilm-9.0.0.jar was patched by ROR 1.63.0
 - file x-pack-security-9.0.0.jar was patched by ROR 1.63.0
 ```

We no longer check for the old-style metadata file, so the absence of new-style metadata either means that the installation is corrupted, or that ES was patched using older ROR.

File: ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala (hunk 31 above)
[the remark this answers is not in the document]
Comment (mgoworko) [reply ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala]: :man_facepalming: 

I added detailed check, for 6.x, 7.x and 8.x patch version difference is allowed, for other versions patch must be equal.

File: ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/filePatchers/JarManifestModifier.scala [hunk 32]
```diff
@@ -0,0 +1,68 @@
...
+import scala.jdk.CollectionConverters.IteratorHasAsScala
+import scala.util.Using
+
+object JarManifestModifier {
+
+  private val patchedByRorVersionPropertyName = "Patched-By-Ror-Version"
+
+  def addPatchedByRorVersionProperty(file: File, rorVersion: String): Unit = {
+    Using(new JarFile(file.toJava)) { jarFile =>
+      val tempJarFile = File(s"temp-${UUID.randomUUID()}.jar")
```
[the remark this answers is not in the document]
mgoworko: I had to revert that change. Using temporary file break the patcher - I did not investigate it, but probably some kind of permissions problem. Temp file is created in /temp directory, but then in this part of code we copy and replace the original jar with this temp file.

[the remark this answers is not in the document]
coutoPL: ok, let's leave the comment

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/base/EsPatchExecutor.scala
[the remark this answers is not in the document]
mgoworko: Ok, I did not know about the requirement - reinstalling the plugins. There is now a simple comparison of full version

coutoPL (approved):
LGTM

PR #1109, opened 2025-05-12
[RORDEV-1461] Docker image patching verification fix

coutoPL:
🐞Fix (ES) Correct patching verification in ROR Docker image entrypoint

### DIFF
--- a/ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
+++ b/ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
@@ -27,9 +27,6 @@ import scala.util.{Failure, Success, Try}
 
 object RorToolsApp extends RorTools {
 
-  // todo:
-  // 1. option: return success when already patched/unpatched
-  // 2. restore backup when fails to patch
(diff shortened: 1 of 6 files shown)


ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -146,15 +143,16 @@ trait RorTools {
   private class VerifyCommandHandler(implicit inOut: InOut) {
     def handle(command: Command.Verify): Result = {
       val esDirectory = esDirectoryFrom(command.customEsPath)
-      handleResult(EsPatchExecutor.create(esDirectory).verify())
+      handleResult(EsPatchExecutor.create(esDirectory).verify(), errorMessagePrefix = "")
```
mgoworko, line 146: This modification may require adjusting other tests too, not the single one that was adjusted (because the verify command may be executed also as an initial step of other commands, and may return errors in those context too)

mgoworko (changes requested):

mgoworko (approved):

PR #1116, opened 2025-05-18
[RORDEV-1414] ES node details audit reporting

mgoworko:
🚀New (ES) Added ES node name and cluster name in the ES cluster audit reporting

## Summary by CodeRabbit

- **New Features**
  - Audit logs now include Elasticsearch node and cluster names for enhanced traceability.
  - Audit log serializers can be switched dynamically, allowing audit entries to reflect different levels of environment detail.

- **Refactor**
  - Internal environment and node settings are now automatically included and passed throughout the system, improving consistency and reducing manual configuration.
  - Audit serializer classes and audit environment context have been modularized and extended for future flexibility.
  - The initialization of environment and node settings has been centralized, removing the need to pass node names explicitly.
  - Audit environment context is now integrated into configuration decoding and core factory initialization.
  - Index-level action filter and plugin components updated to internally derive node and cluster settings.

- **Bug Fixes**
  - None.

- **Chores**
  - Updated plugin version to 1.65.0-pre8.
  - Test suites and utilities updated to support new audit environment context and node settings.
  - Constructor signatures and factory initializations updated to include Elasticsearch node settings for improved test accuracy.
  - Added new helper methods in test utilities to provide consistent node and cluster settings for testing.
  - Integration tests enhanced to verify audit entries with and without environment-aware serializers.


core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -45,7 +45,8 @@ class ReadonlyRest(coreFactory: CoreFactory,
                    val indexConfigManager: IndexConfigManager,
                    val indexTestConfigManager: IndexTestConfigManager,
                    val authServicesMocksProvider: MutableMocksProviderWithCachePerRequest,
-                   val esEnv: EsEnv)
+                   val esEnv: EsEnv,
+                   val esNodeConfig: EsNodeConfig)
```
coutoPL, line 49: Maybe EsEnv should have EsNodeConfig (maybe we should call it EsNodeSettings because in the ROR domain we have configs, but ES has settings)? I think it fits conceptually

core/src/main/scala/tech/beshu/ror/configuration/ReadonlyRestEsConfig.scala
```diff
@@ -23,18 +23,23 @@ import tech.beshu.ror.utils.ScalaOps.*
 
 final case class ReadonlyRestEsConfig(bootConfig: RorBootConfiguration,
                                       sslConfig: RorSsl,
-                                      fipsConfig: FipsConfiguration)
+                                      fipsConfig: FipsConfiguration,
+                                      esNodeConfig: EsNodeConfig)
```
coutoPL, line 27: OTOH, IMO this doesn't fit here.

es67x/src/main/scala/tech/beshu/ror/es/IndexLevelActionFilter.scala
```diff
@@ -68,7 +68,8 @@ class IndexLevelActionFilter(nodeName: String,
   private val ror = ReadonlyRest.create(
     new EsIndexJsonContentService(client),
     auditSinkCreator,
-    EsEnv(env.configFile(), env.modulesFile())
+    EsEnv(env.configFile(), env.modulesFile()),
+    rorEsConfig.esNodeConfig,
```
coutoPL, line 72: In the `IndexLevelActionFilter` class constructor, we have a `nodeName` argument. Let's use EsNodeConfig instead.

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
```diff
@@ -152,13 +153,23 @@ object AuditingTool extends Logging {
...
             rorAuditIndexTemplate = RorAuditIndexTemplate.default,
-            auditCluster = AuditCluster.LocalAuditCluster
+            auditCluster = AuditCluster.LocalAuditCluster,
+            options = Options.default,
           )
+
+          final case class Options(enableReportingEsNodeDetails: Boolean)
+          object Options {
+            val default: Options = Options(
+              enableReportingEsNodeDetails = false,
```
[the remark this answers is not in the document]
coutoPL: Yes, I agree that it's wise to give users an option to fallback to previous behaviour when the described situation takes place. But I'm not sure if the toggle is the best option here.

Currently, the default serializer is `tech.beshu.ror.audit.instances.DefaultAuditLogSerializer` (BTW, the name is a little bit problematic now). Let's assume we created the `tech.beshu.ror.audit.instances.NewerDefaultAuditLogSerializer` (don't get attached to the name) with the additional fields. And starting from now, ROR will use the `NewerDefaultAuditLogSerializer`.

And let's assume that the user configured strict mapping and the error occurs. What could the user do?
IMO, they could configure `DefaultAuditLogSerializer` as a serializer and take advantage of the previous behaviour of auditing. 

----
How should we manage this use case?
1. We should communicate the change in the changelog entry.
2. We should document that, since ROR 1.6x.0, the default serialiser is the NewerDefaultAuditLogSerializer.
3. We should perhaps detect this problem by catching a proper exception when putting the audit document in place, and in the log we should describe what we think happened and link to the relevant document page.

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
coutoPL: I'm not sure.

I mean, yes — the `EsEnv` does describe the environment. To be honest, it's a wrapper for the ES `Environment` class. You can access the ES's `Settings` from the `Environment` instance.

You create an instance of `EsNodeSettings` from the `Settings`.

So, if the `Settings` class is strictly related to the `Environment `class, does that mean that our `EsNodeSettings` class should be part of the `EsEnv` class too?

Have I missed something?

audit/src/main/scala/tech/beshu/ror/audit/enrichers/AuditLogSerializerEnrichedWithEsNodeDetails.scala
```diff
@@ -0,0 +1,31 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.enrichers
+
+import tech.beshu.ror.audit.AuditLogSerializer
+
+class AuditLogSerializerEnrichedWithEsNodeDetails(underlying: AuditLogSerializer,
```
coutoPL, line 21: Currently, we configure serializers like that:
```
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: index
      serializer: tech.beshu.ror.requestcontext.QueryAuditLogSerializer
```

So it'd be nice if users could do the following:
```
serializer tech.beshu.ror.audit.enrichers.AuditLogSerializerEnrichedWithEsNodeDetails
```

mgoworko: Answered in other dicsussion - introduced versioning, and the `DefaultAuditLogSerializer` is an alias for the newest version

audit/src/main/scala/tech/beshu/ror/audit/AuditLogSerializer.scala
```diff
@@ -19,5 +19,6 @@ package tech.beshu.ror.audit
 import org.json.JSONObject
 
 trait AuditLogSerializer {
-  def onResponse(responseContext: AuditResponseContext): Option[JSONObject]
+  def onResponse(responseContext: AuditResponseContext,
```
coutoPL: What about backward compatibility here? The existing custom serializers should work with the new approach too.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -285,27 +285,28 @@ object AuditingSettingsDecoder extends Logging {
...
   private object DeprecatedAuditSettingsDecoder {
-    lazy val instance: Decoder[Option[AuditingTool.Settings]] = Decoder.instance { c =>
+    lazy val instance: Decoder[Option[AuditingTool.AuditSettings]] = Decoder.instance { c =>
       whenEnabled(c) {
         for {
           auditIndexTemplate <- decodeOptionalSetting[RorAuditIndexTemplate](c)("index_template", fallbackKey = "audit_index_template")
           customAuditSerializer <- decodeOptionalSetting[AuditLogSerializer](c)("serializer", fallbackKey = "audit_serializer")
           remoteAuditCluster <- decodeOptionalSetting[AuditCluster.RemoteAuditCluster](c)("cluster", fallbackKey = "audit_cluster")
-        } yield AuditingTool.Settings(
+          enableReportingEsNodeDetails <- c.downField("enable_reporting_es_node_details").as[Option[Boolean]]
```
coutoPL: Do we leave the toggle? Why?

core/src/main/scala/tech/beshu/ror/es/EsNodeSettings.scala
```diff
@@ -0,0 +1,19 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.es
+
+case class EsNodeSettings(nodeName: String, clusterName: String)
```
coutoPL: final

es67x/src/main/scala/tech/beshu/ror/es/utils/EsEnvProvider.scala
```diff
@@ -18,14 +18,19 @@ package tech.beshu.ror.es.utils
...
 object EsEnvProvider {
   def create(environment: Environment): EsEnv = {
+    val settings = environment.settings()
     EsEnv(
       configPath = environment.configFile(),
       modulesPath = environment.modulesFile(),
-      esVersion = EsVersion(major = Version.CURRENT.major, minor = Version.CURRENT.minor, revision = Version.CURRENT.revision)
+      esVersion = EsVersion(major = Version.CURRENT.major, minor = Version.CURRENT.minor, revision = Version.CURRENT.revision),
+      esNodeSettings = EsNodeSettings(
+        nodeName = settings.get("node.name"),
```
coutoPL: Do you confirm these two will always be non-null? Even if the user doesn't set them? (I'm not sure if they are required settings)

gradle.properties
```diff
@@ -1,5 +1,5 @@
 publishedPluginVersion=1.64.2
-pluginVersion=1.65.0-pre1
+pluginVersion=1.65.0-pre2
```
coutoPL: pre5?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/RemoteClusterAuditingToolsSuite.scala
```diff
@@ -61,4 +62,10 @@ class RemoteClusterAuditingToolsSuite
   override protected def baseRorConfig: String = resolvedRorConfigFile.contentAsString
 
   override protected def baseAuditDataStreamName: Option[String] = Option.when(isDataStreamSupported)("audit_data_stream")
+
+  // Adding the ES cluster fields is enabled in the /cluster_auditing_tools/readonlyrest.yml config file (`DefaultAuditLogSerializerV2` is used)
+  override def assertForEveryAuditEntry(entry: JSON): Unit = {
```
coutoPL, line 67: I'm not sure if this test covers this use case too, but I'd be nice to test the scenario as above:
```
1. The user starts ES with ROR with `DefaultAuditLogSerializerV1` configured
2. The user generates some traffic to create some audit documents
3. The user reconfigures ROR to use `DefaultAuditLogSerializerV2` and reloads config
4. The user generates some traffic to create some audit documents
```

The result - there should be no problem with the scenario above.

coutoPL (commented):
Sorry, I forgot about this PR :(

audit/src/main/scala/tech/beshu/ror/audit/AuditLogSerializer.scala
mgoworko: I modified it in a following way:
- the `def onResponse` is restored to its previous state
- the `AuditEnvironmentContext` is instead provided in a Serializer's constructor, with modified class Decoder implementation

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: Sorry, removed. I already removed it everywhere else, it was an unused leftover.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -42,40 +42,42 @@ import scala.util.{Failure, Success, Try}
 
 object AuditingSettingsDecoder extends Logging {
 
-  def instance(esVersion: EsVersion): Decoder[Option[AuditingTool.Settings]] = {
+  def instance(esVersion: EsVersion)
+              (implicit context: AuditEnvironmentContext): Decoder[Option[AuditingTool.AuditSettings]] = {
```
mgoworko, line 46: It seems, that inside this file the Scala 3 syntax is used, but this public method is defined using Scala 2 implicit syntax

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/RemoteClusterAuditingToolsSuite.scala
mgoworko: I added such test in `LocalClusterAuditingToolsSuite`

gradle.properties
mgoworko: pre5 was already released, so pre6

es67x/src/main/scala/tech/beshu/ror/es/utils/EsEnvProvider.scala
mgoworko: I changed it to use

```
        nodeName = Node.NODE_NAME_SETTING.get(settings),
        clusterName = ClusterName.CLUSTER_NAME_SETTING.get(settings).value(),
```

This way of getting settings fallbacks to ES defaults.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/LocalClusterAuditingToolsSuite.scala
```diff
@@ -48,4 +52,53 @@ class LocalClusterAuditingToolsSuite
...
+    entry.obj.get("es_node_name") shouldBe None
+    entry.obj.get("es_cluster_name") shouldBe None
+  }
+
+  "ES" should {
+    "submit audit entries" when {
+      "first request uses V1 serializer, then ROR config is reloaded and second request uses V2 serializer" in {
+        val initialConfig = getResourceContent(rorConfigFileName)
+        rorApiManager.updateRorInIndexConfig(initialConfig).forceOKStatusOrConfigAlreadyLoaded()
+        rorApiManager.reloadRorConfig().force()
```
coutoPL: You don't have to do the reload. The update method reloads the config too

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/LocalClusterAuditingToolsSuite.scala
```diff
@@ -48,4 +52,53 @@ class LocalClusterAuditingToolsSuite
...
+  override def assertForEveryAuditEntry(entry: JSON): Unit = {
+    entry.obj.get("es_node_name") shouldBe None
+    entry.obj.get("es_cluster_name") shouldBe None
+  }
+
+  "ES" should {
+    "submit audit entries" when {
+      "first request uses V1 serializer, then ROR config is reloaded and second request uses V2 serializer" in {
+        val initialConfig = getResourceContent(rorConfigFileName)
+        rorApiManager.updateRorInIndexConfig(initialConfig).forceOKStatusOrConfigAlreadyLoaded()
```
coutoPL: Let's hide some of the implementation details in private methods. Let the test look like this:
```
val initialConfig = getResourceContent(rorConfigFileName
updateRorSettings(...)
doExampleSearchRequest(...)

// here check the audit (we don't hide details so much here, because the reader is interested how the assertion looks like)

val newConfig = changeSerializer(initialConfig)
updateRorSettings(...)
doExampleSearchRequest(...)

// here check the audit (the same comment here)
```

audit/src/main/scala/tech/beshu/ror/audit/AuditLogSerializer.scala
coutoPL: In general, the solution is fine, but it may be complicated for our users. We would have to say to them: "Hey, to create a custom serializer, you have to implement this interface. But when you want to use also ES instance context, you have to create a constructor with param which is the ES instance context" ... or sth like that. Notice that forcing them to use Java or Scala is a big limitation from their perspective. So, IMO it should be as simple as possible. 

So, maybe we should create a separate interface with the method to implement, you proposed earlier? We would write in docs: use the interface A, or if you need es instance-specific context, use B. 

WDYT?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/LocalClusterAuditingToolsSuite.scala
mgoworko: refactored and now looks similar to the suggestion

audit/src/main/scala/tech/beshu/ror/audit/AuditLogSerializer.scala
mgoworko: Done. I added new interface `EnvironmentAwareAuditLogSerializer`. But I did not modify the solution that you reviewed much. I just added adapter, so that this new interface can be used as `AuditLogSerializer`. I also added test checking that this new trait can be used and works correctly.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/RawRorConfigBasedCoreFactory.scala
```diff
@@ -317,7 +319,11 @@ class RawRorConfigBasedCoreFactory(esVersion: EsVersion)
             dynamicVariableTransformationAliases.items.map(_.alias)
           )
         )
-        auditingTools <- AsyncDecoderCreator.from(AuditingSettingsDecoder.instance(esVersion))
+        auditEnvironmentContext = new AuditEnvironmentContext {
```
coutoPL: let's create a non-anonymous class for it (like we did with `AuditRequestContextBasedOnAclResult`)

coutoPL (approved):
LGTM. 
One minor comment - change it if you want to.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/RawRorConfigBasedCoreFactory.scala
mgoworko: done

PR #1122, opened 2025-06-05
[RORDEV-1502] Patching consent in env - tests

mgoworko:


ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/EnvProvider.scala
```diff
@@ -0,0 +1,25 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.tools.core.utils
+
+trait EnvProvider {
```
coutoPL: We have `EnvVarsProvider`. Do you think we could use it here instead?

mgoworko: I think that they may/should stay separate. There are 2 reasons:
- module dependencies - I wouldn't want to introduce any dependency on `core` in the `ror-tools` and I don't think that we should introduce any new common module for both `core` and `ror-tools`
- it is a minor and not very important reason, but we agreed that the consent flag should be case insensitive; so the interface of those 2 env providers are different - in core we get the env with given name, in ror-tools we find the env by case-insensitive name among all envs.

coutoPL: yeah, but:
1. We don't have to introduce a new module. We could move the EnvVarsProvider trait to the ror-tools-core module, and make the core dependent on the ror-tools-core module (all esXX modules already depend on ror-tools-core, so this manipulation is not a big abuse)
2. Do you think the interface has to change? IMO, we could only provide different implementation (eg. `OsEnvVarsCaseSensitiveProvider` or sth like that)

mgoworko: I applied the suggestion in https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1122/commits/192b1f5560d3af905ba3c685787b9a7f61124b0d. But I see mutliple problems with it:
1. In our tests, we use and test the `trait RorTools` and not the executable `object RorToolsApp`. It means, that if we move some logic outside of the `trait RorTools` we no longer test it - because it has to be mocked (provided in a different way than in a real ror-tools app) in tests
2. That is why I think that env provider in ror-tools should be as simple as possible and should just return raw map of env variables. It is not compatilbe with the already existing interface from `core` module
3. The ror-tools and the plugin code are 2 separates use cases with different runtime environment and purposes. I think that a console app may have its own env provider. It has to have a dedicated implementation of the provider either way.
4. After I made the changes, I noticed that the dependency on ror-tools-core util looks weird. I mean, that it raises question why the util from console app is used in plugin, when it has a separate implementation anyway.

As a result I reverted the refactor and only made a small name changes. I really think that the current version is better, but I can restore the refactor with common provider if you still feel strongly about it

coutoPL (approved):
LGTM

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/EnvProvider.scala
coutoPL: ok, then. Let's leave it as it's now :)

PR #1125, opened 2025-06-07
[RORDEV-1502] Fixed ror-tools file permissions

mgoworko:
🐞Fix (ES) Preserve file owner and permissions during ror-tools patch and unpatch operations

---
- The file owner could change, when Jar manifest is modified during patching operation
- but there was also an older, preexisting issue - backup and restore operations did not preserve file owner

## Summary by CodeRabbit

* **New Features**
  * File operations now preserve original file permissions and ownership during backup, restore, and JAR modification processes.

* **Refactor**
  * Improved and modularized handling of file permissions and ownership, introducing new extension methods for easier management.
  * Updated internal logic for modifying files to explicitly manage file metadata, enhancing reliability and maintainability.

* **Chores**
  * Updated plugin version to 1.65.0-pre9.


ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/FileUtils.scala
```diff
@@ -31,12 +31,16 @@ object FileUtils {
...
-    val originalFileOwner = Files.getOwner(jar.toPath)
-    val originalFilePermissions = getOriginalPermissions(jar.toPath)
-    modifyJar(jar)
-    Files.setOwner(jar.toPath, originalFileOwner)
-    setOriginalPermissions(jar.toPath, originalFilePermissions)
+  def modifyFileWithMaintainingOriginalPermissionsAndOwner(file: File)(modifyFile: File => Unit): Unit = {
+    modifyFileWithMaintainingOriginalPermissionsAndOwner[File](file, identity) { f => modifyFile(f); f }
+  }
+
+  def modifyFileWithMaintainingOriginalPermissionsAndOwner[FILE](file: FILE, toJavaFile: FILE => File)(modifyJar: FILE => FILE): Unit = {
```
coutoPL: TBH, I'd be nice if we could get rid of `java.io.File` from our public methods. Better files lib is so nice, it'd be better to use it instead. Mixing these two worlds doesn't look nice.

So, let's modify the previous interface of the method and adapt other places where we are not so consistent with using better files.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/filePatchers/JarManifestModifier.scala
```diff
@@ -28,7 +29,8 @@ object JarManifestModifier {
 
   private val patchedByRorVersionPropertyName = "Patched-By-Ror-Version"
 
-  def addPatchedByRorVersionProperty(file: File, rorVersion: String): Unit = {
+  def addPatchedByRorVersionProperty(file: File,
+                                     rorVersion: String): Unit = modifyFileWithMaintainingOriginalPermissionsAndOwner[File](file, _.toJava) { file =>
```
coutoPL: Ah, it works here, but don't you think it's hard to reason about? Maybe the `modifyFileWithMaintainingOriginalPermissionsAndOwner` method's interface is not so good?

I mean, the method suggests we modify the jar file, but TBH, we don't have to modify it. 
Do you have an idea how to do it better?

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -46,11 +47,19 @@ private[patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
   }
 
   def backup(file: Path): Unit = {
-    os.copy(from = file, to = backupFolderPath / file.last, replaceExisting = true)
+    modifyFileWithMaintainingOriginalPermissionsAndOwner[Path](file, _.toIO) { originalFile =>
+      val backedUpFile = backupFolderPath / originalFile.last
+      os.copy(from = originalFile, to = backedUpFile, replaceExisting = true, copyAttributes = true)
```
coutoPL: we have os.copy in copyToPluginPath too. But it's not used. Maybe we should remove it?

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
```diff
@@ -46,11 +47,19 @@ private[patches] class RorPluginDirectory(val esDirectory: EsDirectory) {
   }
 
   def backup(file: Path): Unit = {
-    os.copy(from = file, to = backupFolderPath / file.last, replaceExisting = true)
+    modifyFileWithMaintainingOriginalPermissionsAndOwner[Path](file, _.toIO) { originalFile =>
```
coutoPL: here the usage of `modifyFileWithMaintainingOriginalPermissionsAndOwner` looks strange too. Becase we don't modify the file

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/filePatchers/JarManifestModifier.scala
mgoworko: I completely refactored the interface. Now there are new extensions method for getting file permissions and owner, setting them and copying them from another, original, file.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/FileUtils.scala
mgoworko: Interface now uses better files File. There are conversion methods from os.Path and java.io.File

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
mgoworko: Refactored, please see the comment in JarManifestModifier and the new implementation.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/RorPluginDirectory.scala
mgoworko: copyToPluginPath is used by CopyTransportNetty4JarToPluginPatch

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/FileUtils.scala
```diff
@@ -31,13 +33,38 @@ object FileUtils {
...
+  given osPathToFile: Conversion[os.Path, File] with
+    def apply(path: os.Path): File = File(path.toString)
+
+  given javaFileToFile: Conversion[java.io.File, File] with
+    def apply(jFile: java.io.File): File = File(jFile.toPath)
+
+  sealed trait FilePermissionsAndOwner
+
+  // The implementation details of FilePermissionsAndOwner should not leak outside of this file
+  private final case class FilePermissionsAndOwnerImpl(filePermissions: Any,
```
coutoPL: maybe `OriginalFilePermissionsAndOwner` instead of Impl-suffix? ;)

coutoPL:
> Plus, we have to answer two questions:
> 
> 1. Why doesn't the issue break anything? Or maybe it does in some circumstances?
> 2. What about the files that now have the wrong owner? How to fix it? Or maybe we don't have to do that?

In general it LGTM, but let's try to answer these two questions before we merge it.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/FileUtils.scala
mgoworko: Sorry, I wrote, but forgot to push my answers and comments before requesting review today. I pushed them now.

About the questions:
1. 
- About preexisting issue (before jar manifest changes) with backup/restore file permissions
  - it only affected those users who upatched ES, and it was done as non-`elasticsearch` user.
  - It is possible, that it was never noticed, because the typical use case is to unpatch, and then immediately upgrade ES, which replaces the affected jar files with correct ones.
- The more recent issue with jar manifest 
  - is a bit different because it affects current jar files used by ES
  - again, it only affect users who patched using non-elasticsearch user
  - see p.2 - it only causes problems if elasticsearch user cannot read jar files owned by root - and my recent (but very limited) test showed that it can read those files and ES starts

2.
- It is hard to analyze, because each installation type is different and we do not have user reports about problems. 
- in my local installation from apt, all jar files are owned by root - and that is ok, as long as elasticsearch user can read them (and it can)
- It is possible that we do not have to fix files already owned by users other than elasticsearch. The owner should be fixed on next ES upgrade, and as long as ES can read those files it will continue to work.

coutoPL: ok, thanks. So, let's see if anyone reports further problems with this.

coutoPL (approved):
LGTM

PR #1127, opened 2025-06-08
[RORDEV-1453] Ubuntu apt-get based installation of ES in tests

mgoworko:
## Summary by CodeRabbit

* **New Features**
  * Added support for running and testing Elasticsearch containers installed via Ubuntu apt as an alternative to the official Docker image.
  * Introduced the ability to specify and handle different Elasticsearch installation types in container management utilities.
  * Added new utilities for capturing and combining Docker container logs during tests.

* **Improvements**
  * Enhanced configuration flexibility for Elasticsearch containers, including dynamic command and configuration directory handling based on installation type.
  * Expanded version-matching patterns for Elasticsearch modules to improve test coverage and granularity.
  * Streamlined integration test execution in the CI pipeline.
  * Extended container creation APIs to optionally accept additional log consumers for enhanced log handling during tests.
  * Improved Docker image building by supporting setting CMD instructions from image descriptions.

* **Bug Fixes**
  * Ensured correct log consumer handling and user context switching during container setup for improved reliability in tests.

* **Tests**
  * Introduced a new test suite to verify ReadonlyREST plugin behavior across different Elasticsearch installation methods and versions.

### DIFF
--- a/ci/run-pipeline.sh
+++ b/ci/run-pipeline.sh
@@ -54,7 +54,7 @@ run_integration_tests() {
   ES_MODULE=$1
 
   echo ">>> $ES_MODULE => Running testcontainers.."
-  ./gradlew ror-tools:test integration-tests:test "-PesModule=$ES_MODULE" || (find . | grep hs_err | xargs cat && exit 1)
+  ./gradlew --no-daemon ror-tools:test integration-tests:test "-PesModule=$ES_MODULE" || (find . | grep hs_err | xargs cat && exit 1)
 }
 
 if [[ -z $TRAVIS ]] || [[ $ROR_TASK == "integration_es90x" ]]; then
--- a/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/DockerImageCreator.scala
+++ b/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/DockerImageCreator.scala
@@ -92,7 +92,7 @@ object DockerImageCreator extends StrictLogging {
     def setCommandFrom(imageDescription: DockerImageDescription): DockerfileBuilder = {
       imageDescription
         .command
-        .foldLeft(builder) { case (b, command) => b.user("elasticsearch").cmd(command) }
+        .foldLeft(builder) { case (b, command) => b.cmd(command) }
     }
 
   }
--- a/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
+++ b/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
@@ -75,9 +75,7 @@ class ReadonlyRestPlugin(esVersion: String,
       .copyFile(esConfig.esConfigDir / "elastic-certificates-cert.pem", fromResourceBy(name = "elastic-certificates-cert.pem"))
       .copyFile(esConfig.esConfigDir / "elastic-certificates-pkey.pem", fromResourceBy(name = "elastic-certificates-pkey.pem"))
       .updateFipsDependencies(esConfig)
-      .user("elasticsearch")
       .copyFile(esConfig.esConfigDir / "readonlyrest.yml", config.rorConfig)
-      .user("root")
       .installRorPlugin()
       .when(performPatching, _.patchES())
   }


integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/support.scala
```diff
@@ -29,51 +33,84 @@ object support {
...
-      with ForAllTestContainer {
+      with ForAllTestContainer
+      with BeforeAndAfterAll {
     this: Suite with EsClusterProvider with ESVersionSupport =>
 
     override lazy val container: EsClusterContainer = clusterContainer
 
+    override protected def afterAll(): Unit = {
+      super.afterAll()
+      pruneDockerImages()
```
mgoworko: Adding the new test with apt-based ES caused the jobs to fail - not enough space on runner. Probably caused by downloading ~700 MB of ES package. In the first step I added cleaning of the resulting Docker images in newly added tests, but it was not enough. So I added pruning unused Docker images after each test suite.

mgoworko: So to sum up - the support for apt installation is added, the test is added and it detects the problematic situation. But there are 2 issues on pipeline:
- test jobs take longer time, even 15 minutes - because of downloading ubuntu, apt ES and building image
- the memory on runner is near 100%, I added pruning images after test suites, but I guess it contributed to a longer run time too - because of time needed for pruning and not reusing parts of images
- Simone mentioned about issue with packages not being available as soon as needed in the apt repo, but I guess we will see when next ES version is released. We will be able to just disable the test until apt package is available

coutoPL: let's add a node why we had to do this

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/support.scala
```diff
@@ -29,51 +33,84 @@ object support {
...
   trait SingleClientSupport extends SingleClient with SingleEsTarget
+
   trait MultipleClientsSupport extends MultipleClients with MultipleEsTargets
 }
+
+private def pruneDockerImages(): Unit = {
+  val logger: Logger = LogManager.getLogger("prune-docker-images")
+  val dockerClient: DockerClient = DockerClientFactory.instance().client()
+  logger.info("Pruning docker images after test suite")
+  val reclaimedSpace = 1.0 * dockerClient.pruneCmd(PruneType.IMAGES).withDangling(false).exec().getSpaceReclaimed / 1024 / 1024
```
coutoPL: Is it too invasive? On the runner - no problem, but when I run tests locally, I don't want to prune all my images. So, maybe we can sth like that?
```
1. list images
2. grep <pattern>
3. remove image by IDs
```

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -77,21 +94,47 @@ class Elasticsearch(esVersion: String,
...
   def toDockerImageDescription: DockerImageDescription = {
-    DockerImageDescription
-      .create(s"docker.elastic.co/elasticsearch/elasticsearch:$esVersion", customEntrypoint)
+    val baseImage = config.esInstallationType match {
+      case EsInstallationType.EsDockerImage =>
+        DockerImageDescription
+          .create(s"docker.elastic.co/elasticsearch/elasticsearch:$esVersion", customEntrypoint)
+      case EsInstallationType.UbuntuDockerImageWithEsFromApt =>
+        val esMajorVersion: String = esVersion.split("\\.")(0) + ".x"
+        DockerImageDescription
```
coutoPL: Let's move this to a private method. It would be nice if it could look like this: 
`DockerImageDescription.createCustomImageWithESInstalledFromApt(esVersion, customEntrypoint)` to fed the need of symmetry ;)

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -77,21 +94,47 @@ class Elasticsearch(esVersion: String,
...
+        destination = configDir(config) / "log4j2.properties",
         file = log4jFileFromResources
       )
       .user("root")
       // Package tar is required by the RorToolsAppSuite, and the ES >= 9.x is based on
       // Red Hat Universal Base Image 9 Minimal, which does not contain it.
-      .runWhen(Version.greaterOrEqualThan(esVersion, 9, 0, 0), "microdnf install -y tar")
-      .run(s"chown -R elasticsearch:elasticsearch ${configDir.toString()}")
+      .runWhen(missingTar, "microdnf install -y tar")
+      .run(s"chown -R elasticsearch:elasticsearch ${esDir.toString()}")
```
coutoPL, line 148: Why do we have to do this?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
```diff
@@ -135,15 +137,15 @@ class ReadonlyRestPlugin(esVersion: String,
...
   private implicit class UpdateFipsDependencies(val image: DockerImageDescription) {
-    def updateFipsDependencies(): DockerImageDescription = {
+    def updateFipsDependencies(esConfig: Elasticsearch.Config): DockerImageDescription = {
       if (isFibsEnabled) {
         image
-          .copyFile(configDir / "additional-permissions.policy", fromResourceBy(name = "additional-permissions.policy"))
-          .copyFile(configDir / "ror-keystore.bcfks", fromResourceBy(name = "ror-keystore.bcfks"))
-          .copyFile(configDir / "ror-truststore.bcfks", fromResourceBy(name = "ror-truststore.bcfks"))
-          .copyFile(configDir / "elastic-certificates.bcfks", fromResourceBy(name = "elastic-certificates.bcfks"))
+          .copyFile(configDir(esConfig) / "additional-permissions.policy", fromResourceBy(name = "additional-permissions.policy"))
```
coutoPL: What do you think about using an extension method instead here? Like this: `esConfig.esConfigDir`?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/logs/DockerLogsToStringConsumer.scala
```diff
@@ -0,0 +1,40 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.utils.containers.logs
+
+import org.testcontainers.containers.output.OutputFrame
+
+import java.util.function.Consumer
+import scala.language.postfixOps
+
+class DockerLogsToStringConsumer() extends Consumer[OutputFrame] {
```
coutoPL: Do we need `()` here?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/support.scala
coutoPL: 1. Do we have ideas how to optimize it?
  - Do we have to use the Ubuntu image? Maybe we can use a thinner alternative
  - Are we sure we don't prune this image between the tests?
  - Any other ideas for current or/and future improvements in this matter?
2. The thing Simone mentioned is important. We want to release a new version as soon as we can, so IMO it's better to address this issue now. Let's compare the dates of releasing the latest versions (the docker images and apt packages). Maybe we could check if there is a way to build the custom ES images, and if it's not, the test could be skipped. WDYT?

mgoworko: 1.

The problem was a  single line in dockerfile (chown). It caused the image size to grow 1GB. After changing to chmod 755 (with comment explaining why we need it at all, details in other discussion) the problem is gone. No pruning required.

2.

I checked it:
- ES 9.0.3 was released on 24 June 2025, our internal notification arrived exactly at 5 PM https://www.elastic.co/blog/elastic-stack-9-0-3-released
- the apt package was available on 24 June 2025 at 11 AM curl -sI https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-9.0.3-amd64.deb | grep -i last-modified
last-modified: Tue, 24 Jun 2025 11:05:55 GMT
- the same applies for 9.0.1 and 9.0.0 - apt package is released on the same day as official announcement

So it seems there should be no ptoblem with that.

mgoworko: I removed the cleaning logic - explanation in the comment obove.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/support.scala
mgoworko: I removed the cleaning logic - explanation in the comment obove.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
mgoworko: I created 2 separate methods, instead of if's and conditions. The syntax is a bit different than suggested - the actual creation of Ubuntu/Official ES docker image is hidden in the util internals, and is not a public/widely used method.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
mgoworko: This single line was responsible for all the pain and suffering in this PR. It caused the built docker image to be 1GB larger that without it. It caused the OOM on pipeline (because we've already used 96% of available storage before this change, this was the last straw)

Changed to 
```
      // ES is started as Docker CMD, so elasticsearch user must have permission to read ES files.
      // In standard Ubuntu with ES from apt it is not necessary, because ES is executed from systemd
      .run(s"chmod 755 ${esDir.toString()}")

File: ci/run-pipeline.sh [hunk 7]
```diff
@@ -54,8 +54,7 @@ run_integration_tests() {
   ES_MODULE=$1
 
   echo ">>> $ES_MODULE => Running testcontainers.."
-  ./gradlew ror-tools:test "-PesModule=$ES_MODULE" || (find . | grep hs_err | xargs cat && exit 1)
-  ./gradlew integration-tests:test "-PesModule=$ES_MODULE" || (find . | grep hs_err | xargs cat && exit 1)
+  ./gradlew ror-tools:test integration-tests:test "-PesModule=$ES_MODULE" || (find . | grep hs_err | xargs cat && exit 1)
```
coutoPL: if the two tasks are merged, it seems we can add `--no-daemon` option

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/DockerImageCreator.scala
```diff
@@ -87,5 +89,11 @@ object DockerImageCreator extends StrictLogging {
         .foldLeft(builder) { case (b, entrypoint) => b.entryPoint(entrypoint.toIO.getAbsolutePath) }
     }
 
+    def setCommandFrom(imageDescription: DockerImageDescription): DockerfileBuilder = {
+      imageDescription
+        .command
+        .foldLeft(builder) { case (b, command) => b.user("elasticsearch").cmd(command) }
```
coutoPL: I'm not sure about it. We quietly change the user without setting the previous one back after the cmd

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
```diff
@@ -66,16 +66,18 @@ class ReadonlyRestPlugin(esVersion: String,
...
-      .copyFile(configDir / "readonlyrest.yml", config.rorConfig)
+      .copyFile(esConfig.esConfigDir / "ror-keystore.jks", fromResourceBy(name = "ror-keystore.jks"))
+      .copyFile(esConfig.esConfigDir / "ror-truststore.jks", fromResourceBy(name = "ror-truststore.jks"))
+      .copyFile(esConfig.esConfigDir / "elastic-certificates.p12", fromResourceBy(name = "elastic-certificates.p12"))
+      .copyFile(esConfig.esConfigDir / "elastic-certificates-cert.pem", fromResourceBy(name = "elastic-certificates-cert.pem"))
+      .copyFile(esConfig.esConfigDir / "elastic-certificates-pkey.pem", fromResourceBy(name = "elastic-certificates-pkey.pem"))
+      .updateFipsDependencies(esConfig)
+      .user("elasticsearch")
+      .copyFile(esConfig.esConfigDir / "readonlyrest.yml", config.rorConfig)
+      .user("root")
```
coutoPL: We don't have to install the plugin as the root user. But we have to patch it as root.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/DockerImageCreator.scala
mgoworko: changing user removed, it was a leftover from some previous version of this PR, where the order of operations was different and it was needed

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/ReadonlyRestPlugin.scala
mgoworko: changing user removed, it was a leftover from some previous version of this PR, where the order of operations was different and it was needed

PR #1130, opened 2025-06-19
[RORDEV-1474] Bump dependency-check plugin version

mateuszkp96:
## Summary by CodeRabbit

- **Chores**
  - Improved CVE database caching in the pipeline for faster and more reliable vulnerability checks.
  - Updated the OWASP Dependency Check Gradle plugin to a newer version.
  - Upgraded the version of the refined_3 library in core and test utilities.
  - Bumped the plugin version to 1.65.0-pre6.
  - Cleaned up and updated CVE suppression rules for more accurate vulnerability reporting.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,8 +4,7 @@ variables:
   isDevelop: $[eq(variables['Build.SourceBranch'], 'refs/heads/develop')]
   isEpic: $[or(contains(variables['Build.SourceBranch'], 'epic/'), contains(variables['System.PullRequest.SourceBranch'], 'epic/'))]
   isPullRequest: $[eq(variables['Build.Reason'], 'PullRequest')]
-  currentDate: $[format('{0:yyyyMM}', pipeline.startTime)]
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -4,6 +4,9 @@ variables:
   isDevelop: $[eq(variables['Build.SourceBranch'], 'refs/heads/develop')]
   isEpic: $[or(contains(variables['Build.SourceBranch'], 'epic/'), contains(variables['System.PullRequest.SourceBranch'], 'epic/'))]
   isPullRequest: $[eq(variables['Build.Reason'], 'PullRequest')]
+  currentDate: $[format('{0:yyyyMM}', pipeline.startTime)]
```
coutoPL: maybe `cacheScopeMonth` or even better (to avoid naming it), let's inline it in the `dependencyCheckCacheKey` value?

coutoPL (approved):
LGTM.
Only one minor comment

PR #1140, opened 2025-07-25
[RORDEV-1481] New audit serializers, including fully configurable serializer

mgoworko:


audit/src/main/scala/tech/beshu/ror/audit/instances/BaseAuditLogSerializer.scala
```diff
@@ -0,0 +1,240 @@
...
+    case object InvolvedIndices extends AuditValuePlaceholder
+
+    case object AclHistory extends AuditValuePlaceholder
+
+    case object ProcessingDurationMillis extends AuditValuePlaceholder
+
+    // Identifiers
+    case object Timestamp extends AuditValuePlaceholder
+
+    case object Id extends AuditValuePlaceholder
```
mgoworko: I think that we should discuss the field names, especially the various id types. At the moment they correspond to the names that already existed, but maybe we could make them more readable.

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
```diff
@@ -17,82 +17,42 @@
...
-      }
-    }
+}
 
-    private def jsonKeys(json: JSONObject) = {
-      Option(JSONObject.getNames(json)).toList.flatten
-    }
-  }
+object DefaultAuditLogSerializerV1 {
+  val defaultV1AuditFields: Map[String, AuditFieldValue] = Map(
```
mgoworko: Each serializer now uses the `BaseAuditLogSerializer.serialize`, only with different arguments:
- fields (serializers have the predefined list of fields, like this one; only `Configurable` serializer gets its list of fields from config
- `AllowedEventSerializationMode` (that was the missing functionality - to log all ALLOWED events, even those for rules with INFO level verbose; there are new serializers `ReportingAllEvents` with and without query; for `Configurable` serializer this setting is defined in config)

integration-tests/src/test/resources/ror_audit/enabled_auditing_tools/readonlyrest.yml
```diff
@@ -6,9 +6,17 @@ readonlyrest:
       - type: index
         index_template: "'audit_index'"
         serializer: "tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV1"
+        serialize_all_allowed_events: false
+        fields:
+          node_name_with_static_suffix: "{ES_NODE_NAME} with suffix"
```
mgoworko: It's maybe an overkill to allow that much configurability, but it basically allows to define a serializer in config, without writing any code. 

Admins will be able to define any fields they want. The values are strings, that may contain any values defined by user, combined with value placeholders provided by ROR. If only our placeholder with number-like value is used, then it is reported as number. Other values and combinations of values are strings.

audit/build.gradle
```diff
@@ -63,13 +63,36 @@ crossBuild {
 
 dependencies {
     implementation                 group: 'org.json',           name: 'json',                version: '20231013'
+    crossBuildV211Implementation   group: 'com.beachape',       name: 'enumeratum_2.11',     version: '1.7.2'
```
coutoPL: TBH, the audit module should be written in Java. For sure, when someone wants to implement their own serializer, it's more likely they know Java than Scala. For that reason, we use `org.json` above ;)

So, let's get rid of the enumeratum and let's use a more java-friendly approach (maybe we don't have to add any other dependency)

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
```diff
@@ -0,0 +1,54 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit
+
+final case class AuditFieldValue(value: List[AuditFieldValuePlaceholder])
```
coutoPL: I think this class (and all other classes related to configuring the audit serializer from ROR settings should be defined in the ROR's core). Notice that users won't be extending these classes (at least I think we should not let them do it).

audit/build.gradle
coutoPL: If we move it to the code, Enumeratum is ok there.

audit/src/main/scala/tech/beshu/ror/audit/BaseAuditLogSerializer.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditResponseContext._
+import tech.beshu.ror.audit.instances.SerializeUser
+
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.FiniteDuration
+
+object BaseAuditLogSerializer {
```
coutoPL: It's rather kind of a serialization helper rather than a base serializer, right?

audit/src/main/scala/tech/beshu/ror/audit/BaseAuditLogSerializer.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  private val timestampFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'").withZone(ZoneId.of("GMT"))
+
+  def serialize(responseContext: AuditResponseContext,
+                environmentContext: AuditEnvironmentContext,
+                fields: Map[String, AuditFieldValue],
```
coutoPL: I'm not sure if it's clear right here:
1. the `String` is a document field name (maybe let's use a type alias)
2. the `AuditFieldValue` is not a document field value, but a document field value descriptor or sth like that (maybe we should rename it)

audit/src/main/scala/tech/beshu/ror/audit/BaseAuditLogSerializer.scala
```diff
@@ -0,0 +1,147 @@
...
+      Map(
+        "@timestamp" -> timestampFormatter.format(requestContext.timestamp)
+      ) ++ fields.map {
+        case (key, field) =>
+          val resolvedValue = field.value.map {
+            resolvePlaceholder(_, environmentContext, matched, finalState, reason, duration, requestContext, error)
+          } match {
+            case Nil => ""
+            case singleElement :: Nil => singleElement
+            case multipleElements => multipleElements.mkString
```
coutoPL: I'm not sure about it. 
I mean, it could be nice (maybe not in the current iteration of the feature) to define some form of field values composition, but this one (concatenation with two strings with no separator) doesn't seem to be useful. Maybe we should get rid of it now?

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
```diff
@@ -17,82 +17,42 @@
...
-import tech.beshu.ror.audit.{AuditLogSerializer, AuditRequestContext, AuditResponseContext}
+import tech.beshu.ror.audit.BaseAuditLogSerializer.AllowedEventSerializationMode
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV1.defaultV1AuditFields
+import tech.beshu.ror.audit._
 
-import java.time.ZoneId
-import java.time.format.DateTimeFormatter
-import scala.collection.JavaConverters._
-import scala.concurrent.duration.FiniteDuration
+class DefaultAuditLogSerializerV1(environmentContext: AuditEnvironmentContext) extends AuditLogSerializer {
```
coutoPL: are we backward compatible?
What if someone extends `DefaultAuditLogSerializerV1` in their custom serializer?

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllEventsAuditLogSerializer.scala
```diff
@@ -0,0 +1,29 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.BaseAuditLogSerializer.AllowedEventSerializationMode
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV2.defaultV2AuditFields
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, BaseAuditLogSerializer}
+
+class ReportingAllEventsAuditLogSerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializer(environmentContext) {
```
coutoPL: maybe `ReportingAllTypesOfEventsAuditLogSerializer`?

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllEventsWithQueryAuditLogSerializer.scala
```diff
@@ -0,0 +1,29 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.BaseAuditLogSerializer.AllowedEventSerializationMode
+import tech.beshu.ror.audit.instances.QueryAuditLogSerializerV2.queryV2AuditFields
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, BaseAuditLogSerializer}
+
+class ReportingAllEventsWithQueryAuditLogSerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializer(environmentContext) {
```
coutoPL: the same here

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,24 +208,61 @@ object AuditingSettingsDecoder extends Logging {
       }
       .decoder
 
+  given auditLogSerializerDecoder(using context: AuditEnvironmentContext): Decoder[Option[AuditLogSerializer]] = Decoder.instance { c =>
+    for {
+      serializeAllAllowedEvents <- c.downField("serialize_all_allowed_events").as[Option[Boolean]]
+      given Option[AllowedEventSerializationMode] = serializeAllAllowedEvents.map(c => if (c) SerializeAllAllowedEvents else SerializeOnlyAllowedEventsWithInfoLevelVerbose)
```
coutoPL: I'm not sure if passing the values using implicits is readable here. Why not declare decoder as a method?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,24 +208,61 @@ object AuditingSettingsDecoder extends Logging {
...
+      fields <- c.downField("fields").as[Option[Map[String, AuditFieldValue]]]
```
coutoPL: Adding "fields" option to any existing serializer is tempting, but when we do that, do we need the `ConfigurableQueryAuditLogSerializer` instance?

audit/src/main/scala/tech/beshu/ror/audit/BaseAuditLogSerializer.scala
mgoworko: 1. Done, added type AuditFieldName
2. I changed responsibility of AuditFieldValue. It is now sealed trait with all the available values + static text.

audit/src/main/scala/tech/beshu/ror/audit/BaseAuditLogSerializer.scala
mgoworko: -> `AuditSerializationHelper`

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
mgoworko: It didn't think of it and it was not compatible because of `environmentContext`. I reverted it to be compatible.

audit/build.gradle
mgoworko: All the code responsible for creating configurable serializer is moved to core. Dependency on enumeratum removed from audit.

audit/src/main/scala/tech/beshu/ror/audit/BaseAuditLogSerializer.scala
mgoworko: This part of code is moved to core and refactored, but It 
- allows to do configure field like that:  `node_name_with_static_suffix: "{ES_NODE_NAME} with suffix"`
- it will be resolved to 
```
AuditFieldValue.Combined(
  List(
    AuditFieldValue.EsNodeName, 
    AuditFieldValue.StaticText(" with suffix")
  )
)
```
- and it will be serialized, for example, to string "ROR_SINGLE with suffix"

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: Decoders are refactored, I think this issue is solved and decoding is more readable.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: It was only decoded to optional value, which was used only for configurable serialized. But it was not clean, decoders are refactored.

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV2.scala
```diff
@@ -17,16 +17,24 @@
...
     )
-    super.onResponse(responseContext)
-      .map(additionalFields.foldLeft(_) { case (soFar, (key, value)) => soFar.put(key, value) })
-  }
+
+}
+
+object DefaultAuditLogSerializerV2 {
+  val defaultV2AuditFields: Map[AuditFieldName, AuditFieldValue] =
+    DefaultAuditLogSerializerV1.defaultV1AuditFields ++ environmentRelatedAuditFields
```
coutoPL: Are you sure we should refer to the V1 fields in the V2 implementation?
Maybe it'd be better to redefine them here?

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllTypesOfEventsAuditLogSerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.AllowedEventSerializationMode
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV2.defaultV2AuditFields
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, AuditSerializationHelper}
+
+class ReportingAllTypesOfEventsAuditLogSerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializer(environmentContext) {
```
coutoPL: I'm still thinking about this name. Because again, I had a problem recalling what "all types of events" mean.

WDYT about sth like this:
1. let's rename DefaultAuditLogSerializerVX `SkipAllowedAuditLogSerializerVX` (or sth like that) serializer
2. let's have "an alias" class DefaultAuditLogSerializerVX that extends SkipAllowedAuditLogSerializerVX
3. let's rename `ReportingAllTypesOfEventsAuditLogSerializer` to  `FullAuditLogSerializer` (or sth like that`

WDYT?

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
```diff
@@ -0,0 +1,141 @@
...
+          json
+      }
+    }
+
+    private def jsonKeys(json: JSONObject) = {
+      Option(JSONObject.getNames(json)).toList.flatten
+    }
+  }
+
+  sealed trait AllowedEventSerializationMode
```
coutoPL: WDYT?
```
sealed trait AllowedEventMode
object AllowedEventMode {
  case object IncludeAll extends AllowedEventMode
  case object Include(eventTypes: NonEmptyList[EventType]) extends AllowedEventMode  
}
```

The current version is too concreted. It'd be hard to maintain backward compatibility using it in the future.

audit/src/main/scala/tech/beshu/ror/audit/EnvironmentAwareAuditLogSerializer.scala
```diff
@@ -17,8 +17,16 @@
 package tech.beshu.ror.audit
 
 import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.AuditFieldName
 
 trait EnvironmentAwareAuditLogSerializer {
```
coutoPL, line 23: I found no implementation of this serializer. 
Now, I'm not sure if we still support the changes described in RORDEV-1414.

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV2.scala
```diff
@@ -17,16 +17,24 @@
...
 
 import org.json.JSONObject
-import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext}
+import tech.beshu.ror.audit._
+import tech.beshu.ror.audit.AuditSerializationHelper.{AllowedEventSerializationMode, AuditFieldName}
+import tech.beshu.ror.audit.EnvironmentAwareAuditLogSerializer.environmentRelatedAuditFields
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV2.defaultV2AuditFields
 
-class DefaultAuditLogSerializerV2(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializerV1 {
+class DefaultAuditLogSerializerV2(environmentContext: AuditEnvironmentContext) extends AuditLogSerializer {
```
coutoPL: this alias could stay, but we should have a better name for this. See the previous, corresponding comment.

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
```diff
@@ -0,0 +1,141 @@
...
+        case (fieldName, fieldValue) =>
+          fieldName.value -> resolvePlaceholder(fieldValue, matched, finalState, reason, duration, requestContext, environmentContext, error)
+      }
+    }
+    resolvedFields
+      .foldLeft(new JSONObject()) { case (soFar, (key, value)) => soFar.put(key, value) }
+      .mergeWith(requestContext.generalAuditEvents)
+  }
+
+  private def resolvePlaceholder(auditValue: AuditFieldValue,
```
coutoPL: Once again, I was fooled by the name: `AuditFieldValue` ;)
For me, it's rather `AuditFieldValueDescriptor`. Because, base on this description and the available data, we create value for the audit event document's filed.

TBH, the `requestContext`, `finalState`, ..., `error` values should be gathered in some container for a sake of readability:

```scala
private def resolve(valueDescriptor: AuditFieldValueDescriptor, eventData: EventData): Any
```

but introducing and creating a new type for a sake of internal implementation may not be a base choice. That's why I propose to do it like this:

```scala
private def resolver(matched: Boolean,
                          finalState: String,
                          reason: String,
                          duration: FiniteDuration,
                          requestContext: AuditRequestContext,
                          environmentContext: Option[AuditEnvironmentContext],
                          error: Option[Throwable]): AuditFieldValueDescriptor => Any
```

Now, it's pretty clear that you create a resolver function to get value from the descriptor.

WDYT?

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
```diff
@@ -0,0 +1,89 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit
+
+private[ror] sealed trait AuditFieldValue
```
coutoPL: `private[ror]`
Is it enough?
When we extend the sealed trait values set in the future, are we sure that users won't have to recompile their custom serializers?

coutoPL (changes requested):

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV2.scala
mgoworko: Ok, good point, it may be better to define them separately.

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
mgoworko: Done, with set of verbosity levels, and adjusted logic in AuditSerializationHelper
```
  object AllowedEventMode {
    case object IncludeAll extends AllowedEventMode

    final case class Include(types: Set[Verbosity]) extends AllowedEventMode
}
```

The config looks like that now
```
      allowed_event_serialization:
          mode: INCLUDE_ALL
```

```
      allowed_event_serialization:
          mode: INCLUDE_WITH_VERBOSITY_LEVELS
          verbosity_levels: [INFO]
```

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
mgoworko: Unless I missed something obvious, we should be fine and it will remain compatible, as long as the `def onResponse(responseContext: AuditResponseContext): Option[JSONObject]` interface remains unchanged. 

I now further moved this trait to `private[ror] object AuditSerializationHelper`. If external developer uses something marked as package-private, IMO they do it on their own risk. 

We have to somehow encapsulate the logic of the serializer. Before this PR, it was done inside DefaultV1 serializer, and all other serializers extended it directly or indirectly. Now it is composition instead of inheritance (IMO it is more clear in this case), but I think it did not change compatibility considerations.

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllTypesOfEventsAuditLogSerializer.scala
mgoworko: 3. Done, `FullAuditLogSerializer` and `FullAuditLogWithQuerySerializer`

1&2. I'm not sure about it. It is a simple change, so I can do it quickly if we agree to do it after this comment, but:
- The name `SkipAllowedAuditLogSerializerVX` would not be fully precise. This serializer does serialize ALLOWED events, for rules with INFO level verbosity. So not all ALLOWED events are skipped, just those with ERROR level verbosity.

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV2.scala
mgoworko: answered in previous comment

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
mgoworko: Done, `private def resolver(eventData: EventData): AuditFieldValueDescriptor => Any`, EventData is a private final case class, only used internally in this file

audit/src/main/scala/tech/beshu/ror/audit/EnvironmentAwareAuditLogSerializer.scala
mgoworko: Yes, nothing has changed about this trait. It is supported. There is a test "custom environment-aware serializer is set and correctly serializes events" in `AuditSettingsTests`, that verifies it works fine.

This trait does not extend `AuditLogSerializer`. Instead, when its implementations are detected in config, they are adapted by `EnvironmentAwareAuditLogSerializerAdapter` to `AuditLogSerializer`. It was done in RORDEV-1414, so that developers can extend simple trait `EnvironmentAwareAuditLogSerializer` instead of creating class with parameter in constructor, like `class MyCustomSerializer(environmentContext: AuditEnvironmentContext) extends AuditLogSerializer`

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
```diff
@@ -17,82 +17,48 @@
...
-      Some(createEntry(matched = false, "INDEX NOT EXIST", "Requested index doesn't exist", responseContext.duration, requestContext, None))
-    case Errored(requestContext, cause) =>
-      Some(createEntry(matched = false, "ERRORED", "error", responseContext.duration, requestContext, Some(cause)))
-  }
+  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] =
+    AuditSerializationHelper.serialize(
+      responseContext = responseContext,
+      environmentContext = None,
+      fields = defaultV1AuditFields,
+      allowedEventMode = AllowedEventMode.Include(Set(Verbosity.Info)),
```
coutoPL: let's import `AllowedEventMode.*`.
I think: `allowedEventMode = Include(Set(Verbosity.Info))` is enought to understand what's this

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV2.scala
```diff
@@ -17,16 +17,49 @@
...
+    AuditFieldName("task_id") -> AuditFieldValueDescriptor.TaskId,
+    AuditFieldName("req_method") -> AuditFieldValueDescriptor.HttpMethod,
+    AuditFieldName("headers") -> AuditFieldValueDescriptor.HttpHeaderNames,
+    AuditFieldName("path") -> AuditFieldValueDescriptor.HttpPath,
+    AuditFieldName("user") -> AuditFieldValueDescriptor.User,
+    AuditFieldName("impersonated_by") -> AuditFieldValueDescriptor.ImpersonatedByUser,
+    AuditFieldName("action") -> AuditFieldValueDescriptor.Action,
+    AuditFieldName("indices") -> AuditFieldValueDescriptor.InvolvedIndices,
+    AuditFieldName("acl_history") -> AuditFieldValueDescriptor.AclHistory
+  ) ++ environmentRelatedAuditFields
```
coutoPL: WDYT about not using `++ environmentRelatedAuditFields` here, but just hardcode:
```
   AuditFieldName("acl_history") -> AuditFieldValueDescriptor.AclHistory,
   AuditFieldName("es_node_name") -> AuditFieldValueDescriptor.EsNodeName,
    AuditFieldName("es_cluster_name") -> AuditFieldValueDescriptor.EsClusterName
)
```
?

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.AllowedEventMode
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV2.defaultV2AuditFields
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, AuditSerializationHelper}
+
+class FullAuditLogSerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializer(environmentContext) {
+
+  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] =
```
coutoPL, line 60: It looks like we shouldn't provide a custom `onResponse` implementation for "an alias" class, don't we?

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.AllowedEventMode
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV2.defaultV2AuditFields
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, AuditSerializationHelper}
+
+class FullAuditLogSerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializer(environmentContext) {
```
coutoPL: We can say that `FullAuditLogSerializer` is "an alias" for `DefaultAuditLogSerializer`.

IMO, the "default sth" naming was not so good. We have to maintain it for the sake of backward compatibility, but IMO we should treat it as an alias for a better-named class. 

So, IMO the `FullAuditLogSerializer` is the better-named class and the `DefaultAuditLogSerializer` should be an alias for it.

WDYT?

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogWithQuerySerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.AllowedEventMode
+import tech.beshu.ror.audit.instances.QueryAuditLogSerializerV2.queryV2AuditFields
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, AuditSerializationHelper}
+
+class FullAuditLogWithQuerySerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializer(environmentContext) {
```
coutoPL: the same here

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
```diff
@@ -0,0 +1,218 @@
...
+ *
+ *    ReadonlyREST is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit
```
coutoPL: maybe we should move it to `tech.beshu.ror.audit.utils` package? To not litter the package with interfaces?

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
```diff
@@ -0,0 +1,218 @@
...
+        "@timestamp" -> timestampFormatter.format(eventData.requestContext.timestamp)
+      ) ++ fields.map {
+        case (fieldName, fieldValue) =>
+          fieldName.value -> resolver(eventData)(fieldValue)
+      }
+    }
+    resolvedFields
+      .foldLeft(new JSONObject()) { case (soFar, (key, value)) => soFar.put(key, value) }
+      .mergeWith(eventData.requestContext.generalAuditEvents)
+  }
```
coutoPL:
```suggestion
  private def createEntry(fields: Map[AuditFieldName, AuditFieldValueDescriptor],
                          eventData: EventData) = {
    val resolveAuditFieldValue =  resolver(eventData)                    
    val resolvedFields: Map[String, Any] = {
      Map(
        "@timestamp" -> timestampFormatter.format(eventData.requestContext.timestamp)
      ) ++ fields.map {
        case (name, valueDescriptor) =>
          name.value -> resolveAuditFieldValue(valueDescriptor)
      }
    }
    resolvedFields
      .foldLeft(new JSONObject()) { case (soFar, (key, value)) => soFar.put(key, value) }
      .mergeWith(eventData.requestContext.generalAuditEvents)
  }
```

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
coutoPL: Could you please manually verify that?

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllTypesOfEventsAuditLogSerializer.scala
coutoPL: You are right. Do you have an idea how to name it better? The "Default..." naming is not so good.
Let's take a perspective of the users. They see ACL with blocks and audit entries. 
Maybe GPT can help with this? ;)

audit/src/main/scala/tech/beshu/ror/audit/EnvironmentAwareAuditLogSerializer.scala
```diff
@@ -17,8 +17,16 @@
...
 import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.{AuditFieldName, AuditFieldValueDescriptor}
 
 trait EnvironmentAwareAuditLogSerializer {
   def onResponse(responseContext: AuditResponseContext,
                  environmentContext: AuditEnvironmentContext): Option[JSONObject]
 }
+
+object EnvironmentAwareAuditLogSerializer {
+  val environmentRelatedAuditFields: Map[AuditFieldName, AuditFieldValueDescriptor] = Map(
```
coutoPL: IMO we don't need it them here

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/ConfigurableAuditLogSerializer.scala
```diff
@@ -0,0 +1,31 @@
...
+package tech.beshu.ror.accesscontrol.audit.configurable
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditSerializationHelper.{AllowedEventMode, AuditFieldName, AuditFieldValueDescriptor}
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializer
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditResponseContext, AuditSerializationHelper}
+
+class ConfigurableAuditLogSerializer(val environmentContext: AuditEnvironmentContext,
+                                     val allowedEventMode: AllowedEventMode,
+                                     val fields: Map[AuditFieldName, AuditFieldValueDescriptor]) extends DefaultAuditLogSerializer(environmentContext) {
```
coutoPL: let's not extend "an alias" class.

coutoPL: BTW, shouldn't we provide here own implementation instead of basing on public implementations?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/AuditFieldValueDeserializer.scala
```diff
@@ -0,0 +1,92 @@
...
+ */
+package tech.beshu.ror.accesscontrol.audit.configurable
+
+import tech.beshu.ror.audit.AuditSerializationHelper.AuditFieldValueDescriptor
+
+object AuditFieldValueDescriptorDeserializer {
+
+  private val pattern = "\\{([^}]+)\\}".r
+
+  def deserialize(str: String): Either[String, AuditFieldValueDescriptor] = {
```
coutoPL: I wondered what it would look like if we used https://typelevel.org/cats-parse/.

GPT says sth like that:
```scala
import cats.data.NonEmptyList
import cats.parse.{Parser => P}
import java.util.Locale

val lbrace = P.char('{')
val rbrace = P.char('}')
val key    = P.charsWhile1(c => c != '{' && c != '}').map(_.toString)

val placeholder: P[Either[String, AuditFieldValueDescriptor]] =
  (lbrace *> key <* rbrace).map(k => lookup.get(k.trim.toUpperCase(Locale.ROOT)).toRight(k))

val text: P[AuditFieldValueDescriptor] =
  P.charsWhile1(_ != '{').map(AuditFieldValueDescriptor.StaticText.apply)

val segment: P[Either[String, AuditFieldValueDescriptor]] =
  placeholder.orElse(text.asRight[String])

val parser: P[List[Either[String, AuditFieldValueDescriptor]]] =
  segment.rep0 <* P.end

def deserialize(s: String) = parser.parseAll(s) match {
  case Left(err) => Left(err.toString)
  case Right(segments) =>
    val (missing, ok) = segments.partitionMap(identity)
    if (missing.nonEmpty)
      Left(s"There are invalid placeholder values: ${missing.distinct.mkString(", ")}")
    else ok match {
      case Nil => Right(AuditFieldValueDescriptor.StaticText(""))
      case single :: Nil => Right(single)
      case many => Right(AuditFieldValueDescriptor.Combined(many))
    }
}
```

I think it's easier to figure out what is parsed here.
WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/AuditFieldValueDeserializer.scala
```diff
@@ -0,0 +1,92 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.accesscontrol.audit.configurable
+
+import tech.beshu.ror.audit.AuditSerializationHelper.AuditFieldValueDescriptor
+
+object AuditFieldValueDescriptorDeserializer {
```
coutoPL: maybe `AuditFieldValueDescriptorParser`?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/AuditFieldValueDeserializer.scala
```diff
@@ -0,0 +1,92 @@
...
+        case Nil => Right(AuditFieldValueDescriptor.StaticText(""))
+        case singleElement :: Nil => Right(singleElement)
+        case multipleElements => Right(AuditFieldValueDescriptor.Combined(multipleElements))
+      }
+      case missingList => Left(s"There are invalid placeholder values: ${missingList.mkString(", ")}")
+    }
+  }
+
+  private def deserializerAuditFieldValueDescriptor(str: String): Option[AuditFieldValueDescriptor] = {
+    str match {
```
coutoPL: maybe `str.toUpper`?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,41 +209,111 @@ object AuditingSettingsDecoder extends Logging {
       }
       .decoder
 
+  given auditLogSerializerDecoder(using context: AuditEnvironmentContext): Decoder[Option[AuditLogSerializer]] = SyncDecoderCreator.from(
+      Decoder.instance[Option[AuditLogSerializer]] { c =>
+        for {
+          configurable <- c.downField("configurable").as[Option[Boolean]]
+          serializer <- if (configurable.contains(true)) {
+            for {
+              allowedEventModeStr <- c.downField("allowed_event_serialization").downField("mode").as[String]
```
coutoPL: maybe we could have `Decoder[AllowedEventMode]`?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,41 +209,111 @@ object AuditingSettingsDecoder extends Logging {
       }
       .decoder
 
+  given auditLogSerializerDecoder(using context: AuditEnvironmentContext): Decoder[Option[AuditLogSerializer]] = SyncDecoderCreator.from(
+      Decoder.instance[Option[AuditLogSerializer]] { c =>
+        for {
+          configurable <- c.downField("configurable").as[Option[Boolean]]
+          serializer <- if (configurable.contains(true)) {
```
coutoPL: the complidate IF's branches suggest that we could have two private methods for these two serializer decoders

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,41 +209,111 @@ object AuditingSettingsDecoder extends Logging {
...
+            for {
+              allowedEventModeStr <- c.downField("allowed_event_serialization").downField("mode").as[String]
+                .left.map(withAuditingSettingsCreationErrorMessage("Configurable serializer is used, but the allowed_event_serialization.mode setting is missing in configuration"))
```
coutoPL: maybe we could have some default here?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,41 +209,111 @@ object AuditingSettingsDecoder extends Logging {
...
+              allowedEventMode <- allowedEventModeStr match {
+                case "INCLUDE_ALL" => Right(AllowedEventMode.IncludeAll)
+                case "INCLUDE_WITH_VERBOSITY_LEVELS" => c.downField("allowed_event_serialization").downField("verbosity_levels").as[Set[Verbosity]].map(AllowedEventMode.Include.apply)
+                case other => Left(DecodingFailure(s"Invalid allowed_event_serialization.mode $other. Allowed values: [INCLUDE_ALL, INCLUDE_WITH_VERBOSITY_LEVELS]", Nil))
+              }
+              fields <- c.downField("fields").as[Map[AuditFieldName, AuditFieldValueDescriptor]]
+                .left.map(withAuditingSettingsCreationErrorMessage("Configurable serializer is used, but the fields setting is missing in configuration"))
```
coutoPL: the 'fields' setting

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
```diff
@@ -373,6 +376,49 @@ class AuditSettingsTests extends AnyWordSpec with Inside {
...
           }
+          "configurable serializer is set" in {
+            val config = rorConfigFromUnsafe(
+              """
+                |readonlyrest:
+                |  audit:
+                |    enabled: true
+                |    outputs:
+                |    - type: log
+                |      configurable: true
```
coutoPL: I'm not sure about this syntax.

Let's take a look at our doc: https://docs.readonlyrest.com/elasticsearch/audit
Let's start from an audit config like this:
```yaml
readonlyrest:
  audit:
    enabled: true
    outputs: 
    - type: index
    - type: log
      enabled: false # by default is true
```

We define outputs, their types, and properties. One of the properties is the serializer.
When we don't specify it we use the default serializer. But we can use different one (or define the default one explicitly):

```yaml
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: index
      serializer: tech.beshu.ror.requestcontext.QueryAuditLogSerializer
      # in case when the `log` type is used
    - type: log
      serializer: tech.beshu.ror.requestcontext.QueryAuditLogSerializer
  ...
```

In this PR we provide a configurable audit log serializer. The idea is as follows: users don't have to implement their own serializer when they want to slightly modify the default behaviour.

So, we are in the context of serializer definition.
It means, IMO, it could be done as follows:

```yaml
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: index
      serializer:
         type: configurable # the other type could be "internal", "static", "defined", or sth like that
        fields: ...
        xxx: ...
    - type: log
      serializer:
         type: static
         class_name: tech.beshu.ror.requestcontext.QueryAuditLogSerializer
```

WDYT?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
```diff
@@ -1242,6 +1288,54 @@ class AuditSettingsTests extends AnyWordSpec with Inside {
               expectedErrorMessage = "The audit 'outputs' array cannot be empty"
             )
           }
+          "configurable serializer is set, but without serialize_all_allowed_events setting" in {
```
coutoPL: hmm, is this `serialize_all_allowed_events` used?

Image

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
```diff
@@ -1242,6 +1288,54 @@ class AuditSettingsTests extends AnyWordSpec with Inside {
...
+          "configurable serializer is set, but without fields setting" in {
+            val config = rorConfigFromUnsafe(
+              """
+                |readonlyrest:
+                |  audit:
+                |    enabled: true
+                |    outputs:
+                |    - type: log
+                |      configurable: true
+                |      allowed_event_serialization:
```
coutoPL: WDYT about sth like this:

```yaml
verbosity_level_serialization_mode: [INFO, ERROR] # default: all
```

So, we don't have to recreate the internal class hierarchy here in the configuraiton

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
mgoworko: It was my mistake, it should not extend Default serializer and it is not an alias. It should just extend base trait `AuditLogSerializer`, nothing more.

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
mgoworko: As I wrote in the discussion above, it was my mistake that `FullAuditLogSerializer` extended `DefaultAuditLogSerializer`. So it was a mistake in code, that I wrote `extends DefaultAuditLogSerializer(environmentContext)`. It should be `extends AuditLogSerializer`. It did not change anything in behavior (because `FullAuditLogSerializer` defines its own `onResponse` impl).


So `FullAuditLogSerializer` is not an alias for Default serializer. It has different behavior - it serializes all ALLOWED events, while `Default` serializer - only those with INFO verbosity level.

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
mgoworko: done

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogWithQuerySerializer.scala
mgoworko: replied in the discussion above

audit/src/main/scala/tech/beshu/ror/audit/EnvironmentAwareAuditLogSerializer.scala
mgoworko: removed after changes from previous comments

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/ConfigurableAuditLogSerializer.scala
mgoworko: It now extends `extends AuditLogSerializer`. It was copy-paste issue. It uses its own implementation - I just replaced incorrect `extends DefaultAuditLogSerializer(environmentContext)` with `extends AuditLogSerializer`

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/AuditFieldValueDeserializer.scala
mgoworko: Ok, done using cats-parse. It is cleaner. It required a bit of refactoring and the syntax from ChatGPT was a bit off, but it worked. Added additional test to verify the error message for invalid placeholder.

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
mgoworko: That is why pipeline failed after recent changes, fixed.

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
mgoworko: done

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: After the change from another comment, the default is now `AllowedEventMode.IncludeAll`

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: Done, extracted 2 private methods

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: done

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
mgoworko: Good idea, done, with adjusted tests.

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
mgoworko: I verified it - I built the serializer from the example (using Scala 2.13):
- this PR does not cause compatibility problems, works just fine 
- but the RORDEV-1414 from June seems to has broken compatibility (the DefaultAuditLogSerializer), perhaps that is what you noticed

What I did:
- the source of all related issues is the handling of `AuditEnvironmentContext`. In 1414 it was done outside of the `AuditResponseContext`. It was my mistake then, I didn't notice that it can be included in `AuditResponseContext`.
- So now I refactored it. The `AuditEnvironmentContext` is now included in the `AuditEnvironmentContext`. That way we do not need serializer constructors with parameters and all compatiblity issues go away.
- As a result, the `EnvironmentAwareAuditLogSerializer` is no longer needed (the neceessary information is already in the `AuditResponseContext`), but is still supported in case anyone extended it.
- The incompatiblity bug means, that if someone built his own serializer, that extends DefaultAuditLogSerializer or QueryAuditLogSerializer againts ROR version from July and August, it may need to be rebuilt.

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllTypesOfEventsAuditLogSerializer.scala
mgoworko: Maybe 'BlockVerbosityAwareAuditLogSerializer` could be an alias for the `Default...`

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
mgoworko: The changes are in a single commit - https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1140/commits/2bb638afa267d936bff319f3910947c47f48ed9d

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+package tech.beshu.ror.audit.instances
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.utils.AuditSerializationHelper.AllowedEventMode.IncludeAll
+import tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV2.defaultV2AuditFields
+import tech.beshu.ror.audit.utils.AuditSerializationHelper
+import tech.beshu.ror.audit.{AuditLogSerializer, AuditResponseContext}
+
+class FullAuditLogSerializer extends AuditLogSerializer {
```
coutoPL, line 58: I've just realised that we could generate some comments that describe the logic of the deserializer in human words. We could obviously reuse these texts in docs. Let's utilize LLM to provide some neat comments.

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] =
+    AuditSerializationHelper.serialize(
+      responseContext = responseContext,
+      fields = defaultV2AuditFields,
```
coutoPL: If it's not an alias, shouldn't we redefine all the fields in FullAuditLogSerializer's companion object?

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogWithQuerySerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+import tech.beshu.ror.audit.instances.QueryAuditLogSerializerV2.queryV2AuditFields
+import tech.beshu.ror.audit.utils.AuditSerializationHelper
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditLogSerializer, AuditResponseContext}
+
+class FullAuditLogWithQuerySerializer extends AuditLogSerializer {
+
+  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] =
+    AuditSerializationHelper.serialize(
+      responseContext = responseContext,
+      fields = queryV2AuditFields,
```
coutoPL: Same here. Maybe it's worth redefining the fields. 

I wonder if we shouldn't do it like this:
1. if this is a separate/new/different serializer, let's define all the fields in a companion object
2. if this is an alias, it should extend some serializer and won't redefine fields (it should not even override onResponse)
3. if this is a serializer that extends another serializer, let's use the fields from the extended serializer and add new ones (but maybe it means that it'd be better to add these fields at the class definition's level, not the companion object's level)

WDYT?
If it makes sense, let's refactor all the serializers to follow these rules. 
Maybe we should also add them to our internal repo (or in the form of an ADR).

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
```diff
@@ -0,0 +1,214 @@
...
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.audit.utils
+
+import org.json.JSONObject
+import tech.beshu.ror.audit.AuditResponseContext._
+import tech.beshu.ror.audit.instances.SerializeUser
+import tech.beshu.ror.audit.{AuditEnvironmentContext, AuditRequestContext, AuditResponseContext}
```
coutoPL: unused import

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
coutoPL: ok, let's not forget to add a changlog entry to the PR's description that we fixed comaptibility of the audit module broken in ROR x.y.z

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
coutoPL: let's move the `SerializeUser` object too (and let's make it `private[ror]`)

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -120,21 +122,21 @@ object AuditingSettingsDecoder extends Logging {
 
     given Decoder[LogBasedSink] = Decoder.instance { c =>
```
coutoPL, line 123: private? here and below

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -205,41 +207,121 @@ object AuditingSettingsDecoder extends Logging {
...
+    for {
+      allowedEventMode <- c.downField("verbosity_level_serialization_mode").as[AllowedEventMode]
+        .left.map(withAuditingSettingsCreationErrorMessage(msg => s"Configurable serializer is used, but the 'verbosity_level_serialization_mode' setting is invalid: $msg"))
+      fields <- c.downField("fields").as[Map[AuditFieldName, AuditFieldValueDescriptor]]
+        .left.map(withAuditingSettingsCreationErrorMessage(msg => s"Configurable serializer is used, but the 'fields' setting is missing or invalid: $msg"))
+      serializer = new ConfigurableAuditLogSerializer(allowedEventMode, fields)
+    } yield Some(serializer)
+  }
+
+  private def classNameBasedSerializerDecoder: Decoder[Option[AuditLogSerializer]] = Decoder.instance { c =>
```
coutoPL: sth is wrong with this decoder...

I think about the serializer decoding like that:

1. it could be a simple definition of an audit module serializer implementation decoder:
```yaml
serializer: tech.beshu.ror.requestcontext.QueryAuditLogSerializer
```

2. it has the extended version too (we used to name it simple-extended in the whole codebase, AFAIR):

```yaml
serializer:
  type: static
  class_name: tech.beshu.ror.requestcontext.QueryAuditLogSerializer
``` 

3. but there is also a decoder of a customizable serializer:

```yaml
serializer:
  type: configurable # the other type could be "internal", "static", "defined", or sth like that
  fields: ...
```

4. and obviously the backward compatible yaml syntax serializer as an alternative for the 1st point.

It would be nice if we could have this split into methods. I imagine that at some point, the serializer decoder should look like this:

```scala
simpleSyntaxSerializerDecoder
  .orElse(extendedSyntaxSerializerDecoder)
  .orElse(customizableSerializerDecoder)
```

or like this:
```
simpleSyntaxSerializerDecoder.orElse(extendedSyntaxSerializerDecoder)
...
def extendedSyntaxSerializerDecoder(...) = {
   staticSerializerDecoder.orElse(customizableSerializerDecoder)
}
```

WDYT?

audit/src/main/scala/tech/beshu/ror/audit/instances/ReportingAllTypesOfEventsAuditLogSerializer.scala
coutoPL: yeah, it's much better. Let's use it

mgoworko: Added aliases for `Default..`, `Default...V1` and `Default...V2`

audit/src/main/scala/tech/beshu/ror/audit/AuditFieldValue.scala
mgoworko: done, added "fix" entry

audit/src/main/scala/tech/beshu/ror/audit/AuditSerializationHelper.scala
mgoworko: done

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
mgoworko: done

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogWithQuerySerializer.scala
mgoworko: I propose (and implemented) something a bit different, IMO more readable and with less duplicates.
- the fields are mostly the same between serializers, and I don't like the idea of duplicating the lists - it creates space for bugs and ommisions
- there are only 3 sets of fields: common fields (most fields, present in all serializers that we provide), ES environment related fields (added in 1414 and present in new, V2 and environment-aware serializers), and full content field
- the serializers do not extend each other now after this PR (there are only name aliases, but not reusing and modifying other class)
- each serializer (provided by us in our codebase) has implementation like shown below
- so it defines which field groups are used, and which allowed events are serialized
- serializers do not extend each other, but instead each uses `AuditSerializationHelper.serialize` with appropriate field groups and allowed event mode

```
  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] =
    AuditSerializationHelper.serialize(
      responseContext = responseContext,
      fieldGroups = Set(CommonFields, EsEnvironmentFields, FullRequestContentFields),
      allowedEventMode = IncludeAll
    )
```

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: it is inside private method

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
mgoworko: There is a problem with the suggested approach - we would loose dedicated decoding errors (in case of failure, we would go to the next decoder instead of throwing the error based on some missing field in first decoder)

Overall, there was already something like you suggested. Perhaps too complicated, so I refactored it. The thing that you probably noticed to be wrong in this decoder is the fact, that is was universal and used for both the serializer class name in simple and extended syntax.

So after refactor I creeated sth like that:
```
    for {
      serializerTypeStr <- c.downField("serializer").downField("type").as[SerializerType]
      result <- serializerTypeStr match {
        case SerializerType.SimpleSyntaxStaticSerializer =>
          c.as[Option[AuditLogSerializer]](simpleSyntaxSerializerDecoder)
        case SerializerType.ExtendedSyntaxStaticSerializer =>
          c.downField("serializer").as[Option[AuditLogSerializer]](extendedSyntaxStaticSerializerDecoder)
        case SerializerType.ExtendedSyntaxConfigurableSerializer =>
          c.downField("serializer").as[Option[AuditLogSerializer]](extendedSyntaxConfigurableSerializerDecoder)
      }
    } yield result
```

The SerializerType decoder is separate and is quite complicated, because of the fact, that "serializer" field may be either String or config section - and that situation has to be handled manually in circe.

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogSerializer.scala
mgoworko: Added LLM-generated comments

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializer.scala
```diff
@@ -16,6 +16,22 @@
...
+ *
+ * This is the recommended serializer for standard audit logging.
+ */
+class BlockVerbosityAwareAuditLogSerializer extends DefaultAuditLogSerializer
 
-class DefaultAuditLogSerializer(environmentContext: AuditEnvironmentContext) extends DefaultAuditLogSerializerV2(environmentContext)
+/**
+ * Base implementation delegating to [[DefaultAuditLogSerializerV2]].
+ *
+ * - Not intended for direct external use.
```
coutoPL: maybe we should annotate it as deprecated instead?

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializer.scala
coutoPL, line 1: should we rename the class name?

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
```diff
@@ -17,82 +17,35 @@
 package tech.beshu.ror.audit.instances
```
coutoPL: the same comments in this file

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
```diff
@@ -17,82 +17,35 @@
...
+ * Serializer for audit events (V1) that is aware of **rule-defined verbosity**.
+ *
+ * - Includes only `CommonFields` (request metadata, user, action, etc.).
+ * - Serializes all non-Allowed events.
+ * - Serializes `Allowed` events only if the corresponding rule
+ * specifies that they should be logged at `Verbosity.Info`.
+ *
+ * Recommended when a minimal, compact audit log is sufficient.
+ */
+class BlockVerbosityAwareAuditLogSerializerV1 extends DefaultAuditLogSerializerV1
```
coutoPL: I don't get why we introduce V1 and V2 for the "new" serializer (to be strict - renamed). 

I took into consideration two things:
1. backward compatibility - we don't have to do that
2. documentation - how are we going to explain to our users (new users, and users who have ROR already installed) in simple words the V1 and V2? - so it may be misleading for them. Do they need it? I suppose - no.

audit/src/main/scala/tech/beshu/ror/audit/instances/BaseAuditLogSerializer.scala
coutoPL: I've added this comment to RORDEV-1574.
let's leave it for now

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializer.scala
```diff
@@ -16,6 +16,22 @@
  */
 package tech.beshu.ror.audit.instances
 
-import tech.beshu.ror.audit.AuditEnvironmentContext
+/**
+ * Serializer for audit events that is aware of **rule-defined verbosity**.
+ *
+ * - Includes `CommonFields` and `EsEnvironmentFields`.
```
coutoPL: CommonFields, EsEnvironmentFields, ... are implementation details. Let's list all the fields in the description (with one line explanation if possible)

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogWithQuerySerializer.scala
coutoPL: > it creates space for bugs and ommisions

Yes, but it also creates a possibility of adding a new field to some serializer unintentionally. 

So, both approaches are error-prone.
But it's ok when we add proper unit tests ;) So, let's add a unit test for each serializer to check what fields it produces. To work as expected (to eliminate or at least lower the possibility of the bug I mentioned above), we should hardcode the list in the test (but hardcoding in tests is not a problem for us, right ;)). 

It may be tedious, so let's try to use LLM to help us with this boring task ;)

coutoPL (changes requested):

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializer.scala
mgoworko: I changed the file name to `BlockVerbosityAwareAuditLogSerializer` - did you mean that? I also moved all deprecated `Default...` serializers to that single file.

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializer.scala
mgoworko: Added `deprecated` annotation

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
mgoworko: So after those 3 comments there is now a single file `BlockVerbosityAwareAuditLogSerializer` file, which contains
```
class BlockVerbosityAwareAuditLogSerializer extends DefaultAuditLogSerializer

@deprecated("Use tech.beshu.ror.audit.instances.BlockVerbosityAwareAuditLogSerializer instead", "1.67.0")
class DefaultAuditLogSerializer extends DefaultAuditLogSerializerV2

@deprecated("Use tech.beshu.ror.audit.instances.BlockVerbosityAwareAuditLogSerializer instead", "1.67.0")
class DefaultAuditLogSerializerV2 extends AuditLogSerializer {...}

@deprecated("Use tech.beshu.ror.audit.instances.BlockVerbosityAwareAuditLogSerializer instead", "1.67.0")
class DefaultAuditLogSerializerV1 extends AuditLogSerializer {...}
```

did you mean something like that?

About V1 and V2 in general (not in this particular case for `BlockVerbosityAwareAuditLogSerializer), but why it was introduced at all):
- we introduced it in RORDEV-1414
- so that user can use the public alias without version suffix - the newest, current version of serializer
- but can also restore previous behavior by using version-numbered serialized
- (by default new environment-related fields are included, but user can restore previous behavior by using version-numbered serializer - that was the idea we discussed then)

audit/src/main/scala/tech/beshu/ror/audit/instances/DefaultAuditLogSerializerV1.scala
mgoworko: please see answer to the comment below

audit/src/main/scala/tech/beshu/ror/audit/instances/FullAuditLogWithQuerySerializer.scala
mgoworko: Added additional tests - one for each serializer - that assert list of fields (and only list of fields, other things, like verbosity level, were already asserted in preexisting tests)

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/LocalClusterAuditingToolsSuite.scala
```diff
@@ -91,20 +89,253 @@ class LocalClusterAuditingToolsSuite
...
+                entry("tid").numOpt.isDefined &&
+                entry("bytes").num == 0
+            ) shouldBe true
+          }
+        }
+        updateRorConfigToUseSerializer("tech.beshu.ror.audit.instances.DefaultAuditLogSerializerV1")
+      }
+      "serialized event contains only expected fields" should {
+        "QueryAuditLogSerializer" in {
+          testSerializerFieldsWithTypes(
```
coutoPL: IMO, we should test these serializers using unit tests. 
Reloading ROR settings for each test is not the way to go in this use case (our integration tests are too long to waste time on such tests.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/LocalClusterAuditingToolsSuite.scala
```diff
@@ -91,20 +89,253 @@ class LocalClusterAuditingToolsSuite
...
+            serializerClassName = "tech.beshu.ror.audit.instances.QueryAuditLogSerializer",
+            expectedFieldsWithTypes = commonFields ++ Map(
```
coutoPL: IMO, we should not use sth like commonField in tests. IMO, we should hardcode all the fields here in this test. We want to prevent unintended extending of the commons fields list in the implementation. Current implementation of these tests won't protect us against this mistake. When the fields are hardcoded, someone should easier notice that the new fields their added, results in the new field in each serializer.

mgoworko: done, hardcoded in each test

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/LocalClusterAuditingToolsSuite.scala
mgoworko: Done, moved (with adjustments) to `AuditSettingsTests`

coutoPL: Let's move this test to the audit module as a separate test suite (because TBH, it's not an audit settings test)

coutoPL: please don't miss this comment too

mgoworko: Done, tests moved to audit module and adjusted, to just create the serializer instance (instead of loading from config). There were no `audit` tests before, so I added scalatest dependency, Gradle plugin and added those tests to "core" tests job.

PR #1141, opened 2025-07-27
[RORDEV-1543] Improved selection logic for user metadata to prioritize matches with a defined Kibana index

coutoPL:
🧐Enhancement (ES) Improved the selection logic for user metadata to prioritize matched blocks with a defined Kibana index when handling login requests.


mateuszkp96 (approved):
LGTM

PR #1144, opened 2025-07-31
[RORDEV-1533] Better validation of ROR patcher arguments

mgoworko:
🧐Enhancement (ES) Patcher - improved handling of patching consent flag provided as environment variable

## Summary by CodeRabbit

* **Refactor**
  * Improved handling of user consent for Elasticsearch patching, allowing inputs from multiple sources and detecting conflicts.
  * Enhanced validation and error reporting for conflicting consent inputs.

* **Tests**
  * Added tests to verify correct handling of multiple and conflicting consent inputs from command-line arguments and environment variables.


ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -53,7 +53,7 @@ trait RorTools {
     OParser.runParser(
       parser,
       allArgs.map(arg => if (arg.startsWith("--")) arg.toLowerCase else arg),
-      Arguments(Command.Verify(None), PatchingConsent.AnswerNotGiven),
+      Arguments(Command.Verify(None), List.empty),
```
coutoPL: Ok, I think it's a little bit complicated here.

This parser should parse the terminal command only. Now, we also include other, non-explicitly-entered values in the arguments list. Now I see, it's misleasing.

My objections:
1. It would be better if we define Arguments as:
```scala
final case class ScriptArguments(command: Option[Command]) {
```
Because the consent makes sense only in the case of the `Patch` command, it should be parsed only when we have to do with it. So:

```scala
final case class Patch(consent: PatchingConsent, customEsPath: Option[os.Path])
object Patch {
  private sealed trait PatchingConsent
  private object PatchingConsent {
    case object Accepted extends PatchingConsent
    case object Rejected extends PatchingConsent
    case object AnswerNotGiven extends PatchingConsent
  }  
}
```

2. When we have the command parsed properly, we can parse the ENV-based arguments (aka `EnvBasedArguments` or sth like that). 
Then (probably) in the `handleParsedArguments` method, we can validate if both arguments make sense from our patcher business logic. 

IMO, thanks to these changes, we will express our business logic clearly in types, and validations will be clearly visible. 

WDYT?

coutoPL (changes requested):

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
mgoworko: I applied most (but not all, will explain) of the suggested changes:
1. Arguments defined as suggested, with  `final case class ScriptArguments(command: Option[Command])`
- thanks to that change it became obvious, that the case when user provides no arguments was not handled properly; I configured it to display usage information and added integration test
2. The `PatchingConsent` is moved to inside the `Command.Patch`. And there is more consistent validation code. But it is not possible to do exactly as suggested. We need to still have list of values, instead of a single value, because:
- The OParser is not very good at applying checks and validations. It allows to validate and check values, but not to modify the result based on them. So when we have multiple values - we have a list of them. We can validate that they are correct, but we cannot transform them into a single value in Oparser result based on validation.
- I understand that you suggested to move the ENV-defined args handling outside of the OParser logic, but that is quite problematic IMO. The values from args and form env would be handled differently. The Oparser-handled  arg would have to be optional (because only env may be defined), we would lose the error messages, displaying of app usage etc.

PR #1151, opened 2025-08-22
POC

mgoworko:


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/implementations/UnboundidLdapUsersService.scala
```diff
@@ -52,7 +52,7 @@ class UnboundidLdapUsersService private(override val id: LdapService#Id,
         LdapUser(
           id = userId,
           dn = Dn(NonEmptyString.unsafeFrom(
-            s"cn=${Filter.encodeValue(userId.value.value)},${userSearchFiler.searchUserBaseDN.value.value}"
+            s"cn=${userId.value.value},${userSearchFiler.searchUserBaseDN.value.value}"
```
mgoworko, line 55: It is a change in production, non-test code. According to LLM, this encoding was not needed, because the entire text is later encoded. It caused encoding issues on Windows - double-encoded value.

core/src/test/resources/current_user_metadata_access_control_tests/ldap_ldap1_user5.ldif
```diff
@@ -13,7 +18,7 @@ objectClass: inetOrgPerson
 cn: user5
 sn: user5
 uid: user5
-userPassword:: e1NNRDV9czdnM0NVekVCMGQxMm5CM0N3VGFrQmp3K0VGMTE3cFg=
+userPassword: user2
```
mgoworko, line 21: The JVM in-memory LDAP used in tests does not understand encoded passwords.

core/src/test/resources/test_example.ldif
```diff
@@ -1,3 +1,14 @@
+dn: dc=example,dc=com
```
mgoworko, line 1: Then in-memory LDAP used in tests on WIndows requires this header - basic config - at the beginning of ldif file

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/LdapSingleContainer.scala
```diff
@@ -0,0 +1,101 @@
...
+import org.testcontainers.containers.{GenericContainer as JavaGenericContainer}
+import tech.beshu.ror.utils.containers.LdapSingleContainer.InitScriptSource
+import tech.beshu.ror.utils.containers.windows.{NonStoppableInMemoryLdapService, WindowsPseudoSingleContainerLdap}
+import tech.beshu.ror.utils.misc.OsUtils
+import tech.beshu.ror.utils.misc.ScalaUtils.*
+
+import scala.concurrent.duration.*
+import scala.language.{implicitConversions, postfixOps}
+
+trait LdapSingleContainer extends SingleContainer[JavaGenericContainer[_]] {
```
mgoworko, line 31: `LdapSingleContainer` is a common interface for both preexisting LdapContainer - testcontainer, and the Windows implementation, that runs in memory.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -35,28 +38,92 @@ object Elasticsearch {
...
-    def updateEsImage(image: DockerImageDescription, config: Config): DockerImageDescription
+    def esUpdateSteps(config: Config): EsUpdateSteps
 
     def updateEsConfigBuilder(builder: EsConfigBuilder): EsConfigBuilder
 
     def updateEsJavaOptsBuilder(builder: EsJavaOptsBuilder): EsJavaOptsBuilder
   }
 
+  object Plugin {
+    final case class EsUpdateSteps(steps: List[EsUpdateStep]) {
```
mgoworko, line 86: In order to execute the preexisting testcontainer update steps on WIndows native ES, I had to separate the definition of those update steps from their execution. `EsUpdateSteps` and related classes are definitions of update steps, that are then executed either on testcontainer image, or Windows ES directory

core/src/test/scala/tech/beshu/ror/utils/TestsUtils.scala
```diff
@@ -397,7 +397,7 @@ object TestsUtils {
   }
 
   def getResourcePath(resource: String): Path = {
-    File(getClass.getResource(resource).getPath).path
+    File(getClass.getResource(resource).toURI).path
```
mgoworko, line 400: On Linux it made no difference if we used getPath or toURI. On Windows we need URI - getPath is Unix-style path.

integration-tests/src/test/resources/current_user_metadata/readonlyrest.yml
```diff
@@ -24,7 +24,7 @@ readonlyrest:
       kibana:
         access: ro
         index: "user2_kibana_index"
-        hide_apps: ["user2_app1", "user2_app2", "/^Analytics\\|(?!(Maps)$).*$/"]
+        hide_apps: ["user2_app1", "user2_app2", "^Analytics\\|(?!Maps$).*"]
```
mgoworko, line 27: I am unsure why it worked fine on Linux, but according to LLM:
```
In ES 8.x on Windows, the bundled GraalJS engine is stricter: it does not like regex written as /…/ strings inside YAML unless they’re strictly valid JS regex literals.
```
It is like that in practice - this regex does not compile since ES 8.x on Windows.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FipsSslSuite.scala
```diff
@@ -57,12 +57,20 @@ class FipsSslSuite
 
   private lazy val rorClusterAdminStateManager = new CatManager(clients.last.adminClient, esVersion = esVersionUsed)
 
-  "Health check" should {
-    "be successful" when {
-      "internode ssl is enabled" in {
-        val response = rorClusterAdminStateManager.healthCheck()
+  // todo: The ES with FIPS does not start correctly when running tests on Windows with ES version lower than 8.18.
```
mgoworko, line 60: It needs further investigation. My suspicion is that it is not an issue with tests, but with ROR itself. So I suspect Fips may not work on Windows for ROR with ES<=8.18. That is why I think fixing it is outside of scope of this task - it may not be test issue, but production code issue, that just never surfaced.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
```diff
@@ -184,7 +185,7 @@ trait BaseAdminApiSuite
           forceReload("/admin_api/readonlyrest_first_update.yml")
 
           // after first reload only dev1 can access indices
-          Thread.sleep(settingsReloadInterval.plus(5 second).toMillis) // have to wait for ROR1_2 instance config reload
+          Thread.sleep(settingsReloadInterval.plus(10 second).toMillis) // have to wait for ROR1_2 instance config reload
```
mgoworko, line 188: On Windows we sometimes need longer intervals in patience config than using Linux testcontainers.

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -246,7 +246,7 @@ trait RorTools {
 
   private lazy val esPathOption =
     opt[String]("es-path")
-      .text(s"Path to elasticsearch directory; default=${EsDirectory.defaultPath}")
+      .text(s"Path to elasticsearch directory; default=/usr/share/elasticsearch")
```
mgoworko, line 249: The path is not correctly serialized on Windows. Even after fixing the format, the default Linux path does not make sense on Windows. So I hardcoded the Linux path in the message.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -496,12 +440,14 @@ class RorToolsAppSuite
            | - file x-pack-core-$esVersionUsed.jar was patched by ROR ${metadata.rorVersion}
            | - file x-pack-ilm-$esVersionUsed.jar was patched by ROR ${metadata.rorVersion}
            | - file x-pack-security-$esVersionUsed.jar was patched by ROR ${metadata.rorVersion}
-           |""".stripMargin
+           |""".stripMargin.replace("\r\n", "\n")
       )
     }
     "Successfully patch, verify and unpatch" in {
+      // This file is created after patching on Windows for ES 8.x
+      val filesExcludedFromHashCalculation = List("plugin-security.policy.tmp")
```
mgoworko, line 448: Seems to be some ES 8.x issue specific to Windows. It does not happen on other ES versions and on Linux.

azure-pipelines.yml
```diff
@@ -190,6 +190,8 @@ stages:
           matrix:
             IT_es91x:
               ROR_TASK: integration_es91x
+            IT_es90x:
```
mgoworko, line 193: At the moment, for test purposes, I configured it to run Windows tests for a few representative modules, and only newest one on Linux.

I propose to always run Windows tests on the newest ES version - even in PR's. It is only one job more, and will allow to catch errors and incompatibilities before merge with develop, WDYT about that?. Which versions should be tested on develop and master pipelines?

PR #1152, opened 2025-08-23
[RORDEV-1564] properly handle `POST /_security/api_key/grant` request

coutoPL:
When ROR doesn't block the request, we xpack security should not block it.

## Summary by CodeRabbit

- New Features
  - Added support for granting API keys via the Security API.
  - Improved compatibility and stability across multiple Elasticsearch versions (7.16+, 8.x, 9.x).

- Tests
  - Added test coverage for API key grant requests.
  - Simplified test DSL and reduced explicit verbosity in test configs.

- Chores
  - Bumped plugin version to 1.66.0-pre3.
  - Updated published metadata URL.


ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/modifiers/bytecodeJars/DummyAuthenticationInAuthenticationChain.scala
```diff
@@ -0,0 +1,297 @@
...
+            case v if v >= es890 =>
+              new SetXpackUserAsAuthenticatedUserForEsGreaterOrEqual890x(super.visitMethod(access, name, descriptor, signature, exceptions))
+            case v if v >= es870 =>
+              new SetXpackUserAsAuthenticatedUserForEsGreaterOrEqual870(super.visitMethod(access, name, descriptor, signature, exceptions))
+            case v if v >= es820 =>
+              new SetXpackUserAsAuthenticatedUserForEsGreaterOrEqual820(super.visitMethod(access, name, descriptor, signature, exceptions))
+            case v if v >= es71713 =>
+              new SetXpackUserAsAuthenticatedUserForEsGreaterOrEqual71713(super.visitMethod(access, name, descriptor, signature, exceptions))
+            case v if v >= es7160 =>
+              new SetXpackUserAsAuthenticatedUserForEsGreaterOrEqual71713(super.visitMethod(access, name, descriptor, signature, exceptions))
```
mateuszkp96: Is the first case needed? Or do the es7160 versions need a dedicated method visitor?

mateuszkp96 (changes requested):

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches/internal/modifiers/bytecodeJars/DummyAuthenticationInAuthenticationChain.scala
coutoPL: Good catch. It figured out that the same patcher can be used for these two modules. I forgot to rename it.

mateuszkp96 (approved):
LGTM

PR #1154, opened 2025-08-31
[RORDEV-1472] Run integration tests on Windows with native Windows version of ES

mgoworko:
🐞Fix (ES) Fixed LDAP Common Name (CN) double encoding issue when ES is installed on Windows

## Summary by CodeRabbit

* **New Features**
  * Expanded Windows support across CI and tests: Windows-specific test runs, pseudo-containers, ES-on-Windows helpers and predefined Windows ports.

* **Bug Fixes**
  * Updated LDAP test data and sample directories; normalized line endings and regexes for cross-platform consistency.

* **Chores**
  * CI gating refined with separate Linux/Windows test actions and updated pipeline parameters.

* **Documentation**
  * CLI help updated to show default ES path /usr/share/elasticsearch.

* **Tests**
  * Added test JSON wrapper, many test imports, OS-aware guards, and stability/timeouts improvements.


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/definitions/ldap/implementations/UnboundidLdapUsersService.scala
```diff
@@ -52,7 +52,7 @@ class UnboundidLdapUsersService private(override val id: LdapService#Id,
         LdapUser(
           id = userId,
           dn = Dn(NonEmptyString.unsafeFrom(
-            s"cn=${Filter.encodeValue(userId.value.value)},${userSearchFiler.searchUserBaseDN.value.value}"
+            s"cn=${userId.value.value},${userSearchFiler.searchUserBaseDN.value.value}"
```
mgoworko, line 55: It is a change in production, non-test code. According to LLM, this encoding was not needed, because the entire text is later encoded. It caused encoding issues on Windows - double-encoded value.

core/src/test/resources/current_user_metadata_access_control_tests/ldap_ldap1_user5.ldif
```diff
@@ -13,7 +18,7 @@ objectClass: inetOrgPerson
 cn: user5
 sn: user5
 uid: user5
-userPassword:: e1NNRDV9czdnM0NVekVCMGQxMm5CM0N3VGFrQmp3K0VGMTE3cFg=
+userPassword: user2
```
mgoworko, line 21: The JVM in-memory LDAP used in tests does not understand encoded passwords.

core/src/test/resources/test_example.ldif
```diff
@@ -1,3 +1,14 @@
+dn: dc=example,dc=com
```
mgoworko, line 1: Then in-memory LDAP used in tests on WIndows requires this header - basic config - at the beginning of ldif file

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/LdapSingleContainer.scala
```diff
@@ -0,0 +1,101 @@
...
+import org.testcontainers.containers.{GenericContainer as JavaGenericContainer}
+import tech.beshu.ror.utils.containers.LdapSingleContainer.InitScriptSource
+import tech.beshu.ror.utils.containers.windows.{NonStoppableInMemoryLdapService, WindowsPseudoSingleContainerLdap}
+import tech.beshu.ror.utils.misc.OsUtils
+import tech.beshu.ror.utils.misc.ScalaUtils.*
+
+import scala.concurrent.duration.*
+import scala.language.{implicitConversions, postfixOps}
+
+trait LdapSingleContainer extends SingleContainer[JavaGenericContainer[_]] {
```
mgoworko: `LdapSingleContainer` is a common interface for both preexisting LdapContainer - testcontainer, and the Windows implementation, that runs in memory.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -35,28 +38,92 @@ object Elasticsearch {
...
-    def updateEsImage(image: DockerImageDescription, config: Config): DockerImageDescription
+    def esUpdateSteps(config: Config): EsUpdateSteps
 
     def updateEsConfigBuilder(builder: EsConfigBuilder): EsConfigBuilder
 
     def updateEsJavaOptsBuilder(builder: EsJavaOptsBuilder): EsJavaOptsBuilder
   }
 
+  object Plugin {
+    final case class EsUpdateSteps(steps: List[EsUpdateStep]) {
```
mgoworko: In order to execute the preexisting testcontainer update steps on WIndows native ES, I had to separate the definition of those update steps from their execution. `EsUpdateSteps` and related classes are definitions of update steps, that are then executed either on testcontainer image, or Windows ES directory

core/src/test/scala/tech/beshu/ror/utils/TestsUtils.scala
```diff
@@ -397,7 +397,7 @@ object TestsUtils {
   }
 
   def getResourcePath(resource: String): Path = {
-    File(getClass.getResource(resource).getPath).path
+    File(getClass.getResource(resource).toURI).path
```
mgoworko, line 400: On Linux it made no difference if we used getPath or toURI. On Windows we need URI - getPath is Unix-style path.

integration-tests/src/test/resources/current_user_metadata/readonlyrest.yml
```diff
@@ -24,7 +24,7 @@ readonlyrest:
       kibana:
         access: ro
         index: "user2_kibana_index"
-        hide_apps: ["user2_app1", "user2_app2", "/^Analytics\\|(?!(Maps)$).*$/"]
+        hide_apps: ["user2_app1", "user2_app2", "^Analytics\\|(?!Maps$).*"]
```
mgoworko, line 27: I am unsure why it worked fine on Linux, but according to LLM:
```
In ES 8.x on Windows, the bundled GraalJS engine is stricter: it does not like regex written as /…/ strings inside YAML unless they’re strictly valid JS regex literals.
```
It is like that in practice - this regex does not compile since ES 8.x on Windows.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FipsSslSuite.scala
```diff
@@ -57,12 +57,20 @@ class FipsSslSuite
 
   private lazy val rorClusterAdminStateManager = new CatManager(clients.last.adminClient, esVersion = esVersionUsed)
 
-  "Health check" should {
-    "be successful" when {
-      "internode ssl is enabled" in {
-        val response = rorClusterAdminStateManager.healthCheck()
+  // todo: The ES with FIPS does not start correctly when running tests on Windows with ES version lower than 8.18.
```
mgoworko, line 61: It needs further investigation. My suspicion is that it is not an issue with tests, but with ROR itself. So I suspect Fips may not work on Windows for ROR with ES<=8.18. That is why I think fixing it is outside of scope of this task - it may not be test issue, but production code issue, that just never surfaced.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAdminApiSuite.scala
```diff
@@ -184,7 +185,7 @@ trait BaseAdminApiSuite
           forceReload("/admin_api/readonlyrest_first_update.yml")
 
           // after first reload only dev1 can access indices
-          Thread.sleep(settingsReloadInterval.plus(5 second).toMillis) // have to wait for ROR1_2 instance config reload
+          Thread.sleep(settingsReloadInterval.plus(10 second).toMillis) // have to wait for ROR1_2 instance config reload
```
mgoworko, line 188: On Windows we sometimes need longer intervals in patience config than using Linux testcontainers.

ror-tools/src/main/scala/tech/beshu/ror/tools/RorToolsApp.scala
```diff
@@ -246,7 +246,7 @@ trait RorTools {
 
   private lazy val esPathOption =
     opt[String]("es-path")
-      .text(s"Path to elasticsearch directory; default=${EsDirectory.defaultPath}")
+      .text(s"Path to elasticsearch directory; default=/usr/share/elasticsearch")
```
mgoworko, line 249: The path is not correctly serialized on Windows. Even after fixing the format, the default Linux path does not make sense on Windows. So I hardcoded the Linux path in the message.

ror-tools/src/test/scala/tech/beshu/ror/tools/RorToolsAppSuite.scala
```diff
@@ -496,12 +440,14 @@ class RorToolsAppSuite
            | - file x-pack-core-$esVersionUsed.jar was patched by ROR ${metadata.rorVersion}
            | - file x-pack-ilm-$esVersionUsed.jar was patched by ROR ${metadata.rorVersion}
            | - file x-pack-security-$esVersionUsed.jar was patched by ROR ${metadata.rorVersion}
-           |""".stripMargin
+           |""".stripMargin.replace("\r\n", "\n")
       )
     }
     "Successfully patch, verify and unpatch" in {
+      // This file is created after patching on Windows for ES 8.x
+      val filesExcludedFromHashCalculation = List("plugin-security.policy.tmp")
```
mgoworko, line 450: Seems to be some ES 8.x issue specific to Windows. It does not happen on other ES versions and on Linux.

azure-pipelines.yml
```diff
@@ -190,6 +190,8 @@ stages:
           matrix:
             IT_es91x:
               ROR_TASK: integration_es91x
+            IT_es90x:
```
mgoworko, line 205: At the moment, for test purposes, I configured it to run Windows tests for a few representative modules, and only newest one on Linux.

I propose to always run Windows tests on the newest ES version - even in PR's. It is only one job more, and will allow to catch errors and incompatibilities before merge with develop, WDYT about that?. Which versions should be tested on develop and master pipelines?

core/src/test/resources/current_user_metadata_access_control_tests/ldap_ldap1_user5.ldif
```diff
@@ -26,7 +31,7 @@ objectClass: groupOfUniqueNames
 cn: ldap1_group1
 uniqueMember: cn=user5,ou=People,dc=example,dc=com
 
-dn: cn=ldap1_group1,ou=Groups,dc=example,dc=com
+dn: cn=ldap1_group2,ou=Groups,dc=example,dc=com
```
coutoPL, line 34: Why did we change this?

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
```diff
@@ -48,15 +50,27 @@ class CurrentUserMetadataAccessControlTests
...
+  private val mappings = List(
     "/current_user_metadata_access_control_tests/wiremock_service1_user5.json",
     "/current_user_metadata_access_control_tests/wiremock_service2_user6.json",
     "/current_user_metadata_access_control_tests/wiremock_service3_user7.json",
-  ))
-  private val ldap1 = LdapContainer.create("LDAP1",
+  )
+
+  private val wiremock =
+    if (OsUtils.isWindows)
```
coutoPL: It'd be nice to hide the OS-related things under some Wiremock adapter skin.

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
```diff
@@ -48,15 +50,27 @@ class CurrentUserMetadataAccessControlTests
...
+      new WindowsPseudoSingleContainerWiremock(8080, mappings)
+    else
+      new WireMockScalaAdapter(WireMockContainer.create(mappings: _*))
+
+  private lazy val (host, port) = wiremock match {
```
coutoPL: Maybe `(wiremockHost, wiremockPort)`?

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
```diff
@@ -74,196 +88,196 @@ class CurrentUserMetadataAccessControlTests
...
+       |    request_timeout_in_sec: 10                                # default 1
+       |    cache_ttl_in_sec: 60                                      # default 0 - cache disabled
+       |    users:
+       |      search_user_base_DN: "ou=People,dc=example,dc=com"
+       |      user_id_attribute: "uid"                                # default "uid"
+       |    groups:
+       |      search_groups_base_DN: "ou=Groups,dc=example,dc=com"
+       |      unique_member_attribute: "uniqueMember"                 # default "uniqueMember"
+       |
+    """.stripMargin.replace("\r\n", "\n").replace("\r", "\n")
```
coutoPL: Do we need it? Only in this test?

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
```diff
@@ -37,7 +38,12 @@ class LdapServerDiscoveryCheckYamlLoadedAccessControlTests
     with ForAllTestContainer
     with Inside {
 
-  override val container: LdapWithDnsContainer = new LdapWithDnsContainer("LDAP1", "test_example.ldif")
+  override val container: Container = if (OsUtils.isWindows) NoOpContainer else new LdapWithDnsContainer("LDAP1", "test_example.ldif")
```
coutoPL: mayeb we could introduce a new tag? 
```scala
object IgnoreOnWindows extends org.scalatest.Tag("tech.beshu.tags.IgnoreOnWindows")
```

Annotate with it the suite and change testOptions to exclude this tag when the tests are run on Windows OS?

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/definitions/ldap/implementations/UnboundidLdapUsersServiceNetworkRelatedTests.scala
```diff
@@ -72,44 +73,47 @@ class UnboundidLdapUsersServiceNetworkRelatedTests
...
-      }
-      "after connection failure and retry" in {
-        val authenticationService = createSimpleAuthenticationService()
-        authenticationService.assertSuccessfulAuthentication
-        ldap1ContainerWithToxiproxy.disableNetwork()
-        authenticationService.assertFailedAuthentication[LDAPSearchException, LDAPSearchException]
-        ldap1ContainerWithToxiproxy.enableNetwork()
-        authenticationService.assertSuccessfulAuthentication
+  // This test suite does not execute on Windows: there is currently no Windows version of ToxiproxyContainer
+  if (!OsUtils.isWindows) {
```
coutoPL: the same here

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
```diff
@@ -1105,35 +1110,38 @@ class LdapServicesSettingsTests private(ldapConnectionPoolProvider: UnboundidLda
...
-          assertion = { definitions =>
-            definitions.items should have size 1
-            val ldapService = definitions.items.head
-            ldapService shouldBe a[ComposedLdapAuthService]
-            ldapService.id should be(LdapService.Name("ldap1"))
-          }
-        )
+      // This test does not execute on Windows: there is currently no Windows version of LdapWithDnsContainer
+      if (!OsUtils.isWindows) {
+        "server discovery is enabled with custom dns and custom ttl" in {
```
coutoPL, line 1113: We should also be able to annotate a single test.

core/src/test/scala/tech/beshu/ror/unit/boot/ReadonlyRestStartingTests.scala
```diff
@@ -1360,7 +1360,7 @@ class ReadonlyRestStartingTests
             val expectedMessage =
               s"""Errors:
                  |Unable to configure audit output using a data stream in local cluster. Details: [Failed to setup ROR audit data stream readonlyrest_audit. Reason: Unable to determine if the index lifecycle policy with ID 'readonlyrest_audit-lifecycle-policy' has been created]
-                 |Unable to configure audit output using a data stream in remote cluster 0.0.0.0. Details: [Failed to setup ROR audit data stream readonlyrest_audit. Reason: Unable to determine if component template with ID 'readonlyrest_audit-mappings' has been created]""".stripMargin
+                 |Unable to configure audit output using a data stream in remote cluster 0.0.0.0. Details: [Failed to setup ROR audit data stream readonlyrest_audit. Reason: Unable to determine if component template with ID 'readonlyrest_audit-mappings' has been created]""".stripMargin.replace("\r\n", "\n")
```
coutoPL: Maybe we could introduce some util (in the form of an extension) for `stripMargin.replace("\r\n", "\n")`?

integration-tests/src/test/resources/current_user_metadata/readonlyrest.yml
coutoPL: Should we improve our docs?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FipsSslSuite.scala
coutoPL: ok, please create a Jira for that. Could you link the current task in Jira and this comment as well?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FipsSslSuite.scala
```diff
@@ -57,12 +57,20 @@ class FipsSslSuite
...
+  //       It needs to be checked further. It is an issue with file and thread operation permissions.
+  if (!(OsUtils.isWindows && (
```
coutoPL: Let's just ignore the whole suite on Windows for now. Add a #todo that it should be unignored when the new Jira is done.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/InMemoryLdapService.scala
```diff
@@ -0,0 +1,169 @@
...
+  extends Startable with LazyLogging {
+
+  private val config = new InMemoryDirectoryServerConfig(defaults.ldap.domainDn)
+  config.setSchema(null)
+  config.addAdditionalBindCredentials(defaults.ldap.bindDn.get, defaults.ldap.adminPassword)
+
+  config.setListenerConfigs(
+    InMemoryListenerConfig.createLDAPConfig("ldap", null, 0, null, false, false),
+    createLDAPSListener("ldaps", 0)
+  )
```
coutoPL:
```suggestion
private val config: InMemoryDirectoryServerConfig = {
    val c = new InMemoryDirectoryServerConfig(defaults.ldap.domainDn)
    c.setSchema(null)
    c.addAdditionalBindCredentials(defaults.ldap.bindDn.get, defaults.ldap.adminPassword)
    c.setListenerConfigs(
      InMemoryListenerConfig.createLDAPConfig("ldap", null, 0, null, false, false),
      createLDAPSListener("ldaps", 0)
    )
    c
  }```

File: tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/InMemoryLdapService.scala [hunk 23]
```diff
@@ -0,0 +1,169 @@
...
+    logger.info(s"LDAP in-memory server '$name' stopped")
+  }
+
+  def ldapPort: Int = server.getListenPort("ldap")
+
+  def ldapSSLPort: Int = server.getListenPort("ldaps")
+
+  def ldapHost: String = "localhost"
+
+  private def initLdapFromFile(): Unit = {
```
coutoPL: This code and the one from the `LdapWaitStrategy` class are similar.
Are we able to move it to some class responsible for such LDAP initiation logic?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/dependencies.scala
```diff
@@ -17,31 +17,55 @@
 package tech.beshu.ror.utils.containers
 
 import monix.eval.Coeval
+import tech.beshu.ror.utils.containers.windows.WindowsPseudoSingleContainerWiremock
+import tech.beshu.ror.utils.misc.OsUtils
 
 object dependencies {
 
   def ldap(name: String, ldapInitScript: String): DependencyDef = {
+    val ldap = LdapSingleContainer.create(name, ldapInitScript)
```
coutoPL: The semantics have changed. Now, Coeval makes no sense in this case.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/LdapContainer.scala
```diff
@@ -26,14 +26,15 @@ import monix.eval.Task
...
+import tech.beshu.ror.utils.containers.LdapSingleContainer.{InitScriptSource, defaults}
+import tech.beshu.ror.utils.containers.LdapWaitStrategy.*
 import tech.beshu.ror.utils.misc.ScalaUtils.*
 
 import java.io.{BufferedReader, InputStreamReader}
 import scala.concurrent.duration.*
 import scala.language.{implicitConversions, postfixOps}
 
-class LdapContainer private[containers] (name: String, ldapInitScript: InitScriptSource)
+class LdapContainer private[containers](name: String, ldapInitScript: InitScriptSource)
```
coutoPL: I think we should do some renaming of the LDAP container-like classes.

This `LdapContainer` is a real LDAP container based on the `openldap` image. Let's change its name.

The `LdapSingleContainer` is a basic trait for sth that acts like LDAP container. Let it be just `LdapContainer`.

The `WindowsPseudoSingleContainerLdap` could be `WindowsPseudoLdapContainer`.

WDYT?

tests-utils/src/main/scala/tech/beshu/ror/utils/files/FileUtils.scala
```diff
@@ -24,11 +24,12 @@ import scala.jdk.CollectionConverters.*
 
 object FileUtils {
 
-  def calculateHash(dir: Path): String = {
+  def calculateHash(dir: Path, excludedFiles: List[String]): String = {
```
coutoPL: `excludedFiles: List[String] = Nil`?

azure-pipelines.yml
coutoPL: Ok, let's do the following:
1. PR should run tests on Windows for the latest ES 8.x and 9.x modules (Elastic develops the latest two versions of ES/KBN)
2. On master | develop, we should run tests on Windows for the first and last module of the major ES version (e.g. es80x and es818x)
3. Let's have a manual job to run tests on Windows for all ES modules
4. Let's modify the name of `run_all_tests` manual job to sth like `run_all_tests_on_linux`
5. Let's add the `run_all_tests` manual job to run all tests on Linux and Windows
 
WDYT?

coutoPL (changes requested):

core/src/test/resources/current_user_metadata_access_control_tests/ldap_ldap1_user5.ldif
mgoworko: It was an invalid value - a duplicate. The group above also has `cn=ldap1_group1`. The testcontainer accepted it, the in-memory LDAP throws exception when it detected duplicates.

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
mgoworko: I though about it before, and even tried again to implement it now after this comment. I think it not viable for some reasons:
- Gradle exlusion tags are not properly handled by IntelliJ . The exluded tests would still be executed on Windows. It is problematic, because then the containers are started, and the containers not supported on Windows are throwing errors. So it will essentially cause the tests to not be executable from IntelliJ on Windows when running multiple tests at once.
- It would still have to be on app level, not gradle level. We have to check os in code, and only then annotate. It is not like the esVersion - which is checked on Gradle level. It is problematic - conditional tagging requires some boilerplate.

So as a result, I created some helper methods `ignoreOnWindows` and `doNotCreateOnWindows` for containers. They are tag-like, but without tag-related difficulties

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/definitions/ldap/implementations/UnboundidLdapUsersServiceNetworkRelatedTests.scala
mgoworko: Changed to:

```
  ignoreOnWindows {
    "An LdapAuthenticationService should" - {
```

Described in the comment above, why not using tag

core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/definitions/LdapServicesSettingsTests.scala
mgoworko: Also changed to `ignoreOnWindows {`  syntax here, for reasons described in another discussion above.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/InMemoryLdapService.scala
mgoworko: Yes, I was able to merge those 2 implementations.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/dependencies.scala
mgoworko: I removed Coeval and the ResolvedDependency class.

core/src/test/scala/tech/beshu/ror/unit/boot/ReadonlyRestStartingTests.scala
mgoworko: Done, `stripMarginAndReplaceWindowsLineBreak`

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
mgoworko: Replaced with stripMarginAndReplaceWindowsLineBreak util, the `replace("\r", "\n")` was redundant

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
mgoworko: It would be the only place where it is used. I replaced it with new util, used like that

```
private val wiremock = OsUtils.currentOs match {
    case CurrentOs.Windows => new WindowsPseudoSingleContainerWiremock(8080, mappings)
    case CurrentOs.OtherThanWindows => new WireMockScalaAdapter(WireMockContainer.create(mappings: _*))
  }
```

integration-tests/src/test/resources/current_user_metadata/readonlyrest.yml
mgoworko: I'm not sure. On Linux both syntaxes work. Probably we should mention it like that:

`"⚠️Windows users: Regex patterns in configuration must use double backslashes (\\) and omit /.../ delimiters to work correctly."`

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/FipsSslSuite.scala
mgoworko: https://readonlyrest.atlassian.net/browse/RORDEV-1605

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/LdapContainer.scala
mgoworko: Done, renamed to:
- base interface `LdapContainer`
- implementations `OpenLdapContainer` and `WindowsPseudoLdapContainer`
- other Windows containers renamed according to `WindowsPseudoXYZContainer` convention too

azure-pipelines.yml
mgoworko: Ok, points 1-4 done. For now I didn't do 5. - I think that having `run_all_tests_on_linux` and `run_all_tests_on_windows` is enough:
- having common `run_all_tests` would make the Azure job condition logic much more complicated
- we would rarely use it, and we can just run both jobs dedicated to Win/Linux

important: after this change in PR pipelines we will now have exactly 11 "long" jobs (not counting those, that are finished in a few minutes, including core tests). Maybe we should have one less, as we have 10 runners.

integration-tests/src/test/resources/current_user_metadata/readonlyrest.yml
coutoPL: ok, let's do that

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
```diff
@@ -37,7 +38,12 @@ class LdapServerDiscoveryCheckYamlLoadedAccessControlTests
     with ForAllTestContainer
     with Inside {
 
-  override val container: LdapWithDnsContainer = new LdapWithDnsContainer("LDAP1", "test_example.ldif")
+  override val container: Container = doNotCreateOnWindows(new LdapWithDnsContainer("LDAP1", "test_example.ldif"))
+
+  private lazy val ldapPort = container match {
+    case ldap: LdapWithDnsContainer => ldap.dnsPort
+    case _ => 1234
```
coutoPL: let's add a comment that this is a random value for the Windows use case

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
```diff
@@ -69,19 +75,28 @@ class LdapServerDiscoveryCheckYamlLoadedAccessControlTests
...
+              blockContext
+            }
           }
         }
       }
     }
   }
 }
+
+object NoOpContainer extends Container {
```
coutoPL: maybe we can move this to test utils?
somewhere close to our containers imeplementations

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RemoteReindexSuite.scala
```diff
@@ -45,7 +47,12 @@ class RemoteReindexSuite
     EsClusterSettings.create(
       clusterName = "ROR_SOURCE_ES",
       nodeDataInitializer = RemoteReindexSuite.sourceEsDataInitializer(),
-      esVersion = EsVersion.SpecificVersion("es67x"),
+      esVersion = OsUtils.currentOs match {
+        case CurrentOs.Windows =>
+          EsVersion.SpecificVersion("es70x") // Integration tests on Windows cannot be run with ES version below 7.0
```
coutoPL, line 52: IMO, we can use es70x for all tests here.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
coutoPL: It's ok. But I'm not sure about naming. We are in the context of the ES plugin. The update steps are installation. 

So, maybe:
1. `def esUpdateSteps(config: Config): EsUpdateSteps` -> `def install(config: Config): PluginInstallationSteps`
2. `final case class EsUpdateSteps(steps: List[EsUpdateStep])` -> `final case class PluginInstallation(steps: List[PluginInstallationStep])`

Or sth like that.

WDYT?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -145,19 +214,33 @@ class Elasticsearch(esVersion: String,
...
-      .run(s"chown -R elasticsearch:elasticsearch ${esDir.toString()}")
+      .run(s"chown -R elasticsearch:elasticsearch ${config.esDir.toString()}")
       .run(s"chown -R elasticsearch:elasticsearch ${config.esConfigDir.toString()}")
       .run("rm /etc/elasticsearch/elasticsearch.keystore")
       .addEnvs(config.envs + ("ES_JAVA_OPTS" -> javaOptsBasedOn(withEsJavaOptsBuilderFromPlugins)))
       .installPlugins()
       .user("elasticsearch")
   }
 
+  def esUpdateSteps: EsUpdateSteps = {
```
coutoPL: maybe instead of exposing this method (it should be renamed too), we should expose list of plugins instead?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -174,23 +257,20 @@ class Elasticsearch(esVersion: String,
       .foldLeft(builder) { case (currentBuilder, update) => update(currentBuilder) }
   }
 
-  private def esConfigFileBasedOn(config: Config,
-                                  withEsConfigBuilder: EsConfigBuilder => EsConfigBuilder) = {
+  def esConfigFile(networkHost: String): File = {
```
coutoPL: I'm not sure about this public method. Above, we have `val config: Config` which describes how the ES config looks (we create the elasticsearch.yml from this description). Can we not add the `networkHost` there?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
coutoPL: Or maybe better: installation & configuration steps?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/InMemoryLdapService.scala
```diff
@@ -0,0 +1,137 @@
...
+import java.security.cert.X509Certificate
+import java.security.{KeyPairGenerator, KeyStore, Security}
+import java.util.Date
+import javax.net.ssl.{KeyManager, KeyManagerFactory}
+import javax.security.auth.x500.X500Principal
+import scala.concurrent.duration.*
+import scala.language.{implicitConversions, postfixOps}
+import scala.util.Using
+
+class InMemoryLdapService(name: String, ldapInitScript: InitScriptSource)
```
coutoPL: maybe private in windows package?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsEsRunner.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    sys.addShutdownHook {
+      logger.info(s"JVM shutting down, stopping Elasticsearch [${config.clusterName}][${config.nodeName}] process...")
+      killEs(process)
+    }
+
+    process
+  }
+
+  def killEs(windowsEsProcess: WindowsEsProcess): Unit = {
```
coutoPL: Let's move this method to the `WindowsEsProcess` class.
Or create a proper extension for this class.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoEsContainer.scala
```diff
@@ -0,0 +1,60 @@
...
+import tech.beshu.ror.utils.containers.windows.WindowsEsSetup.prepareAndStartEs
+
+import java.util.function.Consumer
+
+class WindowsPseudoEsContainer(elasticsearch: Elasticsearch,
+                               waitStrategy: ElasticsearchNodeWaitingStrategy,
+                               additionalLogConsumer: Option[Consumer[OutputFrame]])
+  extends GenericContainer[WindowsPseudoEsContainer]("noop:latest") {
+
+  private var windowsEsProcess: Option[WindowsEsProcess] = None
```
coutoPL: Let's use an atomic ref instead. Let's wrap the process inside some state wrapper (not initialized/intialized/stopped)

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoEsContainer.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  override def start(): Unit = doStart()
+
+  override def doStart(): Unit = {
+    windowsEsProcess match {
```
coutoPL: start only when not initialized. Otherwise throw Invalid State Ex

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoEsContainer.scala
```diff
@@ -0,0 +1,60 @@
...
+      case Some(_) =>
+        ()
+      case None =>
+        windowsEsProcess = Some(prepareAndStartEs(elasticsearch, additionalLogConsumer))
+        waitStrategy.waitUntilReady()
+    }
+  }
+
+  override def stop(): Unit = {
+    super.stop()
```
coutoPL: stop only when initialized. Otherwise throw Invalid State Ex

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoWiremockContainer.scala
```diff
@@ -0,0 +1,81 @@
...
+class WindowsPseudoWiremockContainer(val port: Int, mappings: List[String])
+  extends SingleContainer[GenericContainer[_]] {
+  override val container: WindowsPseudoGenericContainerWiremock = new WindowsPseudoGenericContainerWiremock(port, mappings)
+}
+
+object WindowsPseudoWiremockContainer {
+  class WindowsPseudoGenericContainerWiremock(port: Int, mappings: List[String])
+    extends GenericContainer[WindowsPseudoGenericContainerWiremock]("noop:latest") {
+
+    private var server: Option[WireMockServer] = None
```
coutoPL: the same here (so maybe we need some util trait for that?)

tests-utils/src/main/scala/tech/beshu/ror/utils/elasticsearch/BaseManager.scala
```diff
@@ -94,17 +96,26 @@ abstract class BaseManager(client: RestClient,
...
     private def checkResponseAssertions(): Unit = {
-      if(!isForbidden && esNativeApi) {
-        if (Version.greaterOrEqualThan(esVersion, 7, 14, 0) && Version.lowerThan(esVersion, 7, 16, 0)) {
-          request match {
-            case Some(req) if isExcludedRequest(req) =>
-              // ES [7.14.0,7.16.0) doesn't add the X-elastic-product header to some responses (under some conditions)
-            case Some(_) | None =>
+      OsUtils.currentOs match {
+        case CurrentOs.Windows =>
+          // On Windows the header is not added to some requests in versions <7.14-8.0)
```
coutoPL: This is bad. 
Let's create a jira task for this, and let's add a todo here with the jira ref.

tests-utils/src/main/scala/tech/beshu/ror/utils/misc/OsUtils.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  sealed trait CurrentOs
+
+  object CurrentOs {
+    case object Windows extends CurrentOs
+
+    case object OtherThanWindows extends CurrentOs
+  }
+
+  private object NoOpContainer extends Container {
```
coutoPL: I've already seen sth like this, somewhere in this PR :D

azure-pipelines.yml
```diff
@@ -122,11 +123,19 @@ stages:
         succeeded(),
         or(
           ne(variables['Build.Reason'], 'Manual'),
-          and(eq(variables['Build.Reason'], 'Manual'), eq('${{ parameters.actionToPerform }}', 'run_all_tests'))
+          and(eq(variables['Build.Reason'], 'Manual'), startsWith('${{ parameters.actionToPerform }}', 'run_all_tests'))
```
coutoPL, line 126: why `startsWith`?

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
coutoPL: but cannot we hide all of these?
```scala
  private val wiremock = OsUtils.currentOs match {
    case CurrentOs.Windows => new WindowsPseudoWiremockContainer(8080, mappings)
    case CurrentOs.OtherThanWindows => new WireMockScalaAdapter(WireMockContainer.create(mappings: _*))
  }

  private lazy val (wiremockHost, wiremockPort) = wiremock match {
    case container: WireMockScalaAdapter => (container.getWireMockHost, container.getWireMockPort)
    case _ => ("localhost", 8080)
  }
```

I'd be better if these internal are hidden and look like this:
```
private val wiremock = Wiremock.create(mappings)
private lazy val wiremockHost = wiremock.host // maybe we don't need this, maybe we can just inline it
private lazy val wiremockPort = wiremock.port // maybe we don't need this, maybe we can just inline it
```

WDYT?

core/src/test/scala/tech/beshu/ror/integration/LdapServerDiscoveryCheckYamlLoadedAccessControlTests.scala
mgoworko: Was already in `object OsUtils`, as private util (there is a method `def doNotCreateOnWindows(creator: => Container): Container` there, but the NoOpContainer is private in that util). 

It was a duplicate, removed

tests-utils/src/main/scala/tech/beshu/ror/utils/misc/OsUtils.scala
mgoworko: This is the correct one, used by `def doNotCreateOnWindows`. The second one was unused duplicate, now removed.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsEsRunner.scala
mgoworko: Moved to WindowsEsProcess, which is now normal, non-case class

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoEsContainer.scala
mgoworko: Done. Created new util - WindowsPseudoContainer - encapsulating the pseudo container logic.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/RemoteReindexSuite.scala
mgoworko: It configured it that way, separetely for Win and Linux, on purpose. This version must be the lowest ES version that we run tests on.  On Linux it is 67x. When I set es70x and run tests on 67x one test fails, because the older ES node cannot communicate with newer node.

```
{"error":{"root_cause":[{"type":"exception","reason":"Error parsing the response, remote is likely not an Elasticsearch instance"}],"type":"exception","reason":"Error parsing the response, remote is likely not an Elasticsearch instance","caused_by":{"type":"x_content_parse_exception","reason":"[1:186] [search_response] failed to parse field [hits]","caused_by":{"type":"x_content_parse_exception","reason":"[1:186] [hits] total doesn't support values of type: START_OBJECT"}}},"status":500}
```

I did not find official docs on this subject, but old forum thread suggests it is similar to downgrading and is not supported (the same error message) https://discuss.elastic.co/t/reindex-to-version-6-and-unknown-error-parsing-response/178900

On Windows, it must be es70x, because it is the lowest version tested, and lowest version that can be executed as native Windows ES process.

azure-pipelines.yml
mgoworko: In order to include or exclude the job when running either `run_all_tests_on_linux` or `run_all_tests_on_windows`

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
mgoworko: Refactored to: `def installationSteps(config: Config): PluginInstallationSteps`. 

I think this version is consistent across all usages. Method name `install` would suggest it performs installation, and installation & configuration would be IMO longer and less readable

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
mgoworko: done, list of plugins exposed instead

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
mgoworko: Restored to previous version, not using config. The network host is now replaced in configuration file in the `WindowsEsSetup`, alongside other config adjustments for Windows.

tests-utils/src/main/scala/tech/beshu/ror/utils/elasticsearch/BaseManager.scala
mgoworko: After quick investigation modified comment and created https://readonlyrest.atlassian.net/browse/RORDEV-1630

core/src/test/scala/tech/beshu/ror/integration/CurrentUserMetadataAccessControlTests.scala
mgoworko: done, util used here and for common dependency creation

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoContainer.scala
```diff
@@ -0,0 +1,66 @@
...
+
+abstract class WindowsPseudoContainer[T <: GenericContainer[T], SERVICE] extends GenericContainer[T]("noop:latest") {
+
+  protected def name: String
+
+  protected def prepare(): SERVICE
+
+  protected def awaitReady(): Unit
+
+  protected def destroy(service: SERVICE): Unit
```
coutoPL: Maybe we should introduce an interface for the service instead?
It could have the destroy and getPort methods.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoContainer.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  protected def getPort(service: SERVICE): Int
+
+  private val service: AtomicReference[Option[SERVICE]] = new AtomicReference(Option.empty[SERVICE])
+
+  override final def start(): Unit = doStart()
+
+  override final def doStart(): Unit = {
+    val success = service.compareAndSet(None, Some(prepare()))
```
coutoPL: It's dangerous to do it like that. The `compareAndSet` may be called many times, and we have potential side effect in the `prepare` method.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoContainer.scala
```diff
@@ -0,0 +1,66 @@
...
+    if (!success) throw new IllegalStateException(s"Service $name is already started, cannot start again.")
+    awaitReady()
+  }
+
+  override final def stop(): Unit = {
+    service.getAndUpdate {
+      case Some(service) =>
+        destroy(service)
```
coutoPL: the same here

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoWiremockContainer.scala
```diff
@@ -0,0 +1,74 @@
...
+    override protected def destroy(service: WireMockServer): Unit =
+      service.stop()
+
+    override protected def getPort(service: WireMockServer): Int =
+      service.port()
+
+  }
+}
+
+object WireMockServerCreator {
```
coutoPL: private?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoWiremockContainer.scala
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.File
+
+class WindowsPseudoWiremockContainer(val port: Int, mappings: List[String])
+  extends SingleContainer[GenericContainer[_]] {
+  override val container: WindowsPseudoGenericContainerWiremock = new WindowsPseudoGenericContainerWiremock(port, mappings)
+}
+
+object WindowsPseudoWiremockContainer {
+  class WindowsPseudoGenericContainerWiremock(port: Int, mappings: List[String])
```
coutoPL: private?

tests-utils/src/main/scala/tech/beshu/ror/utils/misc/OsUtils.scala
coutoPL: I'd move the NoOpContainer to tech.beshu.ror.utils.containers package. It may be useful somewhere in the future.

coutoPL (changes requested):

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoContainer.scala
mgoworko: Done, new interface `WindowsPseudoContainer.Service`

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoContainer.scala
mgoworko: You are right. Now I just did it using synchronized method. It is called very rarely and in tests, so performance impact of synchronized doesn't matter.

coutoPL: let's add synchronized to getPort too

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/windows/WindowsPseudoContainer.scala
```diff
@@ -0,0 +1,75 @@
...
+  override final def start(): Unit = doStart()
+
+  override final def doStart(): Unit = synchronized {
+    service.get() match {
+      case Some(_) =>
+        () // The testcontainers library behaves in idempotent way when starting containers. We must adhere to that behavior in the Windows pseudo-container, because the start method can be called more than once.
+      case None =>
+        service.set(Some(prepare()))
+        awaitReady()
+    }
```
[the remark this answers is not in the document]
coutoPL: @mgoworko yes, this makes sense

coutoPL (approved):
LGTM.

But to do left:
1. The mentioned minor changes
2. The docs improvements reminder
3. I'm not sure (I don't want to go through the PR again) - did you fix some Windows-related issues in this PR too? If so, it'd be nice to mention them in the changelog description.

mgoworko:
1. done
2. oneliner PR in docs https://github.com/beshu-tech/readonlyrest-docs/pull/282
3. there is only one thing - fixed double-encoding of LDAP cn (I caused LDAP tests to fail on Windows. I am not sure if it would cause production issues, perhaps it depends on LDAP used. Added to changelog)

PR #1159, opened 2025-09-12
[RORDEV-1610] Fixed: keyword field value suggestions in Discover/Data View filters

coutoPL:
🐞Fix (KBN) [Restored keyword field value suggestions in Discover/Data View filters](https://forum.readonlyrest.com/t/kibana-data-view-filter-not-working-with-keyword/2843)

## Summary by CodeRabbit

- Bug Fixes
  - Preserves indices permissions header during context stashing, improving request handling across supported Elasticsearch versions.
  - Enhances index mapping compatibility by selecting the correct endpoint for ES 6.x vs 7.x+.

- Documentation
  - Updated API schema: ExternalAuthorizationService.type renamed to EXT_AUTHZ.

- Chores
  - Bumped plugin pre-release version to 1.67.0-pre6.


es91x/src/main/scala/tech/beshu/ror/es/utils/ThreadContextOps.scala
```diff
@@ -59,7 +59,7 @@ final class ThreadContextOps(val threadContext: ThreadContext) extends AnyVal {
 
 object ThreadContextOps {
 
-  private val transientHeaderNames: List[String] = "_xpack_security_authentication" :: "_authz_info" :: Nil
+  private val transientHeaderNames: List[String] = "_xpack_security_authentication" :: "_authz_info" :: "_indices_permissions" :: Nil
```
mgoworko, line 62: I know that the idea is to have custom implementation for ES-version specific utils. But I think we can refactor the code and have the `ThreadContextOps` implementation in a single place in core module. From what I see the implementation is the same for all ES modules since es67x, only the ThreadContext  class is of course different (but exposing nearly identical interface).

We can have interface like:
```
trait ThreadContextLike {
  type StoredContextLike <: AutoCloseable
  def stashContext(): StoredContextLike
  def getResponseHeaders: java.util.Map[String, java.util.List[String]]
  def addResponseHeader(key: String, value: String): Unit
  def getTransient[T](key: String): Option[T]
  def putTransient(key: String, value: AnyRef): Unit
  def getHeader(key: String): Option[String]
  def putHeader(key: String, value: String): Unit
}
```

And the util can be generic and defined in core module. If not the full refactor, then at least I would suggest moving `transientHeaderNames` to core.

docs/api/ror-internal-api-swagger.yaml
```diff
@@ -1226,7 +1226,7 @@ components:
         type:
           type: string
           enum:
-            - LDAP
+            - EXT_AUTHZ
```
mgoworko, line 1229: Is this only a fix in swagger? I did not see any corresponding change in code.

ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/utils/EsUtil.scala
```diff
@@ -37,6 +37,7 @@ object EsUtil {
   val es8120: SemVer = SemVer.unsafeParse("8.12.0")
   val es890: SemVer = SemVer.unsafeParse("8.9.0")
   val es870: SemVer = SemVer.unsafeParse("8.7.0")
+  val es850: SemVer = SemVer.unsafeParse("8.5.0")
```
mgoworko: Is it used anywhere?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseXpackApiSuite.scala
```diff
@@ -1080,6 +1081,26 @@ trait BaseXpackApiSuite
...
+      result should have statusCode 200
+      result.terms should be(Set("world"))
+    }
+    "not be allowed for dev2 and test1_index_a" excludeES(allEs6x, allEs7xBelowEs714x) in {
+      val result = dev2XpackApiManager.getTerms("test1_index_a", "hello.keyword")
+
+      result should have statusCode 404
+    }
+    // todo: will be done in separate PR
+    "support filter and fields rule" excludeES(allEs6x, allEs7xBelowEs714x) ignore {
```
mgoworko, line 1097: I think right now it is not clear what the todo is about (what is not yet implemented/fixed and what will be done in next PR). On first glance - the Set.empty response is correct, when there are no docs in `test2_index`

es91x/src/main/scala/tech/beshu/ror/es/utils/ThreadContextOps.scala
coutoPL: I know that exposing our own interfaces and adapters is tempting, but I'm not sure it's worth it.
In the context of `ThreadContextOps`, we could move to the core, the `stashPreservingSomeHeaders` and `putHeaderIfNotPresent` methods. If we want to move `add...AuthenticationHeader` methods, we will have to do sth with the `XPackSecurityAuthenticationHeader` object. 

Notice that each ES could have different headers, different persistent methods of authentication header values, etc.
Obviously, currently we have some duplications, but changes in the ES side won't hurt us.

docs/api/ror-internal-api-swagger.yaml
coutoPL: yes, I noticed this error and fixed it.
Not related to the bug from the PR.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseXpackApiSuite.scala
coutoPL: changed todo description

mgoworko (approved):
LGTM

PR #1163, opened 2025-09-20
[RORDEV-1562] Separate ROR KBN authentication and authorization rules

mgoworko:


core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseRorKbnRule.scala
```diff
@@ -0,0 +1,189 @@
...
+      .toEither
+      .map(Jwt.Payload.apply)
+      .left.map { ex => logger.debug(s"[${requestId.show}] JWT token '${token.show}' parsing error " + ex.getClass.getSimpleName) }
+  }
+
+  private def parser(rorKbn: RorKbnDef) = rorKbn.checkMethod match {
+    case Hmac(rawKey) => Jwts.parser().verifyWith(Keys.hmacShaKeyFor(rawKey)).build()
+    case Rsa(pubKey) => Jwts.parser().verifyWith(pubKey).build()
+    case Ec(pubKey) => Jwts.parser().verifyWith(pubKey).build()
+  }
```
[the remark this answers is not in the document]
mgoworko: No, there is no SignatureCheckMethod.NoCheck

```
  sealed trait SignatureCheckMethod
  object SignatureCheckMethod {
    final case class Hmac(key: Array[Byte]) extends SignatureCheckMethod
    final case class Rsa(pubKey: PublicKey) extends SignatureCheckMethod
    final case class Ec(pubKey: PublicKey) extends SignatureCheckMethod
  }
  ```

File: core/src/test/scala/tech/beshu/ror/unit/acl/factory/decoders/rules/auth/RorKbnAuthorizationRuleSettingsTests.scala [hunk 2]
```diff
@@ -0,0 +1,448 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.unit.acl.factory.decoders.rules.auth
+
+import org.scalatest.matchers.should.Matchers.*
```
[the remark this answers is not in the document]
mgoworko: This module (and entire project, with exception of audit example) is Gradle-defined and uses Scala 3.3.3

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/RorKbnRuleHelper.scala
```diff
@@ -0,0 +1,189 @@
...
+import tech.beshu.ror.accesscontrol.request.RequestContextOps.from
+import tech.beshu.ror.accesscontrol.utils.ClaimsOps.*
+import tech.beshu.ror.accesscontrol.utils.ClaimsOps.ClaimSearchResult.{Found, NotFound}
+import tech.beshu.ror.implicits.*
+import tech.beshu.ror.utils.json.JsonPath
+import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
+
+import scala.util.Try
+
+object RorKbnRuleHelper extends Logging {
```
coutoPL: It'd be great if we could:
1. have a base class for `RorKbnAuthenticationRule` and `RorKbnAuthorizationRule` that contains common things used by both (or depending on what the things are, maybe we replace the base class with some extension class for e.g., Jwt.Token)
2. It'd be nice to reuse the `BaseComposedAuthenticationAndAuthorizationRule` class (like we did in the case of LDAP) for composing logic provided by `RorKbnAuthenticationRule` and `RorKbnAuthorizationRule` classes

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthenticationRule.scala
```diff
@@ -0,0 +1,55 @@
...
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.RorKbnRuleHelper.RorKbnOperation.Authenticate
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.impersonation.{AuthenticationImpersonationCustomSupport, AuthorizationImpersonationCustomSupport}
+import tech.beshu.ror.accesscontrol.blocks.{BlockContext, BlockContextUpdater}
+import tech.beshu.ror.accesscontrol.domain.*
+
+final class RorKbnAuthenticationRule(val settings: Settings,
+                                     override val userIdCaseSensitivity: CaseSensitivity)
+  extends AuthenticationRule
+    with AuthenticationImpersonationCustomSupport
+    with AuthorizationImpersonationCustomSupport
```
coutoPL: to remove?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthorizationRule.scala
```diff
@@ -0,0 +1,58 @@
...
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.RorKbnAuthorizationRule.Settings
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.RorKbnRuleHelper
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.RorKbnRuleHelper.RorKbnOperation.Authorize
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.impersonation.{AuthenticationImpersonationCustomSupport, AuthorizationImpersonationCustomSupport}
+import tech.beshu.ror.accesscontrol.blocks.{BlockContext, BlockContextUpdater}
+import tech.beshu.ror.accesscontrol.domain.GroupsLogic
+
+final class RorKbnAuthorizationRule(val settings: Settings)
+  extends AuthorizationRule
+    with AuthenticationImpersonationCustomSupport
```
coutoPL: to remove?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthRule.scala
```diff
@@ -51,142 +40,25 @@ final class RorKbnAuthRule(val settings: Settings,
...
-      case Found(header) => blockContext.withUserMetadata(_.withUserOrigin(UserOrigin(header.value)))
-      case ClaimSearchResult.NotFound => blockContext
-    }
-  }
 }
 
 object RorKbnAuthRule {
 
-  implicit case object Name extends RuleName[RorKbnAuthRule] {
+  implicit case object Name extends RuleName[RorKbnAuthRule | RorKbnAuthenticationRule] {
```
coutoPL: Why like this? Shouldn't we use `RorKbnAuthRule` here instead?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/users/LocalUsersContext.scala
```diff
@@ -74,7 +74,9 @@ object LocalUsersContext {
     implicit val methodsRule: LocalUsersSupport[MethodsRule] = NotAvailableLocalUsers()
     implicit val repositoriesRule: LocalUsersSupport[RepositoriesRule] = NotAvailableLocalUsers()
     implicit val responseFieldsRule: LocalUsersSupport[ResponseFieldsRule] = NotAvailableLocalUsers()
-    implicit val rorKbnAuthRule: LocalUsersSupport[RorKbnAuthRule] = NotAvailableLocalUsers()
+    implicit val rorKbnAuthRule: LocalUsersSupport[RorKbnAuthRule | RorKbnAuthenticationRule] = NotAvailableLocalUsers()
```
coutoPL: Shouldn't we use `RorKbnAuthRule`?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
```diff
@@ -93,7 +93,9 @@ object VariableContext {
     implicit val maxBodyLengthRule: VariableUsage[MaxBodyLengthRule] = NotUsingVariable()
     implicit val methodsRule: VariableUsage[MethodsRule] = NotUsingVariable()
     implicit val proxyAuthRule: VariableUsage[ProxyAuthRule] = NotUsingVariable()
-    implicit val rorKbnAuthRule: VariableUsage[RorKbnAuthRule] = NotUsingVariable()
+    implicit val rorKbnAuthRule: VariableUsage[RorKbnAuthRule | RorKbnAuthenticationRule] = NotUsingVariable()
+    implicit val rorKbnAuthenticationRule: VariableUsage[RorKbnAuthenticationRule] = NotUsingVariable()
+    implicit val rorKbnAuthorizationRule: VariableUsage[RorKbnAuthorizationRule] = NotUsingVariable()
```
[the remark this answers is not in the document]
coutoPL: I also think we should do:
```
implicit val rorKbnAuthRule: VariableUsage[RorKbnAuthRule] = NotUsingVariable()
```

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRuleDecoders.scala
```diff
@@ -0,0 +1,154 @@
...
+  override protected def decoder: Decoder[RuleDefinition[RorKbnAuthenticationRule]] = {
+    RorKbnAuthRuleDecoder.nameAndGroupsSimpleDecoder
+      .or(RorKbnAuthRuleDecoder.nameAndGroupsExtendedDecoder)
+      .toSyncDecoder
+      .emapE { case (name, groupsLogicOpt) =>
+        rorKbnDefinitions.items.find(_.id === name) match {
+          case Some(rorKbnDef) =>
+            groupsLogicOpt match {
+              case Some(_) =>
+                Left(RulesLevelCreationError(Message(s"Cannot create ${RorKbnAuthenticationRule.Name.name.show}, because there are superfluous groups settings. Remove the groups settings, or use authorization or auth rule, if group settings are required.")))
```
coutoPL: Instead of saying "authorization or auth rule", let's use the rules' names (e.g., `RorKbnAuthorizationRule.Name.name`)

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRuleDecoders.scala
```diff
@@ -0,0 +1,154 @@
...
+
+  override protected def decoder: Decoder[RuleDefinition[RorKbnAuthenticationRule]] = {
+    RorKbnAuthRuleDecoder.nameAndGroupsSimpleDecoder
+      .or(RorKbnAuthRuleDecoder.nameAndGroupsExtendedDecoder)
+      .toSyncDecoder
+      .emapE { case (name, groupsLogicOpt) =>
+        rorKbnDefinitions.items.find(_.id === name) match {
+          case Some(rorKbnDef) =>
+            groupsLogicOpt match {
+              case Some(_) =>
```
coutoPL: Maybe we can avoid nested pattern matching like that:
```
val foundKbnDef = rorKbnDefinitions.items.find(_.id === name)
(foundKbnDef, groupsLogicOpt) match {
  case (Some(_), Some(_)) =>
  ...
}
```
WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRuleDecoders.scala
```diff
@@ -0,0 +1,154 @@
...
+                Left(RulesLevelCreationError(Message(s"Cannot create ${RorKbnAuthenticationRule.Name.name.show}, because there are superfluous groups settings. Remove the groups settings, or use authorization or auth rule, if group settings are required.")))
+              case None =>
+                val settings = RorKbnAuthenticationRule.Settings(rorKbnDef)
+                val rule = new RorKbnAuthenticationRule(settings, globalSettings.userIdCaseSensitivity)
+                Right(RuleDefinition.create(rule))
+            }
+          case None =>
+            Left(RulesLevelCreationError(Message(s"Cannot find ROR Kibana definition with name: ${name.show}")))
```
coutoPL: Maybe we should move the code to a helper method because it's used in several places. It'd be nice to be sure the message is the same.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
coutoPL, line 1: let's rename this file to `RorKbnRulesDecoders`

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRuleDecoders.scala
```diff
@@ -0,0 +1,154 @@
...
+        }
+      }
+      .map(RuleDefinition.create[RorKbnAuthRule | RorKbnAuthenticationRule](_))
+      .decoder
+  }
+}
+
+private object RorKbnAuthRuleDecoder {
```
coutoPL: Maybe we can have `private object RorKbnRulesDecdoersHelper` instead? (like we have in the `LdapRulesDecoders.scala`)

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRuleDecoders.scala
```diff
@@ -0,0 +1,154 @@
...
+      .emapE { case (name, groupsLogicOpt) =>
+        rorKbnDefinitions.items.find(_.id === name) match {
+          case Some(rorKbnDef) =>
+            groupsLogicOpt match {
+              case Some(groupsLogic) =>
+                val settings = RorKbnAuthorizationRule.Settings(rorKbnDef, groupsLogic)
+                val rule = new RorKbnAuthorizationRule(settings)
+                Right(RuleDefinition.create[RorKbnAuthorizationRule](rule))
+              case None =>
+                Left(RulesLevelCreationError(Message(s"Cannot create RorKbnAuthorizationRule - missing groups settings")))
```
coutoPL: let's use `RorKbnAuthorizationRule.Name.name.show` instead

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
```diff
@@ -0,0 +1,154 @@
...
+      }
+      .decoder
+  }
+}
+
+class RorKbnAuthRuleDecoder(rorKbnDefinitions: Definitions[RorKbnDef],
+                            globalSettings: GlobalSettings)
+  extends RuleBaseDecoderWithoutAssociatedFields[RorKbnAuthRule | RorKbnAuthenticationRule] with Logging {
+
+  override protected def decoder: Decoder[RuleDefinition[RorKbnAuthRule | RorKbnAuthenticationRule]] = {
```
coutoPL: Ok, now I understand why I saw `RorKbnAuthRule | RorKbnAuthenticationRule` everywhere before.
We have this in the requirements:
> Because we want to maintain backwards compatibility of our ACL, we have to support it. But when users define it like that, we should parse it to the new ror_kbn_authentication rule class.

and I guess this addresses that. It's ok, but the backward compatibility is at the settings parsing level. Declaring `RuleDefinition[RorKbnAuthRule | RorKbnAuthenticationRule]`, we affect the other parts of the core business logic with sth that from the core's perspective is irrelevant.

core/src/test/scala/tech/beshu/ror/integration/RorKbnAuthenticationYamlLoadedAccessControlTests.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  override protected def configYaml: String =
+    """http.bind_host: _eth0:ipv4_
+      |network.host: _eth0:ipv4_
+      |
+      |http.type: ssl_netty4
+      |#transport.type: local
+      |
+      |readonlyrest:
+      |  ssl:
```
coutoPL: Let's remove from the example the parts that are not used, like "http", "network", "readonlyrest.ssl", etc.

core/src/test/scala/tech/beshu/ror/integration/RorKbnAuthenticationYamlLoadedAccessControlTests.scala
```diff
@@ -0,0 +1,115 @@
...
+      |    - name: kbn3
+      |      signature_key: "1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890.1234567890"
+      |
+    """.stripMargin
+
+  "An ACL" when {
+    "is configured using config above" should {
+      "allow to proceed" when {
+        "JWT token is defined" in {
+          val claims = new DefaultClaims(Map[String, AnyRef]("sub" -> "test", "user" -> "user", "groups" -> "").asJava)
```
coutoPL: In tests, we have `JwtUtils`. Can we also use it here?

core/src/test/scala/tech/beshu/ror/integration/RorKbnAuthnAndAuthzYamlLoadedAccessControlTests.scala
coutoPL, line 1: 1. Can't we merge this test with RorKbnAuthenticationYamlLoadedAccessControlTests.scala? 
2. the same comments from RorKbnAuthenticationYamlLoadedAccessControlTests.scala apply also there

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/RorKbnAuthRuleTests.scala
```diff
@@ -417,24 +416,29 @@ class RorKbnAuthRuleTests
...
-                         configuredGroups: Groups,
+                         configuredGroups: Option[GroupsLogic],
                          tokenHeader: Header,
                          preferredGroupId: Option[GroupId],
                          blockContextAssertion: Option[BlockContext => Unit]) = {
-    val rule = new RorKbnAuthRule(RorKbnAuthRule.Settings(configuredRorKbnDef, configuredGroups), CaseSensitivity.Enabled)
+    val rule = configuredGroups match {
+      case Some(groupsLogic) =>
+        new RorKbnAuthRule(RorKbnAuthRule.Settings(configuredRorKbnDef, groupsLogic), CaseSensitivity.Enabled)
+      case None =>
```
coutoPL: IMO, we shouldn't allow having no configured groups in this test suite.
All the tests that have no groups should be moved to the `RorKbnAuthenticationRuleTests` suite.

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
mgoworko: Yes, it is as you noticed - the union type reflects the situation, where there is a legacy fallback to `RorKbnAuthenticationRule`. It is enforced by the type system. In the current implementation, all the types and core logic are type-invariant.I saw the union type too, and I experimented a bit with type covariance to get rid of it. But it would spread across the entire core module and there would still be some loose ends.

I honestly think, that it is a feature, not a bug or something to fix. We (developers) want to have fallback in our decoder. And Scala "sees" it on the type level and forces us to use union type to reflect that.

It is already limited only to decoding. It does not leak to other parts of the code. The typeclass instances ( `RuleName`, `VariableUsage` etc..) are used only by the decoder. And the type of the decoder is immediately erased (`optionalRuleDecoder.map(_.asInstanceOf[RuleDecoder[Rule]])`). The resulting, decoded rule is of a simple type, not a union type.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthRule.scala
mgoworko: Answered in the discussion below

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/users/LocalUsersContext.scala
mgoworko: Answered in the discussion below

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRuleDecoders.scala
mgoworko: done for all 3 decoders in this file

core/src/test/scala/tech/beshu/ror/integration/RorKbnAuthenticationYamlLoadedAccessControlTests.scala
mgoworko: I copied it from preexisting Auth rule tests. But yes, we can, done.

core/src/test/scala/tech/beshu/ror/integration/RorKbnAuthnAndAuthzYamlLoadedAccessControlTests.scala
mgoworko: 1. The config is different 
- In `RorKbnAuthentication...Tests` there is `ror_kbn_auth` in `users` section
- In `RorKbnAuthnAndAuthz...Tests` there are both `ror_kbn_authentication` and `ror_kbn_authorization` in `users` section
- the config is specific to the entire suite, not a single test
2. Comments applied

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/auth/RorKbnAuthRuleTests.scala
mgoworko: I added RorKbnAuthenticationRuleTests and RorKbnAuthorizationRuleTests - they were missing (though the logic was mostly tested - RorKbnAuthTests verified the common logic, and decoder tests some custom behaviors)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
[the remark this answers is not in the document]
mgoworko: I think that @coderabbitai was wrong about missing `VariableUsage[RorKbnAuthRule]`. In the decoder we use the union type (representing the fallback to authentication rule). And it is the only usage, so we do not need `VariableUsage[RorKbnAuthRule]`.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/RorKbnRuleHelper.scala
mgoworko: 1. I changed `object RorKbnRuleHelper` -> `trait BaseRorKbnRule`. It contains the entire logic, and the authn and authz rules just invoke the `processUsingJwtToken` method of that trait. The rules do not contain anything more than that.
2. Done

coutoPL: ok, it's better now. 
Nevertheless, I would do one more step. 

Currently, the base class assumes there is authentication and authorization. IMO, it should also provide base functionality but e.g, authentication is not a base functionality for the authz rule. 

I see that the common functionality for both (kbn authn and authz rules) is token data extraction. So I see two solutions here: 
1. the base class should have a method like this:
```
protected extractJwtToken[B <: BlockContext: BlockContextUpdater](blockContext: B, rorKbn: RorKbnDef): Either[RuleResult.Rejected[B], TokenData]
```
(where: type TokenData = (Jwt.Payload, ClaimSearchResult[User.Id], ClaimSearchResult[UniqueList[Group]], ClaimSearchResult[Header]))

Implementation of authorization and 

2. we could have an extension method:

```
implicit class KbnJwtDataExtractor(token: AuthorizationToken) extends AnyVal {
   def getTokenDataUsing(rorKbnDef: RorKbnDef): TokenData = ???
}
```

It seems that in the second approach we will need to repeat more code in the imlpementation classes.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
coutoPL: > We (developers) want to have fallback in our decoder. And Scala "sees" it on the type level and forces us to use union type to reflect that.

Yeah, it's good to express things in types, but IMO, in this case, we don't express what we think we express ... hmm, let me explain what I mean by that:

1. we have sth like that:
```
object RorKbnAuthRule {
  implicit case object Name extends RuleName[RorKbnAuthRule | RorKbnAuthenticationRule] {
    override val name = Rule.Name("ror_kbn_auth")
  }
}
```

Someone may ask: 

> We have here the Name definition for RorKbnAuthRule, but I saw the same name has also the RorKbnAuthenticationRule? Why? I've just seen this:
> 
> ```
> implicit case object Name extends RuleName[RorKbnAuthenticationRule] {
>     override val name = Rule.Name("ror_kbn_authentication")
>   }
> ```
> What the heck is the name for the RorKbnAuthenticationRule? 
> 

Yeah ... it can be defended, because indeed some special configuration of the `ror_kbn_auth` section creates RorKbnAuthenticationRule, not RorKbnAuthRule. But do I care about it at this point? Not really. It's a backward compatibility at the settings format level. We don't have to think about it here. That's why it's misleasing.

2. Another places:
```
implicit val rorKbnAuthRule: LocalUsersSupport[RorKbnAuthRule | RorKbnAuthenticationRule] = NotAvailableLocalUsers()
    implicit val rorKbnAuthenticationRule: LocalUsersSupport[RorKbnAuthenticationRule] = NotAvailableLocalUsers()
    implicit val rorKbnAuthorizationRule: LocalUsersSupport[RorKbnAuthorizationRule] = NotAvailableLocalUsers()
```

```
implicit val rorKbnAuthRule: VariableUsage[RorKbnAuthRule | RorKbnAuthenticationRule] = NotUsingVariable()
    implicit val rorKbnAuthenticationRule: VariableUsage[RorKbnAuthenticationRule] = NotUsingVariable()
    implicit val rorKbnAuthorizationRule: VariableUsage[RorKbnAuthorizationRule] = NotUsingVariable()
```

Here, it looks like an error. Someone may think: 

> Totally no idea why we have to do it like that. Why we have `RorKbnAuthenticationRule` in two places.

Obviously, the answer is: because types forced us to do it. But we ended up with sth that is hard to grasp. Maybe there is a problem with the types?

3. And, IMO, we finally reach the point where it, at first glance, looks good, but looking deeper shows us that not really:

```
extends RuleBaseDecoderWithoutAssociatedFields[RorKbnAuthRule | RorKbnAuthenticationRule] with Logging {

  override protected def decoder: Decoder[RuleDefinition[RorKbnAuthRule | RorKbnAuthenticationRule]]
```

When we take a look at `RuleBaseDecoderWithoutAssociatedFields` we will see sth like this:

```
abstract class RuleBaseDecoderWithoutAssociatedFields[T <: Rule : RuleName] extends RuleDecoder[T] {
```

```
sealed abstract class RuleDecoder[T <: Rule : RuleName] extends Decoder[RuleDecoder.Result[T]] {
```

We see that RuleDecoder produces one rule. So, when we write `RuleBaseDecoderWithoutAssociatedFields[RorKbnAuthRule | RorKbnAuthenticationRule]` we declare a decoder that produces sth that is `RorKbnAuthRule` or `RorKbnAuthenticationRule`. There is nothing like this in ROR. There is no rule that acts like `RorKbnAuthRule` or `RorKbnAuthenticationRule`. These are separate rules. 

And THIS IS IT. 

Ok, so how to solve it?
One of the solutions is to provide a new type of decoder:

```
sealed abstract class EitherRuleDecoder[R1 <: Rule : RuleName, R12<: Rule : RuleName](implicit ev: =:=[RuleName[R1], RuleName[R2])
  extends Decoder[Either[RuleDecoder.Result[R1], RuleDecoder.Result[R2]]]
```

At least, it looks like it would.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/BaseRorKbnRule.scala
```diff
@@ -0,0 +1,184 @@
...
+import tech.beshu.ror.accesscontrol.utils.ClaimsOps.ClaimSearchResult.{Found, NotFound}
+import tech.beshu.ror.implicits.*
+import tech.beshu.ror.utils.json.JsonPath
+import tech.beshu.ror.utils.uniquelist.{UniqueList, UniqueNonEmptyList}
+
+import scala.util.Try
+
+trait BaseRorKbnRule extends Logging {
+
+  def processUsingJwtToken[B <: BlockContext : BlockContextUpdater](blockContext: B,
```
coutoPL: protected?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthRule.scala
mgoworko: restored to `RorKbnAuthRule`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/users/LocalUsersContext.scala
mgoworko: restored to RorKbnAuthRule

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/variables/runtime/VariableContext.scala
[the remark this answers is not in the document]
mgoworko: restored to RorKbnAuthRule after second iteration of review

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
mgoworko: Ok, I get the rationale behind it and your explanation. I changed it - described below. I looked at it from a bit different perspective - for me the 
```
  implicit case object Name extends RuleName[RorKbnAuthRule | RorKbnAuthenticationRule] {
    override val name = Rule.Name("ror_kbn_auth")
  }
```
looked great, because it expressed on type level, that rule defined with name ror_kbn_auth could be interpreted as `RorKbnAuthRule`, but also as `RorKbnAuthenticationRule`. I guess that it is not expressive or clear enough, because it tells us nothing about `RorKbnAuthRule` being the main type, and `RorKbnAuthenticationRule` a fallback.


I changed the implementation of the `RorKbnAuthRuleDecoder`. It now extends  `RuleDecoder[Rule]` (generic `Rule`, without specific Rule type. That way the decoder just returns `RuleDefinition[Rule]` and we can return both types of rules - main and fallback. I did not create `EitherRuleDecoder` - it would essentially boil down to what I've done.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/base/RorKbnRuleHelper.scala
mgoworko: I've implemented something similar to option 1. The base trait now exposes a single method

```
protected def processUsingJwtToken[B <: BlockContext](blockContext: B,
                                                        rorKbnDef: RorKbnDef)
                                                       (operation: TokenData => Either[Unit, B]): RuleResult[B]
```

All commons are in the base KBN rule trait, and authenticate/authorize operations specifics are in the rules.

coutoPL (approved):
LGTM

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthenticationRule.scala
```diff
@@ -0,0 +1,85 @@
...
+                .withJwtToken(tokenPayload)
+            )
+          case ClaimSearchResult.NotFound =>
+            blockContext.withUserMetadata(
+              _
+                .withLoggedUser(DirectlyLoggedUser(userId))
+                .withJwtToken(tokenPayload)
+            )
+        }
+        Right(withUserMetadata)
```
mgoworko, line 70: Checking against groups in JWT token is removed in authn rule

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthorizationRule.scala
```diff
@@ -0,0 +1,93 @@
...
+              case Some(matchedGroups) if blockContext.isCurrentGroupEligible(GroupIds.from(matchedGroups)) =>
+                Right(blockContext.withUserMetadata(_.addAvailableGroups(matchedGroups)))
+              case Some(_) | None =>
+                Left(())
+            }
+          case None =>
+            Left(())
+        }
+      case (Found(groups), None) =>
+        UniqueNonEmptyList.from(groups) match {
```
mgoworko: and added as part of authz logic

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
```diff
@@ -0,0 +1,155 @@
...
+      .toSyncDecoder
+      .emapE { case (name, groupsLogicOpt) =>
+        val foundKbnDef = rorKbnDefinitions.items.find(_.id === name)
+        (foundKbnDef, groupsLogicOpt) match {
+          case (Some(rorKbnDef), groupsLogic) =>
+            groupsLogic match {
+              case Some(_) =>
+                ()
+              case None =>
+                logger.warn(s"There are no group mappings configured for rule ${name.value} of type ${RorKbnAuthRule.Name.name.show}. This syntax is deprecated, please change the rule type to ${RorKbnAuthenticationRule.Name.name.show}.")
```
mgoworko: It turns out, that `kbn_auth` rule cannot be simply replaced with `kbn_authentication` rule. It is not a simple fallback/replacement. 

There is a part of authorization logic (checking against groups in JWT token), that must still be executed by `kbn_auth` rule, even when the groups logic is not defined. RorKbnAuthYamlLoadedAccessControlTests verify that behavior

core/src/test/scala/tech/beshu/ror/integration/RorKbnAuthenticationYamlLoadedAccessControlTests.scala
```diff
@@ -95,6 +96,30 @@ class RorKbnAuthenticationYamlLoadedAccessControlTests
             }
           }
         }
+        "JWT token is defined with group that does not match the current group passed in header" in {
```
mgoworko, line 99: Added test - authentication rule should match in that situation.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthorizationRule.scala
```diff
@@ -0,0 +1,80 @@
...
+        }
+      case _ =>
+        RuleResult.Rejected()
+    }
+  }
+
+  private def authorize[B <: BlockContext : BlockContextUpdater](blockContext: B,
+                                                                 result: ClaimSearchResult[UniqueList[Group]],
+                                                                 groupsLogic: GroupsLogic) = {
+    (result, groupsLogic) match {
```
coutoPL: just `result match`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/RorKbnAuthorizationRule.scala
```diff
@@ -0,0 +1,80 @@
...
+      case (NotFound, _) =>
+        Left(())
+      case (Found(groups), groupsLogic) =>
+        UniqueNonEmptyList.from(groups) match {
```
coutoPL: Maybe instead of nested ifs we could use for-comprehension on Option and at the end change to Either?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
coutoPL: IMO, we should change this warning.
User wrote:
```
ror_kbn_auth:
   name: "kbn`
```

And we interpreted that as: `ror_kbn_auth` rule with any groups of [*]. 
We should clearly say it.
And we should add that if the user's intention was authentication only, they should use `ror_kbn_authentication` instead

mgoworko:
```
                logger.warn(
                  s"""There are no group mappings configured for rule ${name.value} of type ${RorKbnAuthRule.Name.name.show}.
                     |The rule is therefore interpreted as ${RorKbnAuthRule.Name.name.show} rule with `any_of: ["*"]` groups logic.
                     |This syntax is deprecated. Either explicitly define groups, or use ${RorKbnAuthenticationRule.Name.name.show} rule instead (when only authentication logic is required).
                     |""".stripMargin
                )
```

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
```diff
@@ -0,0 +1,164 @@
...
+      .toSyncDecoder
+      .emapE { case (name, groupsLogicOpt) =>
+        val foundKbnDef = rorKbnDefinitions.items.find(_.id === name)
+        (foundKbnDef, groupsLogicOpt) match {
+          case (Some(rorKbnDef), Some(groupsLogic)) =>
+            val settings = RorKbnAuthorizationRule.Settings(rorKbnDef, groupsLogic)
+            val rule = new RorKbnAuthorizationRule(settings)
+            Right(RuleDefinition.create[RorKbnAuthorizationRule](rule))
+          case (Some(_), None) =>
+            Left(RulesLevelCreationError(Message(s"Cannot create ${RorKbnAuthorizationRule.Name.name.show} - missing groups settings")))
```
coutoPL: Maybe "missing groups logic"? And maybe it'd be good to add this link to the doc too?

https://github.com/beshu-tech/readonlyrest-docs/blob/master/details/authorization-rules-details.md#checking-groups-logic

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
coutoPL: maybe sth like this?

```
s"""|
   |Missing groups logic settings in ${RorKbnAuthRule.Name.name.show} rule.
   |For old configs, ROR treats this as `groups_any_of: ["*"]`.
   |This syntax is deprecated. Add groups logic ([LINK_TO_PROPER_SECTION_IN_DOCS]), 
   |or use ${RorKbnAuthenticationRule.Name.name.show} if you only need authentication.
   |""".stripMargin
```

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/rules/auth/RorKbnRulesDecoders.scala
mgoworko: Added this link in both places. Is it correct, that this section is not published on docs.readonlyrest.com ? (At least I couldn't find it)

coutoPL: Probably not. During adding the new rules descriptions, could you please check why?

mgoworko: On docs.readonlyrest.com the `authorization-rules-details.md` document is linked as redirection to Github repo blob, not a document exposed on the website. I'm not sure if it is correct, but the links are in this PR are done the same way, so it is consistent.

PR #1165, opened 2025-10-01
[RORDEV-1567] ECS serializer and improved configurable serializer

mgoworko:
🚀New (ES) Audit log serializer compliant with Elastic Common Schema (ECS)
🧐Enhancement (ES) Enable nested field definitions in the configurable audit log serializer for more flexible audit logging

## Summary by CodeRabbit

* **New Features**
  * Added ECS v1 audit serializer and ECS-aware configurable serializer (versioned).
  * Supports nested field paths and persists "@timestamp".
  * New audit value types: boolean, numeric, processing-duration (ms & nanos), and event outcome.
  * Structured final-state values: Allowed, Forbidden, Errored, IndexNotExist.

* **Tests**
  * Added unit and integration tests covering ECS output and nested configurable fields.

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
@@ -40,6 +40,7 @@ import tech.beshu.ror.implicits.*
 class Block(val name: Name,
             val policy: Policy,
             val verbosity: Verbosity,
+            val audit: Audit,
             val rules: NonEmptyList[Rule])
            (implicit val loggingContext: LoggingContext)
   extends Logging {
@@ -101,12 +102,13 @@ object Block {
   def createFrom(name: Name,
                  policy: Option[Policy],
                  verbosity: Option[Verbosity],
+                 audit: Option[Audit],
                  rules: NonEmptyList[RuleDefinition[Rule]])
                 (implicit loggingContext: LoggingContext): Either[BlocksLevelCreationError, Block] = {
     val sortedRules = rules.sorted
     BlockValidator.validate(name, sortedRules) match {
       case Validated.Valid(_) =>
-        Right(createBlockInstance(name, policy, verbosity, sortedRules))
+        Right(createBlockInstance(name, policy, verbosity, audit, sortedRules))
       case Validated.Invalid(errors) =>
         implicit val validationErrorShow: Show[BlockValidationError] = blockValidationErrorShow(name)
         Left(BlocksLevelCreationError(Message(errors.toList.map(_.show).mkString("\n"))))
@@ -116,12 +118,14 @@ object Block {
   private def createBlockInstance(name: Name,
                                   policy: Option[Policy],
                                   verbosity: Option[Verbosity],
+                                  audit: Option[Audit],
                                   rules: NonEmptyList[RuleDefinition[Rule]])
                                  (implicit loggingContext: LoggingContext) =
     new Block(
       name,
       policy.getOrElse(Block.Policy.Allow),
       verbosity.getOrElse(Block.Verbosity.Info),
+      audit.getOrElse(Block.Audit.Enabled),
       rules.map(_.rule)
     )
 
@@ -179,6 +183,16 @@ object Block {
     case object Error extends Verbosity
 
     implicit val eq: Eq[Verbosity] = Eq.fromUniversalEquals
+  }
+
+  sealed trait Audit
+
+  object Audit {
+    case object Enabled extends Audit
+
+    case object Disabled extends Audit
+
+    implicit val eq: Eq[Audit] = Eq.fromUniversalEquals
   }
 
   private class Lifter[B <: BlockContext] {
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/factory/RawRorConfigBasedCoreFactory.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/factory/RawRorConfigBasedCoreFactory.scala
@@ -220,14 +220,22 @@ class RawRorConfigBasedCoreFactory(esVersion: EsVersion)
         case "info" => Right(Verbosity.Info)
         case "error" => Right(Verbosity.Error)
         case unknown => Left(BlocksLevelCreationError(Message(s"Unknown verbosity value: ${unknown.show}. Supported types: 'info'(default), 'error'.")))
+      }.decoder
+    implicit val blockAuditDecoder: Decoder[Block.Audit] =
+      Decoder.instance { c =>
+        for {
+          enabled <- c.downField("enabled").as[Boolean]
+        } yield {
+          if (enabled) Block.Audit.Enabled else Block.Audit.Disabled
+        }
       }
-        .decoder
     Decoder
       .instance { c =>
         val result = for {
           name <- c.downField(Attributes.Block.name).as[Block.Name]
           policy <- c.downField(Attributes.Block.policy).as[Option[Block.Policy]]
           verbosity <- c.downField(Attributes.Block.verbosity).as[Option[Block.Verbosity]]
+          audit <- c.downField(Attributes.Block.audit).as[Option[Block.Audit]]
           rules <- rulesNelDecoder(definitions, globalSettings, mocksProvider)
             .toSyncDecoder
             .decoder
@@ -235,9 +243,11 @@ class RawRorConfigBasedCoreFactory(esVersion: EsVersion)
               _.mapObject(_
                 .remove(Attributes.Block.name)
                 .remove(Attributes.Block.policy)
-                .remove(Attributes.Block.verbosity))
+                .remove(Attributes.Block.verbosity)
+                .remove(Attributes.Block.audit)
+              )
             ))
-          block <- Block.createFrom(name, policy, verbosity, rules).left.map(DecodingFailureOps.fromError(_))
+          block <- Block.createFrom(name, policy, verbosity, audit, rules).left.map(DecodingFailureOps.fromError(_))
         } yield BlockDecodingResult(
           block = block,
           localUsers = rules.map(localUsersForRule).combineAll,
@@ -541,6 +551,7 @@ object RawRorConfigBasedCoreFactory {
       val name = "name"
       val policy = "type"
       val verbosity = "verbosity"
+      val audit = "audit"
     }
 
   }
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AccessControlListLoggingDecorator.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/logging/AccessControlListLoggingDecorator.scala
@@ -105,14 +105,30 @@ class AccessControlListLoggingDecorator(val underlying: AccessControlList,
       import tech.beshu.ror.accesscontrol.logging.AccessControlListLoggingDecorator.responseContextShow
       logger.info(responseContextShow[B].show(responseContext))
     }
-    auditingTool.foreach {
-      _
-        .audit(responseContext, auditEnvironmentContext)
-        .runAsync {
-          case Right(_) =>
-          case Left(ex) =>
-            logger.warn(s"[${responseContext.requestContext.id.show}] Auditing issue", ex)
+    blockAuditSettings(responseContext) match {
+      case Some(Block.Audit.Disabled) =>
+        ()
+      case None | Some(Block.Audit.Enabled) =>
+        auditingTool.foreach {
+          _
+            .audit(responseContext, auditEnvironmentContext)
+            .runAsync {
+              case Right(_) =>
+              case Left(ex) =>
+                logger.warn(s"[${responseContext.requestContext.id.show}] Auditing issue", ex)
+            }
         }
+    }
+  }
+
+  private def blockAuditSettings[B <: BlockContext](responseContext: ResponseContext[B]): Option[Block.Audit] = {
+    responseContext match {
+      case AllowedBy(_, block, _, _) => Some(block.audit)
+      case Allow(_, _, block, _) => Some(block.audit)
+      case ForbiddenBy(_, block, _, _) => Some(block.audit)
+      case Forbidden(_, _) => None
+      case RequestedIndexNotExist(_, _) => None
+      case Errored(_, _) => None
     }
   }
 
--- a/core/src/test/scala/tech/beshu/ror/unit/acl/EnabledAccessControlListTests.scala
+++ b/core/src/test/scala/tech/beshu/ror/unit/acl/EnabledAccessControlListTests.scala
@@ -176,6 +176,7 @@ class EnabledAccessControlListTests extends AnyWordSpec with MockFactory with In
       Block.Name(name),
       policy,
       Block.Verbosity.Info,
+      Block.Audit.Enabled,
       NonEmptyList.of(
         new RegularRule {
           override val name: Rule.Name = Rule.Name("auth")
--- a/core/src/test/scala/tech/beshu/ror/unit/acl/blocks/BlockTests.scala
+++ b/core/src/test/scala/tech/beshu/ror/unit/acl/blocks/BlockTests.scala
@@ -55,6 +55,7 @@ class BlockTests extends AnyWordSpec with BlockContextAssertion with Inside with
           name = blockName,
           policy = Block.Policy.Allow,
           verbosity = Block.Verbosity.Info,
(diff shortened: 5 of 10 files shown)


core/src/main/scala/tech/beshu/ror/accesscontrol/audit/ecs/EcsV1AuditLogSerializer.scala
```diff
@@ -0,0 +1,83 @@
...
+class EcsV1AuditLogSerializer(val allowedEventMode: AllowedEventMode) extends AuditLogSerializer {
+
+  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] = {
+    AuditSerializationHelper.serialize(responseContext, fields, allowedEventMode)
+  }
+
+}
+
+object EcsV1AuditLogSerializer {
+  private val fields: Map[AuditFieldName, AuditFieldValueDescriptor] = Map(
```
mgoworko: The main problem with adding ECS serializer, is that it is highly permissive, without strict guidelines. Some of our field names can be adapted to ECS in multiple ways.

This serializer, implemented in codebase, is a base, ECS-compliant implementation, that should be sufficient. But at the same way, I think that users should be able to define their own, as `configurable` serializer. I added possibility of configuring nested fields (ECS requires support for nested fields) and numeric/boolean fields.

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
```diff
@@ -223,8 +236,18 @@ private[ror] object AuditSerializationHelper {
 
     final case class StaticText(value: String) extends AuditFieldValueDescriptor
 
+    final case class BooleanValue(value: Boolean) extends AuditFieldValueDescriptor
+
+    final case class NumericValue(value: BigDecimal) extends AuditFieldValueDescriptor
+
     final case class Combined(values: List[AuditFieldValueDescriptor]) extends AuditFieldValueDescriptor
 
+    final case class Nested(values: Map[AuditFieldName, AuditFieldValueDescriptor]) extends AuditFieldValueDescriptor
```
coutoPL: It's one level of nesting. Let's do it in a more generic way (to allow many nesting levels). 

BTW. We define fields like that:
```
fields: Map[AuditFieldName, AuditFieldValueDescriptor]
```

So, maybe we should not add Nested as a descriptor, but maybe we should modify the `AuditFieldName` definition instead.

E.g.
`AuditFieldName` is:
* a string OR
*  is a composition of `AuditFieldName`s (or strings)?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/AuditFieldValueDescriptorParser.scala
```diff
@@ -27,7 +27,7 @@ object AuditFieldValueDescriptorParser {
   private val key = P.charsWhile(c => c != '{' && c != '}')
 
   private val placeholder: P[Either[String, AuditFieldValueDescriptor]] =
-    (lbrace *> key <* rbrace).map(k => deserializerAuditFieldValueDescriptor(k.trim.toUpperCase).toRight(k))
+    (lbrace *> key <* rbrace).map(k => deserializerAuditFieldValueDescriptor(k.trim).toRight(k))
```
coutoPL, line 31: Now, are the fields case sensitive?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/ecs/EcsV1AuditLogSerializer.scala
```diff
@@ -0,0 +1,86 @@
...
+    AuditFieldName("ecs") -> AuditFieldValueDescriptor.Nested(
+      // Schema defined by EcsV1AuditLogSerializer is ECS 1.4.0 compliant and does not use newer features introduced by later versions
```
coutoPL: Where can I find what's inside the ECS 1.4.0? Could you please link a source?

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
coutoPL: oh, I see I was wrong about the nesting.
Nevertheless, let's consider the change above.

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
```diff
@@ -1956,6 +2064,10 @@ class AuditSettingsTests extends AnyWordSpec with Inside {
     }
   }
 
+  def circeJsonWithIgnoredTimestamp(json: Option[JSONObject]): Option[Json] = {
```
coutoPL: hmm, I think we need here instead two test util methods:
1. Convert JsonObject to Circe Json
2. extension to ignore fields in Circe JSON

Using these two, we could implement this one. 
And let's make the `circeJsonWithIgnoredTimestamp` method private

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/configurable/AuditFieldValueDescriptorParser.scala
mgoworko: This part of code is about placeholders (values like `{LOGGED_USER}` in the field value, that are resolved in runtime). This parser only parses raw value, and passed it to the `deserializerAuditFieldValueDescriptor` method. There was doubled `.toUpperCase`. I left it in the implementation

```
  private def deserializerAuditFieldValueDescriptor(str: String): Option[AuditFieldValueDescriptor] = {
    str.toUpperCase match {
```
So I removed it here, because it was doubled, and the better place for it is in the method.

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
mgoworko: Actually that was my first version of implementation - the field name as list of strings, representing nesting. I have it in git history, so I can do it with no problems if we so decide. But I honestly think that the current version is better, because:
- it represents 1:1 the structure of the JSON, so is easier to grasp
- it forces to defined all fields inside the section, which improves readability
- it is especially better when user defines "configurable" serializer:

Example from tests:
```
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: log
      serializer:
        type: configurable
        verbosity_level_serialization_mode: [INFO]
        fields:
          custom_section:
            nested_text: "nt"
            nested_number: 123
            nested_boolean: true
            double_nested:
              double_nested_next: "dnt"
              triple_nested:
                triple_nested_next: "tnt"
          node_name_with_static_suffix: "{ES_NODE_NAME} with suffix"
          another_field: "{ES_CLUSTER_NAME} {HTTP_METHOD}"
          tid: "{TASK_ID}"
          bytes: "{CONTENT_LENGTH_IN_BYTES}"
```

With the approach based on the list of strings, it would look like that:
```
fields:
  - path: ["custom_section", "nested_text"]
    value: "nt"

  - path: ["custom_section", "nested_number"]
    value: 123

  - path: ["custom_section", "nested_boolean"]
    value: true

  - path: ["custom_section", "double_nested", "double_nested_next"]
    value: "dnt"

  - path: ["custom_section", "double_nested", "triple_nested", "triple_nested_next"]
    value: "tnt"

  - path: ["node_name_with_static_suffix"]
    value: "{ES_NODE_NAME} with suffix"

  - path: ["another_field"]
    value: "{ES_CLUSTER_NAME} {HTTP_METHOD}"

  - path: ["tid"]
    value: "{TASK_ID}"

  - path: ["bytes"]
    value: "{CONTENT_LENGTH_IN_BYTES}"
```

I think that the nested approach is more readable in the code too.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/ecs/EcsV1AuditLogSerializer.scala
mgoworko: Added link to the ECS docs in the comment. After one of AI comments I used ECS 1.6 specification with `action` field

But as I wrote in other comment - ECS schema is extremely flexible and permissive. I'm guessing, that the ability to define nested fields and configure own ECS serializer in the config (as `configurable` serializer) may be more useful to users than the implementation provided by us. 

That is because there is no guarantee, that the field mappings provided by us will be fully compatible with some external tools the user might want to use to digest the ECS data.There seem to be a number of issues and confusion over what the correct format of ECS data is. And the answer is, that there is no single correct field mapping, "it depends" :)

`https://github.com/elastic/ecs/issues/516`

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/ecs/EcsV1AuditLogSerializer.scala
```diff
@@ -0,0 +1,86 @@
...
+        AuditFieldName("name") -> AuditFieldValueDescriptor.ImpersonatedByUser,
+      ),
+    ),
+    AuditFieldName("event") -> AuditFieldValueDescriptor.Nested(
+      AuditFieldName("id") -> AuditFieldValueDescriptor.Id,
+      AuditFieldName("action") -> AuditFieldValueDescriptor.Action,
+      AuditFieldName("type") -> AuditFieldValueDescriptor.Type,
+      AuditFieldName("reason") -> AuditFieldValueDescriptor.FinalState,
+      AuditFieldName("duration") -> AuditFieldValueDescriptor.ProcessingDurationNanos,
+    ),
```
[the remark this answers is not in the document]
mgoworko: I refactored the `action` section and used ECS 1.6 specification

```
 AuditFieldName("event") -> AuditFieldValueDescriptor.Nested(
      AuditFieldName("id") -> AuditFieldValueDescriptor.Id,
      AuditFieldName("dataset") -> AuditFieldValueDescriptor.Action, // ROR Action describes the resource being accessed, so it is dataset in ECS
      AuditFieldName("action") -> AuditFieldValueDescriptor.Type, // ROR Type describes operation that is executed, so it is ECS action
      AuditFieldName("reason") -> AuditFieldValueDescriptor.FinalState, // ROR Final state described the outcome, so it is the ECS reason of event happening, the ROR Reason is a custom detailed_reason label
      AuditFieldName("duration") -> AuditFieldValueDescriptor.ProcessingDurationNanos,
    ),
```

The ROR "reason" is now one of custom ECS labels:
```
    AuditFieldName("labels") -> AuditFieldValueDescriptor.Nested(
      ...
      AuditFieldName("detailed_reason") -> AuditFieldValueDescriptor.Reason,
    ),
```

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
coutoPL: I'm not sure if I agree. I see it as follows:
1. we have some representation of the fields and value descriptors in the ROR's YAML
2. then we have our internal (domain) representation of it
3. finally, we serialize the representation to the ES index document

And the only thing that is JSON here is dthe ocument. 
The internal representation is not a JSON, and the "fields" representation in ROR's YAML is not a JSON either (we don't support arrays).

But I don't want you to change the YAML's representation. It's fine. But obivously we should detect invalid settings like:

1. 

```yaml
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: log
      serializer:
        type: configurable
        verbosity_level_serialization_mode: [INFO]
        fields: 1 # or string, or bool, or array
```

2. 

```yaml
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: log
      serializer:
        type: configurable
        verbosity_level_serialization_mode: [INFO]
        fields:
          custom_section:
            - sth: 1
            - sth: 2
```

3. 

```yaml
readonlyrest:
  audit:
    enabled: true
    outputs:
    - type: log
      serializer:
        type: configurable
        verbosity_level_serialization_mode: [INFO]
        fields:
          custom_section: {}
          custom_section2: null
```

So, when we take into consideration the limitations, we could say that we have a map of simplified field path keys (without arrays) and value descriptor leafs. 

In code, we represent it as `Map[AuditFieldName, AuditFieldValueDescriptor]`, where 
* `AuditFieldValueDescriptor` is always a leaf (of type string, int, bool ... maybe null?!) 
* `AuditFieldName` is a path to this leaf (e.g., `custom_section.double_nested.double_nested_next` or `node_name_with_static_suffix`)

That's why I propose to define `AuditFieldName` (in the domain ... maybe we should rename it to AuditFieldPath) like this:

`type AuditFieldPath = NonEmptyList[String]`.

And `AuditFieldValueDescriptor` will be free of sth like `Nested`.

WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/ecs/EcsV1AuditLogSerializer.scala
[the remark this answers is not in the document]
coutoPL: @mgoworko is it ok now?

coutoPL (commented):
one comment left

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
mgoworko: I did it as suggested:
- there is now `AuditFieldPath`, that stores `NonEmptyList[String]` 
- AuditFieldValueDescriptor.Nested is removed
- the YAML representation is not changed - so the structure in yaml is nested, but it is decoded to flat Map[AuditFieldPath, AuditFieldValueDescriptor]

It required a bit of boilerplate and utils (because the structure in config is no longer 1:1 mirror of the domain representation in code). Invalid settings are detected. Cases 1 and 3 were already handled before review, case 2 is now handled when converting from nested structure in yaml to domain representation.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/ecs/EcsV1AuditLogSerializer.scala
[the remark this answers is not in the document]
mgoworko: As I wrote in earlier discussion, there is no single correct solution. The ECS schema does not fit our use case perfectly and is very ambiguous. But I made some adjustments based on that comment:
1. The ROR `FinalState`:
- it was previously the `event.reason` ECS field
- I put it as `labels.ror_final_state` - custom field
- I started to use `event.outcome` - it can be only `success`/`failure`/`unknown` - I map the `FinalState` to those 3 values
2. The ROR action is mapped as ECS field `event.action`
3. The ROR type (the request class) is not `event.reason`

audit/src/main/scala/tech/beshu/ror/audit/utils/AuditSerializationHelper.scala
```diff
@@ -153,7 +194,41 @@ private[ror] object AuditSerializationHelper {
     final case class Include(types: Set[Verbosity]) extends AllowedEventMode
   }
 
-  final case class AuditFieldName(value: String)
+  final case class AuditFieldPath(path: NonEmptyList[String])
```
coutoPL: Don't add additional dependencies (especially from the Scala's world) to the audit module. 

We can live without NonEmptyList here.
Just close constructor and add proper smart constructor/s.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/decoders/AuditingSettingsDecoder.scala
```diff
@@ -337,15 +375,62 @@ object AuditingSettingsDecoder extends Logging {
...
+          .map(pairs => DecodedAuditFieldValue.Nested(pairs.toMap))
+    )
   }
 
-  given auditFieldValueDecoder: Decoder[AuditFieldValueDescriptor] = {
-    SyncDecoderCreator
-      .from(Decoder.decodeString)
-      .emap(AuditFieldValueDescriptorParser.parse)
-      .decoder
+  private sealed trait DecodedAuditFieldValue
```
coutoPL: I think we need a better names here. These are misleading.
AFAIU, it's some intermediate structures. 
I wonder if we couldn't use the final stuctures here instead.

mgoworko: I refactored it to remove the intermediate type. The entire logic of decoding nested fields is now in the decoder implementation.

PR #1222, opened 2026-03-15
[RORDEV-1861] Failover mode and cluster health check added for remote audit clusters

mateuszkp96:
```yaml
- type: new
  components: [es]
  text: "Audit: a remote audit cluster can now run in `failover` mode, and ROR checks the cluster before it starts auditing."
```

## Summary

This is an addition to RORDEV-2190. It gives the remote audit cluster two things: a `failover` mode, and a connectivity check at startup.

Before this change, a remote audit cluster had only one mode: `round-robin`. ROR sent audit events to all nodes in turn. If a node was down, the ES `RestClient` marked it dead and kept it dead with a growing backoff. During that time ROR could drop audit events. ROR also accepted any cluster settings without a check. A typo in a node URL, or nodes from two different clusters, was only found later in the logs.

## Failover mode

In `failover` mode ROR uses one client per node. It sends every audit event to the first node. It moves to the next node only when the current node fails.

- A failure that can be retried (an `IOException`, or a 502/503/504 response) opens the circuit of that node. ROR then tries the next node.
- A fatal failure (any other response error) stops the request. ROR does not try the other nodes, because the other nodes will answer the same.
- An open circuit keeps a node out of the selection for `1000 * 1.5^(failures-1)` milliseconds, up to 30 minutes. This is the same upper limit as the ES `RestClient` uses.
- A successful request closes the circuit of that node.
- When all circuits are open, ROR tries the node whose circuit opens again first. This keeps the client usable instead of dropping the event.

`round-robin` stays the default. It is unchanged.

## Connectivity check

When ROR creates an audit output for a remote cluster, it now sends `GET /` to every configured node, in parallel, with three retries. The results decide the outcome:

| Result | Outcome |
|---|---|
| No node answers | Settings are rejected |
| Nodes answer, but with different `cluster_uuid` | Settings are rejected |
| Some nodes answer, all from the same cluster | ROR logs a warning and starts |
| All nodes answer, all from the same cluster | ROR starts |

Two different clusters in one audit output is always an error. One audit output can write to one cluster only.

The new `ignore_es_connectivity_problems` flag lets you accept an unreachable cluster:

```yaml
readonlyrest:
  audit:
    enabled: true
    outputs:
      - type: index
        cluster:
          nodes: [ "http://audit1:9200", "http://audit2:9200" ]
          mode: failover                          # or round-robin (default)
          ignore_es_connectivity_problems: false  # default
```

The flag covers a connectivity problem only. It does not cover a wrong cluster. Note also that a `data_stream` output must read or create the data stream when it starts, so that output still fails when no node answers.

## Tests

- `FailoverClientTests` — 15 unit tests for the failover client: node selection, the two failure classes, circuit timing, the 30-minute limit, and the all-circuits-open case.
- `ReadonlyRestStartingTests` — startup with mixed clusters, with all nodes down, and with `ignore_es_connectivity_problems` on. Three WireMock stubs give two nodes of one cluster and one node of another.
- `AuditingConfigTests` — decoding of `mode: failover` and of `ignore_es_connectivity_problems`.
- `RemoteClusterAuditingToolsSuite` — rewritten. It now tests both modes against Toxiproxy, and the settings reload with the nodes down. Each test forces a fresh engine, so no circuit state moves between tests.

## Scope

- Core: the failover client, the connectivity check, the settings decoder, and the audit output creation.
- All 34 `es*x` modules: the same four files in each — the failover client build-up, the retryable-response test, and the `protected` creator methods.

## For the reviewer

- `IndexBasedAuditOutputServiceCreator.index` and `DataStreamBasedAuditOutputServiceCreator.dataStream` are now `protected`. Callers use `createIndexService` / `createDataStreamService`, which run the check first. This makes it impossible to build an output that skips the check.
- The connectivity check runs on every settings reload, so a reload now costs one HTTP round trip per audit node.
- The docs need a new page section for `mode: failover` and `ignore_es_connectivity_problems`. The error message already points to `docs.readonlyrest.com/elasticsearch/audit#custom-audit-cluster`.

## Summary by CodeRabbit

* **New Features**
  * Added support for remote audit clusters with round-robin and failover behavior.
  * Added a connectivity check for remote audit setups, with an option to continue when problems are ignored.
  * Improved audit output handling for both index-based and data-stream-based configurations.

* **Bug Fixes**
  * Remote audit connections now handle partial node outages more gracefully.
  * Retry behavior was improved for temporary Elasticsearch gateway errors.


mateuszkp96:
@coderabbitai full review

es79x/src/main/scala/tech/beshu/ror/es/services/RestClientAuditSinkService.scala
```diff
@@ -96,34 +101,54 @@ final class RestClientAuditSinkService private(client: RestClient, inFlightReque
...
-          client = createRestClient(remoteCluster),
-          inFlightRequestSemaphore = new Semaphore(remoteCluster.maxInflightRequests)
-        )
+        val restClient = createRestClient(remoteCluster, hosts)
+        val clusterAwareClient = new RoundRobinClient(restClient)
+        createService(remoteCluster, clusterAwareClient, hosts)
+      case ClusterMode.Failover =>
+        val clientsPerNode = hosts.map(host => createRestClient(remoteCluster, NonEmptyList.one(host)))
+        val clusterAwareClient = FailoverClient.create(clientsPerNode)
+        createService(remoteCluster, clusterAwareClient, hosts)
```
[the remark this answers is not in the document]
mateuszkp96: HttpHost has no option to set a path. In boot tests, different paths were needed for WireMock.

es810x/src/main/scala/tech/beshu/ror/es/services/RestClientAuditSinkService.scala
```diff
@@ -96,34 +101,54 @@ final class RestClientAuditSinkService private(client: RestClient, inFlightReque
...
+    hosts
+      .map { host =>
+        Resource.make(Task.delay(createRestClient(remoteCluster, NonEmptyList.one(host))))(client => Task.delay(client.close()))
+      }
+      .sequence
+      .map {
+        _.map(client => new RestClientDataStreamService(client))
+      }
+      .map(AuditDataStreamCreator.remote(_, remoteCluster.ignoreClusterConnectivityProblems))
+  }
```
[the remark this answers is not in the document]
mateuszkp96: DataStreamCreator attempts to set up a data stream on each node, and this should be done regardless of cluster mode (to write a document to the data stream, it needs to be set up before)

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+
+final class AuditRemoteClusterHealthcheck(httpClientsFactory: HttpClientsFactory) extends RequestIdAwareLogging {
+
+  import AuditRemoteClusterHealthcheck.*
+  import AuditRemoteClusterHealthcheck.given Show[AuditClusterNode]
+
+  def check(cluster: RemoteAuditCluster): Task[Either[Error, Unit]] = {
+    implicit val requestId: RequestId = RequestId(UUID.randomUUID().toString)
+    val auditClusterNodes = cluster.nodes.toNonEmptyList
+    createHttpClient(auditClusterNodes)
```
coutoPL: I wonder if we can create the HTTP client once. I see its pool is related to the node count, but do we need to set it? I also see that the check method is used only at the ROR startup, but IMO, we may do it for the sake of clarity (in case of any future reuse of this class).

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+      .use { httpClient =>
+        Task.parTraverseUnordered(auditClusterNodes.toList.zipWithIndex) { case (node, idx) =>
+          withRetries(getNodeInfo(cluster, httpClient, node))
+            .map(response => (response, idx))
+        }
+      }
+      .map { auditNodesInfo =>
+        val (maybeConnectionErrors, maybeAuditNodeInfos) = {
+          auditNodesInfo
```
coutoPL: Maybe like this?
```
auditNodesInfo
            // sorting to have deterministic ordering of errors
            .sortBy((_, idx) => idx)
            .partitionMap((response, _) => response)
            .bimap(NonEmptyList.fromList, NonEmptyList.fromList)
```

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+
+  import AuditRemoteClusterHealthcheck.*
+  import AuditRemoteClusterHealthcheck.given Show[AuditClusterNode]
+
+  def check(cluster: RemoteAuditCluster): Task[Either[Error, Unit]] = {
+    implicit val requestId: RequestId = RequestId(UUID.randomUUID().toString)
+    val auditClusterNodes = cluster.nodes.toNonEmptyList
+    createHttpClient(auditClusterNodes)
+      .use { httpClient =>
+        Task.parTraverseUnordered(auditClusterNodes.toList.zipWithIndex) { case (node, idx) =>
```
coutoPL: let's move this code to well name private method

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+          withRetries(getNodeInfo(cluster, httpClient, node))
+            .map(response => (response, idx))
+        }
+      }
+      .map { auditNodesInfo =>
+        val (maybeConnectionErrors, maybeAuditNodeInfos) = {
```
coutoPL: let's move this code to well name private method

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+            }
+            .partitionMap {
+              case (response, idx) => response
+            }
+            .bimap(NonEmptyList.fromList, NonEmptyList.fromList)
+        }
+        val detailedConnectionErrors = maybeConnectionErrors.map(_.toList.map(_.show).show)
+        (for {
+          auditNodeInfos <- maybeAuditNodeInfos.toRight(s"No health node detected in remote cluster. ${detailedConnectionErrors.get}")
+          _ <- detailedConnectionErrors.toLeft(())
```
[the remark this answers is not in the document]
coutoPL: @mateuszkp96, this is a valid point.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+      case None => url.show
+    }
+  }
+
+  private final case class RetryConfig(initialDelay: FiniteDuration, backoffScaler: Int, maxRetries: Int)
+
+  private final case class AuditNodeInfo(node: AuditClusterNode, info: ClusterInfoResponse)
+
+  private final case class ClusterInfoResponse(nodeName: String,
+                                       clusterName: String,
```
coutoPL: formatting

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
```diff
@@ -0,0 +1,187 @@
...
+import tech.beshu.ror.utils.RefinedUtils.positiveFiniteDuration
+import tech.beshu.ror.utils.RequestIdAwareLogging
+import tech.beshu.ror.utils.ScalaOps.retryBackoffEither
+
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+import scala.util.control.NonFatal
+
+final class AuditRemoteClusterHealthcheck(httpClientsFactory: HttpClientsFactory) extends RequestIdAwareLogging {
```
coutoPL: This is more like `AuditRemoteClusterConnectivityCheck` ... or maybe we should leave the name as it is, but change the endpoint to `/_cluster/health`?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/AuditDataStreamCreator.scala
```diff
@@ -45,13 +46,18 @@ final class AuditDataStreamCreator(services: NonEmptyList[DataStreamService]) ex
...
+          Task.delay(noRequestIdLogger.info(s"Data stream ${dataStreamName.dataStream.show} already exists."))
             .as(Valid(()))
-        case false =>
+        case Right(false) =>
           val settings = defaultSettingsFor(dataStreamName.dataStream)
           setupDataStream(service, settings)
+        case Left(ex) =>
+          val errorMessage = s"Unable to determine if data stream ${dataStreamName.dataStream.show} exists."
+          Task.delay(noRequestIdLogger.info(errorMessage, ex))
+            .as(Validated.cond(ignoreEsConnectivityProblems, (), ErrorMessage(errorMessage)))
```
coutoPL: It's suspicious here. 
I mean, we don't use the "ignoreEsConnectivityProblems" in this file, but we assume that the exception (ex) is related to connectivity problems. 

It looks like it should be done on different level or sth.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
```diff
@@ -299,17 +313,29 @@ object AuditingTool extends RequestIdAwareLogging {
...
         EsDataStreamBasedAuditSink.create(
           config.logSerializer,
           config.rorAuditDataStream,
           auditSinkService,
           config.auditCluster
-        ).map(_.leftMap(error => CreationError(error.message)).toValidated)
+        ).map(_.leftMap(error => CreationError(error.message)))
       }
 
+  private def optionalClusterHealthCheck(cluster: AuditCluster,
```
coutoPL: I wonder if the cluster health check shouldn't be a part of the cluster creation. Because at this level, we shouldn't be bothered by the health checking. The problem I've just mentioned is highlighted by two places where we use this method - it's easy to forget to add it.

core/src/test/scala/tech/beshu/ror/unit/es/services/DataStreamServiceTest.scala
```diff
@@ -541,8 +541,8 @@ class DataStreamServiceTest
     service
   }
 
-  extension (c: AuditDataStreamCreator.type) {
-    def create(service: DataStreamService): AuditDataStreamCreator = AuditDataStreamCreator(NonEmptyList.of(service))
+  extension (c : AuditDataStreamCreator.type) {
+    def create(service: DataStreamService): AuditDataStreamCreator = AuditDataStreamCreator(NonEmptyList.of(service), true)
```
coutoPL: Please use named args here

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterHealthcheck.scala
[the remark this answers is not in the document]
coutoPL: Maybe, when `ignore_es_connectivity_problems: true` we should still check the connectivity but instead of falling, we could write a warn log?
And when `ignore_es_connectivity_problems: false` (default) we could add to the log info that the connectivity check can be disabled.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/AuditSinkServiceCreator.scala
```diff
@@ -16,22 +16,68 @@
...
+      case AuditCluster.LocalAuditCluster =>
+        create.map(Right(_))
+      case remote: RemoteAuditCluster =>
+        new AuditRemoteClusterConnectivityCheck(httpClientsFactory)
+          .check(remote)
+          .flatMap {
+            case Right(()) =>
+              create.map(Right(_))
+            case Left(error: AuditRemoteClusterConnectivityCheck.Error.ConnectivityError) if remote.ignoreClusterConnectivityProblems =>
+              Task.delay(noRequestIdLogger.warn(s"Audit cluster connectivity check failed, but 'ignore_es_connectivity_problems: true' is set, so auditing will proceed: ${error.message}"))
```
coutoPL: Maybe the info level is better here? If the user sets `ignore_es_connectivity_problems: true`, they shouldn't be bothered by warning logs.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/AuditSinkServiceCreator.scala
```diff
@@ -16,22 +16,68 @@
...
+                                               httpClientsFactory: HttpClientsFactory,
+                                               create: Task[A]): Task[Either[AuditSinkServiceCreator.InitializationError, A]] =
+    AuditSinkServiceCreator.withConnectivityCheck(cluster, httpClientsFactory, create)
+}
+
+object AuditSinkServiceCreator extends RequestIdAwareLogging {
+
+  final case class InitializationError(message: String)
+
+  private def withConnectivityCheck[A](cluster: AuditCluster,
```
coutoPL: 1) Why is this private method defined in the companion object? Maybe we can just move it to the trait?

Or maybe it could stay where it's and the 
```
protected final def withConnectivityCheck[A](cluster: AuditCluster,
                                               httpClientsFactory: HttpClientsFactory,
                                               create: Task[A]): Task[Either[AuditSinkServiceCreator.InitializationError, A]]
```
method is not needed in the trait?

2) let's rename `A` to `SERVICE` or `AUDIT_SERVICE`

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterConnectivityCheck.scala
```diff
@@ -0,0 +1,213 @@
...
+          case Right(()) =>
+            logger.warn("Some audit cluster nodes are unreachable, but auditing will proceed using the remaining nodes")
+            Right(())
+          case Left(details) =>
+            Left(Error.ConfigurationError(s"Audit cluster healthcheck failed for remote cluster ${cluster.show}. Details: $details"))
+        }
+    }
+  }
+
+  private def checkClusterUuids(infos: NonEmptyList[AuditNodeInfo]): Either[String, Unit] = {
```
coutoPL: maybe `ensureNodesFromSameCluster`?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterConnectivityCheck.scala
```diff
@@ -0,0 +1,213 @@
...
+object AuditRemoteClusterConnectivityCheck {
+
+  sealed trait Error { def message: String }
+
+  object Error {
+    final case class ConnectivityError(message: String) extends Error
+    final case class ConfigurationError(message: String) extends Error
+  }
+
+  private given Show[AuditClusterNode] = Show.show { n =>
```
coutoPL: I know that this show instance is used only in this file, but the `AuditClusterNode` is the audit-related public domain, so maybe let's follow the convention and move the definition to the `LogsShowInstances` trait. WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/remote/AuditRemoteClusterConnectivityCheck.scala
```diff
@@ -0,0 +1,213 @@
...
+    val url = n.toUrl
+    url.authorityOption match {
+      case Some(authority) => url.withAuthority(authority.copy(userInfo = None)(using UriConfig.default)).toUrl.show
+      case None => url.show
+    }
+  }
+
+  private given Show[RemoteAuditCluster] = Show.show { cluster =>
```
coutoPL: the same here

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/AuditDataStreamCreator.scala
```diff
@@ -39,19 +39,28 @@ final class AuditDataStreamCreator(services: NonEmptyList[DataStreamService]) ex
       .toList
       .map(createIfNotExists(_, dataStreamName))
       .sequence
-      .map(_.map(_.leftMap(NonEmptyList.one)).combineAll.toEither)
+      .map { results =>
+        val (errors, successes) = results.map(_.toEither).separate
+        if (successes.nonEmpty) Right(())
```
coutoPL: We ignore the fact that there could be a non-empty error list here. Do you think it's safe? I guess the decision may be related to the fact that somewhere there is the optional connectivity check, but at this level, I'm not sure if we can assume that safely.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
```diff
@@ -299,16 +287,13 @@ object AuditingTool extends RequestIdAwareLogging {
...
-      .flatMap { auditSinkService =>
-        EsDataStreamBasedAuditSink.create(
-          config.logSerializer,
-          config.rorAuditDataStream,
-          auditSinkService,
-          config.auditCluster
-        ).map(_.leftMap(error => CreationError(error.message)).toValidated)
-      }
+                                   service: DataStreamBasedAuditSinkService): Task[Either[CreationError, SupportedAuditSink]] =
+    EsDataStreamBasedAuditSink.create(
```
coutoPL: maybe formatting like this?
```scala
EsDataStreamBasedAuditSink
  .create(
      config.logSerializer,
      config.rorAuditDataStream,
      service,
      config.auditCluster
   )
   .map(_.leftMap(error => CreationError(error.message)))
```

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
```diff
@@ -318,6 +303,19 @@ object AuditingTool extends RequestIdAwareLogging {
     case _: EsDataStreamBasedAuditSink => "data_stream"
   }
 
+  extension (creator: AuditSinkServiceCreator) {
+    private def asIndexBased: IndexBasedAuditSinkServiceCreator = creator match {
```
coutoPL: This is really weird to me. For some reason, we can safely cast AuditSinkServiceCreator to IndexBasedAuditSinkServiceCreator. 

And below we have an invalid state ...

It looks like we need to pass two creators to the `AuditingTool#create` method. 
Or have a creator that could create an index-based audit service and a data-stream audit service (optionally). 

I understand that some older ES versions don't support data streams. But we should express it at the core interface level. 

I understand that currently it's determined at the ES module level, and that's why the code needs casting and the illegal state (when there is some misconfiguration at the ES module level).

WDYT about improving it?

coutoPL (changes requested):

coutoPL:
@mateuszkp96, and please take a look at [this test](https://dev.azure.com/beshu-tech/ReadonlyREST%20for%20Elasticsearch/_build/results?buildId=10395&view=ms.vss-test-web.build-test-results-tab&runId=16414&resultId=100863&paneView=debug). I noticed it sometimes fails. There is some non-determinism in it.

coutoPL:
> @mateuszkp96, and please take a look at [this test](https://dev.azure.com/beshu-tech/ReadonlyREST%20for%20Elasticsearch/_build/results?buildId=10395&view=ms.vss-test-web.build-test-results-tab&runId=16414&resultId=100863&paneView=debug). I noticed it sometimes fails. There is some non-determinism in it.

@mateuszkp96, please take a look at [Simone's attempt to fix this](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/commit/fb290cec2879ebe85643e88a2e91a5c71ee20cd8).

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
mateuszkp96: I completely rebuilt the audit sink creation process and the illegal states are gone.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/AuditDataStreamCreator.scala
mateuszkp96: I've narrowed this down to just one ES node. If the nodes belong to the same cluster, all changes propagate to other nodes accordingly. We already made the single-cluster assumption when introducing the round-robin client.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
```diff
@@ -212,209 +215,261 @@ final class AuditingTool private (private[ror] val sinks: List[AuditSink])(
...
-      object Config {
-
-        final case class EsIndexBasedSink(
-            serializer: JsonAuditSerializer,
-            rorAuditIndexTemplate: RorAuditIndexTemplate,
-            auditCluster: AuditCluster
-        ) extends Config
-
-        object EsIndexBasedSink {
+  object Mode {
```
coutoPL, line 220: I'm not sure about that. 
Do we have sth like Standard or Legacy mode in the auditing domain?
I understand we have sth like legacy and standard configuration in case of the auditing. But then? Do we need it when the auditing tool is created?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/EsAuditCapabilities.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import tech.beshu.ror.accesscontrol.audit.sink.{
+  DataStreamBasedAuditSinkServiceCreator,
+  IndexBasedAuditSinkServiceCreator
+}
+
+sealed trait EsAuditCapabilities
+
+object EsAuditCapabilities {
+  final class Index(val creator: IndexBasedAuditSinkServiceCreator) extends EsAuditCapabilities
```
coutoPL, line 27: `IndexOnly` and `IndexOrDataStream`?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/CoreFactory.scala
```diff
@@ -70,8 +71,43 @@ trait CoreFactory {
...
+  ): Task[Either[NonEmptyList[CoreCreationError], CoreCreationResult]]
+
+}
+
+object CoreFactory {
+
+  final class CoreCreationResult(val core: Core, val auditSetup: AuditSetup)
+
+  object CoreCreationResult {
+    sealed trait AuditSetup
```
coutoPL, line 85: This hierarchy mirrors the EsAuditCapabilities hierarchy which is suspicious.

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -249,21 +251,39 @@ class ReadonlyRest(
       }
   }
 
-  private def createAuditingTool(auditingConfig: AuditingConfig)(
+  private def createAuditingTool(result: CoreCreationResult, engineResources: EngineResources)(
```
coutoPL, line 254: we don't need the whole CoreCreationResult here

core/src/main/scala/tech/beshu/ror/es/services/MultiNodeRestClient.scala
```diff
@@ -0,0 +1,236 @@
...
+import tech.beshu.ror.accesscontrol.domain.RequestId
+import tech.beshu.ror.es.services.FailoverClient.*
+import tech.beshu.ror.es.services.MultiNodeRestClient.*
+import tech.beshu.ror.utils.RequestIdAwareLogging
+
+import java.time.Clock
+import java.util.concurrent.ConcurrentHashMap
+import scala.jdk.CollectionConverters.*
+
+trait MultiNodeRestClient[Req, Resp] {
```
coutoPL, line 30: Why the javish api design?

coutoPL (changes requested):

coutoPL (approved):
LGTM

PR #1227, opened 2026-03-22
[RORDEV-1961] Improve consistency of groups rules settings and impersonation settings

mgoworko:
🧐 Enhancement - Improve consistency of groups rules settings - the `users` section can only be present in the config, when there is at least one groups rule that uses it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Config validation now rejects configurations where a defined users section is not referenced by any groups rule.
  * Validation reports clear errors for missing or empty ACL sections and for duplicate block names.
  * Error messages for misconfigured ACL blocks have been improved to make fixes clearer.

* **Chores**
  * Build tooling version updated.


core/src/main/scala/tech/beshu/ror/accesscontrol/factory/CoreFactory.scala
```diff
@@ -410,6 +401,51 @@ class RawRorSettingsBasedCoreFactory(esEnv: EsEnv)
     }
   }
 
+  private def extractAndValidateAclBlocks(blockDecodingResult: FieldListResult[BlockDecodingResult],
```
coutoPL: We have sth like BlockValidator. Maybe we should introduce AclValidator and move these validations there? 
WDYT?

mgoworko: Refactored in order to add dedicated validator, added additional tests.

gradle.properties
```diff
@@ -1,5 +1,5 @@
-publishedPluginVersion=1.68.0
-pluginVersion=1.69.0
+publishedPluginVersion=1.69.0
+pluginVersion=1.70.0-pre1
```
coutoPL: let's bump the pre version

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/AclValidator.scala
```diff
@@ -0,0 +1,74 @@
...
+
+import cats.data.{NonEmptyList, ValidatedNel}
+import tech.beshu.ror.accesscontrol.blocks.Block
+import tech.beshu.ror.accesscontrol.blocks.definitions.UserDef
+import tech.beshu.ror.accesscontrol.blocks.rules.auth.base.BaseGroupsRule
+import tech.beshu.ror.accesscontrol.factory.decoders.definitions.*
+import tech.beshu.ror.implicits.*
+import tech.beshu.ror.utils.ScalaOps.findDuplicates
+
+abstract class AclValidator[B] {
```
coutoPL: I'm not sure we need the abstract class here. There is only one implementation and I cannot imagine we will need another one.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/AclValidator.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  protected def extract(blockRepresentation: B): Block
+
+  def validate(blockRepresentationsOpt: Option[List[B]],
+               userDefs: Definitions[UserDef],
+               aclAttributeName: String): ValidatedNel[String, NonEmptyList[B]] = {
```
coutoPL: Let's remove this param and use `Attributes.acl` in this file

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/AclValidator.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  private def validateThatUsersConfigSectionIsUsedWhenPresent(blocks: List[Block],
+                                                              userDefs: Definitions[UserDef]): ValidatedNel[String, Unit] = {
+    val thereAreUserDefinitions = userDefs.items.nonEmpty
+    lazy val thereIsGroupsRule = blocks.flatMap(_.rules.toList).collect {
+      case rule: BaseGroupsRule[_] => rule
+    }.nonEmpty
+
+    if (thereAreUserDefinitions && !thereIsGroupsRule) {
+      "The `users` config section is defined, but there is no groups rule that uses it. Either remove the `users` section in the config, or add the groups rule in the ACL.".invalidNel
```
coutoPL: Shouldn't we use reference to the "users" section name here too?

coutoPL (changes requested):

coutoPL (approved):
LGTM

PR #1233, opened 2026-04-03
[RORDEV-1986] Optimize ACL request processing time and CPU usage

coutoPL:
* **🧐 Enhancement** (ES) Reduced CPU usage and response latency for requests involving index/alias resolution             
* **🧐 Enhancement** (ES) Improved throughput under concurrent load by avoiding unnecessary locking when a    
  cached result is already available  

## Summary

 Targeted CPU and allocation optimizations across the index/alias resolution pipeline, reducing redundant work 
  per request.x
                                                                                                                
  - Cache fast-path before semaphore (AsyncCacheableActionWithKeyMapping): check cache before acquiring/creating
   a semaphore, avoiding unnecessary Semaphore allocation and task scheduling on cache hits. Also avoid creating
   a new semaphore when one already exists.                                                                     
  - Memoize cluster state per request (LocalIndicesManager, RemoteIndicesManager): all derived sets (indices,
  aliases, data streams, etc.) are now computed at most once per request via lazy val (local) and Task.memoize  
  (remote), instead of re-traversing the cluster state on every method call.
  - Pre-computed snapshot types (EsClusterService): introduced LocalIndicesSnapshot and LocalDataStreamsSnapshot
   that eagerly compute flat sets (indicesAndAliases, indices, aliases) when created, so repeated lookups within
   a request hit already-computed values.
  - Mutable builder for alias maps (EsClusterService): replaced foldLeft + immutable Map.combine (via Monoid)   
  with a mutable.Map builder, eliminating intermediate map allocations when building the alias-to-indices       
  reverse mapping.
  - CovariantSet consolidation: replaced many separate implicit wrapper classes with a single Scala 3 extension 
  block (reduces wrapper object allocations). toCovariantSet now avoids copying when the input is already a Set.
   Added a shared cached Monoid instance.
  - lazy val → val in FullLocalDataStreamWithAliases / FullRemoteDataStreamWithAliases: fields are always       
  accessed, so lazy initialization was pure overhead.                                                           
  - EnabledAccessControlList: refactored deniedResultFrom to compute denial causes once (was computed 3 times in
   separate methods). Simplified gatherGroupMetadataPreservingOrder from nested foldLeft to flatMap.


core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditRequestContextBasedOnAclResult.scala
```diff
@@ -77,7 +77,7 @@ private[audit] class AuditRequestContextBasedOnAclResult[B <: BlockContext](requ
         blockName -> matchedAndCause
       }
       .toMap
-  override val content: String = requestContext.restRequest.content
+  override lazy val content: String = requestContext.restRequest.content
   override val contentLength: Integer = requestContext.restRequest.contentLength.toBytes.toInt
```
mateuszkp96: Let's make this field lazy too - it is based on restRequest.content

es717x/src/main/scala/tech/beshu/ror/es/services/EsNodeClusterService.scala
```diff
@@ -557,8 +491,104 @@ class EsNodeClusterService(nodeName: String,
...
+            backingIndices = backingIndices
+          )
+        }
+        .toCovariantSet
+    }
+
+    private def aliasesPerDataStreamFrom(metadata: Metadata): Map[DataStreamName.Full, Set[DataStreamName.Full]] = {
+      val result = mutable.HashMap.empty[DataStreamName.Full, mutable.Builder[DataStreamName.Full, Set[DataStreamName.Full]]]
+      val dataStreamAliases = metadata.dataStreamAliases()
+      dataStreamAliases.keySet().asScala.foreach { aliasName =>
```
mateuszkp96: Maybe we should use Set.forEach from Java instead of converting to a Scala Set?

core/src/main/scala/tech/beshu/ror/es/services/EsClusterService.scala
```diff
@@ -35,14 +38,132 @@ trait EsClusterService {
...
+    builder.view.mapValues(_.toCovariantSet).toMap
+  }
+
+  private def indices(filteredBy: Set[IndexAttribute])
+                     (implicit requestId: RequestId) = {
+    allIndicesAndAliases
+      .filter(i =>
+        if (filteredBy.nonEmpty) filteredBy.contains(i.attribute)
+        else true
+      )
```
mateuszkp96: If the filteredBy param is empty, we don't need to filter

core/src/main/scala/tech/beshu/ror/es/services/EsClusterService.scala
```diff
@@ -35,14 +38,132 @@ trait EsClusterService {
...
+      }
+  }
+
+  private def remoteIndices(filteredBy: Set[IndexAttribute])
+                           (implicit id: RequestId) = {
+    allRemoteIndicesAndAliases
+      .map(_.filter(i =>
+        if (filteredBy.nonEmpty) filteredBy.contains(i.attribute)
+        else true
+      ))
```
mateuszkp96: If the filteredBy param is empty, we don't need to filter

core/src/main/scala/tech/beshu/ror/es/services/EsClusterService.scala
```diff
@@ -35,14 +38,132 @@ trait EsClusterService {
...
+    builder.view.mapValues(_.toCovariantSet).toMap
+  }
+
+  private def dataStreams(filteredBy: Set[IndexAttribute])
+                         (implicit requestId: RequestId) = {
+    allDataStreamsAndAliases
+      .filter(ds =>
+        if (filteredBy.nonEmpty) filteredBy.contains(ds.attribute)
+        else true
+      )
```
mateuszkp96: And here

core/src/main/scala/tech/beshu/ror/es/services/EsClusterService.scala
```diff
@@ -35,14 +38,132 @@ trait EsClusterService {
...
+  }
+
+  private def remoteDataStreams(filteredBy: Set[IndexAttribute])
+                               (implicit id: RequestId) = {
+    allRemoteDataStreamsAndAliases
+      .map(_.filter(ds =>
+        if (filteredBy.nonEmpty) filteredBy.contains(ds.attribute)
+        else true
+      ))
+  }
```
mateuszkp96: And here

mateuszkp96 (changes requested):

es67x/src/main/scala/tech/beshu/ror/es/utils/CallActionRequestAndHandleResponse.scala
```diff
@@ -20,49 +20,35 @@ import monix.eval.Task
...
+  extension [RESPONSE <: ActionResponse](caller: ActionListener[RESPONSE] => Unit)
+    def execute[R](f: RESPONSE => R): Task[R] = {
+      val listener = new GenericResponseListener[RESPONSE]()
+      caller(listener)
+      listener.result(f)
+    }
 }
 
-private final class GenericResponseListener[RESPONSE <: ActionResponse] extends ActionListener[RESPONSE] {
+private final class GenericResponseListener[RESPONSE <: ActionResponse] extends ActionListener[RESPONSE]:
```
mateuszkp96: Let's use curly braces instead of `:`

mateuszkp96 (approved):
LGTM with one minor comment

PR #1237, opened 2026-04-21
[RORDEV-1646] Making ROR bootstrap configuration consistent between `elasticsearch.yml` and Java properties

coutoPL:
🧐 Enhancement (ES) ROR bootstrap settings in `elasticsearch.yml` are now configured via proper nested YAML
  blocks under `readonlyrest.*` keys.
  
# Documentation:
https://github.com/beshu-tech/readonlyrest-docs/pull/312

# What's new / changed:                                                                                        
   
  - readonlyrest.settings.* — configure where ROR looks for ACL settings: custom index name, custom file path, 
  and maximum allowed settings size loaded from the index.             
  - readonlyrest.load_from_index.* — control how ROR polls the index for ACL settings changes and how it
  retries during startup if the index is not yet available.                                                    
  - readonlyrest.force_load_from_file — force ROR to ignore the index entirely and load ACL settings from file
  only (recovery mechanism for corrupted in-index settings).                                                   
  - readonlyrest.not_started_response_code / readonlyrest.failed_to_start_response_code — now documented under 
  their correct nested YAML form (previously only flat dot-notation was shown).
  - All node settings can also be provided as JVM system properties using the same dot-separated key.          
                                                                       
  Backward compatibility with the deprecated readonlyrest.settings_index key is preserved but not documented.  
                                               
#   Technical changes                                                                                            
                                                                                                               
  - RorSettingsSourcesConfig.scala — replaced flat key lookups with YamlLeafOrPropertyOrEnvDecoder; new nested 
  key structure under readonlyrest.settings.*; legacy keys kept for backward compat.
  - RorCoreSettingsLoadingStrategy.scala — new readonlyrest.load_from_index.* block; refreshInterval field
  renamed to pollInterval; retry strategy decoded from nested YAML.
  - RorBootSettings.scala — not_started_response_code / failed_to_start_response_code migrated to              
  YamlLeafOrPropertyDecoder.                  
  - RorSslSettings.scala — internal refactoring: String key constants replaced with NonEmptyString; decoders   
  migrated to YamlLeafOrPropertyOrEnvDecoder. No key name changes.     
  - YamlLeafOrPropertyDecoder.scala (new) — unified decoder supporting nested YAML, flat dot-notation leaf, and
   JVM system properties, replacing the removed YamlKeyDecoder.scala.  
  - RorProperties.scala — removed; responsibilities distributed into the new decoder infrastructure.           
  - ElasticsearchConfigLoader.scala (renamed from previous) — minor cleanup.
  - Test fixture elasticsearch.yml files updated to use the new nested key structure throughout.


core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala
```diff
@@ -0,0 +1,288 @@
...
+import eu.timepit.refined.types.string.NonEmptyString
+import io.circe.{ACursor, Json, JsonNumber}
+import tech.beshu.ror.providers.EnvVarProvider.EnvVarName
+import tech.beshu.ror.providers.PropertiesProvider.PropName
+import tech.beshu.ror.providers.{EnvVarsProvider, PropertiesProvider}
+import tech.beshu.ror.utils.FromString
+
+import scala.language.implicitConversions
+
+trait YamlLeafOrPropertyOrEnvDecoder[T] {
```
mgoworko, line 31: Maybe just RorConfigValueDecoder ?

core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+private final class OptionalListYamlLeafOrPropertyOrEnvDecoder[T](path: NonEmptyList[NonEmptyString])
+                                                                 (implicit valueCreator: FromString[T],
+                                                                  propertiesProvider: PropertiesProvider,
+                                                                  envVarsProvider: EnvVarsProvider)
+  extends YamlLeafOrPropertyOrEnvDecoder[Option[Set[T]]] {
+
+  private val propName   = NonEmptyString.unsafeFrom(JsonPathOps.pathAsString(path))
```
mgoworko: Let's get rid of `NonEmptyString.unsafeFrom` calls - we can rather safely encapsulate them in JsonPathOps methods.

core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  def whenSectionPresent[T](sectionPath: NonEmptyList[NonEmptyString])
+                           (inner: YamlLeafOrPropertyOrEnvDecoder[Option[T]])
+                           (implicit propertiesProvider: PropertiesProvider,
+                            envVarsProvider: EnvVarsProvider): YamlLeafOrPropertyOrEnvDecoder[Option[T]] =
+    json => {
+      val inYaml       = JsonPathOps.focusAt(json, sectionPath).exists(j => !j.isNull)
+      val inProperties = propertiesProvider.hasPropertyWithPrefix(JsonPathOps.pathAsString(sectionPath) + ".")
+      val envPrefix    = JsonPathOps.pathPrefixToEnvVarPrefix(sectionPath)
+      val inEnv        = envVarsProvider.hasEnvMatching(k => k.startsWith(envPrefix) && (k.length == envPrefix.length || k.charAt(envPrefix.length) != '_'))
```
mgoworko: I was suspicious about the `k.length == envPrefix.length`, Claude tells me I'm correct that it is a dead code (because we check for prefix, and this checks if the prefix is at the same time the entire name of the config settings, please check if that's correct.

mgoworko: Conditions `is...` in this method are sophisticated enough to be extracted to methods)

mgoworko: Maybe there should be helper methods, that create prefix path (add ".") for json, something like `hasEnvWithPrefix` for envs, and `jsonHasPath` instead of `JsonPathOps.focusAt(json, sectionPath).exists(j => !j.isNull)` ?

core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
```diff
@@ -84,10 +84,15 @@ class AuditSettingsTests extends AnyWordSpec with Inside {
     "audit is disabled" should {
       "be disabled" when {
         "one line audit format" in {
-          val settings = rorSettingsWithAuditUnsafe(
+          val settings = rorSettingsFromUnsafe(
             """
+              |readonlyrest.audit.enabled: false
```
mgoworko, line 81: Are there tests that verify that the old syntax works?

core/src/test/scala/tech/beshu/ror/utils/TestsUtils.scala
```diff
@@ -444,6 +445,22 @@ object TestsUtils {
       ))
   }
 
+  private given Using.Releasable[File] = _.delete(swallowIOExceptions = true)
```
mgoworko, line 448: I see mixed Scala 2/3 styles in this PR. On the other hand there are some (but not many) givens and other Scala 3 syntax parts. On the other hand, new typeclasses and utils use implicits and implicit classes for extensions

core/src/main/scala/tech/beshu/ror/settings/es/RorCoreSettingsLoadingStrategy.scala
```diff
@@ -69,48 +71,174 @@ object RorCoreSettingsLoadingStrategy extends YamlFileBasedSettingsLoaderSupport
...
+
+    object defaults {
+      val coreRefreshSettings: CoreRefreshSettings = CoreRefreshSettings.Enabled((5 second).toRefinedPositiveUnsafe)
+      val loadingDelay: LoadingDelay = LoadingDelay((5 second).toRefinedNonNegativeUnsafe)
+      val loadingAttemptsCount: LoadingAttemptsCount = LoadingAttemptsCount(Refined.unsafeApply(5))
+      val loadingAttemptsInterval = LoadingAttemptsInterval((5 second).toRefinedNonNegativeUnsafe)
+    }
+
+    private object legacyConsts {
+      val refreshInterval: NonEmptyString = NonEmptyString.unsafeFrom("com.readonlyrest.settings.refresh.interval")
```
mgoworko: We have `inline def nes` in RefinedUtils, that creates NonEmptyString safely from literals

core/src/test/scala/tech/beshu/ror/utils/TestsEnvVarsProvider.scala
```diff
@@ -0,0 +1,36 @@
...
+  override def getEnv(name: EnvVarName): Option[String] = envMap.get(name)
+
+  override def hasEnvMatching(predicate: String => Boolean): Boolean =
+    envMap.keys.exists(k => predicate(k.value.value))
+}
+
+object TestsEnvVarsProvider {
+  def default: TestsEnvVarsProvider = new TestsEnvVarsProvider(Map.empty)
+
+  def usingMap(map: Map[String, String]): TestsEnvVarsProvider = new TestsEnvVarsProvider(
```
mgoworko: I think the argument should be map: Map[NonEmptyString, String], wchich will allow to avoid yet another `unsafeFrom`

core/src/main/scala/tech/beshu/ror/settings/es/RorCoreSettingsLoadingStrategy.scala
```diff
@@ -69,48 +71,174 @@ object RorCoreSettingsLoadingStrategy extends YamlFileBasedSettingsLoaderSupport
...
-          for {
-            loadingRetryStrategySettings <- loadLoadingRetryStrategySettings(systemContext.propertiesProvider)
-            coreRefreshIntervalSettings <- loadCoreRefreshSettings(systemContext.propertiesProvider)
-          } yield RorCoreSettingsLoadingStrategy.LoadFromIndexWithFileFallback(
-            loadingRetryStrategySettings, coreRefreshIntervalSettings
-          )
-      }
+
+    object defaults {
+      val coreRefreshSettings: CoreRefreshSettings = CoreRefreshSettings.Enabled((5 second).toRefinedPositiveUnsafe)
```
mgoworko: Same here, we have helpers (inline and safe) for creating refined values from literals

core/src/main/scala/tech/beshu/ror/settings/ror/RawRorSettingsYamlParser.scala
```diff
@@ -33,7 +33,7 @@ class RawRorSettingsYamlParser(maxSize: Information) {
 
   def fromString(content: String): Either[ParsingRorSettingsError, RawRorSettings] = {
     handleParseResult(yamlParser.parse(content))
-      .map(RawRorSettings(_, content))
+      .map(json => RawRorSettings(YamlOps.jsonWithOneLinerKeysToRegularJson(json), content))
```
mgoworko: Nitpick: better readibility with

```
  def fromString(content: String): Either[ParsingRorSettingsError, RawRorSettings] = {
    handleParseResult(yamlParser.parse(content))
      .map(YamlOps.jsonWithOneLinerKeysToRegularJson)
      .map(RawRorSettings(_, content))
  }
```

core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala
```diff
@@ -0,0 +1,288 @@
...
+                           (implicit propertiesProvider: PropertiesProvider,
+                            envVarsProvider: EnvVarsProvider): YamlLeafOrPropertyOrEnvDecoder[Option[String]] =
+    createOptionalValueDecoder(path, FromString.string)
+
+  def optionalBooleanDecoder(path: NonEmptyList[NonEmptyString])
+                            (implicit propertiesProvider: PropertiesProvider,
+                             envVarsProvider: EnvVarsProvider): YamlLeafOrPropertyOrEnvDecoder[Option[Boolean]] =
+    createOptionalValueDecoder(path, FromString.boolean)
+
+  def createLegacyPropertyDecoder[T](legacyKey: NonEmptyString, decoder: FromString[T])
```
mgoworko, line 85: The old implementation created log when using default:
```noRequestIdLogger.info(s"No '${name.show}' property found. Using default: ${default.show}")```

**Review by mgoworko — CHANGES_REQUESTED** [review]

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 1 above)
Comment (coutoPL) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala:31]: `RorConfigValueDecoder` is misleading (because it's rather ROR config but defined in elasticsearch.yml, not in readonlyrest.yml/ROR index), and the name tells nothing about it. 

The current name is a little bit verbose (this is true), but it tells a lot about the semantics of the decoding.

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 3 above)
Comment (coutoPL) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala]: I'm not sure if I understand what to do here.

File: core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala (hunk 4 above)
Comment (coutoPL) [reply core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala:81]: What old syntax are you talking about?

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 10 above)
Comment (coutoPL) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala:85]: We don't need this log anymore. Now we read from file, with fallback to property, with fallback to env. Old approach was only reading from the property.

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 3 above)
Comment (mgoworko) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala]: Sorry, I just reread my comment—it’s indeed a bit of gibberish.

I meant, that in the lines 63-66 (where we check the existence of the config section) there is a logic, that defines how the prefix looks like. And that logic can be extracted:
- for json `focusAt(...).exists(j => !j.isNull)` - we could create `def JsonPathOps.hasPath`
- for propertiesProvider we could create `def hasPropertyInSection(sectionPath)`
- for envVarsProvider we could create `def hasEnvWithPrefixMatching(sectionPath)` instead of using `hasEnvMatching` and `val envPrefix`

That would extract the rather complicated logic of [checking nulls/adding dots/checking lenght and last char] to well named helpers.

File: core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala (hunk 4 above)
Comment (mgoworko) [reply core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala:81]: I meant the change from multi-line config to single-line config names. The previous config structure (with yaml sections and subsections) has been replaced in tests with the single-line config setting names (with "." separator). I was just asking if there is a test, that verifies the old syntax still works.

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 10 above)
Comment (mgoworko) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala:85]: We still have defaults, for example:
```
    object defaults {
      val coreRefreshSettings: CoreRefreshSettings = CoreRefreshSettings.Enabled((5 second).toRefinedPositiveUnsafe)
      val loadingDelay: LoadingDelay = LoadingDelay((5 second).toRefinedNonNegativeUnsafe)
      val loadingAttemptsCount: LoadingAttemptsCount = LoadingAttemptsCount(Refined.unsafeApply(5))
      val loadingAttemptsInterval = LoadingAttemptsInterval((5 second).toRefinedNonNegativeUnsafe)
    }
```

Before this change when we used the default value, that fact was logged. Now it is not. I'm not sure if we need this log, but I just want to check that this change in behavior was intentional.

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 3 above)
Comment (coutoPL) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala]: I'm not sure about it. I don't want to extend PropertiesProvider and EnvVarsProvider with sth like a section or sth. There are already methods that work on prefix - IMO this is enough. But I extracted private methods, like you advised in one of the comments above - please take a look and tell me WDYT

File: core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala (hunk 10 above)
Comment (coutoPL) [reply core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala:85]: I've added additional logging. I think it solves the problem. WDYT?

File: core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala (hunk 4 above)
Comment (coutoPL) [reply core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala:81]: I've added more tests

Comment (coutoPL) [reply core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala:81]: BTW, this is not an old and new syntax, but YAML should support both

File: core/src/main/scala/tech/beshu/ror/settings/es/EsConfigBasedRorSettings.scala [hunk 11]
```diff
@@ -29,16 +31,27 @@ final case class EsConfigBasedRorSettings(settingsSource: RorSettingsSourcesConf
...
 
   def from(esEnv: EsEnv)
           (implicit systemContext: SystemContext): Task[Either[LoadingError, EsConfigBasedRorSettings]] = {
-    val result = for {
+    for {
+     _ <- Task.delay(noRequestIdLogger.info("Loading ROR node settings ..."))
+     result <- doLoad(esEnv).value
+     _ <- result match {
+       case Right(settings) => Task.delay(noRequestIdLogger.debug(s"Loaded ROR node settings:\n${settings.show}"))
+       case Left(_) => Task.unit
```
mgoworko, line 43: Perhaps it's a good place to log the error too

core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala
mgoworko: Ok, now it looks much more readable to me :+1:

mgoworko (approved):
LGTM
(with one small comment, also I think claude may be right about detecting prefix, the change I suggested might have caused that other problem)

core/src/main/scala/tech/beshu/ror/settings/es/EsConfigBasedRorSettings.scala
coutoPL: The error is already passed to the exception that is thrown when the settings cannot be parsed

core/src/main/scala/tech/beshu/ror/utils/yaml/YamlLeafOrPropertyDecoder.scala
coutoPL: "One Claude will say yes, and another Claude will say no" ;)

BTW, I made the condition a little bit more readable than before and added a proper test to test it.

PR #1238, opened 2026-04-21
[RORDEV-2006] Replace AsyncHttpClient with Apache HttpClient

mgoworko:
🚨 Security Fix (ES) Fixed vulnerability CVE-2026-40490

Apache HttpClient is now used when connecting to external authentication, external JWT verification and external groups provider services, instead of AsyncHttpClient.

## Summary by CodeRabbit

* **Dependencies**
  * Replaced AsyncHttpClient with Apache HTTPClient 5.6.1 and added explicit Netty transport components.

* **Bug Fixes**
  * HTTP client shutdown now runs asynchronously and logs close failures for safer shutdown.

* **Refactor**
  * Reworked HTTP client factory and implementations for improved resource management and lifecycle handling.

* **New Features**
  * Added a domain-specific response status type for main settings endpoints.


core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -17,34 +17,27 @@
...
-import scala.jdk.DurationConverters.*
-import scala.jdk.FutureConverters.*
+import scala.jdk.CollectionConverters.IteratorHasAsScala
 import scala.language.postfixOps
 
 trait HttpClientsFactory {
 
   def create(config: Config): HttpClient
-  def shutdown(): Unit
+  def shutdown(): Task[Unit]
```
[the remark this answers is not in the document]
mgoworko: In test it is fixed, but in runAsyncAndForget seems to be the common method to close resources in ReadonlyRest

core/build.gradle
```diff
@@ -66,6 +65,7 @@ dependencies {
     api group: 'commons-codec',                 name: 'commons-codec',                       version: '1.15'
     api group: 'com.beachape',                  name: 'enumeratum_3',                        version: '1.9.2'
     api group: 'com.hrakaroo',                  name: 'glob',                                version: '0.9.0'
+    api group: 'org.apache.httpcomponents.client5', name: 'httpclient5',                     version: '5.6.1'
```
mgoworko, line 68: I didn't want to obscure git history of dependencies and their versions by reformatting (indentation) the entire list.

core/src/main/scala/tech/beshu/ror/boot/ReadonlyRest.scala
```diff
@@ -134,7 +135,7 @@ class ReadonlyRest(coreFactory: CoreFactory,
   private[ror] def loadRorEngine(settings: RawRorSettings,
                                  settingsIndex: RorSettingsIndex)
                                 (implicit requestId: RequestId): Task[Either[StartingFailure, Engine]] = {
-    val httpClientsFactory = new AsyncHttpClientsFactory
+    val httpClientsFactory = new DefaultHttpClientsFactory
```
mgoworko: I'm wondering if maybe we should preserve the old AsyncHttpClient for another ROR version (under feature flag, to allow switching to it in case any user reports problems with the new client). BUt on the other hand, the CVE will be still reported in our pipeline, and potentially in clients' security reports.

coutoPL: No, if sth like this happens, we will recommend a downgrade as a workaround, and we will deliver a patch as soon as possible.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -82,46 +75,31 @@ object HttpClientsFactory {
 }
 
 // todo: remove synchronized, use more sophisticated lock mechanism
-class AsyncHttpClientsFactory extends HttpClientsFactory with RequestIdAwareLogging {
+class DefaultHttpClientsFactory extends HttpClientsFactory with RequestIdAwareLogging {
```
coutoPL: let's rename it - we need the client name in the factory name. 
"Default" is too generic IMO.

coutoPL: We have a file `HttpClientsFactory` where we define the factory interface and the client interface. 
Let's have a symmetric file with the implementation in which we will have:
1. The client implementation
2. The factory implementation

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -82,46 +75,31 @@ object HttpClientsFactory {
...
-    } catch {
-      case ex: Throwable =>
-        noRequestIdLogger.error(s"Failed to create AsyncHttpClient", ex)
-        throw ex
+  override def shutdown(): Task[Unit] = {
+    val clients = synchronized {
+      isWorking.set(false)
+      existingClients.iterator().asScala.toList
     }
+    Task.parSequenceUnordered(clients.map(_.close().onErrorHandleWith { e =>
```
coutoPL: 1. let's format it better
2. I wonder if we shouldn't have the log in the client impl. WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -82,46 +75,31 @@ object HttpClientsFactory {
 }
 
 // todo: remove synchronized, use more sophisticated lock mechanism
```
coutoPL: ok, let's remove this comment. I don't see it's worth improving. 
Just introduce:
`private object lock` within the class and use this object for sychronization

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/ApacheBasedSimpleHttpClient.scala
```diff
@@ -0,0 +1,124 @@
...
+      case Method.Post => "POST"
+    }
+    val httpRequest = SimpleHttpRequest.create(method, request.url.toStringRaw)
+    request.headers.foreach { case (k, v) => httpRequest.setHeader(k, v) }
+
+    Task.deferFuture {
+      val promise = Promise[SimpleHttpResponse]()
+      client.execute(
+        httpRequest,
+        new FutureCallback[SimpleHttpResponse] {
```
coutoPL: Let's introduce an extension for the client that handles the callback -> Task conversion inside some new method.

core/build.gradle
```diff
@@ -77,6 +77,9 @@ dependencies {
         exclude group: 'org.apache.logging.log4j', module: 'log4j-core'
     }
     api group: 'io.monix',                      name: 'monix_3',                             version: '3.4.1'
+    api group: 'io.netty',                      name: 'netty-codec-http2',                   version: '4.1.132.Final'
```
coutoPL, line 80: Do we need Netty now in the constants section? 
Please review the other constraints section entries too

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
mgoworko: In my view, the HttpClientsFactory implementation (currently with name `DefaultHttpClientsFactory`) should not be directly linked with a specific http client backend implementation. That is why I left it here and I still think it should stay in the `HttpClientsFactory` as a common part, which remains the same regardless of underlying client.

I think that the missing part is something which can be called `SimpleHttpClientCreator`

```
trait SimpleHttpClientCreator {
  def create(config: Config): HttpClient
}
```

After changes now we have a separate file with SimpleHttpClient and SimpleHttpClientCreator trait definitions, and a symmetric file, which implements the client and its creator (but not the HttpClientsFactory, which remains the same regardless of http client backend).

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
mgoworko: Log moved to the client implementation, formatting improved in both places

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
mgoworko: Please see the answer to the second comment. I'd like to still have a common DefaultHttpClientsFactory, which is not tied to any specific http client impl - in order to isolate the dependency on Apache to a single file and not add "Apache" to the factory name.

```
class DefaultHttpClientsFactory(simpleHttpClientCreator: SimpleHttpClientCreator)
```

core/build.gradle
mgoworko: Yes, we do. The only usage at the moment is the ssl implementation `object SSLCertHelper`. The dependencies and constrains are also transitively used by the ES-specific modules. Without the constrains, ROR for older ES modules uses older netty artifact (observed locally, when I removed those constrains because I thought they are not needed anymore)

The `netty-codec-http2` is even more problematic, because it is not needed in compile time, but ROR for ES >= 8 fails in runtime on missing classes.

I wonder if maybe we should have a separate module for ssl commons, which would define netty versions and dependencies separately from core.

As for other constrains: I removed io.spray - it seems completely unused.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/SimpleHttpClient.scala
```diff
@@ -0,0 +1,31 @@
...
+
+trait SimpleHttpClient[F[_]] {
+  def send(request: HttpClient.Request)
+          (implicit requestId: RequestId): F[HttpClient.Response]
+
+  def close(): F[Unit]
+}
+
+trait SimpleHttpClientCreator {
+  def create(config: Config): HttpClient
```
coutoPL: `type HttpClient = SimpleHttpClient[Task]` this is defined in HttpClientsFactory compation object. I see the following problems:
1. HttpClientsFactory uses SimpleHttpClientCreator. It's not so good that SimpleHttpClientCreator has to know anything about HttpClientsFactory
2. The same about Config definition
3. I wonder if we shouldn't declare it like this:
```
trait SimpleHttpClientCreator[F[_]] {
  def create(config: Config): SimpleHttpClient[F[_]]
}
```

or maybe even like that:
```
trait SimpleHttpClientCreator[F[_], C <: SimpleHttpClient[F]] {
  def create(config: Config): C
}
```

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -79,49 +72,36 @@ object HttpClientsFactory {
     )
   }
 
+  def default() = new DefaultHttpClientsFactory(ApacheBasedSimpleHttpClientCreator)
+
 }
 
-// todo: remove synchronized, use more sophisticated lock mechanism
-class AsyncHttpClientsFactory extends HttpClientsFactory with RequestIdAwareLogging {
+private class DefaultHttpClientsFactory(simpleHttpClientCreator: SimpleHttpClientCreator)
```
coutoPL: just `httpClientCreator`

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -17,34 +17,27 @@
...
 
 import java.util.concurrent.{CopyOnWriteArrayList, TimeUnit}
-import scala.concurrent.duration.*
-import scala.jdk.CollectionConverters.*
-import scala.jdk.DurationConverters.*
-import scala.jdk.FutureConverters.*
+import scala.jdk.CollectionConverters.IteratorHasAsScala
 import scala.language.postfixOps
 
 trait HttpClientsFactory {
```
coutoPL: but do we need this trait now? When we have the creator trait, we don't need this one IMO.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
coutoPL: When we remove the trait, we may have just HttpClientsFactory

coutoPL: ok, it's a good approach. Let's do it like that

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/HttpClientsFactory.scala
```diff
@@ -65,63 +49,35 @@ object HttpClientsFactory {
...
-          .setUseInsecureTrustManager(!config.validate)
-          .build()
-      }
-    } catch {
-      case ex: Throwable =>
-        noRequestIdLogger.error(s"Failed to create AsyncHttpClient", ex)
-        throw ex
+  def shutdown(): Task[Unit] = {
+    val clients = lock.synchronized {
+      isWorking.set(false)
```
coutoPL: I've just noticed we didn't protect ourselves against a double shutdown. It's a minor thing, but maybe we can check the `isWorking` flag and do nothing when it's not.

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/ApacheBasedSimpleHttpClient.scala
```diff
@@ -0,0 +1,139 @@
...
+
+      val client = HttpAsyncClients.custom()
+        .setConnectionManager(connManager)
+        .setDefaultRequestConfig(requestConfig)
+        .build()
+
+      client.start()
+      client
+    } catch {
+      case ex: Throwable =>
```
coutoPL: Let's catch NonFatal` only

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/ApacheBasedSimpleHttpClient.scala
```diff
@@ -0,0 +1,139 @@
...
+          .setHostVerificationPolicy(HostnameVerificationPolicy.CLIENT)
+          .setHostnameVerifier(NoopHostnameVerifier.INSTANCE)
+          .buildAsync()
+        connManagerBuilder.setTlsStrategy(tlsStrategy)
+      }
+
+      val connManager = connManagerBuilder.build()
+
+      val requestConfig = RequestConfig.custom()
+        .setConnectionRequestTimeout(Timeout.ofMilliseconds(config.requestTimeout.value.toMillis))
```
coutoPL: AFAIU, when there is no connection in the pool to lease, it waits for `config.requestTimeout`, and if within this time the connection can be leased, it waits `config.requestTimeout` for the request handling. 
Maybe it'd be better to fall immediately when there is no free connection in the pool (at least in the context of the settings we expose). I guess we have to set 0 or a small value here.

coutoPL (approved):
LGTM.

1. just minor comments left
2. please bump pre-version

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/ApacheBasedSimpleHttpClient.scala
mgoworko: Changed to 1ms (there is even a predefined constant in Apache for 1ms, because 0 means infinite wait.

PR #1239, opened 2026-04-23
[RORDEV-2008] Use HTTP Maven repo for public artifact storage instead of Gradle s3 transport

coutoPL:
Switched public artifact repository resolution from Gradle s3:// transport to a plain HTTP(S) Maven URL built from ROR_LIBS_STORE_URL and BUCKET.

This avoids Gradle S3 authentication requirements for publicly accessible artifacts and makes resolution work consistently with both native public S3 endpoints and S3-compatible public storage. The repo path remains /<bucket>/ror/libs, and dependency resolution continues to use artifact-based metadata lookup.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to source S3 repository settings from environment variables, enabling more flexible deployment setups
  * Simplified repository endpoint configuration logic

### DIFF
--- a/build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
+++ b/build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
@@ -24,30 +24,16 @@ project.ext {
     publishedPluginVersion = rootProject.properties['pluginVersion']
     pluginVersion = rootProject.properties['pluginVersion']
     esVersion = project.hasProperty('esVersion') ? project.property('esVersion') : project.property('latestSupportedEsVersion')
-    s3BucketName = System.env.BUCKET ?: "beshu"
+    s3BucketName = System.getenv('BUCKET') ?: 'beshu'
+    s3StoreAddress = System.getenv('ROR_LIBS_STORE_URL') ?: 'https://dgp.serve.beshu.tech'
 }
 
 group = 'org.elasticsearch.plugin'
-version = pluginVersion + '_es' + project.ext.esVersion
-
-def storeAddress = System.env.ROR_LIBS_STORE_URL ?: ""
-if (storeAddress.startsWith("http")) {
-    System.setProperty("org.gradle.s3.endpoint", storeAddress)
-}
+version = "${pluginVersion}_es${project.ext.esVersion}"
 
 repositories {
     maven {
-        url = "s3://" + project.s3BucketName + "/ror/libs"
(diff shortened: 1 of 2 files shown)


build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
```diff
@@ -24,30 +24,16 @@ project.ext {
     publishedPluginVersion = rootProject.properties['pluginVersion']
```
mgoworko, line 24: 1. 
Using this config, Gradle sync still hangs in my environment when running locally in Intellij without credentials to s3 configured. 

I researched the issue. It seems, that the culprit is `org.bouncycastle:bctls-fips:2.1.22`. Its POM file specifies, that it needs `bcutil-fips` in range version [2.1.5-2.2.0) (instead of a single, defined version). As a result, Gradle queries all the maven repos for versions in this range, including s3. 

I added explicit dependencies locally, which solves the issue. Do you think we can add it?
```
    api group: 'org.bouncycastle',              name: 'bcutil-fips',                         version: '2.1.6'
    api group: 'org.bouncycastle',              name: 'bc-fips',                             version: '2.1.2'
```

I don't know what changed and caused that issue (Gradle resolution mechanism / IntelliJ behavior / bctls-fips POM file in the repo, though that is less likely).

EDIT: actually, I think the second issue with HTTP 403 on s3 may have triggered that issue, but I'm not sure

2. 
For es80x module I still get
```
Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not HEAD 'https://dgp.serve.beshu.tech/beshu/ror/libs/org/elasticsearch/plugin/transport-netty4/8.0.1/transport-netty4-8.0.1.jar'. Received status code 403 from server: Forbidden
```

mgoworko (changes requested):
I added comment about 2 issues with local development I still encounter

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
coutoPL: 1. done
2. Please check now

mgoworko: I confirm it works now, transport-netty4 is downloaded and Gradle does not freeze :+1:

mgoworko (approved):
LGTM

PR #1241, opened 2026-04-30
[RORDEV-2017] CVE-2026-42583, CVE-2026-42587, CVE-2026-42584, CVE-2026-42580, CVE-2026-42585, CVE-2026-42581, CVE-2026-41417, CVE-2026-34479, CVE-2026-34480

mgoworko:
🚨Security Fix (ES) 
- netty-related: CVE-2026-42583, CVE-2026-42587, CVE-2026-42584, CVE-2026-42580, CVE-2026-42585, CVE-2026-42581, CVE-2026-41417
- `log4j-core`-related (updated in our test code, but managed directly by ES, so suppressed): CVE-2026-34479, CVE-2026-34480

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Log4j vulnerability suppression: broadened suppressed CVEs and tightened artifact version matching.

* **Chores**
  * Adjusted build and dependency handling for logging libraries and swapped the shadowed libs module for a managed libs module.
  * Simplified CI triggers to run automatically only on master, develop, and epic branches.
  * Added internal logging utilities to standardize project logging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1241)


azure-pipelines.yml
```diff
@@ -16,9 +16,6 @@ trigger:
     include:
       - master
       - develop
-      - '*/epic/*'
```
mgoworko, line 19: Current situation:
- CI triggered for every branch pushed to the repo
- for PRs there are 2 CI runs for each commit, one of them runs the BUILD stage, the other does not

After this proposed change:
- branches without PRs do not trigger CI
- there is only one CI run per commit in PRs

core/build.gradle
```diff
@@ -72,7 +73,6 @@ dependencies {
     api group: 'io.jsonwebtoken',               name: 'jjwt-impl',                           version: '0.13.0'
     api group: 'io.jsonwebtoken',               name: 'jjwt-orgjson',                        version: '0.13.0'
     api group: 'org.jooq',                      name: 'joor',                                version: '0.9.15'
-    api group: 'org.apache.logging.log4j',      name: 'log4j-api-scala_3',                   version: '13.1.0'
```
mgoworko, line 75: The log4j-api-scala is now archived and no longer maintained. We use 3 files from that dependency. I copied those files to a new module 'logging' (with preserved apache license and with copyright link to github repo). They cannot be placed in the 'core' module, because of the Scala macro (the macro needs to be placed in different module than its usages, otherwise it causes compilation issues).

There are 2 alternatives:
- adding CVE suppression (but it is not a long-term solution - the dependency is no longer maintained since January, and the problems with it would return sooner or later)
- removing our dependency on that code and moving to Java APIs (but that would take longer than just quickly fixing the CVE warning)

mgoworko: Maybe I changed too much in this PR (more than is needed to fix, or rather suppress the warning). Especially the new module is a rather important side effect.  I'm open to discussion and making a minimal fix (just adding suppression)

suppressions_cve.xml
```diff
@@ -106,15 +106,18 @@ False positive: CVE-2021-34364 is for "Refined GitHub" (browser extension), not
...
     <suppress>
         <notes><![CDATA[
-   file name: log4j-api-2.22.1.jar
-   CVE-2025-68161 affects log4j-core's Socket Appender (TLS hostname verification), not log4j-api.
-   The plugin does not bundle log4j-core - it is provided by Elasticsearch at runtime.
-   log4j-api:2.22.1 is a transitive dependency from log4j-api-scala_3:13.1.0.
-   The actual vulnerability exists in Elasticsearch's log4j-core (version 2.19.0), which the plugin cannot fix.
-   Upgrading log4j-api to 2.25.3 without matching log4j-core causes runtime compatibility issues.
-   This must be fixed by Elasticsearch upgrading to log4j-core:2.25.3+.
+   file name: log4j-api-2.11.1.jar
```
mgoworko, line 109: now this suppression is about the CVEs from the real dependency (2.11.1), and not the transitive version (2.22.1) pulled because of `log4j-api-scala_3`

settings.gradle
```diff
@@ -28,6 +28,7 @@ assert JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_17) :
 rootProject.name = 'readonlyrest'
 include 'ror-shadowed-libs'
 include 'audit'
+include 'logging'
```
coutoPL: Maybe instead of creating another module (we have many of them ;)) we could have one for the libs we have to manage ourselves? 

Or even we could add this to the `ror-shadowed-libs` module (obviously after renaming it). WDYT?

azure-pipelines.yml
coutoPL: TBH, I don't understand this change. Sth has changed in the Azure pipeline?

settings.gradle
mgoworko: I tried to rename `ror-shadowed-libs` to `ror-managed-libs` and add those 3 logging-related classes there. But I simply cannot make it work. I encounter the same compilation issues as when I tried to put  those classes directly in the `core` module. 

The Scala compiler in our configuration can't handle the macros and throws multiple compilation errors (cyclic references/unresolved paths etc.). Claude was thinking for 40 minutes and burned 100k tokens (in 3 separate sessions) - also nothing useful came out of it. There are also additional issues when we mix shadowed and non-shadowed code in the same module (but those issues are resolvable).

azure-pipelines.yml
mgoworko: I don't know when it started (I guess some month ago), but there are now 2 separate CI runs for each commit (one triggered for branch commit, and second one triggered for commit in the PR - but it is the same commit). And we have limited resources for running jobs.

You can see in the list of CI runs on Azure, that for each commit there are 2 pipelines, (only one of them runs the build stage).

In this change I propose to run pipeline `only` for PR's and master and develop branches. That way there would be no doubled pipelines for PRs and the CI will not be triggered for branches without PRs. Perhaps this change should be in a separate PR...

coutoPL: ok, I found [this](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops#pr-triggers).

> If no pr triggers appear in your YAML file, pull request validations are automatically enabled for all branches

So, the change is ok. Just please revert this one: `- '*/epic/*'`.
Sometimes we have epic branches that should be tested in the same way as master and develop are.

settings.gradle
coutoPL: ok, what about renaming the `logging` module to `ror-managed-libs`? We could put similar cases there in the future. Would it work?

coutoPL (commented):
commented

settings.gradle
mgoworko: 1. Ultimately I managed to do it using a single module (`ror-managed-libs`) that contains both shadowed libs and logging classes and macros. It turns out, that there was a combination of some wrong Gradle options, local cache and CI issues with es94x module that made me think that is was impossible before, but after some fixes, full clean of all caches and rebase I managed to make it work

2. There were some more CVEs iuntroduced in the meantime. I added suppression for CVE-2026-34479 and bumped netty.

coutoPL (approved):
LGTM!

Please add a changelog-style entry to the PR description with all the CVEs fixed.

PR #1244, opened 2026-05-08
[RORDEV-1837] Handle global_checkpoint_sync as an internal action

mgoworko:
🐞 Fix (ES) The ES action `indices:admin/seq_no/global_checkpoint_sync` is now treated as an internal action and bypasses ACL evaluation. This action is dispatched by Elasticsearch internally after write operations and should never
  require explicit user permissions.

## Summary by CodeRabbit

* **New Features**
  * Added support for Elasticsearch global checkpoint synchronization as an internal operation, enhancing access control handling for background maintenance activities.

* **Tests**
  * Expanded test coverage for action classification to verify proper identification of internal operations, including global checkpoint synchronization and retention lease actions.


coutoPL (approved):
LGTM

PR #1246, opened 2026-05-10
[RORDEV-1895] Add support for traditional EC private key format in SSL cert loading

mateuszkp96:
🧐 Enhancement (ES) External SSL now supports EC private keys produced by dehydrated and similar ACME clients

Previously, SslContext only handled PKCS#8 and RSA formats.
This change adds support for the EC private key format
(BEGIN EC PRIVATE KEY) produced by dehydrated and openssl ecparam commands.

## Summary by CodeRabbit

* **New Features**
  * Improved SSL/TLS certificate handling: accepts multiple PEM private-key encodings for broader compatibility.

* **Documentation**
  * Added instructions for generating SSL test keys and certificates (EC and RSA variants).

* **Tests**
  * Added tests and fixtures validating loading of various PEM key formats and successful SSL context creation.

* **Chores**
  * Bumped tooling plugin version and updated ignore rules to track PEM test files.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1246)


core/src/test/scala/tech/beshu/ror/unit/settings/es/RorSslSettingsTest.scala
```diff
@@ -663,6 +721,28 @@ class RorSslSettingsTest
     }
   }
 
+  private def withForceLoad[A](esConfigYaml: String,
```
coutoPL: I'm not sure if we need this method. IMO, instead, we should do the following:
1. add `additionalFiles` (empty by default) param to `forceLoad` and `load` methods
2. get rig of `f: RorSslSettings => A` because AFAIS, it doesn't bring much

mateuszkp96: The withEsEnv method creates a temporary directory that is deleted when the block exits. I used a new method with a callback to keep the directory open and run assertions against private key files. Now I changed the tests to use absolute paths for resource files, so the assertions are not tied to the temporary directory's lifecycle.

coutoPL (approved):
LGTM

PR #1247, opened 2026-05-12
[RORDEV-2038] Optimizations

coutoPL:
* **🧐Enhancement** (ES) Slashed ACL evaluation overhead for wildcard index patterns: 49× more throughput, 98% lower p99 latency, and 50% less CPU compared to previous ROR version.
 
* **🐞Fix** (ES) Fixed `expand_wildcards` parameter being ignored during index resolution: ROR now correctly filters indices and aliases by their open/closed state when resolving wildcard patterns, preventing closed indices from leaking into rewritten requests
 
 ---
 
This PR focuses on reducing CPU usage and latency during ACL request evaluation, particularly for index-resolution-heavy workloads.

**Key optimizations**
  
**GlobPatternsMatcher — fast-path classification** 
Patterns are now classified at construction time into exact / prefix / suffix / infix / complex buckets. Simple patterns use string operations (equals, startsWith, endsWith, contains) instead of the full glob engine. A matchAll short-circuit is added for the wildcard-only case.

**LocalIndicesManager — per-request snapshot caching** 
A LocalIndicesSnapshot / LocalDataStreamsSnapshot is built once per request from the cluster state and exposes lazily-computed views (indices, aliases, open/closed, per-alias maps, etc.). Previously every call to allIndices, allAliases, allDataStreams, etc. re-fetched and re-filtered the same raw cluster state data independently.
  
**BaseIndicesProcessor — single-pass resolution**
Index/alias/data-stream names relevant to the current request are now resolved in a single resolveRequestedNames pass rather than firing 12 independent Task.sequence calls that each repeated the full index scan.

**Pre-computed hashCode on index domain classes**
IndexName.Full/Pattern, ClusterIndexName.Local/Remote, and cluster name types now cache their hashCode eagerly. These objects appear at high frequency in Set and Map operations.

**Set.sized / Set.mapFrom builders**
New pre-sized set construction helpers avoid repeated internal resize/rehash cycles when building sets of
known-size collections (e.g., index + aliases).

**FullLocalIndexWithAliases / FullLocalDataStreamWithAliases — eager fields**
Changed lazy val fields to val and use the new Set.mapFrom / Set.sized helpers for one-pass set construction.


sscarduzio:
## Performance impact estimate

Hot path targeted: per-request index/alias/data-stream filtering in `BaseIndicesProcessor` + `EsClusterService`. Runs on every indices-aware request.

### Per-commit gains

| Commit | Change | Est. impact |
|---|---|---|
| opt1 | `LocalIndicesSnapshot` precomputes opened/closed views + per-alias maps once per request | **High** — eliminates ~270K mutable-set ops/request on a 5K-index cluster |
| opt2 | Memoize `hashCode` on `ClusterIndexName.{Local,Remote}` | **Medium** — 2–5% CPU on hash-heavy paths |
| opt3 | `val` → `lazy val` in snapshots | Low — saves work when not all projections are used |
| opt4 | Memoize `hashCode` on `IndexName.Full/Pattern`, `ClusterName.Full/Pattern` | **Medium** — compounds with opt2 |
| opt5 | `mutable.Set` → `Set.newBuilder`, drop `view.mapValues.toMap` | Low — ~5–10% off the builder phase |
| opt6 | Try `ArrayBuffer` accumulator | (reverted by opt10) |
| opt7 | Sized `Set.newBuilder` in `FullLocal/RemoteIndexWithAliases` | Low–medium — 3–8% off snapshot construction |
| opt8 | **Big refactor:** resolve requested names once, skip 4 always-empty tasks, spawn only relevant ones | **Very high** — ~50% fewer `Task` allocations, **15–30%** off per-request ACL time |
| opt9 | Short-circuit empty `requestedAliases`/`requestedDataStreams` | **Medium** — 5–15% on index-only requests |
| opt10 | Revert opt6 | net zero |
| opt11 | Memoize per-request alias maps in `RemoteIndicesManager` + fast-path for `allIndexAttributes` | **High** (CCR only) — comparable to opt1 for remote indices |

### Bottom line

Realistic projection for a busy local-only cluster (~5K indices, ~3 aliases each, alias-heavy queries):

- **ROR ACL eval time / request: ~30–50% reduction** (compounding opt1 + opt2/4 + opt8 + opt9)
- **CPU at p99 under load: ~15–25% reduction** of the ROR overhead specifically
- **Wall-clock latency: ~5–15%** on indices-heavy ops (depends on ACL share of total request time)
- **GC pressure: meaningfully lower** — fewer transient `Map`/`Set` per request, 4 no-op `Task`s/req gone

For CCR / cross-cluster workloads, opt11 roughly doubles the effective gain on those paths. Pure-local workloads see no change from opt11.

### Caveats

- Estimates are static analysis only — no JMH/load test numbers yet
- The 2 tip `fix` commits and the `wip` commit should be reviewed; they may walk back some gains for correctness
- `hashCode` memoization adds ~4 bytes/instance — trivial heap overhead
- Snapshot `lazy val`s in opt1 precompute partitioned views eagerly on first access; ideal would be even more on-demand if some projections are rarely used

To turn estimates into numbers: a JMH harness around `BaseIndicesProcessor.canPass` for representative cluster sizes (1K / 5K / 20K indices, with/without aliases, with/without remote) would let us validate each commit's contribution in isolation.

coutoPL:
@sscarduzio analysis of results from my docker-based tests:

```
Latency (steady state)

  ┌────────┬─────────┬─────────┬─────────────┐
  │ Metric │ Old ROR │ New ROR │ Improvement │
  ├────────┼─────────┼─────────┼─────────────┤
  │ mean   │ ~407ms  │ ~96ms   │ −76% (4.2×) │
  ├────────┼─────────┼─────────┼─────────────┤
  │ p50    │ ~335ms  │ ~80ms   │ −76% (4.2×) │
  ├────────┼─────────┼─────────┼─────────────┤
  │ p95    │ ~983ms  │ ~195ms  │ −80% (5.0×) │
  ├────────┼─────────┼─────────┼─────────────┤
  │ p99    │ ~1440ms │ ~335ms  │ −77% (4.3×) │
  └────────┴─────────┴─────────┴─────────────┘
  
  Throughput

  Window req/s: ~50 req/s → ~170 req/s — 3.4× more throughput with the same 20 workers.
  
  CPU
  
  ┌────────┬─────────┬─────────┬─────────────┐
  │ Metric │ Old ROR │ New ROR │ Improvement │
  ├────────┼─────────┼─────────┼─────────────┤
  │ mean   │ ~586%   │ ~479%   │ −18%        │
  ├────────┼─────────┼─────────┼─────────────┤
  │ p50    │ ~588%   │ ~474%   │ −19%        │
  ├────────┼─────────┼─────────┼─────────────┤
  │ p95    │ ~599%   │ ~550%   │ −8%         │
  └────────┴─────────┴─────────┴─────────────┘

  (Values are % of one core — Docker on macOS convention, so ~600% ≈ 6 cores saturated.)
  
  CPU efficiency (req/s per CPU%)
  
  Old: 50 / 586 = 0.085
  New: 170 / 479 = 0.355 → 4.2× better CPU efficiency

  The new ROR does 4× the work at 18% lower CPU. The latency distribution also tightened significantly — the old p99 (1440ms) is worse
  than the new p95 (195ms). 
```

sscarduzio:
Impressive @coutoPL! This would make it a great blog post :) 
Reminds me of the back & forth of the early days I had with CERN guys

coutoPL:
@claude review please

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/indices.scala
```diff
@@ -570,21 +582,53 @@ final class FullLocalIndexWithAliases(val indexName: IndexName.Full,
...
+  def from(expandWildcardsOpen: Boolean, expandWildcardsClosed: Boolean): IndexAttributeFilter = {
+    (expandWildcardsOpen, expandWildcardsClosed) match {
+      case (true, true) => IndexAttributeFilter.All
+      case (false, false) => IndexAttributeFilter.All
+      case (true, false) => IndexAttributeFilter.Opened
+      case (false, true) => IndexAttributeFilter.Closed
+    }
+  }
+
+  implicit val monoid: Monoid[IndexAttributeFilter] = Monoid.instance(
```
mgoworko: I got suspicious about this monoid. I think that the neutral (identity) element of this monoid may be incorrect. I think that there is missing NoWildcards (Neutral) case, or something like that. It is more visible in the def from method above, where we have now case:
case (false, false) => IndexAttributeFilter.All - IMO it should be IndexAttributeFilter.NoWildcards

```
  implicit val monoid: Monoid[IndexAttributeFilter] = Monoid.instance(
    emptyValue = IndexAttributeFilter.NoWildcards,
    cmb = {
      case (x, y) if x == y  => x
      case (NoWildcards, y)  => y
      case (x, NoWildcards)  => x
      case _                 => All
    }
  )
```

I think that the newly added tests verify the new behavior, not the one from before this PR

--------------------

**I asked Claude for an in-depth analysis of this case and my proposed fix:**

```The regression requires a MultiSearch where at least one sub-request has both expandWildcardsOpen=false and expandWildcardsClosed=false (i.e. expand_wildcards=none), while another has a real filter like expand_wildcards=open.```

Image

File: core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/LocalIndicesManager.scala [hunk 2]
```diff
@@ -29,43 +29,33 @@ class LocalIndicesManager(requestContext: RequestContext,
...
-    if (indexAttributesFromRequest.isEmpty) indicesSnapshot.indicesAndAliases
-    else attributeFilteredIndices.flatMap(_.all)
-  }
-  private lazy val cachedAllIndices: Set[LocalIndexName] = {
-    if (indexAttributesFromRequest.isEmpty) indicesSnapshot.indices
-    else attributeFilteredIndices.map(_.index)
-  }
-  private lazy val cachedAllAliases: Set[LocalIndexName] = indicesSnapshot.aliases
+  private lazy val cachedAllIndices: Set[LocalIndexName] =
+    indicesSnapshot.indicesFor(indexAttributesFromRequest)
```
mgoworko, line 39: Aliases were not filtered before this PR. Why has it changed? Is it not a change in behavior?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/indices/clusterindices/LocalIndicesManager.scala
```diff
@@ -29,43 +29,33 @@ class LocalIndicesManager(requestContext: RequestContext,
...
-    if (indexAttributesFromRequest.isEmpty) dataStreamsSnapshot.dataStreams
-    else attributeFilteredDataStreams.map(_.dataStream)
-  }
-  private lazy val cachedAllDataStreamAliases: Set[LocalIndexName] = dataStreamsSnapshot.dataStreamAliases
+  private lazy val cachedAllDataStreamsAndAliases: Set[LocalIndexName] =
+    dataStreamsSnapshot.dataStreamsAndAliasesFor(indexAttributesFromRequest)
+  private lazy val cachedAllDataStreams: Set[LocalIndexName] =
+    dataStreamsSnapshot.dataStreamsFor(indexAttributesFromRequest)
+  private lazy val cachedAllDataStreamAliases: Set[LocalIndexName]
+  = dataStreamsSnapshot.dataStreamAliasesFor(indexAttributesFromRequest)
```
mgoworko: same here

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/indices.scala
```diff
@@ -231,7 +242,9 @@ object ClusterIndexName {
         val local: Full = Full(NonEmptyString.unsafeFrom("(local)"))
       }
 
-      final case class Pattern private(value: NonEmptyString) extends ClusterName
+      final case class Pattern private(value: NonEmptyString) extends ClusterName {
+        override val hashCode: Int = scala.runtime.ScalaRunTime._hashCode(this)
```
mgoworko: Perhaps some helper trait with runtime optimizations can be added? It would encapsulate the static hash logic

mgoworko (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/indices.scala
coutoPL: you, are right. I get rid of the monoid.

But, this is done by purpose:
```
// expand_wildcards=none: ROR must still resolve concrete names against all indices (open+closed) for ACL evaluation
      case (false, false) => IndexAttributeFilter.All
```

I've added a comment with the description of why

sscarduzio (commented):
Great work on this PR — the benchmark numbers are impressive and the architectural direction (snapshot caching, single-pass resolution, fast-path pattern classification) is exactly right. A few actionable suggestions for the top items I noticed:

**1. `buildAliasMap` is defined three times**
The same mutable-HashMap-to-immutable-Map helper appears identically in `EsClusterService` companion, `LocalIndicesSnapshot`, and `RemoteIndicesManager`. Moving it to the `EsClusterService` companion (or a small private utility) and delegating from the other two would remove the duplication.

**2. `LocalIndicesSnapshot` scans `raw` twice for open/closed**
```scala
private lazy val opened = raw.filter(_.attribute == IndexAttribute.Opened)
private lazy val closed = raw.filter(_.attribute == IndexAttribute.Closed)
```
A single `partition` call would do both in one pass:
```scala
private lazy val (opened, closed) = raw.partition(_.attribute == IndexAttribute.Opened)
```

**3. `ScalaRunTime._hashCode` is an internal API**
```scala
override val hashCode: Int = scala.runtime.ScalaRunTime._hashCode(this)
```
This works reliably today, but `_`-prefixed Scala runtime methods are implementation details that can change between Scala versions. A short comment explaining why this is safe (or a link to the Scala source) would help future readers, or you could compute it explicitly per class to be fully safe.

**4. Formatting glitch in `LocalIndicesManager`**
```scala
private lazy val cachedAllDataStreamAliases: Set[LocalIndexName]
= dataStreamsSnapshot.dataStreamAliasesFor(indexAttributesFromRequest)
```
The `=` ended up on a new line — the rest of the file keeps it on the same line as the declaration.

**5. Consider a unit test for `GlobPatternsMatcher.Kind.of` edge cases**
The property-based tests cover overall matcher correctness well, but a small targeted test for the dispatch logic itself (e.g. `"**"`, `"a**b"`, single-char patterns like `"*"` vs `"?"`) would make future changes to the fast-path classification safer to refactor.

Overall this is a high-quality, well-tested PR — the new `LocalIndicesManagerTest`, `RemoteIndicesManagerTest`, and snapshot tests are excellent additions.

sscarduzio:
## Re-review

All five points addressed, three of them better than my original suggestion:

| # | Concern | Resolution |
|---|---|---|
| 1 | Triple `buildAliasMap` duplication | ✅ Extracted as a Scala 3 extension `drainToMap` on `mutable.HashMap[K, mutable.Builder[V, C]]` in `ScalaOps.scala`. More general than what I asked for — value type is generic `C` rather than fixed `Set[V]`. Both `EsClusterService` and `RemoteIndicesManager` call it. Third copy is gone. |
| 2 | Double scan for opened/closed | ✅ Single pass via `foldLeft` (see note below) |
| 3 | `ScalaRunTime._hashCode` is internal API | ✅✨ Excellent fix — new private `EagerHashCode` trait using the public `MurmurHash3.productHash`. Removes the internal API, dedupes the override across six case classes, and matches exactly what scalac generates for case classes. Strictly better than the comment-and-leave-it option I suggested. |
| 4 | `=` on new line | ✅ |
| 5 | Tests for `Kind.of` edge cases | ✅ Added `**`, `a**`, `a**b`, `?` (positive + two negatives), `*?` (positive + negative on empty). Good coverage. |

**Plus** the new `expand_wildcards` integration test in `SearchApiSuite` exercising `idx-open`/`idx-closed` with `expand_wildcards=open` — that validates the second headline fix in the PR description, which was previously only covered indirectly. Nice addition.

### One small note on #2

```scala
private lazy val (opened, closed) =
  raw.foldLeft((Set.empty[FullLocalIndexWithAliases], Set.empty[FullLocalIndexWithAliases])) {
    case ((o, c), index) => index.attribute match {
      case IndexAttribute.Opened => (o + index, c)
      case IndexAttribute.Closed => (o, c + index)
    }
  }
```

Functionally correct and runs in a single pass — good. Two minor points:

- `Set + element` on an immutable `Set` allocates a new Set per element. For a 5K-index cluster that's 5K intermediate Sets being garbage-collected. `raw.partition(_.attribute == IndexAttribute.Opened)` would be the simpler one-liner; for max throughput, `mutable.Set.Builder` + `.result()` at the end is the Scala-idiomatic high-performance version. Given that the *purpose* of this PR is throughput, the `partition` form (or builders) seems more on-message.
- The match drops any `IndexAttribute` that is neither `Opened` nor `Closed`. Safe iff the sum type is sealed with exactly those two — true today, but a `case other => …` (or sealed-exhaustivity assertion) would future-proof against silent loss if a third state is ever added.

### Also noticed

The `IndexAttributeFilter` monoid → `from(IterableOnce)` swap is a small cleanup that removes a `cats.Monoid` import without changing semantics — fine. The new comment `// expand_wildcards=none: ROR must still resolve concrete names against all indices (open+closed) for ACL evaluation` is a great piece of institutional knowledge to capture; that's exactly the kind of "why" that gets lost otherwise.

LGTM once CI is green. The `foldLeft` → `partition` (or builder) suggestion is small and non-blocking.

sscarduzio (approved):

mgoworko (approved):
LGTM!

PR #1249, opened 2026-05-18
Claude GH Actions jobs improvements

coutoPL:
## Summary

- **`claude.yml`**: Added `author_association` check to every trigger branch (issue comment, PR review comment, PR review, issue open/assign). Without this, any public GitHub user could invoke `@claude` and get a shell session backed by `CLAUDE_CODE_OAUTH_TOKEN`. Allowed associations: `OWNER`, `MEMBER`, `COLLABORATOR`.
- **Both workflows**: Replaced `Bash(gh:*)` with specific subcommand allowlists to close prompt-injection exfiltration paths (`gh auth token`, `gh secret`, `gh workflow run`, etc.):
  - `claude.yml` (interactive): `Bash(gh:pr*),Bash(gh:issue*),Bash(gh:run view*),Bash(gh:run list*)`
  - `claude-code-review.yml` (automated review): `Bash(gh:pr view*),Bash(gh:pr diff*),Bash(gh:pr comment*)`

## Threat model

Before this change, a PR body or issue comment containing a prompt-injection payload could trigger an unrestricted `gh` shell in the CI runner, then call `gh auth token` to print the OAuth token and exfiltrate it (e.g. via a `gh issue comment`). The author-association gate is the primary defence; the narrowed allowlist is a second layer that limits blast radius even if injection bypasses the gate.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -43,4 +43,6 @@ jobs:
           show_full_output: true
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
+          # Restrict to gh CLI only — prevents prompt injection in PR diffs from running arbitrary shell commands
+          claude_args: '--allowedTools Bash(gh:*)'
 
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -43,6 +43,6 @@ jobs:
           show_full_output: true
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
-          # Restrict to gh CLI only — prevents prompt injection in PR diffs from running arbitrary shell commands
-          claude_args: '--allowedTools Bash(gh:*)'
+          # Restrict to read PR diff and post review comments only — prevents prompt injection in PR diffs from running arbitrary shell commands
+          claude_args: '--allowedTools Bash(gh:pr view*),Bash(gh:pr diff*),Bash(gh:pr comment*)'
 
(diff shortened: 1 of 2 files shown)


sscarduzio:
Fix makes sense — the original `claude.yml` trigger had no author check, so any GitHub user could `@claude` on a public repo and run a Claude session with our OAuth token + Bash access. Prompt-injecting via PR/issue body to exfiltrate the token was very feasible. The author-association allowlist + `--allowedTools Bash(gh:*)` together close the primary hole and add a sensible second layer. 👍

A few actionable points:

1. **PR description is empty.** For a security fix this should spell out the threat model (anyone-can-trigger → prompt-injection → token exfiltration) so the rationale isn't lost. Future maintainers will otherwise be tempted to relax the allowlist or widen `--allowedTools`.

2. **`fetch-depth: 0` in `claude-code-review.yml` is unrelated to the security fix.** Either move to a separate PR or mention it in the description (why does the reviewer need full history?).

3. **`Bash(gh:*)` is still pretty broad.** It permits `gh api …` (arbitrary API calls under whatever scopes the OAuth token has), `gh auth token` (prints the token to stdout — easy injection-exfil sink), `gh secret`, `gh workflow run`, etc. Worth narrowing to something like `Bash(gh pr:*, gh issue:*, gh api repos/${{ github.repository }}/*:*)` or similar — whatever the actual Claude review workflows need.

4. **Nit: redundant guard in `claude-code-review.yml`.** That file is triggered by `pull_request` and already gates on `pull_request.author_association`. The new `claude_args` line is the only real change there; flagging in case the symmetry with `claude.yml` masked a no-op.

Questions:

- Do we have any other workflows in this repo (or sibling repos) using `pull_request_target` with a similar `@claude` trigger? That event runs with secrets even on forks — if any exist, they need the same treatment.
- What scopes does `CLAUDE_CODE_OAUTH_TOKEN` actually have? If it's narrow (read-only PRs/issues), `gh:*` is lower-risk; if it can write to repos, point #3 becomes more important.
- Did you confirm `author_association: CONTRIBUTOR` is correctly excluded by the allowlist? GitHub assigns that to anyone with a merged PR, so it's the boundary case worth a quick test.

coutoPL:
@sscarduzio 
1. & 3. - addressed
2. no changes
4. yes, it's redundant, but I think we should leave it as it's

PR #1260, opened 2026-06-07
[RORDEV-2075] [perf] Benchmarked ACL hot-path optimizations: static-config pre-resolution (incl. indices rule), evaluation short-circuit, allocation-free matching

sscarduzio:
🧐**Enhancement** (ES) faster ACL request evaluation, especially for configurations with many blocks, static rule values, header rules and JWT authentication

Jira: RORDEV-2075. Follow-up to #1247 / #1233. Every row was benchmark-validated during review (JMH + `-prof gc`; the benchmark suite now lives in #1269, which lands before this PR).

## Optimizations ranked by per-request impact — opt any row out by reverting its commit

Sorted by the first column — estimated µs saved per typical request\*.

| Saved per request\* | Optimization | Better by (benchmarked) | Alloc saved | Risk | Opt-out |
|---|---|---|---|---|---|
| **~6.4 µs** | **`IndicesRule` pre-resolution** (2.12 µs × 3 blocks) | ~12,500% per evaluation | −26 KB | medium surface; 211/211 tests | `5dfa9e422` |
| **~4.1 µs** | Rule/block evaluation short-circuit | ~50% end-to-end | −13.0 KB | low — history proven identical | `4590a98d6` |
| **~2.2 µs** | `basicAuth` decoded once per request | ~400% | −11.1 KB | low | `0bbd75863` |
| **~2.1 µs** | `Header.Name` cached lowercase (2 scans/req) | ~5,600% per scan | −4.6 KB | low | `993c0b8fb` |
| **~1.5 µs** | Static resolve bypass: `users`/`data_streams`/`repositories`/`snapshots` | ~4,400–7,800% per evaluation | → 0 | none — dynamic path untouched | round 1 |
| **~1 µs** (only FLS configs) | `fields`/`response_fields` pre-resolution | ~4,400–7,800% per evaluation (identical code path as the row above) | → 0 | low | `4c49fa8eb` |
| **sub-µs** | Glob match inner loop | ~30–150% | ≈ 0 match-path | none | round 1 |
| **sub-µs** | `headers_and`/`or` compiled matchers | ~10% | −55% | none | round 1 |
| **sub-µs** (+ correctness) | Kibana regex `Pattern.quote` + cache | regex compile off hot path | cached | ⚠️ only semantic change (see note) | round 1 |
| **~0.2 µs** per JWT block | JWT/ROR-KBN parser hoist | ~2–4% | ~1% | ~zero | `8aa1cfa0d` |
| **±, workload-dependent** | `EagerHashCode` | ~87% better hot lookups / ~15% worse create-once | = | 🤔 **discuss** — cheap to drop | `aa4e5bb1c` (r2 part) |
| n/a | JMH benchmarks (6 classes) | evidence only | n/a | none | `088088f99` |

\* Request model: 10-block mostly-static config, 20 request headers, ~5 `auth_key`-family blocks, indices rule in ~3 blocks (Kibana-style traffic). Apple Silicon, JMH f=1 wi=3 i=4; ratios are the signal.

**Bottom line: ~16 µs and ~55 KB allocation saved per typical request** (sum of point estimates above) — roughly 2× on ACL evaluation overhead for static configs, larger under GC pressure.

## Future-work evidence (NOT in this PR)

- **Sync-first/no-WriterT rewrite**: hand-rolled evaluator does the same 10-block decision in 2.71 µs vs 7.89 (**~190% better**, −65% alloc); the per-request scheduler hop costs another **~10.6 µs** → ~580% bound combined (`AclEvaluationBenchmark`).
- **JWT claims cache**: signature verification is **5–45 µs per block per request** (`JwtParserBenchmark`) — needs security sign-off.

## Behavior note (the only semantic change; core 2521/0 green)

Kibana non-strict-path regex now `Pattern.quote`s the index name: stricter for `.` in index names; regex-metacharacter names that previously failed to compile (never matched) now match literally.


sscarduzio:
## Addressed review feedback in c3b7bbc17

### ⚠️ `resolveAll` still firing per-request on the static path (Claude + CodeRabbit)

Fixed in **DataStreamsRule**, **RepositoriesRule**, **SnapshotsRule**. The root cause was exactly as Claude framed it: the precomputed matcher was only consulted *inside* `checkAllowed*`, while the caller still ran `resolveAll(...).toCovariantSet` unconditionally — allocating a `List` + `Set` per request that the static path never used, and running the wildcard sentinel check against the dynamically-resolved set.

All three now mirror the `UsersRule` shape — `resolveAll` is gated behind `getOrElse`, so the static path does **zero** per-request resolution:

```scala
val allowedSnapshots = staticAllowedSnapshots.getOrElse {
  AllowedSnapshots.from(resolveAll(settings.allowedSnapshots.toNonEmptyList, blockContext).toCovariantSet)
}
```

I went with Claude's "bypass entirely" approach rather than CodeRabbit's "keep `resolveAll`, add a second cached `Set`" — the wildcard short-circuit flag and the matcher are bundled into a small `Allowed*` holder computed once for static configs. This also removes the secondary issue Claude noted (the wildcard `contains` check now runs against a precomputed flag, not a freshly-allocated set). The holder's `matcher` is `lazy`, so a static *wildcard* config never pays to build a matcher it won't use.

> **Open question 1 — intentional or oversight?** Oversight. It was a partial optimization; now consistent with `UsersRule`.

### Nit: non-atomic cache population in `BaseKibanaRule`

Fixed — switched to Caffeine's atomic `get(key, loader)`. As Claude noted, `None` (the compile-failure sentinel) is a valid non-null value so it round-trips through the cache fine. Compile-once is now guaranteed under concurrent load. (The pre-existing `kibanaIndicesRegexesCache` in `indices.scala` still uses the older `getIfPresent`/`put` shape; left out of scope here.)

### Open question 2 — benchmark Scala compiler / Gradle resolution

Confirmed **not a build defect**. I ran it locally end-to-end on this branch:

```
./gradlew :benchmarks:jmh -PjmhArgs="GlobPatternsMatcher -f 1 -wi 1 -i 1 -r 1s -w 1s"
→ BUILD SUCCESSFUL — all 4 benchmarks executed and reported numbers
```

Gradle's `scala` plugin resolves the Scala 3 compiler automatically from the `scala3-library_3:3.3.7` dependency declared in `benchmarks/build.gradle` (same mechanism every other module uses via `base-common-conventions`). The JDK 17 toolchain is consistent with the rest of the repo (core=11, audit=8), provisioned via the `foojay-resolver-convention`.

The "can't run the benchmark" in the PR description was an **environment** limitation, not a config bug: the original authoring sandbox blocked the Gradle plugin portal / foojay toolchain download. With network access it builds and runs cleanly, so no change to `benchmarks/build.gradle` was needed.

### Validation

- `:core:compileScala` — clean (incl. `-Xfatal-warnings`)
- `DataStreamsRuleTests`, `RepositoriesRuleTests`, `SnapshotsRuleTests`, `UsersRuleTests` — 42 passed
- `KibanaAccessRuleTests`, `KibanaUserDataRuleTests` — 62 passed

No behavior change: structural equality and ACL decisions are unchanged; only per-request work is reduced.

sscarduzio:
## Quantified: the `resolveAll`-bypass optimization

Added `RuleStaticResolutionBenchmark` (commit c3ade5303) to put a number on the per-request win for **statically configured** `DataStreamsRule`/`RepositoriesRule`/`SnapshotsRule`. It exercises the two real code paths through the production helpers (`resolveAll`, `staticallyResolvedValues`, `PatternsMatcher.create`) — no rule/mock machinery, because a static `AlreadyResolved.resolve` never touches the `BlockContext`.

- **old path** — what the rules did before: `resolveAll(...).toCovariantSet` + wildcard check + `PatternsMatcher.create(set)` **on every request**
- **new path** — what they do now: reuse the matcher precomputed once at construction; per request is a field read + a `match` call

### Results (JDK 25, `-f 2 -wi 5 -i 5 -prof gc`)

| static config | old µs/op | new µs/op | speedup | old B/op | new B/op | alloc cut |
|---|--:|--:|--:|--:|--:|--:|
| few patterns (3) | 0.402 | 0.009 | **~45×** | 2,288 | ~0 | **−100%** |
| many patterns (30) | 4.426 | 0.056 | **~79×** | 14,120 | ~0 | **−100%** |

So the static per-request path drops from **~0.4–4.4 µs to single-digit nanoseconds**, and from **2.3–14 KB/op of garbage to zero**. The matcher build (glob classification + compilation) and the intermediate `List`/`Set` allocations now happen once at rule construction instead of on every request.

---

### For completeness — the `GlobPatternsMatcher` rewrite (commit 69cbe40c3, `GlobPatternsMatcherBenchmark`)

Measured the same way (before = `Vector` + `exists(closure)`, after = `Array` + `while`):

| benchmark (2000 candidates) | before µs/op | after µs/op | speedup | alloc cut |
|---|--:|--:|--:|--:|
| `matchEach_fewPatterns` (3, CS) | 48.1 | 19.2 | **2.50×** | −100% (253 KB→0.1 B) |
| `matchEach_manyPatterns` (30, CS) | 144.8 | 109.8 | **1.32×** | −100% (227 KB→0.6 B) |
| `matchEach_manyPatterns` (30, CI) | 172.4 | 132.3 | **1.30×** | −100% (227 KB→0.8 B) |
| `filterAll_manyPatterns` (30, CS) | 192.7 | 149.0 | **1.29×** | −78% (293 KB→66 KB) |

Both optimizations attack the same thing from different angles — per-request allocation on the ACL hot path — and both drive it to ~zero for the common static-config case.

> Numbers are from a dev laptop (JDK 25); treat the ratios as the signal, not the absolute µs.

sscarduzio:
## Addressed the latest review findings in 0c2495b0f

Thanks for the thorough re-reviews. Walking through each:

| Finding | Action |
|---|---|
| **`GlobPatternsMatcher.Compiled` is a `case class` with `Array` fields** (broken derived `equals`/`hashCode`/`copy`) | ✅ Fixed — converted to a plain `final class`. It's created once and never compared/copied so there was no active bug, but the value-like façade was a latent trap; a plain class makes the intent explicit. |
| **`BaseHeaderRule.isFulfilled` silently falls back to per-request matcher construction** | ✅ Fixed — dropped the `getOrElse(..., matcherFor(...))` in favor of a direct `compiledMatcherByRequirement(req)` lookup. A subclass passing an un-enumerated requirement is now a fail-fast `NoSuchElementException` instead of a silent perf regression. |
| **`kibanaIndicesRegexesCache` uses non-atomic `getIfPresent`/`put`** (pre-existing) | ✅ Fixed — switched to Caffeine's atomic `get(key, loader)`, matching the example this PR set in `BaseKibanaRule`. |
| **`UsersRule` static path doesn't `.toSet`** (inconsistent with sibling rules; possible duplicate patterns) | ✅ Fixed — `.toSet` before `PatternsMatcher.create`, now consistent with the dynamic path and the DataStreams/Repositories/Snapshots holders. |
| **`ToBeResolved` (dynamic) path untested in the refactored rules** | ✅ Added coverage — `DataStreamsRuleTests` now has a permit + deny case using an `@{user}` runtime variable, exercising the `getOrElse { resolveAll(...) }` fallback end-to-end (the path the static-precompute optimization bypasses). DataStreams is representative; Repositories/Snapshots share the identical holder pattern. |
| **`benchmarks/build.gradle` pulls `jitpack.io` + gradle libs-releases** unnecessarily | ✅ Trimmed — removed both; no declared or transitive dependency needs them. Kept `mavenCentral` + the S3 bucket (required for `:core`'s transitive ES/shaded artifacts). Verified `:benchmarks:jmh` still resolves and runs. |
| **Missing trailing newline in `RepositoriesRule.scala` / `SnapshotsRule.scala`** | ✅ Fixed. |

### Rejected / not changed (with reason)

- **`benchmarks` compiles on every `./gradlew build`** — intentional. Keeping the benchmark in the build graph means an API break in the benchmark surfaces immediately rather than rotting; the `jmh` task itself is not wired into `check`/`assemble` so benchmarks don't *run* in CI.
- **`nonStrictAllowedPathsPatternCache` `maximumSize(1000)`** — left as-is; it mirrors the sibling `kibanaIndicesRegexesCache` bound, and the headroom is harmless (a handful of distinct Kibana index names in practice).
- **`KibanaIndexName` missing `EagerHashCode`** (open question) — intentionally out of scope for this PR; its `hashCode` already chains through `ClusterIndexName.Local` (which has `EagerHashCode`), so the cost is already low. Candidate for a future sweep.

### Validation
`:core:compileScala` clean (incl. `-Xfatal-warnings`); `PatternsMatcherTest`, `HeadersAndRuleTests`, `UsersRuleTests`, `KibanaAccessRuleTests`, `KibanaUserDataRuleTests`, `DataStreamsRuleTests` (now 14 incl. the 2 dynamic cases) all green; `:benchmarks:jmh` resolves + runs after the repo trim.

sscarduzio:
## Addressed the latest review + ran a security analysis — `a44d3a0da`

I ran a multi-agent **security analysis** of this PR (threat model: any optimization that changes ACL matching/authorization *semantics* = a vulnerability). Bottom line: **all shipped changes are semantics-preserving and safe** — the static-precompute is correctly gated by `staticallyResolvedValues` (returns `None` if *any* variable is `ToBeResolved`, so nothing `@{...}`/user/group-dependent is ever frozen), `EagerHashCode` caches the exact `MurmurHash3.productHash` so it can't diverge from `equals`, the glob rewrite is a faithful translation, and the cache key fully determines its value. No blockers. The follow-ups below address the residual hardening points from the review:

| Review finding | Action in `a44d3a0da` |
|---|---|
| **`Repositories`/`Snapshots` dynamic path untested** | ✅ Added permit + deny `ToBeResolved` (`@{user}`) tests to both, mirroring `DataStreamsRuleTests`. The `getOrElse { resolveAll(...) }` fallback in all three rules is now covered — a regression that froze a dynamic config to a static matcher would now fail a test. |
| **`BaseHeaderRule` direct `Map.apply` is an opaque runtime trap** | ✅ Replaced the bare `NoSuchElementException` with `getOrElse(throw new IllegalStateException("Precompilation miss for … must enumerate every requirement …"))`. Fail-fast intent preserved, failure now legible. Behavior unchanged for both existing subclasses (lookup always hits). |
| **Open Q1: `@kibana_index` substituted into regex without `Pattern.quote`** | ✅ Good catch — fixed. `@kibana_index` only appears as a literal path segment, so the name must match literally; without quoting, the `.` in `.kibana` acted as a wildcard and a name with regex metachars would compile wrong or fail. Now `Pattern.quote(kibanaIndexName.stringify)`. All Kibana suites still green. **Scope note:** the sibling `kibanaIndicesRegexesCache` in `indices.scala` has the same pre-existing pattern across ~20 interpolations — left for a separate focused change to avoid per-line regex surgery in this perf PR. |

### Rejected / not changed
- **Nit: `AllowedXxx` retains the backing `Set` in the wildcard case** — it's a one-element set for typical wildcard configs (`*`/`_all`); negligible, and removing it would complicate the lazy-matcher holder. Left as-is.
- **Nit: `matchesExact` redundant `nonEmpty` guard** — harmless; `Set.contains` on empty is O(1) anyway. Not worth the churn.
- **Open Q2: benchmark needs S3 for `:core` transitives at compile time** — yes, intentional; the benchmark depends on `:core`, and the S3 bucket is the same one every module resolves against. Kept `mavenCentral` + S3 (trimmed jitpack/gradle-libs in the prior commit).

### Security tests on the roadmap
The analysis specified must-have gates for the *future* optimization roadmap (esp. JWT): any new precompute must gate on `staticallyResolvedValues` and **never** be applied to group logic (per-group `@{acl:current_group}` re-resolution must not be frozen); JWT work must re-verify per-rule/per-key and never turn a parse-skip into a verify-skip. Those tests will land with that work, not here.

Validation: `:core` compiles clean (`-Xfatal-warnings`); 127 tests across the touched rule, header, and Kibana suites pass (incl. 6 new dynamic-path cases).

core/src/main/scala/tech/beshu/ror/accesscontrol/utils/RuntimeMultiResolvableVariableOps.scala
```diff
@@ -33,4 +35,20 @@ object RuntimeMultiResolvableVariableOps {
...
   }
+
+  /**
+   * Returns the fully resolved values when every variable is a static constant
+   * (i.e. does not depend on the request/block context). In that case the result
+   * is identical for every request, so structures derived from it (e.g. a
+   * `PatternsMatcher`) can be built once up front instead of on every request.
+   * Returns `None` when at least one value must be resolved per request.
+   */
+  def staticallyResolvedValues[T](variables: NonEmptyList[RuntimeMultiResolvableVariable[T]]): Option[List[T]] =
```
coutoPL: In this codebase, we prefer well-named methods and meaningful types over detailed descriptions in the comments. The reader should be able to figure out what the method does by reading its name, parameter names and types, and return type. 

In this example, we could improve two things:
1. the method name. If would propose: `resolveAllIfPreResolved` 
2. the return type should be `Option[NonEmptyList[T]]`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/UsersRule.scala
```diff
@@ -27,14 +27,20 @@ import tech.beshu.ror.accesscontrol.blocks.variables.runtime.RuntimeMultiResolva
...
-import tech.beshu.ror.accesscontrol.utils.RuntimeMultiResolvableVariableOps.resolveAll
+import tech.beshu.ror.accesscontrol.utils.RuntimeMultiResolvableVariableOps.{resolveAll, staticallyResolvedValues}
 
 class UsersRule(val settings: Settings,
                 implicit val userIdCaseSensitivity: CaseSensitivity)
   extends RegularRule {
 
   override val name: Rule.Name = UsersRule.Name.name
 
+  // Built once when the allowed user ids are statically configured (no runtime
```
coutoPL: In this codebase, we prefer easy-to-read code over comments explaining the code in the human language. IMO, the code below is so easy to understand that comments in this style are not needed.
Nevertheless, it would be nice to leave a comment that this is an optimization done on purpose. We add a little bit of complexity intentionally to optimize the performance.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/UsersRule.scala
```diff
@@ -27,14 +27,20 @@ import tech.beshu.ror.accesscontrol.blocks.variables.runtime.RuntimeMultiResolva
...
+  // variables); otherwise the matcher is created per request from the resolved values.
+  private val staticUserIdsMatcher: Option[PatternsMatcher[User.Id]] =
+    staticallyResolvedValues(settings.userIds.toNonEmptyList)
+      .map(values => PatternsMatcher.create(values.toSet)) // `.toSet` to dedup, consistent with the dynamic path below
```
coutoPL: the same about this comment

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/DataStreamsRule.scala
```diff
@@ -45,6 +45,14 @@ class DataStreamsRule(val settings: Settings)
     new ZeroKnowledgeIndexFilter(true)
   )
 
+  // Built once when the allowed data streams are statically configured (no runtime
```
coutoPL: The same comment applies here.
I won't add any other comment like this. Consider if this applies to other comments added in this PR.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/DataStreamsRule.scala
```diff
@@ -102,4 +113,17 @@ object DataStreamsRule {
   }
 
   final case class Settings(allowedDataStreams: NonEmptySet[RuntimeMultiResolvableVariable[DataStreamName]])
+
+  // Bundles the wildcard short-circuit flag with the matcher so both can be precomputed
+  // once for static configurations. The matcher is lazy so the wildcard path (which
+  // short-circuits before matching) never pays for building it.
+  private final class AllowedDataStreams(val hasWildcard: Boolean, allowedDataStreams: Set[DataStreamName]) {
```
coutoPL: It should have a private constructor.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/RepositoriesRule.scala
```diff
@@ -54,10 +62,15 @@ class RepositoriesRule(val settings: Settings)
     }
   }
 
+  private def allowedRepositoriesFor(blockContext: BlockContext): AllowedRepositories =
```
coutoPL: The definition of this method is placed above the method it's used in. For the sake of readability of the code (when well-named), it should be defined after the code it's used in (in this example, after the methods `checkRepositories` and `checkSnapshotRepositories`)

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/RepositoriesRule.scala
```diff
@@ -115,4 +128,17 @@ object RepositoriesRule {
   }
 
   final case class Settings(allowedRepositories: NonEmptySet[RuntimeMultiResolvableVariable[RepositoryName]])
-}
\ No newline at end of file
+
+  // Bundles the wildcard short-circuit flag with the matcher so both can be precomputed
+  // once for static configurations. The matcher is lazy so the wildcard path (which
+  // short-circuits before matching) never pays for building it.
+  private final class AllowedRepositories(val hasWildcard: Boolean, allowedRepositories: Set[RepositoryName]) {
```
coutoPL: The same comment about the constructor. 
I won't add any other comment in this style. Review the changed code and all places like that and fix it.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
```diff
@@ -26,24 +26,43 @@ import tech.beshu.ror.syntax.*
 private[http] abstract class BaseHeaderRule
   extends RegularRule {
 
+  // The configured header requirements (and therefore their value-matching globs) are
+  // static for the lifetime of the rule. Precompile each requirement's value matcher once
+  // at construction instead of building a fresh single-element `PatternsMatcher` for every
+  // (requirement x request header) pair on every request.
+  protected def headerAccessRequirements: Iterable[AccessRequirement[Header]]
```
coutoPL: I see that both implementations (HeadersOrRule and HeadersAndRule) define settings that look the same. IMO, instead of creating the protected method, we should:
1. create a settings definition in this class companion object
2. get rid of settings definitions from implementations companion object
3. use the settings directly instead of the `headerAccessRequirements`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
```diff
@@ -26,24 +26,43 @@ import tech.beshu.ror.syntax.*
...
                             requestHeaders: Set[Header]): Boolean = {
+    // Direct lookup (no silent `getOrElse(matcherFor(...))` fallback): a subclass that passes a
+    // requirement it did not enumerate in `headerAccessRequirements` is a programming error, and we
+    // want it to fail fast here rather than silently rebuild a matcher per request (the very cost
+    // this precompilation removes). The explicit message makes that contract violation legible
+    // instead of surfacing as a bare `NoSuchElementException`. With the current subclasses this
+    // lookup always hits.
+    val matches = compiledMatcherByRequirement.getOrElse(
+      accessRequirement,
+      throw new IllegalStateException(
```
coutoPL: In this repo, we follow the rule "make illegal states impossible" whenever it's easy to achieve. In this example, we could easily get rid of the IllegalStateException in favour of a compiler check.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/EagerHashCode.scala
```diff
@@ -0,0 +1,31 @@
...
+ * Caches a case class's `hashCode` eagerly. Mixed into small immutable domain
+ * identities that appear at high frequency in `Set`/`Map` operations on the hot
+ * ACL request-evaluation path, so their `hashCode` is computed once instead of
+ * being recomputed on every lookup/insertion.
+ *
+ * The cached value is identical to the compiler-generated case class `hashCode`
+ * (`MurmurHash3.productHash`), so it stays consistent with structural `equals`
+ * and changes no equality semantics — only when the hash is computed.
+ */
+private[domain] trait EagerHashCode { this: Product =>
```
coutoPL: Maybe we should move it to `misc.scala`? WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/matchers/GlobPatternsMatcher.scala
```diff
@@ -39,14 +39,69 @@ private[matchers] class GlobPatternsMatcher[A: Matchable](val values: Iterable[A
...
+  // closures used. This drops the per-call closure allocations and Vector iterator
+  // overhead while leaning on those String methods (JIT intrinsics) — matching
+  // semantics are unchanged.
+
+  private def matchesExact(norm: String): Boolean = {
+    val exact = compiled.exact
+    exact.nonEmpty && exact.contains(norm)
+  }
+
+  private def matchesPrefix(norm: String): Boolean = {
```
coutoPL: All of the method above looks almost the same. 
IMO, in ScalaOps.scala we should create:

```
extension [T: ClassTag](array: Array[T])
    inline def existsWith(inline pred: T => Boolean): Boolean = {
      var i = 0
      while (i < array.length) {
        if (pred(array(i))) return true
        i += 1
      }
      false
    }
```

And use it like this here:
```
  private def matchesPrefix(norm: String): Boolean = compiled.prefixes.existsWith(norm.startsWith)
  private def matchesSuffix(norm: String): Boolean = compiled.suffixes.existsWith(norm.endsWith)
  private def matchesInfix(norm: String): Boolean  = compiled.infixes.existsWith(norm.indexOf(_) >= 0)
  private def matchesComplex(raw: String): Boolean = compiled.complex.existsWith(_.matches(raw))
```

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/elasticsearch/DataStreamsRuleTests.scala
```diff
@@ -147,32 +150,72 @@ class DataStreamsRuleTests extends AnyWordSpec with MockFactory {
         )
       }
     }
+    "match a runtime variable" when {
+      // Dynamic path: the configured value is a `ToBeResolved` runtime variable (`@{user}`), so the
```
coutoPL: We don't need comments like this in tests, too.

benchmarks/README.md
```diff
@@ -0,0 +1,60 @@
+# benchmarks
```
coutoPL: The benchmarks module is a good foundation for automated regression detection. It would be worth integrating it into the develop pipeline in a follow-up change. 

Requirements:  
  - Benchmark results from the current branch are compared against a stored baseline (e.g. master or the last develop run)
  - Regressions are surfaced as a visible warning on the pipeline run
  - A regression must not fail the pipeline — detection only, human decision required
  - Same behaviour as the CVE vulnerability check: warn but don't block

coutoPL (changes requested):
Additionally:
1. Add Jira number to the PR title. If there is no Jira for this specific change, please add the related Jira number
2. Create concise changelog entry (this style: https://github.com/beshu-tech/readonlyrest-docs/blob/master/changelog.md) and put in the PR's description at the top

core/src/main/scala/tech/beshu/ror/accesscontrol/utils/RuntimeMultiResolvableVariableOps.scala
sscarduzio: Done in 4700f444d — renamed to `resolveAllIfPreResolved`, return type is now `Option[NonEmptyList[T]]`, docstring dropped.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/UsersRule.scala
sscarduzio: Done in 4700f444d — trimmed to a one-line "optimization done on purpose" note.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/auth/UsersRule.scala
sscarduzio: Done in 4700f444d (same as above).

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/DataStreamsRule.scala
sscarduzio: Done in 4700f444d — applied here and reviewed all similar comments across the PR.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/DataStreamsRule.scala
sscarduzio: Done in 4700f444d — `AllowedDataStreams` now has a private constructor (built only via `from`).

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/RepositoriesRule.scala
sscarduzio: Done in 4700f444d — moved `allowedRepositoriesFor` below its callers.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/elasticsearch/RepositoriesRule.scala
sscarduzio: Done in 4700f444d — private constructors on AllowedRepositories and AllowedSnapshots.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
sscarduzio: Done in 4700f444d — `Settings` moved to `BaseHeaderRule` companion; subclasses take `BaseHeaderRule.Settings`; per-subclass Settings + `headerAccessRequirements` indirection removed.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
sscarduzio: Done in 4700f444d — IllegalStateException replaced by a compile-time guarantee: each requirement is precompiled with its matcher (`CompiledRequirement`), so no runtime lookup can miss.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/EagerHashCode.scala
sscarduzio: Done in 4700f444d — moved `EagerHashCode` into `misc.scala`.

core/src/main/scala/tech/beshu/ror/accesscontrol/matchers/GlobPatternsMatcher.scala
sscarduzio: Done in 4700f444d — added `Array.existsWith` to ScalaOps and collapsed the four matches* methods to one-liners as suggested.

core/src/test/scala/tech/beshu/ror/unit/acl/blocks/rules/elasticsearch/DataStreamsRuleTests.scala
sscarduzio: Done in 4700f444d — removed the comment.

benchmarks/README.md
sscarduzio: Agreed — warn-only benchmark regression detection (compare to a baseline, surface as a non-blocking warning like the CVE check) is a great fit, but it needs pipeline wiring + a stored baseline, so I will keep it as a separate follow-up PR rather than expand this one. Leaving this thread open to track it.

sscarduzio:
Thanks @coutoPL — addressed all of the review feedback in `4700f444d`:

- **Naming/types** — `staticallyResolvedValues` → `resolveAllIfPreResolved`, returns `Option[NonEmptyList[T]]`.
- **Comments** — trimmed the explanatory comments across UsersRule/DataStreamsRule/RepositoriesRule/SnapshotsRule (kept a one-line "intentional optimization" note) and removed the test comment; reviewed all similar comments in the PR.
- **Private constructors** — `AllowedDataStreams`/`AllowedRepositories`/`AllowedSnapshots` are now built only via `from`.
- **Method ordering** — `allowedRepositoriesFor` moved below its callers.
- **`existsWith`** — added the `Array.existsWith` inline extension to `ScalaOps` and collapsed the four `matches*` methods to one-liners.
- **`EagerHashCode`** — moved into `misc.scala`.
- **`BaseHeaderRule`** — `Settings` hoisted to the base companion (subclasses now take `BaseHeaderRule.Settings`), dropped the `headerAccessRequirements` indirection, and replaced the `IllegalStateException` with a compile-time guarantee: each requirement is precompiled together with its matcher (`CompiledRequirement`), so a missing matcher is impossible by construction.

All touched rule/matcher/decoder/settings unit suites pass (incl. the case-insensitive header-name matching — which caught an `Eq` regression mid-refactor and is now green).

The benchmark-regression-in-pipeline idea I've left as an open thread to track as a **separate follow-up PR** (it needs pipeline wiring + a stored baseline; warn-only like the CVE check).

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
```diff
@@ -16,34 +16,49 @@
...
+object BaseHeaderRule {
+
+  final case class Settings(headerAccessRequirements: NonEmptySet[AccessRequirement[Header]])
 
-  private def matches(pattern: Header, header: Header) = {
-    if (pattern.name === header.name) {
+  // A header requirement paired with its precompiled value matcher. Header-name comparison is
+  // case-insensitive (per `Header.Name`'s `Eq`) and the value comparison is case-sensitive — exactly
+  // the original per-request matching semantics.
+  final case class CompiledRequirement(accessRequirement: AccessRequirement[Header], matches: Header => Boolean)
```
coutoPL: I think we should:
1. Get rid of the case class here. It's not needed.
2. Make the constructor private
3. rename `CompiledRequirement` to `CompiledHeaderRequirementMatcher`
4. add `def isFulfilledBy(requestHeaders: Set[Header]): Boolean` method remove `protected def isFulfilled(requirement: CompiledRequirement, requestHeaders: Set[Header]): Boolean` protected method from the `BaseHeaderRule` class
5. the `CompiledHeaderRequirementMatcher` class and companion object should have `private [http]` access
6. `Settings` class should take `CompiledHeaderRequirementMatcher`s as param. It should have a smart constructor with `headerAccessRequirements: NonEmptySet[AccessRequirement[Header]]` and should compile the requimemnts in it and create the class instance
7. Because `CompiledHeaderRequirementMatcher` is not a case class, the `Settings` class should not be a case class too

coutoPL (changes requested):

coutoPL:
this has not been fixed yet:
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1260#pullrequestreview-4448991493

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
sscarduzio: Done in 47d41b337:
- `CompiledRequirement` → `CompiledHeaderRequirementMatcher`, now a plain class (not case class) with a private constructor and `private[http]` (object too).
- Logic moved into `CompiledHeaderRequirementMatcher.isFulfilledBy(requestHeaders: Set[Header])`; the protected `isFulfilled` is removed from `BaseHeaderRule`.
- `Settings` is no longer a case class; its smart constructor takes `NonEmptySet[AccessRequirement[Header]]` and compiles the matchers internally. Kept a derived `headerAccessRequirements` view for logging/tests.

All 37 header rule + settings tests pass (incl. the case-insensitive header-name matching).

benchmarks/README.md
sscarduzio: Agreed — this is a good follow-up. As you noted it should be detection-only (warn, never fail; same model as the CVE check), comparing against a stored baseline. It's out of scope for this PR; suggest tracking it as a separate RORDEV ticket so it gets its own review. Leaving this thread for you to resolve/convert as you see fit.

sscarduzio:
@coutoPL ready for re-review. Your two comments are addressed:

- **BaseHeaderRule** — full refactor done (47d41b337): `CompiledRequirement` → `CompiledHeaderRequirementMatcher` (plain class, private ctor, `private[http]`); logic moved to `isFulfilledBy`; protected `isFulfilled` removed; `Settings` no longer a case class, smart constructor takes `NonEmptySet[AccessRequirement[Header]]` and compiles internally. All 37 header rule/settings tests pass. Thread resolved.
- **benchmarks README** (pipeline regression-detection) — replied; suggest tracking as its own ticket since it's a separate, detection-only follow-up. Left that thread for you.

Note: CLA still needs signing on my side before merge.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
```diff
@@ -16,34 +16,53 @@
...
-      case AccessRequirement.MustBeAbsent(forbiddenHeader) =>
-        requestHeaders.forall(!matches(forbiddenHeader, _))
-    }
+  protected val compiledRequirements: NonEmptyList[CompiledHeaderRequirementMatcher] =
+    settings.compiledRequirements
+}
+
+object BaseHeaderRule {
+
+  // Requirements are compiled once in the smart constructor (they are static for the rule's lifetime)
```
coutoPL: A comment explaining easy-to-read code. We don't need it.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
```diff
@@ -16,34 +16,53 @@
...
 
-  protected def isFulfilled(accessRequirement: AccessRequirement[Header],
-                            requestHeaders: Set[Header]): Boolean = {
-    accessRequirement match {
-      case AccessRequirement.MustBePresent(requiredHeader) =>
-        requestHeaders.exists(matches(requiredHeader, _))
-      case AccessRequirement.MustBeAbsent(forbiddenHeader) =>
-        requestHeaders.forall(!matches(forbiddenHeader, _))
-    }
+  protected val compiledRequirements: NonEmptyList[CompiledHeaderRequirementMatcher] =
```
coutoPL: We don't need this protected val. In the implementation class, the value can be accessed by directly accessing `settings` which are available at the implementation classes' level.

settings.gradle
coutoPL, line 1: Add Jira number to the Github PR title. If there is no Jira for this specific change, please add the related Jira number

coutoPL, line 1: Create concise changelog entry (this style: https://github.com/beshu-tech/readonlyrest-docs/blob/master/changelog.md) and put in the Github PR's description at the top

coutoPL (changes requested):

sscarduzio:
@coutoPL all four open review items addressed in `19218989e`:
- `BaseHeaderRule`: `protected val compiledRequirements` removed — subclasses read `settings.compiledRequirements` directly
- Explanatory comments removed (and the test-comment cleanup propagated to `RepositoriesRuleTests`/`SnapshotsRuleTests`)
- Changelog entry added at the top of the description
- Title/description refreshed: stale `staticallyResolvedValues` references fixed, EagerHashCode demoted to 'unmeasured/structural', and the Kibana `Pattern.quote` edge-case fix is now called out explicitly instead of claiming 'no behavior changes'

Previous CI red was the dgp outage (private Maven repo down), since recovered — this push revalidates.

sscarduzio:
Second pass pushed (`aa4e5bb1c`..`5dfa9e422`, 7 commits, +161/−119): deep-review findings implemented — IndicesRule pre-resolution (the headline: most-used rule, was re-resolving + compiling 3–4 matchers per request), JWT/ROR-KBN parser hoisted to the definition, rule/block evaluation short-circuits after Denied/Permitted, fields/response_fields pre-resolution, basicAuth once per request, Header.Name lowercase caching. Full core suite green (2521/0, IndicesRuleTests 211/211). Description updated with the breakdown. Deliberately excluded pending security sign-off: JWT claims cache, unix-crypt verdict cache.

sscarduzio:
@coutoPL description restructured as a per-optimization decision table — 13 rows, each with measured gain/alloc/risk and (for round 2) the single commit SHA to revert if you want any row out. Two rows marked 'discuss' (EagerHashCode — real 1.87× on hot lookups but −15% on create-once objects; cheap to drop). Everything else is benchmark-backed keep.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/http.scala
```diff
@@ -42,9 +42,12 @@ object CorrelationId {
   implicit val show: Show[CorrelationId] = Show.show(_.value.value)
 }
 
-final case class Header(name: Header.Name, value: NonEmptyString)
+final case class Header(name: Header.Name, value: NonEmptyString) extends EagerHashCode
 object Header {
-  final case class Name(value: NonEmptyString)
+  final case class Name(value: NonEmptyString) extends EagerHashCode {
+    // Precomputed so the case-insensitive `Eq` below does not lowercase both sides on every comparison.
+    val lowerCased: String = value.value.toLowerCase(Locale.US)
```
coutoPL: it should be `lazy val`

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/JwtParserBenchmark.scala
coutoPL, line 1: I see no sense in this benchmark. This is the code that has to be maintained, and the optimization is clear, it doesn't have to be measured. IMO, to be removed.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/IndicesRuleResolutionBenchmark.scala
```diff
@@ -0,0 +1,102 @@
...
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(2)
+class IndicesRuleResolutionBenchmark {
+
+  // Static (AlreadyResolved) config: resolveAll never touches the block context, so null is safe.
+  private val noBlockContext: BlockContext = null
+
+  private val configuredPatterns = List(
```
coutoPL: very very basic example.
the value of this benchmark is low

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/HeaderRuleMatchBenchmark.scala
```diff
@@ -0,0 +1,120 @@
...
+    required.forall { r =>
+      requestHeaders.exists(h => h.name === r.name && PatternsMatcher.create(r.value.value :: Nil).`match`(h.value.value))
+    }
+
+  private def newPath(precompiled: Vector[(Header, PatternsMatcher[String])], requestHeaders: Vector[Header]): Boolean =
+    precompiled.forall { case (r, valueMatcher) =>
+      requestHeaders.exists(h => h.name === r.name && valueMatcher.`match`(h.value.value))
+    }
+
+  @Benchmark
```
coutoPL: The benchmarks should go to the beginning of the class. The other things are just implementation details. The readability is low now, because technical things go first

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/HeaderNameEqBenchmark.scala
coutoPL, line 1: low value of the benchmark. IMO, to be removed

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/EagerHashCodeBenchmark.scala
coutoPL, line 1: The same here. Optimisation is obvious. The benchmark proves nothing in the long term. IMO - to be removed

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/BasicAuthDecodeBenchmark.scala
```diff
@@ -0,0 +1,80 @@
...
+  private def decode(): Option[BasicAuth] =
+    headers
+      .to(LazyList)
+      .map(BasicAuth.fromHeader)
+      .find(_.isDefined)
+      .flatten
+
+  @Benchmark
+  def oldPath_decodePerBlock(bh: Blackhole): Unit = {
+    var i = 0
```
coutoPL: Imperative code? We may do it more idiomatically. Here and in the other benchmarks too

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/AclEvaluationBenchmark.scala
```diff
@@ -0,0 +1,354 @@
...
+        usersDefinitionDuplicateUsernamesValidationEnabled = false
+      ),
+      Set.empty
+    )
+  )
+
+  // --- evaluator 1b: current production scaffolding, measured without the doPrivileged hop ------
+
+  // Verbatim production `executeBlocksForRegularRequest`: real current rule loop inside the block.
+  private def currentExecuteBlock(block: Block, context: RequestContext.Aux[BC]): WriterT[Task, Vector[BlockHistory[BC]], Decision[BC]] =
```
coutoPL: No, no. Why do we even need to write this code here?!
IMO, we shoudn't do that.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/AclEvaluationBenchmark.scala
```diff
@@ -0,0 +1,354 @@
...
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.AverageTime, Mode.Throughput))
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(2)
+class AclEvaluationBenchmark {
+
+  private type BC = GeneralNonIndexRequestBlockContext
```
coutoPL: Readability of the benchmark is low. The important things are hidden at the end, and the technical details are in the front. It's not a good way to structure a class.

settings.gradle
```diff
@@ -29,6 +29,7 @@ rootProject.name = 'readonlyrest'
 include 'ror-managed-libs'
```
coutoPL, line 29: The changelog entry added in the PR's description is not good:
1. it's too long
2. it's too technical - it should be written from the perspective of the user
3. it doesn't follow the convention (https://github.com/beshu-tech/readonlyrest-docs/blob/master/changelog.md)
4. it should be ready to copy&paste when the changelog is being created during the release

coutoPL (changes requested):

benchmarks/README.md
sscarduzio: Done as a dedicated PR: #1269 (RORDEV-2080) — advisory allocation gate on PRs, nightly KPI time series on a git orphan branch, dashboard. Sequenced to land BEFORE this PR so the series captures this PR's bump.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
sscarduzio: Removed in 19218989e.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/http/BaseHeaderRule.scala
sscarduzio: Removed in 19218989e — implementations read settings.compiledRequirements directly.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/http.scala
sscarduzio: Done in 810f7e2dd — lazy val.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/JwtParserBenchmark.scala
sscarduzio: Agreed — removed in 810f7e2dd. The whole benchmarks module moved out of this PR; long-lived KPI benchmarks live in #1269 (entry-point based, no old-vs-new replicas to maintain).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/IndicesRuleResolutionBenchmark.scala
sscarduzio: Removed in 810f7e2dd with the module. #1269's replacement is parameterized to realistic scale (patterns 10/100/500 x requested indices 1/50) and measures the real IndicesRule.check entry point.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/HeaderRuleMatchBenchmark.scala
sscarduzio: Module removed from this PR in 810f7e2dd; your structure guidance (benchmarks first, plumbing last) will be applied to the #1269 classes.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/HeaderNameEqBenchmark.scala
sscarduzio: Agreed — removed in 810f7e2dd.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/EagerHashCodeBenchmark.scala
sscarduzio: Agreed — removed in 810f7e2dd (and the measurement honestly showed it is workload-dependent: +87% hot lookups, -15% create-once; documented in the PR description decision table).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/BasicAuthDecodeBenchmark.scala
sscarduzio: Module removed from this PR in 810f7e2dd; #1269's benchmarks will get an idiom pass per this guidance.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/AclEvaluationBenchmark.scala
sscarduzio: Agreed — removed in 810f7e2dd. Replicating production code in benchmarks is exactly what we are dropping; #1269 measures production entry points only.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/AclEvaluationBenchmark.scala
sscarduzio: Removed in 810f7e2dd; structure guidance carried to #1269.

settings.gradle
sscarduzio: Replaced with a one-line user-facing entry following the readonlyrest-docs convention (see updated description) — copy-paste ready for the release changelog.

settings.gradle
sscarduzio: Done — title carries RORDEV-2075.

sscarduzio: Done — description now opens with a one-line convention-style changelog entry, copy-paste ready.

coutoPL (approved):
LGTM

coutoPL:
Please bump the ROR version to `1.71.0-pre1` before merging. 
Remember to use the "squash and merge" option

sscarduzio:
@coutoPL done — bumped to `1.71.0-pre1` in cce5822e6 (also moved `publishedPluginVersion` 1.69.1 → 1.70.1 to match develop's current released version). Will use **squash and merge** as you noted. CI is green (the earlier `IT_es70x` red was the known `RemoteClusterAuditingToolsSuite` round-robin audit flake — re-ran, passed).

sscarduzio:
Addressed the claude[bot] review findings in `4d9c7650a`:

**✅ Fixed — `kibanaIndicesRegexesCache` raw index-name interpolation** (`domain/indices.scala`)
`createKibanaRelatedIndicesRegexes` interpolated `kibanaIndex.stringify` raw into 20 compiled regexes. An index name with regex metacharacters (the default `.kibana` already has a `.`) would match wrongly or throw `PatternSyntaxException`. Now `Pattern.quote`'d once and reused — the exact fix already applied to `nonStrictAllowedPathsPatternFor` in `BaseKibanaRule`. The `\d+\.\d+` version suffixes stay real regex. Not deferred after all — it was a quick, self-contained fix.

**✅ Fixed — `resolveAllIfPreResolved` undocumented all-or-nothing contract** (`RuntimeMultiResolvableVariableOps`)
Added a comment: returns `Some` only if ALL variables are `AlreadyResolved`; one `ToBeResolved` yields `None` → full per-request resolution. Behaviour unchanged, just no longer a trap for the next reader who reuses it.

**❌ Not a bug — `.executor(global)` "inconsistent with `kibanaIndicesRegexesCache`"**
The review's premise is incorrect: `kibanaIndicesRegexesCache` (`indices.scala:92`) *does* set `.executor(global)`. The two caches already match, so dropping it from one would *create* the inconsistency. Left as-is.

The other notes (`matchesInfix` `_`-binding style, `lazy val parser` `@volatile` FYI, and the open questions on `basicAuth` lazy-val freshness / cache lifetime across reloads) were verified correct in the review itself — no action.

Validated: `:core:compileScala` green under `-Xfatal-warnings`; `IndicesRuleTests` 211/211 (covers the `.custom_kibana*` related-index matching that exercises the quoted path); Kibana access tests 62/62.

sscarduzio:
Addressed the second-round review in `f63d5ca37`:

**✅ Fixed — IndicesRule dynamic-variable test gap** (the Warning)
Added `"match a runtime variable"` / `"not match a runtime variable"` cases to `IndicesRuleLocalIndexTests` driving a `@{user}` (`ToBeResolved`) config — mirroring the DataStreams/Repositories/Snapshots pattern. The match case sets a logged user via `modifyBlockContext` so `@{user}` resolves to `test1` and matches; the deny case resolves to a different index. `BaseIndicesRuleTests.assertRuleForIndexRequest` now asserts the *actual* resolved `loggedUser` (it was hardcoded `None`) — a no-op for the 211 existing static tests. **IndicesRuleTests: 213/213.**

**✅ Fixed — `FieldsRule` / `ResponseFieldsRule` unreachable `None` branch** (the Nit)
`resolveAllIfPreResolved` yields a `NonEmptyList`, so the unique set is never empty. Replaced `.flatMap(UniqueNonEmptyList.from)` with `.map(UniqueNonEmptyList.fromNonEmptyList)` — the guarantee is now explicit at the type level (no Option to thread).

**❌ Skipped — `.map(_.flatten)` → `.flatMap(identity)`** (the other Nit)
Purely stylistic (the review said so). `.map(_.flatten)` reads more clearly than `.flatMap(identity)`, so I left it.

Open questions: (1) no `integration-tests/` suite exercises an `indices` rule with `@{user}`/`@{jwt}` values today — so the unit coverage just added is the relevant guard. (2) the flatten spelling — see above.

Validated: `:core:compileScala` + `:core:compileTestScala` green under `-Xfatal-warnings`.

PR #1261, opened 2026-06-07
[RORDEV-2016] ACL logging and audit unification, replaced verbosity mechanism, simplified audit config

mgoworko:
<details>
  <summary><strong>🚨Warning</strong> (ES) Block-level <code>audit: {enabled: false}</code> now suppresses the default ACL log in addition to custom audit sinks</summary>

  Previously, disabling audit at the block level (`audit: {enabled: false}`) only suppressed writes to configured ES audit sinks. The built-in ACL log (the human-readable decision line written to Elasticsearch logs)
  continued to fire. It now also suppresses the ACL log. Installations that relied on the ACL log always being present regardless of block-level audit settings should use `audit: {enabled_audit_sinks: [default_acl_log]}`
  to keep ACL logging active for that block, or set `audit: {disabled_audit_sinks: [my-index-sink]}` instead of disabling audit entirely.
  </details>

  <details>
  <summary><strong>🚀New</strong> (ES) <a href="https://docs.readonlyrest.com/elasticsearch/audit#block-level-audit-control">Per-block audit sink routing via named sinks, <code>enabled_audit_sinks</code>, and
  <code>disabled_audit_sinks</code></a></summary>

  Audit outputs can now be assigned a `name` in the global `audit.outputs` list. Each ACL block can then declare `enabled_audit_sinks` (whitelist) or `disabled_audit_sinks` (blacklist) to control which sinks receive events
  when that block is matched. This allows, for example, routing high-sensitivity blocks exclusively to a dedicated secure index sink while noisy service-account blocks skip the default ACL log, all without separate ROR
  instances.
  </details>

  <details>
  <summary><strong>🚀New</strong> (ES) <a href="https://docs.readonlyrest.com/elasticsearch/audit#the-default-acl-log">The built-in ACL log is now a configurable, named sink (<code>default_acl_log</code>) controlled by
  <code>audit.default_acl_log_enabled</code></a></summary>

  The human-readable ACL decision log that has always been written to Elasticsearch logs is now a first-class named sink exposed under the reserved name `default_acl_log`. It can be targeted in block-level
  `enabled_audit_sinks` / `disabled_audit_sinks` routing, and globally controlled via the new `audit.default_acl_log_enabled` setting (default: `true`, preserving existing behaviour). To replace the built-in ACL log with a
  custom format or destination, set `default_acl_log_enabled: false` and add a `log` output with `serializer: {type: acl}`.
  </details>

  <details>
  <summary><strong>🚀New</strong> (ES) <a href="https://docs.readonlyrest.com/elasticsearch/audit#built-in-rolling-file-appender">Built-in rolling file appender for <code>log</code> audit outputs via
  <code>file_appender</code></a></summary>

  The `log` audit output type now accepts a `file_appender` sub-section (`file_path`, `max_file_size`, `max_files`) that configures a rolling file appender internally, without requiring changes to `log4j2.properties`. This
  simplifies deployments where audit events need to be written to a dedicated file with automatic rotation.
  </details>
  
  <details>
  <summary><strong>🚀New</strong> (ES) <a href="https://docs.readonlyrest.com/elasticsearch/audit#acl-serializer">New <code>acl</code> serializer type for <code>log</code> audit outputs</a></summary>

  A new `serializer: {type: acl}` option is available for `log`-type audit outputs. It reproduces the human-readable format of the built-in ACL log, making it easy to redirect or duplicate ACL logging to a custom logger, a
  separate file, or a named sink that can be included or excluded per block.
  </details>
  
  <details>
  <summary><strong>🧐Enhancement</strong> (ES) Block-level <code>log_allowed_events</code> replaces <code>verbosity</code>; <code>allowed_events_serialization_mode</code> replaces
  <code>verbosity_level_serialization_mode</code> in serializers</summary>

  The block-level `verbosity: error/info` setting is superseded by `audit: {log_allowed_events: false/true}`, which makes the intent explicit and integrates cleanly with the new per-sink routing. The old `verbosity` key
  continues to work as a backward-compatible alias. Similarly, the `verbosity_level_serialization_mode` setting in configurable and ECS serializers is replaced by `allowed_events_serialization_mode`, accepting `"always"`
  (serialize all allowed events) or `"based_on_block_settings"` (serialize only events from blocks with `log_allowed_events: true`).
  </details>

## Summary by CodeRabbit

* **New Features**
  * Rolling file-based audit output with configurable rotation (size/retention) and ACL-aware formatting.
  * New audit configuration model: per-block sink selection with explicit enabled/disabled sink lists and validated sink names.
  * File-size parsing/validation for log rotation settings.

* **Refactor**
  * Auditing is now always wired through core/bootstrap and sinks are named and validated.

* **Tests**
  * Expanded coverage for decoding, sink validation, rolling-file behavior, and audit-output scenarios.


core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/BaseAuditSink.scala
```diff
@@ -18,10 +18,12 @@ package tech.beshu.ror.accesscontrol.audit.sink
 
 import monix.eval.Task
 import org.json.JSONObject
+import tech.beshu.ror.accesscontrol.blocks.Block
 import tech.beshu.ror.accesscontrol.domain.RequestId
 import tech.beshu.ror.audit.{AuditLogSerializer, AuditResponseContext}
 
-private[audit] abstract class BaseAuditSink(auditLogSerializer: AuditLogSerializer) {
+private[audit] abstract class BaseAuditSink(val name: Block.SinkName,
```
coutoPL: I'm not sure if the SinkName domain object should be declared in Block class. Maybe audit.scala will be a better place for it?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/LogBasedAuditSink.scala
```diff
@@ -19,17 +19,25 @@ package tech.beshu.ror.accesscontrol.audit.sink
...
+private[audit] final class LogBasedAuditSink(sinkName: Block.SinkName,
+                                             serializer: AuditLogSerializer,
+                                             loggerName: RorAuditLoggerName) extends BaseAuditSink(sinkName, serializer) {
 
   private val logger: Logger = LogManager.getLogger(loggerName.value.value)
 
   override protected def submit(event: AuditResponseContext, serializedEvent: JSONObject)
                                (implicit requestId: RequestId): Task[Unit] = Task {
-    logger.info(serializedEvent.toString)
+    serializer match {
```
coutoPL: Sth smells here.
I didn't analyse it better, but we:
1. use a (almost) private method here.
2. have to check the serializer type and sth special for it

Maybe we cannot use BaseAuditSink as a base class here? 
Or maybe there is sth wrong with the serializers' responsibility? 
Or maybe all of these?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/RollingFileBasedAuditSink.scala
```diff
@@ -0,0 +1,125 @@
...
+
+      val appender = Option(RollingFileAppenderFactory.create(
+        s"RorAuditFile-${loggerName.value.value}",
+        config.filePath.toString,
+        config.filePath.toString + ".%i",
+        layout,
+        triggeringPolicy,
+        rolloverStrategy,
+        log4jConfig
+      )).getOrElse(throw new IllegalStateException("Appender builder returned null"))
```
coutoPL: maybe better message here?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AclAuditLogSerializer.scala
```diff
@@ -0,0 +1,56 @@
...
+import tech.beshu.ror.utils.RefinedUtils.nes
+
+class AclAuditLogSerializer extends AuditLogSerializer {
+
+  override def onResponse(responseContext: AuditResponseContext): Option[JSONObject] = {
+    val suppress = responseContext match {
+      case allowed: AuditResponseContext.Allowed => allowed.verbosity != Verbosity.Info
+      case _                                     => false
+    }
+    if (suppress) None else Some(new JSONObject())
```
coutoPL: yeah, this is weird too.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AclAuditLogSerializer.scala
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  private[audit] def formatMessage(responseContext: AuditResponseContext, debugEnabled: Boolean): String = {
+    responseContext.requestContext match {
+      case ctx: AuditRequestContextBasedOnAclResult[?] =>
+        given Show[Header] = if (debugEnabled) headerShow else obfuscatedHeaderShow(ctx.loggingContext.obfuscatedHeaders)
+        ctx.aclMessageShow(debugEnabled).show(ctx.responseContext)
+      case ctx =>
+        // AuditRequestContext is an open trait in the published audit module, so a third-party
```
coutoPL: There is no way to protect against it?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AclAuditLogSerializer.scala
```diff
@@ -0,0 +1,56 @@
...
+        s"[unknown-context:${ctx.getClass.getSimpleName}] id=${ctx.id} action=${ctx.action} uri=${ctx.uriPath}"
+    }
+  }
+}
+
+object AclAuditLogSerializer {
+  // Preserved from the class that originally emitted ACL log entries, so existing
+  // log4j configurations targeting that logger name continue to work unchanged.
+  val defaultLoggerName: RorAuditLoggerName =
+    RorAuditLoggerName(nes("tech.beshu.ror.accesscontrol.logging.AccessControlListLoggingDecorator"))
```
coutoPL: Should we hardcode it? Cannot we do sth like `classOf[AccessControlListLoggingDecorator].name` or similar?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditingTool.scala
```diff
@@ -40,24 +42,43 @@ import tech.beshu.ror.utils.RequestIdAwareLogging
...
-  def close(): Task[Unit] = {
-    auditSinks
-      .parTraverse(_.close())
-      .map((_: NonEmptyList[Unit]) => ())
+  private def activeSinksFor[B <: BlockContext](responseContext: ResponseContext[B]): List[Block.AuditSink] = {
+    responseContext match {
+      case AllowedBy(_, blockContext, _) =>
+        blockContext.block.auditSinks
+      case Allowed(_, userMetadata, _) =>
+        userMetadata match {
```
coutoPL: Maybe we can flatten to one case level? WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
```diff
@@ -38,16 +41,31 @@ import tech.beshu.ror.utils.RequestIdAwareLogging
...
-class Block(val name: Name,
-            val policy: Policy,
-            val verbosity: Verbosity,
-            val audit: Audit,
-            val rules: NonEmptyList[Rule])
+class Block(val name: Block.Name,
+            val policy: Block.Policy,
+            val rules: NonEmptyList[Rule],
+            val audit: Block.Audit,
+            val auditSinks: List[Block.AuditSink])
```
coutoPL: it would be nice if we could add `auditSinks` to `audit`. Because there should be no audit sinks where audit is disabled.

core/src/main/scala/tech/beshu/ror/accesscontrol/domain/audit.scala
```diff
@@ -188,4 +188,27 @@ object AuditCluster {
...
   val default: RorAuditLoggerName = RorAuditLoggerName(nes("readonlyrest_audit"))
-}
\ No newline at end of file
+}
+
+final case class FileSize private(value: String)
+object FileSize {
+  def from(raw: String): Either[CreationError, FileSize] = {
+    val trimmed = raw.trim
+    if (trimmed.matches("(?i)^\\d+\\s*(B|KB|MB|GB|TB)?$")) {
```
coutoPL: Please take a look at `val information: FromString[Information]` instance.
IMO, we should be consistent in our settings.

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/LogBasedAuditSink.scala
mgoworko: I think that the underlying cause of that smell, is the fact that we have `audit` module, which is treated as a public API, with `trait AuditLogSerializer`. 

That module is rarely and reluctantly modified. And the core module uses that API directly, using raw types and instances from `audit` module, despite the fact, that the `core` module adds a lot functionality and requires some additional metadata/enrichments.

I created new trait `AuditSerializer` in `core` module. It both: wraps the AuditLogSerializer instances when needed, and also defines core-specific implementations: ECS, ACL, Configurable.

I think this refactor made it cleaner and eliminated a lot of smell.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AclAuditLogSerializer.scala
mgoworko: Answered in first comment, now AclAuditLogSerializer returns the String-typed log, instead of JsonObject hack.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AclAuditLogSerializer.scala
mgoworko: This issue could be fixed with an incompatible change in the `audit` module - parameterizing `sealed trait AuditResponseContext` by adding `[+RC <: AuditRequestContext]`. But IMO it is not worth it. The change would propagate across audit and core modules, including Java audit API. 

I modified the comment to better describe this situation:
```
        // AuditRequestContext is an open trait in the audit module, which is a public API.
        // AuditRequestContextBasedOnAclResult is the only implementation of that trait in the production codebase.
        // As a result, this case should never be executed, but it is not guaranteed on the type level.
```

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
mgoworko: I refactored it, to add auditSinks to Enabled Block.Audit. It improved the situation a bit. But generally, it is a very complex issue:
- it concerns the order of config decoding and initialization of ROR components
- the problem is, that ACL needs to be loaded before the auditing tool is created, and we can resolve the auditSinks only after auditing tools is created - it is a sort of an initialization loop
- I tried to refactor and untangle this loop, but it has proven to be very complex and I abandoned it - it would be too huge in an already big audit-related changes, with no obvious benefits

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/RollingFileBasedAuditSink.scala
mgoworko: `Could not create RollingFileAppender`

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditSerializer.scala
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.accesscontrol.audit
+
+import tech.beshu.ror.audit.AuditLogSerializer
+import tech.beshu.ror.audit.utils.AuditSerializationHelper.{AllowedEventMode, AuditFieldPath, AuditFieldValueDescriptor}
+
+sealed trait AuditSerializer
```
coutoPL, line 25: Yes, this hierarchy is really nice. But WDYT about having the serialization logic close to this hierarchy? By:
1. Defining the `serialize` method OR
2. Defining extension method `serialize`

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditSerializer.scala
```diff
@@ -0,0 +1,36 @@
...
+
+object AuditSerializer {
+  final case class Delegating(serializer: AuditLogSerializer) extends AuditSerializer
+
+  case object Acl extends AuditSerializer
```
coutoPL: I see that we currently have:
1. TextBasedAuditSink
2. BaseAuditSink

Looking at the implementation, the second one could be named: `JsonBasedAuditSink`. 

But IMO the fact that sth is serialized as JSON or text is a Serializer characteristic. So, maybe the AuditSerializer should have two sealed traits: JsonAuditSerializer and TextAuditSerializer`. And the `Acl` would be the TextAuditSerializer.

Then in code like this:
```scala
  private def safeRunSerializer(context: AuditResponseContext) = {
    auditSerializer match {
      case AuditSerializer.Delegating(serializer) =>
        Task.delay(serializer.onResponse(context))
      case AuditSerializer.Acl =>
        Task.delay(None)
      case AuditSerializer.EcsV1(allowedEventMode, includeFullRequestContent) =>
        Task.delay(EcsV1AuditLogSerializer.onResponse(context, allowedEventMode, includeFullRequestContent))
      case AuditSerializer.Configurable(allowedEventMode, fields) =>
        Task.delay(ConfigurableAuditLogSerializer.onResponse(context, allowedEventMode, fields))
    }
  }
```

We don't have to return sth like "undefined" (Task.delay(None)) for the non-fitting thing.

WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/BaseAuditSink.scala
```diff
@@ -18,10 +18,15 @@ package tech.beshu.ror.accesscontrol.audit.sink
...
-import tech.beshu.ror.audit.{AuditLogSerializer, AuditResponseContext}
+import tech.beshu.ror.accesscontrol.audit.AuditSerializer
+import tech.beshu.ror.accesscontrol.audit.configurable.ConfigurableAuditLogSerializer
+import tech.beshu.ror.accesscontrol.audit.ecs.EcsV1AuditLogSerializer
+import tech.beshu.ror.accesscontrol.blocks.Block
+import tech.beshu.ror.accesscontrol.domain.{RequestId, SinkName}
+import tech.beshu.ror.audit.AuditResponseContext
 
-private[audit] abstract class BaseAuditSink(auditLogSerializer: AuditLogSerializer) {
+private[audit] abstract class BaseAuditSink(val name: SinkName, auditSerializer: AuditSerializer)
```
coutoPL: As mentioned earlier, if we introduce the richer hierarchy of the AuditSerializer, we would probably need to rename this class and change the type of the `auditSerializer` param.

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/TextBasedAuditSink.scala
```diff
@@ -0,0 +1,56 @@
...
+      case AuditSerializer.Configurable(allowedEventMode, fields) =>
+        ConfigurableAuditLogSerializer.onResponse(event, allowedEventMode, fields).foreach(log)
+      case AuditSerializer.Acl =>
+        AclAuditLogSerializer
+          .format(event, logger.isDebugEnabled)
+          .foreach(msg => logger.info(s"[${requestId.value}] $msg"))
+    }
+  }
+
+  private def log(json: JSONObject): Unit = logger.info(json.toString)
```
coutoPL: WDYT about using ' .toString (0) ' here explicitly?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
```diff
@@ -189,41 +210,16 @@ class Block(
...
-    val sortedRules = rules.sorted
-    BlockValidator.validate(name, sortedRules) match {
-      case Validated.Valid(_) =>
-        Right(createBlockInstance(name, policy, verbosity, audit, sortedRules))
-      case Validated.Invalid(errors) =>
-        implicit val validationErrorShow: Show[BlockValidationError] = blockValidationErrorShow(name)
-        Left(BlocksLevelCreationError(Message(errors.toList.map(_.show).mkString("\n"))))
-    }
-  }
+  trait AuditSink {
```
coutoPL: let's move AuditSink trait to `tech.beshu.ror.accesscontrol.audit.sink` package

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
```diff
@@ -22,34 +22,55 @@ import monix.eval.Task
...
 )(
     implicit val loggingContext: LoggingContext
 ) extends RequestIdAwareLogging {
 
   import Lifter.*
 
+  def withResolvedAuditSinks(allSinks: List[Block.AuditSink]): Block = {
+    val newAudit = audit match {
+      case Audit.Disabled =>
+        noRequestIdLogger.warn(
```
coutoPL: I wonder if this should be a warning level. I guess someone intentionally disabled the audit for this block. Maybe we could put it in the debug log? And maybe one log is enough (not a log per block). WDYT?

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
```diff
@@ -254,19 +250,61 @@ object Block {
...
 
   sealed trait Audit
 
   object Audit {
-    case object Enabled extends Audit
+
+    final case class Enabled(
+        logAllowedEvents: Boolean = true,
+        enabledAuditSinks: EnabledAuditSinks = EnabledAuditSinks.All,
+        precomputedAuditSinks: PrecomputedAuditSinks = PrecomputedAuditSinks.NotAvailable,
```
coutoPL, line 276: Maybe we have to explain in comment why we need these two and how they differ?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/CoreFactory.scala
```diff
@@ -244,34 +253,53 @@ class RawRorSettingsBasedCoreFactory(esEnv: EsEnv)(
...
     Decoder
       .instance { c =>
         val result = for {
           name <- c.downField(Attributes.Block.name).as[Block.Name]
           policy <- c.downField(Attributes.Block.policy).as[Option[Block.Policy]]
-          verbosity <- c.downField(Attributes.Block.verbosity).as[Option[Block.Verbosity]]
-          audit <- c.downField(Attributes.Block.audit).as[Option[Block.Audit]]
+          legacyVerbosityAudit <- c.as[Option[Block.Audit]](legacyVerbosityAuditDecoder)
+          auditFromConfig <- c.downField(Attributes.Block.audit).as[Option[Block.Audit]]
+          // explicit audit section takes precedence over the legacy verbosity key
```
coutoPL: WDYT about forbidding the use of verbosity when we use the new approach?

core/src/main/scala/tech/beshu/ror/accesscontrol/factory/CoreFactory.scala
```diff
@@ -283,16 +311,82 @@ class RawRorSettingsBasedCoreFactory(esEnv: EsEnv)(
...
+    }
+  }
+
+  private def auditSinkNamesDecoder(
+      blocksNel: NonEmptyList[BlockDecodingResult],
+      auditingConfig: AuditingTool.AuditingConfig
+  ): AsyncDecoder[Unit] =
+    AsyncDecoderCreator.instance[Unit] { _ =>
+      Task.now {
+        val configuredSinkNames: scala.collection.Set[SinkName] = auditingConfig.outputsConfig match {
```
coutoPL, line 380: it looks like we need some new private methods here

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditSerializer.scala
mgoworko: done, extension method

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/AuditSerializer.scala
mgoworko: Done. Introducing JsonAuditSerializer/TextAuditSerializer division made it simpler and more type safe in the decoder and logic.

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
mgoworko: Replaced with a single log (with multiple block names if needed) in the `coreDecoder`

core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/Block.scala
mgoworko: added detailed description in the comment

core/src/main/scala/tech/beshu/ror/accesscontrol/audit/sink/AuditSink.scala
```diff
@@ -0,0 +1,32 @@
...
+import tech.beshu.ror.audit.AuditResponseContext
+
+trait AuditSink {
+  def name: SinkName
+
+  def submit(event: AuditResponseContext)(
+      implicit requestId: RequestId
+  ): Task[Unit]
+
+  def close(): Task[Unit] = Task.unit
```
coutoPL: Maybe we should remove the default implementation? When the closing is not needed, the implementation will override and return Task.unit. Leaving it as an abstract method, we give the implementer a chance to think whether the closing is needed or not.

coutoPL (approved):
LGTM. One minor comment left

PR #1262, opened 2026-06-08
[RORDEV-2076] [ci] Self-contained build image + auto-cancel superseded runs (eliminate failable downloads & queue saturation)

sscarduzio:
Jira: [RORDEV-2076](https://readonlyrest.atlassian.net/browse/RORDEV-2076)

## Why

Two chronic CI problems, one PR:

1. **Flaky build-time downloads.** The build depended on several external CDNs at build time (JDK toolchains via api.foojay.io/adoptium, the Gradle distribution via services.gradle.org, apt, Maven Central/plugin portal) — all of which produced transient `504`s that failed builds even though the code was fine.
2. **Queue saturation.** Every new commit to a PR spawned a fresh full run (10–34 IT legs) **without cancelling the previous one** — a few quick commits piled multiple complete matrices onto the 20 hosted runners (observed: 50+ job backlog, multiple live runs for the same PR).

### 1. Self-contained CI toolchains image (`ci/toolchains/JdkToolchains.Dockerfile`)

One image bakes everything the download-free CI tasks need:
- **JDK 8/11/17/21 toolchains** (8→`:audit`, 11→`:core`+ES 6.7–7.x, 17→ES 8.x, 21→ES 9.x) — recorded in the image's `GRADLE_USER_HOME/gradle.properties`, NOT the project file (absolute Linux paths there abort Gradle init on Windows agents)
- **The Gradle distribution** + all plugin/dependency caches
- **Every module's compile classpath** — including the per-ES-version private artifacts (e.g. all 22 `transport-netty4` versions)
- **core/audit test classpaths** and the audit cross-build (2.11/2.12/2.13/3) **including the Scala compiler bridges** (resolved via a detached configuration only when a compile actually runs — a resolve-only prime misses them)
- A **static, SHA256-pinned docker CLI** (publish stages talk to the agent host's daemon via the mounted socket)

**The image build self-validates**: it re-runs the primed CI tasks offline against the baked cache (with core/audit outputs wiped first, so compiler-bridge resolution is genuinely exercised) and only then writes a `.ror-ci-baked` sentinel. An incomplete cache fails the image build — never a CI run.

At CI time the warm cache covers the common case with zero network calls; a cache miss (e.g. a dep bumped since the last image rebuild) simply falls back to the network, and the **weekly scheduled image rebuild** keeps that stale window short (per review, no `--offline` hard-enforcement: a stale cache must not block every CI run). Integration tests stay online by design (ES binaries + private libs at runtime). The nested Tooling-API build that packages the plugin for the "load ROR plugin" suites reuses the baked home when the sentinel is present — it used to download the Gradle distribution on every IT leg.

### 2. Pipeline on the image + queue fixes (`azure-pipelines.yml`)

- All Linux jobs run in the image via a `resources.containers` alias (the `container:` property cannot expand `$(variables)`); redundant Gradle `Cache@2` restores removed; `TOOLCHAINS_VERIFY` is a fast sanity check of the baked Gradle home
- **`BUILD_TOOLCHAINS_IMAGE` runs weekly on a schedule** (master, Mon 03:00 UTC) to refresh the baked dependency cache without manual action — in isolation: every other stage is gated on `Build.Reason != 'Schedule'`. Still manually triggerable (`actionToPerform=build_toolchains_image`) for JDK/Gradle changes that also need an image-tag bump
- **Windows jobs clear `GRADLE_USER_HOME`/`GRADLE_OPTS` at JOB scope** — step-level `env:` does not reliably override pipeline variables for child processes on Windows (validated: all 3 WIN_IT legs green)
- **`pr.autoCancel: true` + `trigger.batch: true`** — a new commit cancels the superseded in-progress PR run instead of stacking (verified live, repeatedly); top-level `epic/*` added to triggers (`*/epic/*` never matched the repo's actual epic branches)
- **Publish/release stages run in the container on the self-hosted pool** (adjacent to the artifact store; all JDKs + baked cache — resolves the bare-agent JDK-discovery and lost-cache review findings)
- **Fixed: silently-skipped Maven publishing** — the pre-check probed decommissioned `oss.sonatype.org` (neither 200 nor 404 → every publish step skipped, stage green). Removed; `run-pipeline.sh` performs the correct existence check
- **Fixed: `ES_S3_UP` gating** now accepts `SucceededWithIssues` (its upload is `continueOnError`; a partial failure must not silently skip the whole test pipeline)

## Operational notes for reviewers

- **When dependencies change** (new ES version, bumped lib): nothing to do — cache misses fall back to the network and the weekly scheduled `BUILD_TOOLCHAINS_IMAGE` rebuild re-bakes the cache. **JDK set or Gradle version changes** still require a manual image rebuild + tag bump (`ci/toolchains/JdkToolchains.Dockerfile` header documents both paths).
- **Validated on real runs:** compile/audit/unit/license green on the baked cache; all 10 Linux IT legs green (one ScalaTest infra flake on es80x, unrelated); **all 3 Windows IT legs green**; auto-cancel verified on consecutive pushes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,3 +1,11 @@
+# Single source of truth for the CI toolchains image (bump the tag here on a JDK/Gradle change).
+# Declared as a container resource (NOT a variable) because the job-level `container:` property is
+# resolved at compile time and does not expand $(variables).
+resources:
+  containers:
+    - container: rortoolchains
+      image: beshultd/ror-ci-toolchains:jdk-8-11-17-21-gradle-9.2.1
+
 variables:
   system.debug: 'true'
   ssh_file: '~/.ssh/gh_deploy_key.priv'
@@ -8,18 +16,47 @@ variables:
   isNewEsBranchOrPr: $[or(contains(variables['Build.SourceBranch'], 'newes/'), contains(variables['System.PullRequest.SourceBranch'], 'newes/'))]
   dependencyCheckCacheKey: $[format('dependency-check-v1-{0:yyyyMM}', pipeline.startTime)]
   dependencyCheckDataDir: '$(Pipeline.Workspace)/dependency-check-data/v1'
-  GRADLE_USER_HOME: '$(Pipeline.Workspace)/.gradle'
+  # Use the Gradle home baked into the CI image (ci/Dockerfile.toolchains): it already contains the
+  # Gradle distribution + resolved plugins/dependencies, so download-free tasks run with `--offline`
+  # and never hit a flaky external CDN. (Override per-build with ROR_GRADLE_OFFLINE=false to refresh
+  # the cache when dependencies change.)
+  GRADLE_USER_HOME: '/opt/gradle-home'
+  # Disable toolchain auto-download (fail fast instead of hitting the adoptium CDN). The CI-image
+  # JDK locations are NOT set here — they live in the image's baked GRADLE_USER_HOME/gradle.properties
+  # (reliably honoured, unlike installations.paths via GRADLE_OPTS), so they apply only to the Linux
+  # container jobs and never leak to the bare Windows jobs (which use a workspace-local GRADLE_USER_HOME).
+  GRADLE_OPTS: '-Dorg.gradle.java.installations.auto-download=false'
 
 trigger:
-  batch: false
+  # Coalesce rapid pushes into one run (Azure has no auto-cancel for branch runs).
+  batch: true
   branches:
     include:
       - master
       - develop
+      - 'epic/*'
       - '*/epic/*'
   tags:
     exclude:
       - '*'
+  paths:
+    include:
+      - '*'
+    exclude:
+      - 'docs/*'
+      - '*.md'
+      - '*/*/*.md'
+
+pr:
+  # A new commit cancels the prior in-progress PR run instead of stacking another
+  # full run on the queue (GitHub default; set explicitly so it is version-controlled).
+  autoCancel: true
+  branches:
+    include:
+      - master
+      - develop
+      - 'epic/*'
+      - '*/epic/*'
   paths:
     include:
       - '*'
@@ -84,54 +121,35 @@ stages:
       )
     jobs:
       - job: WARMUP
-        displayName: 'Download Gradle and toolchains'
-        container: eclipse-temurin:17-jdk
-        timeoutInMinutes: 30
+        displayName: 'Verify baked Gradle home and toolchains'
+        container: rortoolchains
+        timeoutInMinutes: 10
         steps:
           - checkout: self
             fetchDepth: 1
             clean: false
-          - task: Cache@2
-            displayName: 'Restore Gradle cache'
-            inputs:
-              key: 'gradle | "$(Agent.OS)" | **/gradle/wrapper/gradle-wrapper.properties | **/settings.gradle* | **/build.gradle* | **/gradle.properties'
-              restoreKeys: |
-                gradle | "$(Agent.OS)"
-                gradle
-              path: $(GRADLE_USER_HOME)
-              cacheHitVar: GRADLE_CACHE_HIT
+          # Verify the baked offline Gradle home is present/usable (failure => rebuild the image).
           - bash: |
-              set +e
-              apt-get update && apt-get install -y git curl
-              echo "Warming up Gradle and toolchains..."
-              ./gradlew --no-daemon -q help
-              exit 0
-            displayName: 'Download Gradle and toolchains (best effort)'
-            condition: ne(variables.GRADLE_CACHE_HIT, 'true')
+              set -e
+              echo "Baked GRADLE_USER_HOME=$GRADLE_USER_HOME"
+              ls "$GRADLE_USER_HOME/wrapper/dists" || { echo "No baked Gradle dist in image"; exit 1; }
+              ./gradlew --no-daemon --offline -q help
+            displayName: 'Verify offline Gradle home (no downloads)'
             env:
               GRADLE_USER_HOME: $(GRADLE_USER_HOME)
 
   - stage: OPTIONAL_CHECKS
     displayName: 'Optional checks'
     dependsOn: [ ES_S3_UP, GRADLE_WARMUP ]
-    condition: and(in(dependencies.ES_S3_UP.result, 'Succeeded', 'Skipped'), in(dependencies.GRADLE_WARMUP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'), ne(variables['Build.Reason'], 'Manual'))
+    condition: and(in(dependencies.ES_S3_UP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'), in(dependencies.GRADLE_WARMUP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'), ne(variables['Build.Reason'], 'Manual'))
     jobs:
       - job: CVE_CHECK
-        container: eclipse-temurin:17-jdk
+        container: rortoolchains
         steps:
           - checkout: self
             fetchDepth: 1
             clean: false
             persistCredentials: true
-          - task: Cache@2
-            displayName: 'Restore Gradle cache'
-            inputs:
-              key: 'gradle | "$(Agent.OS)" | **/gradle/wrapper/gradle-wrapper.properties | **/settings.gradle* | **/build.gradle* | **/gradle.properties'
-              restoreKeys: |
-                gradle | "$(Agent.OS)"
-                gradle
-              path: $(GRADLE_USER_HOME)
-              cacheHitVar: GRADLE_CACHE_HIT
           - task: Cache@2
             displayName: 'Restore CVE DB from cache'
             inputs:
@@ -176,27 +194,18 @@ stages:
     dependsOn: [ ES_S3_UP, GRADLE_WARMUP ]
     condition:
       and(
-        in(dependencies.ES_S3_UP.result, 'Succeeded', 'Skipped'),
+        in(dependencies.ES_S3_UP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
         in(dependencies.GRADLE_WARMUP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
         ne(variables['Build.Reason'], 'Manual')
       )
     jobs:
       - job:
-        container: eclipse-temurin:17-jdk
+        container: rortoolchains
         steps:
           - checkout: self
             fetchDepth: 1
             clean: false
             persistCredentials: true
-          - task: Cache@2
-            displayName: 'Restore Gradle cache'
-            inputs:
-              key: 'gradle | "$(Agent.OS)" | **/gradle/wrapper/gradle-wrapper.properties | **/settings.gradle* | **/build.gradle* | **/gradle.properties'
-              restoreKeys: |
-                gradle | "$(Agent.OS)"
-                gradle
-              path: $(GRADLE_USER_HOME)
-              cacheHitVar: GRADLE_CACHE_HIT
           - script: |
               export ROR_LIBS_STORE_URL=$VAR_ROR_LIBS_STORE_URL
               export ROR_LIBS_STORE_ACCESS_KEY_ID=$VAR_ROR_LIBS_STORE_ACCESS_KEY_ID
@@ -224,7 +233,7 @@ stages:
     dependsOn: [ ES_S3_UP, GRADLE_WARMUP ]
     condition:
       and(
-        in(dependencies.ES_S3_UP.result, 'Succeeded', 'Skipped'),
+        in(dependencies.ES_S3_UP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
         in(dependencies.GRADLE_WARMUP.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
         or(
           ne(variables['Build.Reason'], 'Manual'),
@@ -242,22 +251,13 @@ stages:
             not(and(eq(variables['Build.Reason'], 'Manual'), eq('${{ parameters.actionToPerform }}', 'run_all_tests_on_windows')))
           )
         displayName: 'Unit tests'
-        container: eclipse-temurin:17-jdk
+        container: rortoolchains
         timeoutInMinutes: 30
         steps:
           - checkout: self
             fetchDepth: 1
             clean: false
             persistCredentials: true
-          - task: Cache@2
-            displayName: 'Restore Gradle cache'
-            inputs:
-              key: 'gradle | "$(Agent.OS)" | **/gradle/wrapper/gradle-wrapper.properties | **/settings.gradle* | **/build.gradle* | **/gradle.properties'
-              restoreKeys: |
-                gradle | "$(Agent.OS)"
-                gradle
-              path: $(GRADLE_USER_HOME)
-              cacheHitVar: GRADLE_CACHE_HIT
           - script: |
               echo "[TEST] executing ROR_TASK = $ROR_TASK"
               ci/run-pipeline.sh
@@ -285,7 +285,7 @@ stages:
               and(eq(variables['Build.Reason'], 'Manual'), eq('${{ parameters.actionToPerform }}', 'run_all_tests_on_linux'))
             )
           )
-        container: eclipse-temurin:17-jdk
+        container: rortoolchains
         timeoutInMinutes: 120
         steps:
           - template: ci/azure-templates/integration-test-steps.yml
@@ -315,7 +315,7 @@ stages:
               and(eq(variables['Build.Reason'], 'Manual'), eq('${{ parameters.actionToPerform }}', 'run_all_tests_on_linux'))
             )
           )
-        container: eclipse-temurin:17-jdk
+        container: rortoolchains
         timeoutInMinutes: 120
         steps:
           - template: ci/azure-templates/integration-test-steps.yml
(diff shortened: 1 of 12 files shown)


sscarduzio:
@coutoPL ready for review. Two themes:

1. **Self-contained build image** — bakes JDK 8/11/17/21 + Gradle dist + primed dep cache + Scala compiler bridges; download-free tasks run `--offline`. Image rebuilt & pushed; `COMPILE_CODEBASE_CHECK`/unit/audit green on it, IT legs run real suites. (Found & fixed two real gaps while validating: the Scala cross-build bridge wasn't primed, and a stale `Cache@2` in the IT template clobbered the baked `/opt/gradle-home`.)
2. **Queue-saturation fix** — `pr.autoCancel: true` + `trigger.batch: true` so a new commit cancels the prior in-progress PR run instead of stacking another full matrix. Verified live (each push auto-cancelled the previous run).

Also: `UPLOAD_PRE_ROR`/`RELEASE_ROR`/`PUBLISH_MVN_ARTIFACTS` now run on the self-hosted pool inside the container (next to the artifact storage; all JDKs + baked cache), which resolves the bare-agent JDK-discovery / lost-cache concerns from the bot reviews.

sscarduzio:
## ✅ Full realistic pipeline green — run [10349](https://dev.azure.com/beshu-tech/ReadonlyREST%20for%20Elasticsearch/_build/results?buildId=10349) on `bfdfb34f5`

All 23 jobs succeeded with **zero retries**: 10 Linux IT legs, 3 Windows IT legs, 4 BUILD assembly legs, unit tests, audit cross-build — every test/build leg first-try green on the baked toolchains image.

Only non-green mark: the optional `CVE_CHECK` leg (`continueOnError`) flags two pre-existing upstream netty 4.1.133 CVEs (CVE-2026-42587, CVE-2026-48043) — unrelated to this PR, same on develop. Skipped stages are the publish stages that never run on PR builds.

Ready for review.

sscarduzio:
@coutoPL full realistic pipeline is now green — run [10349](https://dev.azure.com/beshu-tech/ReadonlyREST%20for%20Elasticsearch/_build/results?buildId=10349): all 23 jobs (10 Linux IT + 3 Windows IT + 4 BUILD + unit/audit), zero retries. Details in the comment above. Ready when you are.

ci/azure-templates/integration-test-steps.yml
```diff
@@ -21,20 +21,13 @@ steps:
...
-  - task: Cache@2
-    displayName: 'Restore Gradle cache'
-    inputs:
-      key: 'gradle | "$(Agent.OS)" | **/gradle/wrapper/gradle-wrapper.properties | **/settings.gradle* | **/build.gradle* | **/gradle.properties'
-      restoreKeys: |
-        gradle | "$(Agent.OS)"
-        gradle
-      path: $(GRADLE_USER_HOME)
-      cacheHitVar: GRADLE_CACHE_HIT
+  # No Cache@2 restore: the Gradle cache is baked into the image at GRADLE_USER_HOME.
```
coutoPL: Comment on the removed code?
Let's remove it.

ci/toolchains/JdkToolchains.Dockerfile
coutoPL, line 1: Rename the file to `JdkToolchains.Dockerfile`

ci/resolve-deps.init.gradle
```diff
@@ -0,0 +1,26 @@
+// Init script used ONLY when priming the CI image's Gradle cache (ci/Dockerfile.toolchains).
```
coutoPL: no ReadonlyREST header in the file

ci/run-pipeline.sh
```diff
@@ -40,9 +40,19 @@ echo ">>> ($0) RUNNING CONTINUOUS INTEGRATION; task? $ROR_TASK"
 # Log file friendly Gradle output
 export TERM=dumb
 
+# Run download-free tasks against the image's pre-seeded Gradle home so they never touch the
+# network (no Gradle dist / JDK / plugin / dependency download -> no flaky-CDN build failures).
+# Default on; export ROR_GRADLE_OFFLINE=false to opt out (e.g. when bumping a dependency, so the
+# cache can be refreshed). NOT applied to tasks that legitimately fetch from the network: the CVE
+# check (NVD data) and anything pulling ES binaries (integration tests, buildRorPlugin, publish).
+OFFLINE_FLAG=""
```
coutoPL: Does --offline add real value here? Both core and integration tests resolve deps from the Gradle cache — but --offline is only applied to some tasks, making the coverage inconsistent. Meanwhile, the flag doesn't protect against the actual risk: if the cache is incomplete, the task fails regardless of whether --offline is set. The image bake step is the right place to guarantee cache completeness; the flag itself just converts a potential network fetch into a hard error with a less helpful message. No visible improvements in the processing time in pipeline too.

ci/toolchains/JdkToolchains.Dockerfile.dockerignore
coutoPL, line 1: This file is used only in the context of toolchain Docker image creation. Let's:
1) create subfolder in `ci` for toolchain-related docker things
2) move the Dockerfile and `resolve-depts.init.grade` there
3) move this file there, too

azure-pipelines.yml
```diff
@@ -84,54 +121,35 @@ stages:
       )
     jobs:
       - job: WARMUP
```
coutoPL: Let's rename it to TOOLCHAINS_VERIFY?

azure-pipelines.yml
```diff
@@ -1,3 +1,11 @@
+# Single source of truth for the CI toolchains image (bump the tag here on a JDK/Gradle change).
+# Declared as a container resource (NOT a variable) because the job-level `container:` property is
+# resolved at compile time and does not expand $(variables).
+resources:
+  containers:
+    - container: rortoolchains
```
coutoPL: maybe `ror-toolchains`?

ci/Dockerfile.toolchains
```diff
@@ -0,0 +1,121 @@
...
+# Why these JDKs (one Java toolchain per supported target):
+#   * JDK  8  -> :audit            (cross-built lib published to Maven Central; Java-8 floor)
+#   * JDK 11  -> :core + ES 6.7-7.x adapters + ror-tools*/tests-utils (ES 6.7-7.x bundle Java 11)
+#   * JDK 17  -> ES 8.x adapters + :benchmarks + :integration-tests   (ES 8.x bundles Java 17)
+#   * JDK 21  -> ES 9.x adapters                                      (ES 9.x bundles Java 21)
+#
+# Build context MUST be the repo root (the build needs settings.gradle/*/build.gradle to
+# resolve dependencies during the priming step):
+#
+#   docker build -f ci/Dockerfile.toolchains -t beshultd/ror-ci-toolchains:jdk-8-11-17-21-gradle-9.2.1 .
```
coutoPL: Shouldn't we have a stage in which the image is built?
I don't see when it's done now. Looks like a manual job.

ci/Dockerfile.toolchains
```diff
@@ -0,0 +1,121 @@
...
+ && apt-get install -y --no-install-recommends git curl ca-certificates unzip \
+ && rm -rf /var/lib/apt/lists/*
+
+# Docker CLIENT only (static binary — distro-agnostic, no apt repo/codename needed):
+# UPLOAD_PRE_ROR/RELEASE_ROR run `docker login` / `publishEsRorDockerImage` from inside this
+# container and talk to the agent host's Docker daemon over the bind-mounted /var/run/docker.sock
+# (the self-hosted az-ror-es agents mount it).
+ARG DOCKER_CLI_VERSION=27.3.1
+ARG DOCKER_CLI_SHA256=9b4f6fe406e50f9085ee474c451e2bb5adb119a03591f467922d3b4e2ddf31d3
+RUN curl -fsSL "https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_CLI_VERSION}.tgz" -o /tmp/docker.tgz \
```
coutoPL: maybe we can do it like this?
```
  FROM docker:27.3.1-cli AS docker-cli

  COPY --from=docker-cli /usr/local/bin/docker /usr/local/bin/docker
```

ci/toolchains/JdkToolchains.Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+# Bake the image's JDK locations into the GRADLE_USER_HOME gradle.properties. This is read reliably
+# at CI time (installations.paths via GRADLE_OPTS is NOT reliably honoured), applies only to jobs
+# that use this baked Gradle home (the Linux container jobs), and is never seen by the bare Windows
+# jobs (they use a workspace-local GRADLE_USER_HOME) — so the Linux paths can't break Windows.
+RUN mkdir -p /opt/gradle-home \
+ && printf '%s\n' \
+      'org.gradle.java.installations.auto-download=false' \
+      'org.gradle.java.installations.paths=/opt/java/jdk8,/opt/java/jdk11,/opt/java/openjdk,/opt/java/jdk21' \
+      > /opt/gradle-home/gradle.properties
+COPY . /tmp/ror-src
```
coutoPL, line 85: This part can also be done in multi-stage. We need this to fetch dependencies, but in the final layer, we don't need it.

ci/Dockerfile.toolchains
```diff
@@ -0,0 +1,121 @@
...
+ #   audit_build_check      -> `audit:crossBuildAssemble` (2.11/2.12/2.13/3 bridges + cross deps)
+ #   core_tests             -> `core:test audit:test`     (so prime their TEST classpaths too)
+ # `testClasses` pulls test-only deps into the cache without running the tests.
+ # Integration tests are NOT offline (they fetch ES binaries + private libs at runtime), so not primed here.
+ && ./gradlew --no-daemon --console=plain --max-workers=1 \
+      -Dorg.gradle.parallel=false -Dorg.gradle.workers.max=1 \
+      -Dorg.gradle.java.installations.auto-download=false \
+      -Dorg.gradle.java.installations.paths=/opt/java/jdk8,/opt/java/jdk11,/opt/java/openjdk,/opt/java/jdk21 \
+      classes \
+      :core:testClasses :audit:testClasses \
```
coutoPL: We use "classes" (root module), so let's use "testClasses" too, instead of calling it on selective modules. Maintainance will be easier

ci/Dockerfile.toolchains
```diff
@@ -0,0 +1,121 @@
...
+      :audit:crossBuildAssemble \
+ # Self-validate OFFLINE the exact classpaths the --offline CI tasks need, so an incomplete cache fails
+ # THIS image build instead of a CI run later. core/audit build outputs are wiped first so the offline
+ # pass genuinely RE-COMPILES (exercising offline Scala compiler-bridge resolution) instead of
+ # short-circuiting on UP-TO-DATE outputs from the online prime above.
+ && rm -rf /tmp/ror-src/core/build /tmp/ror-src/audit/build \
+ && ./gradlew --no-daemon --console=plain --offline --max-workers=1 \
+      -Dorg.gradle.parallel=false -Dorg.gradle.workers.max=1 \
+      classes :core:testClasses :audit:testClasses :audit:crossBuildAssemble \
```
coutoPL: the same here

coutoPL (changes requested):

ci/azure-templates/integration-test-steps.yml
sscarduzio: Done in 70482fd6d — comment removed.

ci/toolchains/JdkToolchains.Dockerfile
sscarduzio: Done in 149f0acf0 — renamed and moved to `ci/toolchains/JdkToolchains.Dockerfile` (combined with the subfolder request below).

ci/resolve-deps.init.gradle
sscarduzio: Done in 149f0acf0 — header added.

ci/toolchains/JdkToolchains.Dockerfile.dockerignore
sscarduzio: Done in 149f0acf0 — Dockerfile, `resolve-deps.init.gradle` and the ignore file all live in `ci/toolchains/` now. The ignore file became the BuildKit sibling `JdkToolchains.Dockerfile.dockerignore`, so the repo root stays clean and it can't affect other docker builds.

azure-pipelines.yml
sscarduzio: Done in 70482fd6d — stage and job renamed to TOOLCHAINS_VERIFY (dependsOn/conditions updated).

azure-pipelines.yml
sscarduzio: Done in 70482fd6d — alias renamed to `ror-toolchains` everywhere.

ci/Dockerfile.toolchains
sscarduzio: Done in 70482fd6d — added a manual-only BUILD_TOOLCHAINS_IMAGE stage (run the pipeline with actionToPerform=build_toolchains_image); it builds the image from `ci/toolchains/JdkToolchains.Dockerfile` and pushes the tag read from `resources.containers`. Kept manual on purpose: rebuilds are rare (JDK/Gradle/dep drift) and the tag bump in this file stays the single switch-over point.

ci/Dockerfile.toolchains
sscarduzio: Done in 149f0acf0 — docker CLI now comes from `docker:27.3.1-cli` via multi-stage COPY (same static binary, no curl/sha dance).

ci/toolchains/JdkToolchains.Dockerfile
sscarduzio: Done in 149f0acf0 — priming now runs in a `gradle-prime` stage; the final image only COPYs the baked /opt/gradle-home, so the repo sources never become a layer of the shipped image.

ci/Dockerfile.toolchains
sscarduzio: Done in 149f0acf0 — root `testClasses` used in both the prime and the offline-validate pass.

ci/Dockerfile.toolchains
sscarduzio: Done in 149f0acf0 (same change in the offline-validate pass).

ci/run-pipeline.sh
sscarduzio: I'd keep it — the point isn't speed, it's enforcement. With a warm cache Gradle is quiet either way, but any cache miss (e.g. a dep bumped without rebaking the image) silently falls back to the network: CI stays green via the flaky CDNs this PR removes, until one of them 504s again. `--offline` turns that regression into an immediate, clear failure (with ROR_GRADLE_OFFLINE=false as the escape hatch) and also stops dynamic/changing-module metadata re-checks from hitting the network. Coverage is intentionally "every task that *can* run offline" — ITs/plugin builds can't, since they fetch ES binaries at runtime. If you still prefer to drop it after this, happy to remove.

coutoPL: The goal is pipeline reliability, not cache enforcement. --offline solves the wrong problem: it doesn't prevent CDN flakiness — it just makes stale-cache situations fail harder. When a dependency is bumped without rebuilding the image, --offline breaks every CI run until someone manually triggers a rebuild. That's less reliable than the occasional CDN 504.

Proposed approach:
1. Keep the baked cache — warm cache hits require no network calls, which is the common case and already solves the CDN flakiness problem.
2. Remove --offline — when the cache is stale (dep bumped), Gradle falls back to the network naturally. The fallback occasionally hits a CDN 504, but it's rare and self-healing.
3. Schedule automatic cache image rebuilds (daily or weekly) — keeps the stale window short without requiring any manual intervention or awareness from the developer who added the dependency.
  
With --offline + manual rebuild, a single dep bump breaks CI for everyone until someone diagnoses and triggers a rebuild. Without --offline + scheduled rebuild, the same scenario results in a short window of occasional network fallbacks — far less disruptive. The scheduled rebuild also fails in isolation (the rebuild job), rather than spreading the failure across every CI run.

coutoPL (changes requested):

ci/run-pipeline.sh
sscarduzio: Fair enough — done in 784f8231c: `--offline` removed everywhere (run-pipeline.sh, the nested IT plugin build, TOOLCHAINS_VERIFY), and BUILD_TOOLCHAINS_IMAGE now also runs on a weekly schedule (develop, Mon 03:00 UTC) in isolation from regular CI runs.

coutoPL (changes requested):
The code is ok for me.
The last request: please change the target branch to master.

sscarduzio:
Revalidation fully green — run [10378](https://dev.azure.com/beshu-tech/ReadonlyREST%20for%20Elasticsearch/_build/results?buildId=10378), 23/23 succeeded on the latest sha (`--offline` removal + weekly toolchains rebuild per review).

One observation from the run: the WIN leg's only real flake was a **foojay JDK-download failure** — Windows legs still fetch toolchains at runtime (the baked image covers Linux container jobs only). Pre-installing JDKs via `JavaToolInstaller` on Windows is a cheap future hardening — follow-up, not this PR.

sscarduzio:
Done — target branch changed to `master` (clean retarget: same merge-base, diff unchanged). Follow-up in 79e9908d8: the weekly `schedules.branches.include` now points at `master` too, so the scheduled toolchains rebuild runs against the branch this yaml actually lands on (develop's yaml has no `Schedule` gating, so a scheduled run there would have executed the full pipeline). PR description updated to match.

azure-pipelines.yml
```diff
@@ -74,7 +74,7 @@ schedules:
     displayName: 'Weekly CI toolchains image rebuild'
     branches:
       include:
-        - develop
+        - master
```
coutoPL: The develop branch was a better choice. In the master branch, there are fewer pipeline runs than on the develop. Master is updated once every one or two months.

coutoPL (changes requested):

azure-pipelines.yml
sscarduzio: Done in 7f3b9be6e — schedule back on develop. (The earlier move to master was my over-correction during the retarget; develop is indeed where the runs and dependency drift are.)

coutoPL (approved):
LGTM

PR #1264, opened 2026-06-09
[RORDEV-1229] pre-build Docker image publishing CI: optional tag + skip-if-unchanged 

coutoPL:
Two additions to the manual pre-build publishing pipeline, plus a refactor:
  
  - **Optional `tag`**: also publishes each image as `<esVersion>-ror-<tag>` pointing at the current commit, so
  consumers can fetch a specific build.
  - **Skip unchanged builds**: tags every build with `<esVersion>-ror-<gitShortSha>` and skips the Gradle
  build+push if that tag already exists (`forceRebuild` overrides).
  - Pipeline made manual-only (`pr: none`); publish logic moved to `ci/ci-lib.sh`.

### DIFF
--- a/.azure/publish-pre-builds.yml
+++ b/.azure/publish-pre-builds.yml
@@ -9,6 +9,10 @@ parameters:
     type: string
     displayName: 'Branch to build from (falls back to develop if not found, builds from current branch if empty)'
     default: ' '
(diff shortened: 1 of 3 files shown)


Dzuming (approved):
LGTM

PR #1267, opened 2026-06-10
[RORDEV-2079] [ci] Authenticated Docker Hub pulls in IT legs (kill the anonymous rate-limit lottery)

sscarduzio:
**Changelog:** [ci] integration-test legs authenticate Docker Hub pulls via `DOCKER_AUTH_CONFIG` (testcontainers + docker-java read it natively), switching from the per-IP anonymous quota — shared across ALL Azure hosted tenants — to a per-account quota.

Two IT legs died to this lottery today alone (`toomanyrequests` on es810x, and the same ror-tools test lost es816x to an apt flake — baking those images is a separate follow-up).

- No-op until the `DOCKER_HUB_RO_TOKEN` (secret) + `DOCKER_HUB_USER` pipeline variables are set — safe to merge independently
- Windows legs untouched (native ES, no Docker)
- Known gap (follow-up): `container:` job images (e.g. `eclipse-temurin`) are pulled by the agent before steps run — covering those needs a Docker registry service connection on named container resources

## Summary by CodeRabbit

* **Chores**
  * Updated the CI integration test steps to optionally enable authenticated Docker Hub pulls when read-only Docker Hub credentials are provided, reducing failures caused by anonymous pull rate limits on hosted agents.


ci/azure-templates/integration-test-steps.yml
```diff
@@ -34,6 +34,12 @@ steps:
       export ROR_LIBS_STORE_URL=$VAR_ROR_LIBS_STORE_URL
       export ROR_LIBS_STORE_ACCESS_KEY_ID=$VAR_ROR_LIBS_STORE_ACCESS_KEY_ID
       export ROR_LIBS_STORE_ACCESS_KEY_SECRET=$VAR_ROR_LIBS_STORE_ACCESS_KEY_SECRET
+
+      # Authenticated Docker Hub pulls (testcontainers/docker-java read DOCKER_AUTH_CONFIG):
+      # avoids the per-IP anonymous rate limit shared across all Azure hosted agents.
+      if [ -n "$VAR_DOCKER_HUB_RO_TOKEN" ]; then
+        export DOCKER_AUTH_CONFIG="{\"auths\":{\"https://index.docker.io/v1/\":{\"auth\":\"$(printf '%s:%s' "$VAR_DOCKER_HUB_USER" "$VAR_DOCKER_HUB_RO_TOKEN" | base64 -w0)\"}}}"
+      fi
```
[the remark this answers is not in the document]
sscarduzio: Valid — a token without a user would produce a malformed `:token` auth entry and silent anonymous fallback. Condition now requires both `VAR_DOCKER_HUB_USER` and `VAR_DOCKER_HUB_RO_TOKEN`. Done in 51b9e2d91.

sscarduzio:
Addressed in `601d7f232`:

**✅ Nit — no confirmation log when auth is active.** Added an `[TEST]`-convention line: `Docker Hub authenticated pulls ENABLED (user '…')` when both vars are set, `DISABLED (anonymous, rate-limited)` otherwise — so flaky-pull investigations can confirm whether auth was in effect.

**❌ Nit — `DOCKER_AUTH_CONFIG` unmasked.** Won't-fix: the script never echoes it (added an explicit comment to that effect), and the value is base64 of a **read-only** pull token, not the raw secret — Azure already masks the source `DOCKER_HUB_RO_TOKEN` var. There's no exposure path here; masking a value that's never printed adds nothing.

sscarduzio:
Reversed my earlier won't-fix and added the masking in `6b324a28e`.

You're right that the real surface is the **agent-level** `system.debug:true` (line 10) machinery, not my script — and `DOCKER_AUTH_CONFIG` is a plain env var, not a pipeline secret, so the agent could surface it in verbose output. Added `##vso[task.setvariable variable=DOCKER_AUTH_CONFIG;isSecret=true]` right after the export: registers the literal for redaction in all subsequent log output, while the child Gradle/testcontainers JVM (launched later in the same step) still reads the real env value. Low-cost given the global debug flag.

ci/azure-templates/integration-test-steps.yml
```diff
@@ -26,13 +26,29 @@ steps:
       export ROR_LIBS_STORE_ACCESS_KEY_ID=$VAR_ROR_LIBS_STORE_ACCESS_KEY_ID
       export ROR_LIBS_STORE_ACCESS_KEY_SECRET=$VAR_ROR_LIBS_STORE_ACCESS_KEY_SECRET
 
+      # Authenticated Docker Hub pulls (testcontainers/docker-java read DOCKER_AUTH_CONFIG):
```
coutoPL: It's a great improvement.
Can we do it in a separate pipeline step?

ci/azure-templates/integration-test-steps.yml
coutoPL, line 1: 1. Don't we need the same improvements in the context of the integration tests on Windows?
2. We do use the test-containers also in case of unit tests

coutoPL (changes requested):

sscarduzio:
Addressed both threads in `aff3ff38c`:

**Thread 1 — "do it in a separate pipeline step?"** ✅ Done. Extracted into a reusable steps-template `ci/azure-templates/docker-hub-auth-steps.yml`. As a *separate prior* step it publishes `DOCKER_AUTH_CONFIG` as a job-scoped variable that the test step inherits — actually cleaner than the same-process `export` it replaces, and the `isSecret` masking still applies.

**Thread 2 — Windows + unit tests:**
- **Unit tests** ✅ — you're right, `core_tests` starts testcontainers too (LDAP/WireMock; `core/build.gradle` pulls `testcontainers-scala`), same per-IP rate limit. Wired the shared step into the `UNIT` job.
- **Windows IT** — deliberately left out: those jobs run ES **natively**, and their Linux-image testcontainers deps (openldap/wiremock/toxiproxy) are `IgnoreOnWindows` because Windows Docker can't run Linux images — so they never pull from Docker Hub. The auth would be dead config there. Happy to revisit if Windows ever grows Linux-container deps.

YAML validated; template paths resolve.

sscarduzio:
Good catch in the latest review — that was a real regression I introduced, fixed in `24066dc14`:

**✅ Warning — secret var didn't propagate after the step extraction.** Correct: `isSecret=true` job variables are **not** auto-injected into subsequent steps' `env` (Azure security behavior). The inline version worked via same-shell `export`; my refactor silently fell back to anonymous pulls. Now explicitly mapped `DOCKER_AUTH_CONFIG: $(DOCKER_AUTH_CONFIG)` in **both** consuming steps (the IT test script and the UNIT job). This is exactly the kind of silent half-broken state worth catching before merge — thank you.

**✅ Nit (Q2) — `$(DOCKER_HUB_USER)` literal expansion when unset.** Hardened the guard: it now also rejects the unexpanded `'$(NAME)'` literal for both user and token, so auth only activates when both are genuinely set (verified all 4 cases: both-real→ENABLED, empty/literal→DISABLED).

The CI run on this commit will confirm the propagation end-to-end. @coutoPL — your two original concerns (separate step + unit-test wiring) were addressed in `aff3ff38c`; this commit fixes the propagation bug that refactor introduced. Ready for re-review when you have a moment.

ci/azure-templates/integration-test-steps.yml
```diff
@@ -21,6 +21,8 @@ steps:
     fetchDepth: 1
     clean: false
     persistCredentials: true
+  # Authenticated Docker Hub pulls (separate, reusable step — shared with the unit-test job).
```
coutoPL: IMO, this comment is not needed. It's described in depth inside docker-hub-auth-steps.yml what this is for.

azure-pipelines.yml
```diff
@@ -308,12 +308,17 @@ stages:
             fetchDepth: 1
             clean: false
             persistCredentials: true
+          # core_tests starts testcontainers too (LDAP/WireMock) — same Docker Hub rate limit as IT.
```
coutoPL: IMO, this comment is not needed. It's described in depth inside docker-hub-auth-steps.yml what this is for.

coutoPL (approved):
Two small, minor comments left. Please fix them, but in general, LGTM.

sscarduzio:
Thanks both — addressed in `a48695bfd`:

**✅ Warning (claude[bot]) — non-JSON `DOCKER_AUTH_CONFIG` when auth is disabled.** Real bug, good catch: with no credentials the `else` branch set no variable, so `$(DOCKER_AUTH_CONFIG)` in the consuming `env:` blocks expanded to the literal `"$(DOCKER_AUTH_CONFIG)"` — non-JSON that docker-java would choke on for every credential-less run (my "harmless" comment was wrong). The `else` branch now publishes an **empty secret**, so the var is always defined → expands to empty → docker-java treats it as "no auth". This is the consuming-side mirror of the input-side dual-guard.

**✅ @coutoPL — redundant comments.** Removed the duplicated `DOCKER_AUTH_CONFIG` explanation at both consuming sites; the full rationale lives in `docker-hub-auth-steps.yml`, with just a one-line pointer left behind.

Re-review when convenient — should be good to merge after this.

sscarduzio:
Dug deeper before merging and found the previous fix (`a48695bfd`, empty-secret in the else branch) was **still broken** — replaced with a robust approach in `a7186e17c`.

**Why the empty-secret approach didn't work** (verified against testcontainers 2.0.3 source + Azure docs):
- testcontainers `RegistryAuthLocator` does `if (getenv("DOCKER_AUTH_CONFIG") != null) ObjectMapper.readTree(it)` — **null-check only, no blank guard**. So it throws a JSON parse error on *both* an empty string `""` **and** the literal `"$(DOCKER_AUTH_CONFIG)"`. The only clean-skip path is the env var being **genuinely absent** (`getenv` → null).
- An Azure cross-step variable **can't express "absent"**: an empty `task.setvariable` is ambiguous/undocumented, and an undefined `$(VAR)` maps to its literal. So *any* `env: { DOCKER_AUTH_CONFIG: $(DOCKER_AUTH_CONFIG) }` mapping risks feeding `RegistryAuthLocator` a value it chokes on — breaking every credential-less CI run.

**The fix:** dropped the pipeline-step template; the auth is now a **sourceable script** (`ci/docker-hub-auth.sh`) that the test step `source`s in its own shell (same process as `run-pipeline.sh`):
- both creds real → `export DOCKER_AUTH_CONFIG=<valid JSON>` (reaches the JVM in-process) + `isSecret` masking
- otherwise → sets nothing → the var stays **unset** → `RegistryAuthLocator` gets `null` → skips cleanly

Still reusable across the IT + unit-test legs (@coutoPL's separate/reusable ask), and now correct by construction. Verified by sourcing in all three scenarios: real→valid JSON, no-creds→unset, Azure-literal→unset.

Re-review appreciated — this one's worth a look since the mechanism changed.

coutoPL (approved):
LGTM

PR #1269, opened 2026-06-10
[RORDEV-2080] [ci] JMH benchmark suite for ACL hot paths: 21 elected KPIs, manually-run before/after tooling (perf-CI legs parked)

sscarduzio:
**Changelog:** [ci] JMH benchmark suite for the ACL hot paths, with a reviewed KPI list (`benchmarks/kpis.yml`). Manually-run tooling: measure before/after numbers for performance-relevant PRs.

## What this PR provides

A new `benchmarks` module that measures how fast the ROR ACL really is, using the same code paths production uses.

**What you can do with it:**

1. **Measure the ACL hot paths** — `./gradlew :benchmarks:jmh` runs all benchmarks: end-to-end ACL evaluation, the indices/groups/actions/headers rules, JWT verification, glob matching, basic-auth decoding.
2. **Check a PR for performance changes** — `./gradlew :benchmarks:jmhSmoke` runs the most important benchmarks (~7 minutes) with allocation profiling. Run it on `develop` and on your branch, compare the two tables, paste them in the PR. `benchmarks/README.md` describes this workflow.
3. **Agree on what we track** — `benchmarks/kpis.yml` is the reviewed list of tracked numbers (21 KPIs in two tiers). Changing what we track = changing this file in a PR. The smoke run automatically covers every tier-1 entry, and `./gradlew :benchmarks:verifyKpis` fails if an entry points to a benchmark that no longer exists.

**Why the numbers can be trusted:** the benchmarks call the real production entry points (`EnabledAccessControlList.handleRegularRequest`, `rule.check`, `PatternsMatcher.create`), with realistic request shapes (20 headers, ~100 ACL blocks, ~100 groups — sizes reported from the field).

## What this PR does NOT do

There is no CI gate and no automated tracking. An automated version (PR gate + nightly time series with alerts) was built, reviewed, and shelved — it needs a dedicated benchmark machine and its alerting logic needs rework first. It lives on the `parked/perf-nightly-tools` branch; `benchmarks/README.md` explains what has to happen before reviving it.

## Footprint

Everything is inside `benchmarks/` plus one `include` line in `settings.gradle`. The module uses plain Gradle tasks (no plugin), so it cannot affect the build of anything else.

Validation: compile clean under `-Xfatal-warnings`, `formatCodeCheck` clean, `verifyKpis` green, `jmhSmoke` runs end-to-end.

https://claude.ai/code/session_01GSyUyjGmV6wXtn3TPbxUbh


benchmarks/baselines/alloc-baseline.json
```diff
@@ -0,0 +1,42 @@
...
+   "host": "leibniz.local",
+   "jdk": "openjdk version \"25.0.2\" 2026-01-20",
+   "os": "Darwin 24.6.0"
+  },
+  "jmhArgs": "-f 1 -wi 2 -i 2 -prof gc (seed)",
+  "srcSha": "df81468daa07fa876c9dd38f11febdbad40b4946"
+ },
+ "toleranceAbsBytes": 64,
+ "tolerancePct": 5.0
+}
```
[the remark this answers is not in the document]
sscarduzio: Partially valid: the gate is advisory (`continueOnError: true`) and the baseline is explicitly `provisional: true` — the documented plan is to regenerate it from a CI agent run (`compare.py --write-baseline`) before flipping the gate to blocking, so the arm64/Darwin seed values are intentional placeholders. The missing piece (no env-mismatch detection) is real though: compare.py now warns when the run arch/OS differs from `baseline.source.env`, pointing at `--write-baseline`. Done in 37e8d5584.

benchmarks/tools/nightly_judge.py
```diff
@@ -0,0 +1,174 @@
...
+            f.write(markdown)
+
+    for message in warnings:
+        print(f"##vso[task.logissue type=warning]{message}" if on_azure else f"WARNING: {message}")
+
+    if violations:
+        for kpi_id, median, new_value in violations:
+            message = (f"KPI regression: {kpi_id} {median:.2f} -> {new_value:.2f} us/op "
+                       f"({(new_value - median) / median * 100:+.1f}% vs rolling median)")
+            print(f"##vso[task.logissue type=error]{message}" if on_azure else f"ERROR: {message}",
```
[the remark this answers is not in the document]
sscarduzio: Valid — with a zero median the MAD threshold is also 0, so any positive value becomes a violation and the % formatting would raise ZeroDivisionError. Guarded as suggested. Done in 37e8d5584.

sscarduzio:
Addressed the second-round review in f23818126.

**W1 — provisional baseline compares wrong-platform numbers.** `compare.py` now short-circuits after the env-mismatch warning when the baseline is `provisional` **and** its `source.env` arch/os differ from the run: it prints `allocation gate SKIPPED …` and returns 0 instead of emitting spurious REGRESSION/improved rows against the arm64/Darwin seed. Same-env comparisons are untouched (verified: same-env big regression still exits 1). The gate activates automatically on the first `--write-baseline` re-seed on the CI agent.

**W2 — enterprise deny path had no PR alloc coverage.** Added `EnterpriseScenarioBenchmark.denyPath` to `ci-smoke.txt` so its allocations are measured on PRs. I deliberately did **not** add a `b_op` baseline value or an `acl.enterprise.deny.alloc` KPI entry — those must be measured on the CI agent, not seeded from the arm64 dev box, so until the authoritative `--write-baseline` runs there it surfaces as `NEW (not in baseline)` (informational, non-gating). README documents this and the follow-up to add the baseline+KPI entry in that re-seed PR. (Answers open Q3: the omission was about *seeding honesty*, not smoke budget.)

**W3 — `null` block stored in the block context.** It was in **five** files, not three. Replaced all sites with a single documented `BenchmarkSupport.noBlock` sentinel explaining that `blockContext.block` is unused on the measured ACL path (only audit/logging reads it) and that a real `Block` would need a `NonEmptyList[Rule]` + implicit `LoggingContext` for a never-read field — so the NPE-if-someone-dereferences-it risk now lives in one obvious place.

**Nit 1 — unescaped `.` in `ci-smoke.txt`.** Dots are now escaped (`\.`) at every package separator.

**Nit 2 — `</script>` in embedded JSON.** `render_dashboard.py` routes both JSON embeds through a `script_json()` helper that neutralises `</` → `<\/` (verified: `</script>` → `<\/script>`).

**Open Q2 — `EsServices(emptyClusterService, null, null)`.** Annotated: the two nulls are `serviceAccountTokenService` and `apiKeyService`, untouched by any rule the KPI benchmarks exercise.

**Open Q1 — re-seed plan.** It's a manual operator step (run `compare.py --write-baseline` on the first real PERF_SMOKE run on the CI image), gated by the `provisional` flag now clearing on `--write-baseline`. Tracked in the README follow-ups; not automated to keep baseline changes reviewed-like-code.

Validation: `:benchmarks:compileScala` green under `-Xfatal-warnings`; all Python tools `py_compile` + behaviorally tested (W1 skip/control, Nit2 escape, dashboard render); every `alloc-baseline.json` key still has a matching `kpis.yml` entry; all `ci-smoke.txt` regexes compile.

The two earlier CodeRabbit inline threads (env-mismatch warning, zero-median guard) were already resolved in 37e8d5584.

sscarduzio:
Addressed the round-3 review in `9d15aaa64`:

**✅ W1 — bare `null` in `RuleStaticResolutionBenchmark`.** `block = null` → `noBlock`. It was the only remaining raw `null` in benchmark construction; now in the same NPE-together safety net as the other 5 sites (`noBlock` was already in scope via the `BenchmarkSupport.*` import).

**✅ W2 — `last_good_sha` is the last-run SHA, not last-good.** Renamed → `last_run_sha` with a comment: it's the most recent recorded SHA (possibly itself violating), so `SUSPECT_RANGE` is a **lower bound** that can under-shoot across multi-night regressions. No false "good" claim. (The precise per-KPI last-passing SHA would need persisted pass/fail history — noted as a follow-up.)

**✅ N1 — dead `improvements` list.** Removed; the `"improved"` status already shows in the markdown row, and improvements never gate the build.

**✅ N2 — `--label performance || true` can silently drop the issue.** Now `--create-label` (creates the label on first use) + a loud `##vso[task.logissue type=warning]` if the create still fails — a regression can no longer go silently UNFILED.

**✅ N3 — `ci-smoke.txt` read at config time.** Moved the read + `args` assembly into `doFirst`, so a missing file no longer `FileNotFoundException`s every `./gradlew` in `:benchmarks`.

Validated: `:benchmarks:compileScala` green; `py_compile` on both tools.

sscarduzio:
Addressed round-4 in `d7fc29e53`:

**✅ W1 — committed `.pyc` bytecode.** Good catch — CPython 3.14 `__pycache__/*.pyc` slipped in from a local run. `git rm`'d both, and added `__pycache__/` + `*.pyc`/`*.pyo` to `.gitignore` so it can't recur (verified `git check-ignore`).

**✅ W2 — push failure suppresses the regression check.** Added `continueOnError: true` to the "Commit run record to benchmark-history" step — a double-network-error push must not skip the downstream judge/dashboard (`condition: succeeded()`) and silently drop the regression. `HISTORY_DIR` is published before the push, so the judge runs against the local worktree record.

**✅ N1 — `jmhCompare` after a failed `jmhSmoke`.** Added `onlyIf { smokeRaw.get().asFile.exists() }` — if `jmhSmoke` crashes before writing the JSON, `jmhCompare` skips cleanly instead of throwing a confusing second `FileNotFoundError`.

**✅ N2 — imports inside `build_panels()`.** Hoisted `calendar` / `datetime` to module top.

Validated: `py_compile` all three tools; `:benchmarks` Gradle config still evaluates.

ci/run-pipeline.sh
```diff
@@ -79,6 +79,7 @@ if [[ $ROR_TASK == "core_tests" ]]; then
   ./gradlew --no-daemon --stacktrace core:test audit:test
 fi
 
+
```
coutoPL: not needed

.gitignore
```diff
@@ -9,6 +9,7 @@ tmp/*
 *.log
 */bin
 
+
```
coutoPL: not needed

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/AclEvaluationBenchmark.scala
```diff
@@ -0,0 +1,128 @@
...
+      s"expected permit in block$blocks, got: $permitResult"
+    )
+    val denyResult = acl.handleRegularRequest(denyContext).runSyncUnsafe()._1
+    require(
+      denyResult.isInstanceOf[RegularRequestResult.ForbiddenByMismatched[?]],
+      s"expected deny, got: $denyResult"
+    )
+  }
+
+  @Benchmark
```
coutoPL, line 69: let's move the benchmarks to the top of the class for the sake of readability. 
The expected order:
1. setup
2. benchmarks
3. util methods

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/AclEvaluationBenchmark.scala
```diff
@@ -0,0 +1,128 @@
...
+        AccessRequirement.MustBePresent(Header(Header.Name(nes("X-Custom-1")), nes("value-1")))
+      )))
+    )
+    // .sorted = production RuleOrdering (what Block.createFrom applies); validation is config-time-only.
+    new Block(Block.Name(s"block$i"), Block.Policy.Allow, Block.Verbosity.Info, Block.Audit.Enabled, rules.sorted)
+  }
+
+  @Setup(Level.Trial)
+  def setup(): Unit = {
+    val allBlocks = NonEmptyList.fromListUnsafe((1 to blocks).map(blockOf).toList)
```
coutoPL: Let's create some well-named private methods. E.g., setupAcl, assertAllowedRequest, assertForbiddenRequest.

The readability of the setup will be improved. The details will be in the private methods at the end of the file.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/AclEvaluationBenchmark.scala
```diff
@@ -0,0 +1,128 @@
...
+  @Param(Array("10", "50", "100", "150"))
+  var blocks: Int = scala.compiletime.uninitialized
+
+  private implicit val loggingContext: LoggingContext = LoggingContext(Set.empty)
+
+  private var acl: EnabledAccessControlList = scala.compiletime.uninitialized
+  private var permitContext: NonIndexRequestContext = scala.compiletime.uninitialized
+  private var denyContext: NonIndexRequestContext = scala.compiletime.uninitialized
+
+  private def blockOf(i: Int): Block = {
```
coutoPL: rename to `createExampleBlock(blockIndex: Int)`

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
```diff
@@ -0,0 +1,177 @@
...
+ * configured groups (user has 100 groups), an indices rule with 120 patterns, an actions rule
+ * and a headers rule, evaluated against a realistic 20-header search request.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(2)
+class EnterpriseScenarioBenchmark {
```
coutoPL, line 52: Similar comments also apply to this benchmark and any other benchmark in this PR

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
```diff
@@ -0,0 +1,177 @@
...
+  private implicit val loggingContext: LoggingContext = LoggingContext(Set.empty)
+
+  private var acl: EnabledAccessControlList = scala.compiletime.uninitialized
+  private var permitContext: IndexRequestContext = scala.compiletime.uninitialized
+  private var denyContext: IndexRequestContext = scala.compiletime.uninitialized
+
+  private def groupId(j: Int): GroupId = GroupId(nes(s"dept-$j"))
+
+  private def groupsRuleOf(i: Int): AnyOfGroupsRule = {
+    val permittedGroupIds = UniqueNonEmptyList.unsafeFrom(
```
coutoPL: createGroup private method needed

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
```diff
@@ -0,0 +1,177 @@
...
+      (0 until configuredGroupCount)
+        .map(j => AlreadyResolved(NonEmptyList.one(groupId(j): GroupIdLike)): RuntimeMultiResolvableVariable[GroupIdLike])
+    )
+    val userDef = UserDef(
```
coutoPL: createUserDef private method needed

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
```diff
@@ -0,0 +1,177 @@
...
+    val requested = (0 until 5)
+      .map(j => RequestedIndex(ClusterIndexName.fromString(s"logs-app-$j-2026").get, excluded = false))
+      .toCovariantSet
+    new IndexRequestContext(realisticHeaders(credentials), requested)
+  }
+
+  @Setup(Level.Trial)
+  def setup(): Unit = {
+    val allBlocks = NonEmptyList.fromListUnsafe((1 to blockCount).map(blockOf).toList)
+    acl = new EnabledAccessControlList(
```
coutoPL: I saw a similar part of the code in the previous benchmark. Maybe we should create a BenchmarkAclUtils file and put there methods that help to create ACL-related objects for benchmark setup purposes. WDYT?

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/domain/BasicAuthDecodeBenchmark.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  @Setup(Level.Trial)
+  def setup(): Unit = {
+    headers = realisticHeaders(Credentials(User.Id(nes("admin")), PlainTextSecret(nes("secret-password"))))
+  }
+
+  @Benchmark
+  def decodePerRequest(bh: Blackhole): Unit = {
+    val context = new NonIndexRequestContext(headers)
+    var i = 0
```
coutoPL: Use more Scala-idiomatic constructions. E.g.
```
(1 to authBlocksPerRequest).foreach(_ => bh.consume(context.basicAuth))
```

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/matchers/GlobPatternsMatcherBenchmark.scala
```diff
@@ -0,0 +1,84 @@
...
+    val rnd = new Random(1234)
+    candidates = Array.tabulate(2000) { _ =>
+      if (rnd.nextInt(100) < 30) s"index-${rnd.nextInt(manyPatterns)}-${rnd.nextInt(1000)}"
+      else s"unrelated-${rnd.nextInt(100000)}"
+    }
+    candidatesList = candidates.toList
+  }
+
+  private def matchAll(matcher: PatternsMatcher[String], bh: Blackhole): Unit = {
+    var i = 0
```
coutoPL: scala-idimatic here too. And in other places of the PR too.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/matchers/GlobPatternsMatcherBenchmark.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  @Setup(Level.Trial)
+  def setup(): Unit = {
+    def patternsOf(n: Int) = (0 until n).map(i => s"index-$i-*")
+    fewCaseSensitive = PatternsMatcher.create[String](patternsOf(fewPatterns))(using Matchable.caseSensitiveStringMatchable)
+    manyCaseSensitive = PatternsMatcher.create[String](patternsOf(manyPatterns))(using Matchable.caseSensitiveStringMatchable)
+    manyCaseInsensitive = PatternsMatcher.create[String](patternsOf(manyPatterns))(using Matchable.caseInsensitiveStringMatchable)
+
+    // ~30% of candidates match a prefix; ~70% miss and scan all buckets.
+    val rnd = new Random(1234)
```
coutoPL: needed some private, well-named method

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/rules/ActionsRuleBenchmark.scala
```diff
@@ -0,0 +1,68 @@
...
+@Fork(2)
+class ActionsRuleBenchmark {
+
+  private var rule: ActionsRule = scala.compiletime.uninitialized
+  private var matchingContext: GeneralNonIndexRequestBlockContext = scala.compiletime.uninitialized
+  private var missingContext: GeneralNonIndexRequestBlockContext = scala.compiletime.uninitialized
+
+  @Setup(Level.Trial)
+  def setup(): Unit = {
+    val actions = (1 to 9).map(i => Action(s"indices:admin/custom-$i/*")).toList :+ searchAction
```
coutoPL: Some new private, well-named methods needed. Here and in other benchmarks too.

benchmarks/build.gradle
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+ext {
+    jmhVersion = '1.37'
+}
+
+dependencies {
+    implementation project(':core')
+    implementation group: 'org.scala-lang',   name: 'scala3-library_3',         version: '3.3.7'
```
coutoPL: order dependencies by name

benchmarks/ci-smoke.txt
coutoPL, line 1: Do we need to have two files? The one and kpis.yml? Maybe we could read all the tier 1 benchmarks' names from there? And use them for a smoke test?

benchmarks/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+- **Old-path replicas live only while their PR is open.** A PR that optimizes a hot path may
+  add `oldPath_*` replica methods as review evidence, but they are pruned when the PR merges.
+- KPI benchmarks pin the canonical size at field max (~100 blocks / ~100 groups); `@Param`
+  axes around it make scaling curvature visible without exponent series.
+- Blocks are built with the production `RuleOrdering` (`rules.sorted`), and the indices KPI has
+  both a concrete-name variant and a wildcard-expansion variant (the wildcard path is the
+  production hot path for `logs-*`-style requests).
+
+## Parked CI legs
```
coutoPL: I struggled to understand this section. I guess you wanted to write sth like that: "we already tried to build the automated version, here's why it's shelved, and here's what has to happen before anyone revives it". 

It should be rewritten in simple, easy-to-grasp form. Currently, it's written from the perspective of the one who implemented the solution. Be empathic and think from the perspective of a reader. 

No need to write an essay about it. Just clear, simple form.

coutoPL (changes requested):
The description should be fixed. Too many technical details.
Should be focused on what functionalities the PR provides. 
Clear, simple form, taking into account the reader perspective. 
Jargon and methaphors should be limited.

ci/run-pipeline.sh
sscarduzio: Outdated — this PR no longer touches ci/run-pipeline.sh (removed in 949d46a; footprint is benchmarks/ + one settings.gradle line).

.gitignore
sscarduzio: Outdated — .gitignore is no longer touched by this PR (removed in 949d46a).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/AclEvaluationBenchmark.scala
sscarduzio: Done in 1f5b952 — every benchmark class now reads setup -> @Benchmark methods -> private util methods.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/AclEvaluationBenchmark.scala
sscarduzio: Done in 1f5b952: setup() is now 6 lines — createAcl(...) + createExampleBlock(blockIndex) + assertRequestAllowedBy/assertRequestForbiddenBy (the assert helpers live in BenchmarkAclUtils so all benchmarks share them).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/AclEvaluationBenchmark.scala
sscarduzio: Renamed to createExampleBlock(blockIndex: Int) in 1f5b952.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
sscarduzio: Applied to every benchmark in 1f5b952 — same order and private-creator style throughout.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
sscarduzio: Done in 1f5b952 — createGroup lives in the shared BenchmarkAclUtils (used by both this benchmark and GroupsRuleBenchmark).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
sscarduzio: Done in 1f5b952 — createUserDef in BenchmarkAclUtils, shared with GroupsRuleBenchmark.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/acl/EnterpriseScenarioBenchmark.scala
sscarduzio: Agreed and done in 1f5b952 — new support/BenchmarkAclUtils with the shared ACL-object creators (createAcl, createAuthKeyRule, createGroupId/createGroup/createUserDef, createAnyOfGroupsRule) plus the assertion helpers. AclEvaluationBenchmark, EnterpriseScenarioBenchmark and GroupsRuleBenchmark all use it.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/domain/BasicAuthDecodeBenchmark.scala
sscarduzio: Done in 1f5b952 — exactly that form: (1 to authBlocksPerRequest).foreach(_ => bh.consume(context.basicAuth)).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/matchers/GlobPatternsMatcherBenchmark.scala
sscarduzio: Done across the PR in 1f5b952 — the remaining var/while loops are gone (candidates.foreach(...) here).

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/matchers/GlobPatternsMatcherBenchmark.scala
sscarduzio: Done in 1f5b952 — createMatcher(patternCount, matchable) and createCandidates() private methods.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/rules/ActionsRuleBenchmark.scala
sscarduzio: Done everywhere in 1f5b952 — createActionsRule/createBlockContext here; every benchmark got the same treatment.

benchmarks/build.gradle
sscarduzio: Done in 1f5b952 — ordered by name (jmh-core, jmh-generator-bytecode, scala3-library_3; the :core project dependency stays first).

benchmarks/ci-smoke.txt
sscarduzio: Great idea — done in 1f5b952. ci-smoke.txt is deleted; jmhSmoke now parses kpis.yml and runs every tier-1 benchmark (deduped across .time/.alloc pairs). One reviewed file owns both what we track and what the smoke run covers. Net effect on the subset: gains rsa2048 (tier 1), drops actionMissed (tier 2). Verified end-to-end: jmhSmoke runs the 9 tier-1 methods.

benchmarks/README.md
sscarduzio: Rewritten in 1f5b952 as "The automated version (shelved)": what we built, the two plain-language reasons it is shelved (no dedicated machine; alerting logic that could silently normalize a slow regression), and the 3 concrete steps before anyone revives it.

sscarduzio:
Round-7 review addressed in 1f5b952 (all threads replied + resolved):
- **PR description rewritten** — functionality-first, plain language, jargon cut.
- **Structure**: every benchmark class is now setup → benchmarks → private util methods; well-named private creators everywhere (`createExampleBlock(blockIndex)`, `createIndicesRule`, …).
- **Shared `BenchmarkAclUtils`** — ACL-object creation (`createAcl`, `createGroup`, `createUserDef`, `createAnyOfGroupsRule`) + assertion helpers, used by all three ACL-heavy benchmarks.
- **Scala-idiomatic loops** — no var/while left.
- **`ci-smoke.txt` deleted** — `jmhSmoke` now runs the tier-1 entries parsed from `kpis.yml`; one reviewed file owns what we track AND what the smoke covers. `verifyKpis` still guards against renames.
- **README**: shelved-automation section rewritten reader-first; deps ordered by name.

Verified locally: compile clean under `-Xfatal-warnings`, `formatCodeCheck` clean, `verifyKpis` green (15 entries), `jmhSmoke` end-to-end green (9 tier-1 methods, `-prof gc`). Ready for another look.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/support/BenchmarkSupport.scala
coutoPL, line 1: I wonder if we shouldn't merge BenchmarSupport with BenchmarkAclUtils. It looks like both are util-like classes. Please decide.

coutoPL (approved):
LGTM.

One minor comment left. Fix if you think it's worth doing that.

benchmarks/src/main/scala/tech/beshu/ror/benchmarks/support/BenchmarkSupport.scala
sscarduzio: Decided: keep them separate. They split on a real axis — BenchmarkSupport is the fake WORLD the benchmarks run in (request contexts, ES-cluster stubs, header shapes), BenchmarkAclUtils builds the production ACL objects UNDER measurement (rules/blocks/ACLs) + decision asserts. Merging would make one ~280-line util grab-bag — the shape round-7 just refactored away. fadd1a6e makes the split self-documenting: each file's scaladoc now names its axis and points at the other, so the next reader sees it is deliberate, not accidental.

coutoPL: Ok, but do you think the names of the classes are not misleading? They look similar.

PR #1272, opened 2026-06-12
[RORDEV-2081] Always return Kibana index in user metadata response

coutoPL:
The `/_readonlyrest/metadata/user` endpoint did not always include the kibana.index field. When a user was authenticated by a block that did not explicitly declare a Kibana index (for example, a block with only `auth_key` and no Kibana rule), the response omitted the index entirely:

```
  "kibana": {
    "access": "unrestricted"
  }
```

This was inconsistent with ROR's actual behavior: when no Kibana index is declared, ROR already falls back to the default index (.kibana) internally. Only the metadata response failed to reflect that, leaving the ROR Kibana plugin without an index to rely on.

With this change, kibana.index is always present in the response, defaulting to .kibana when not explicitly configured. The API contract has been updated to mark the field as required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Kibana index metadata is now always included in access control responses and authorization results; defaults to ".kibana" when not specified.

* **New Features**
  * Selection now prefers the first allowed block that establishes a Kibana policy when resolving the Kibana index.

* **Refactor**
  * Kibana policy/index handling standardized across evaluation paths and rules.

* **Documentation**
  * Internal API version and changelog updated to mark kibana.index as required.

* **Tests**
  * Test expectations updated for guaranteed kibana.index presence and new selection behavior.

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/EnabledAccessControlList.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/EnabledAccessControlList.scala
@@ -166,7 +166,7 @@ class EnabledAccessControlList(val blocks: NonEmptyList[Block],
     if (ignoreGroupsHandling || withGroups.isEmpty) {
       val firstFittingMatch = withLoggedUsers
         .allowedThroughFirstForbidden()
-        .firstAllowedWithKibanaIndexOrHead()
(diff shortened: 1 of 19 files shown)


mgoworko (approved):
LGTM

PR #1274, opened 2026-06-17
[RORDEV-2087] Fix TSVB visualizations not showing for kibana_access: ro users (ES/Kibana 9.x)

coutoPL:
🐞Fix (ES/KBN) Fix TSVB visualizations not showing for `kibana_access: ro` users for ES/KBN 9.x

---

## Problem
On all ES/Kibana 9.x versions, TSVB / legacy visualization panels are not shown for read-only Kibana users (`kibana_access: ro`). The panel renders empty and Kibana reports `"Could not find the data view"`. RW/admin users are unaffected.

## Root cause
Kibana's TSVB backend route `/internal/metrics/vis/data` issues an `indices:data/write/bulk` against the `.kibana` index while rendering (to resolve/persist the data view). `BaseKibanaRule` only permits RO writes to `.kibana` whose REST path (`/.kibana/url/...`, `/doc/index-pattern...`) and can never match a batched `POST /_bulk`. ROR denies the write (`FORBIDDEN by default`); "Could not find the data view" is just the secondary symptom Kibana surfaces.
  
## Fix
Treat the TSVB write as another non-strict location RO may write to: recognize the originating Kibana endpoint via the trusted `x-ror-kibana-request-path` header and OR it into `isRoNonStrictCase`. The allowance stays narrow — only `.kibana`, only RO(`ROStrict` still locked), only `indices:data/write/*`, and only from that specific endpoint. No new matcher, no extra `shouldMatch` branch, no ES-module changes; both `KibanaAccessRule` and `KibanaUserDataRule` inherit the fix.

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/kibana/BaseKibanaRule.scala
+++ b/core/src/main/scala/tech/beshu/ror/accesscontrol/blocks/rules/kibana/BaseKibanaRule.scala
@@ -104,8 +104,24 @@ abstract class BaseKibanaRule(val settings: Settings)
     isTargetingKibana &&
       isAccessOtherThanRoStrictConfigured &&
       kibanaCannotBeModified &&
-      isNonStrictAllowedPath &&
+      (isNonStrictAllowedPath || isMetricsVisDataRequest) &&
       isNonStrictAction
+  }
+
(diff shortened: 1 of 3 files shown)


coutoPL:
Fixed on the ROR KBN side: https://github.com/sscarduzio/readonlyrest_kbn/pull/931

PR #1275, opened 2026-06-18
[RORDEV-2098] Fix CVE_CHECK: switch to NVD data feeds and suppress false positives 

mateuszkp96:
## Summary by CodeRabbit

* **Chores**
  * Updated the dependency vulnerability scanning plugin to a newer version for improved coverage and reporting.
  * Tuned vulnerability database/NVD behavior by disabling automatic NVD updates and adjusting fetch and retention settings.
  * Re-enabled the optional CVE-check pipeline job so it runs when the surrounding pipeline conditions allow, including loading/saving the CVE cache.
  * Refreshed CVE suppression rules to reflect updated dependency versions and added/adjusted several CVE entries to better handle false positives.


coutoPL (approved):
LGTM

PR #1279, opened 2026-06-30
[RORDEV-2074] Building and publishing optimizations

coutoPL:
🧐 Enhancement (ES) Smaller ES with ReadonlyREST Docker images
🧐 Enhancement (ES) Faster Docker image pulls across Elasticsearch versions

## Summary                                                                                                                                                 

Eliminates redundant Scala compilation and Docker image pushes when distributing the ROR plugin across all ES patch versions within a single module's range.                                                                                                         
                                                                                                                                                                       
### Build model: compile once, repackage many                                                                                                                        
                                                                                                                                                                       
Each ES module now compiles Scala **once**, at its oldest (baseline) supported ES version. All other patch versions are derived by repackaging without recompiling:            
- `plugin-descriptor.properties` — `elasticsearch.version` is patched to the target version (ES enforces it matches the node version)                                                                                                                              
- `ror-build-info.properties` — `es_version` inside the fat ROR jar is updated (boot log only)                                                                       
- ES-versioned dependency jars (e.g. `elasticsearch-rest-client-9.x.y`) — fetched via `downloadEsVersionedDependencyJars` and swapped in                                                                                                                 
                                                                                                                                                                       
Everything else in the ZIP is identical across patch versions and is reused unchanged.                                                                               

### Bytecode safety guard                                                                                                                                            
                                                                                                                                                                       
`verifyRepackageBytecodeNewest` compiles the module at its **newest** ES version and byte-diffs the class files against the base. If any class differs — meaning the compiler produced version-dependent output — the task fails loudly before any repackage step runs. This makes the "compile once, reuse" assumption an enforced invariant rather than a convention.                                  

### Docker: shared fat layer across patch versions                                                                                                                   
                                                                                                                                                                       
The ES+ROR Docker image is now split into two `COPY --link` stages:                                                                                                  
- **`payload/`** — the fat, version-independent layer (~100 MB): all ROR jars whose bytecode is the same across a module's patch range. Because `--link` builds this layer independently of the ES base chain, its content digest is identical for every ES patch version of a module. The registry stores this blob once; each per-version image references it.                                                                                              
- **`per-version/`** — the thin, per-version layer (a few MB): `plugin-descriptor.properties`, the ES-version-stamped dependency jars, and the renamed module jar. This is the only part that changes between patch versions.                                                                                                                                    
                                                                                                                                                                       
`gosu` is now sourced from the pinned `tianon/gosu:1.17@sha256:...` upstream image rather than rebuilt via alpine + GPG fetch/verify on every image build.

### Convention plugin consolidation                                                                                                                                  
                                                                                                                                                                       
`readonlyrest.plugin-common-conventions` now applies the three specialized sub-conventions (`verify-repackage-bytecode`, `scala-compile-conventions`, `docker-image-conventions`) itself. Each ES module's `build.gradle` is reduced to declaring its JDK version and its own dependencies.                                                                    
                                                                                                                                                                       
### CI publish pipeline                                                                                                                                              
                                                                                                                                                                       
- `printEsModules` and `printEsVersionsForModule` emit to stdout; callers pass `--quiet` to suppress Gradle's own output, eliminating temp file intermediaries.                                                                                                
- `es_jars_dir` tmpdir is scoped to each module invocation (owned and cleaned by `publish_module_versions`) rather than the outer publish loop.                                                                                                     
- Per-version ZIPs are deleted immediately after upload to cap peak disk usage at ~1 ZIP at a time.                                                                                                                                                              
- Build and Docker cleanup in `publish_module_versions` is guarded to release mode only.   

### RESULTS                                                                                                                                              
                                                                                                                                                                               
  Per-version release time dropped from ~6m15s to ~1m30s per plugin (~4.2x faster), by reusing a single built zip and repackaging it per target version instead of doing a full
  build for every ES patch version.                                                                                                                                            
                                                                                                                                                                               
  Before (build #10368, publishRorPlugin + publishEsRorDockerImage, one full build per version):                                                                               
  - ~231s to build & upload the artifact                                                                                                                                       
  - ~144s to build & push the Docker image                                                                                                                                     
  - ~375s (6m15s) per plugin version                                                                                                                                           
  - This run was actually cancelled after ~3h having only gotten through 28/29 attempted versions — at that rate a full ES 8.x release (107 versions) would take an estimated  
  ~11h                                                                                                                                                                         
                                                                                                                                                                               
  After (build #10655, repackageRorPluginForVersion -PreusePackagedZip + pushRorDockerImage, build zip once per module then repackage per version):                            
  - ~19s to repackage & upload the artifact                                                                                                                                    
  - ~64s to build & push the Docker image                                                                                                                                      
  - ~90s (1m30s) per plugin version                                                                                                                                            
  - Full ES 8.x release (107 versions) completed in 2h45m                                                                                                                      
                                                                                                                                                                               
  Net effect: ~75% reduction in release time for the ES 8.x job alone, saving roughly 8.5 hours per full release run.


sscarduzio:
Adversarial + strict-correctness review (swarm, findings verified against the diff). Net: the consolidation is a genuine, correct win — version coverage, distJars exclusions, bytecode-guard wiring, and packaging parity all check out. **One blocker**, the rest are confirm-with-author nits.

## 🚨 Blocker — JDK-8287073 Corretto swap dropped from Docker images that still bundle the crashing JDK

The rewritten Dockerfiles for **es714x, es716x, es80x, es81x, es82x, es83x, es84x** (and the `--add-modules=jdk.net` write for **es717x, es82x, es83x, es84x**) lost the per-version JDK swap. `FROM …elasticsearch:${ES_VERSION}` is unchanged, so these images ship the affected bundled JDK.

Verified: develop's `es82x/Dockerfile` carries `# ES 8.2.x bundles JDK 18.0.x which crashes on cgroup v2 hosts (JDK-8287073) … Corretto 19.0.0`; the new file has no swap block. Same for the whole set (`corretto|8287073|add-modules` hit count new=0 across es714x–es84x vs old=3–6; es85x+ correctly 0). The affected ranges match the repo's own **unchanged** `Elasticsearch.scala::hasBuggyBundledJdk` (7.15.1–7.17.6, 8.0.0–8.4.x) and `needsCorretto19` / `--add-modules=jdk.net`. `ror-tools patch` patches ES bytecode, not the JDK — no compensating mechanism exists on the branch.

Concrete failure: published images for these ranges NPE at JVM start (`CgroupV2Subsystem.getInstance()`) on any cgroup-v2 host (modern Docker/K8s) → **ES never boots**; es717x/es82x/es83x/es84x additionally hit `NoClassDefFoundError: jdk.net.Sockets` from HttpClient 5 (still a live `core` dep) at ROR external-auth runtime. Aggravating: develop built only `latestSupportedEsVersion` per module, this PR builds/publishes the whole range → these broken lower-patch images are **net-new artifacts**.

Fix: re-introduce the es717x-style PATCH-gated conditional Corretto swap in the affected Dockerfiles + the `jvm.options.d/ror.options` `--add-modules=jdk.net` write. If intentional, state why the base image no longer ships the affected JDK and why HttpClient 5 no longer needs `jdk.net`.

## Nits (confirm, non-blocking)

- **`suppressions_cve.xml:38,79`** — newly suppresses `CVE-2021-22135` + `CVE-2021-22137` (Elasticsearch CVEs) inside the existing false-positive block. Confirm each is genuinely N/A to the shipped ES range, not silencing a live vuln to pass the gate.
- **`ci/publish-ror-plugins.sh:41`** — `es_jars_dir=$(mktemp -d)` is `rm -rf`'d only on the success path; five early `return 1`s (47/54/59/71/90) leak it. Add `trap 'rm -rf "$es_jars_dir"' RETURN` — counter to the PR's disk-exhaustion goal otherwise.
- **`.zip.sha1` sidecar dropped** (`plugin-common-conventions.gradle`, dropped `ant.checksum … sha1`; publish script uploads only zip + `.sha512`). Fine if no external download page / mirror fetches `<zip>.sha1`; confirm none does. (In-zip `.sha512` unchanged and correct.)
- **`ci/publish-ror-plugins.sh`** — S3 upload runs unconditionally while docker re-tag is `checkTagNotExist`-gated; a re-run of an already-released version re-PUTs every zip (benign — zips are reproducible — but real cost). Gate S3 on the same check if idempotent re-runs are wanted.
- **`verify-repackage-bytecode.gradle:117` + `docker-image-conventions.gradle`** — bytecode guard samples only `esVersions.newest`, and the Docker fat/thin layer hardcodes ES-versioned jar globs while the zip repackage discovers them dynamically. Both latent (today's modules agree); worth a comment that the guard samples endpoints, or loop it over `esVersions.all`.

**Verdict: approve after fixing the blocker.** The dedup itself is well-earned (−6489/+3086, `EsVersions` correctly replaces the unstable `.internal.VersionNumber` API, exclusions byte-identical) — no over-engineering flagged.

sscarduzio (approved):
## Approved

### What I checked
- All 14 build-base Java/Groovy files: BytecodeComparison, EsModuleFinder, EsVersions, PluginArtifact, PrintEsModulesTask, convention plugins (verify-repackage-bytecode, scala-compile-conventions, docker-image-conventions, plugin-common-conventions)
- All 35 es*/Dockerfile changes (gosu pinning, COPY --link fat/thin split, tini paths, USER directives, Corretto patch stages)
- All 35 es*/build.gradle and es*/gradle.properties (consistency, supportedEsVersions as single source of truth)
- ci/publish-ror-plugins.sh (repackage flow, tmpdir lifecycle, disk management, base zip preservation)
- ci/ci-lib.sh, ci/run-pipeline.sh (tag function, integration test flow)
- azure-pipelines.yml (stage conditions, dependencies)
- Unit tests: BytecodeComparisonTest, EsModuleFinderTest, EsVersionsTest
- Cross-module consistency verified across all 35 ES modules

### What looks good
- **BytecodeComparison** is clean — SHA-256 digest approach is sound, toHex is correct, tests cover diverging/non-diverging/ignored-entry cases
- **Compile-once-repackage-many** architecture is well-structured: base and target compiles produce differently-named jars so both coexist for comparison
- **Cross-module consistency** — all 35 modules have identical build.gradle pattern, correct gosu digest pinning, consistent COPY --link fat/thin split
- **CI pipeline** — tmpdir lifecycle is correct (RETURN trap + mktemp -d), base zip preserved during version loop, per-version ZIPs deleted after upload
- **Unit tests are genuine** — BytecodeComparisonTest tests actual byte differences, EsVersionsTest covers pre-release ordering edge cases
- **4.2x faster releases** — measured 6m15s → 1m30s per version

### Non-blocking suggestions for follow-up
1. **[MAJOR] docker-image-conventions.gradle:88-108** — The fat/thin layer split uses hardcoded jar name globs (e.g. `elasticsearch-rest-client-*.jar`) instead of dynamic discovery from `downloadEsVersionedDependencyJars`. If a future module adds an ES-versioned dependency with a different name pattern, it would contaminate the fat layer silently. Consider adding a validation step that asserts the globs match the dynamically discovered artifacts.
2. **[MAJOR] plugin-common-conventions.gradle:264-272** — `repackageRorPluginForVersion` jar-swap: the `delete()` return value is not checked. If the base jar isn't found (artifact name changed between versions), the delete silently no-ops and the zip could contain BOTH base and target jars → jar-hell at runtime. Consider asserting `delete() == true`.

Neither of these is triggered by current data — they're forward-looking hardening suggestions.

PR #1284, opened 2026-07-02
[RORDEV-2107] Return MainSettingsNotFound for missing in-index settings document

coutoPL:
The provide-in-index-settings endpoint only treated a missing settings index as "not found" (status "empty"); a missing settings document (index present, doc absent) fell through to a generic failure (status "ko"). Map DocumentNotFound to MainSettingsNotFound as well so both cases report status "empty". Updated BaseAdminApiSuite accordingly. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved settings lookup behavior so missing settings are handled correctly in more cases, returning a clearer “not found” outcome.
  * Updated related test expectations to match the revised response format.

* **Chores**
  * Bumped the build plugin version to the latest pre-release.


Dzuming (approved):
LGTM

PR #1263, opened 2026-06-08
[RORDEV-2077] [test] Faster, reliable integration tests on free hosted runners: Docker layer sharing (−16%/leg), disk fixes, opt-in sharding

sscarduzio:
## What

Make the Elasticsearch **integration-test suite fast and reliable on the free Microsoft-hosted Azure runners** — no runner resize, no paid runners, full test coverage preserved (all 34 ES versions, every suite). What started as "real parallel fork JVMs" converged, after measurement, on a tighter set of wins: a Docker-layer-sharing rebuild that cut per-leg wall-clock **~16%**, the disk-exhaustion fixes that let the heavy ES 8.x legs fit at all, and a batch of correctness fixes for races/hangs/leaks uncovered along the way. Suite **sharding** stays as an opt-in knob (`-PshardCount`) rather than the default.

## Headline results — before vs after (per-leg wall-clock, shardCount=1)

Measured on CI: master baseline (build 10531) vs this PR (build 10541), every IT leg.

| ES leg | before (master) | after (this PR) | Δ |
|---|--:|--:|--:|
| es67x | 61m | 55m | **-10%** |
| es70x | 60m | 54m | **-11%** |
| es710x | 70m | 61m | **-12%** |
| es711x | 72m | 63m | **-13%** |
| es714x | 85m | 74m | **-12%** |
| es716x | 74m | 66m | **-11%** |
| es717x | 75m | 67m | **-10%** |
| es72x | 62m | 53m | **-15%** |
| es73x | 59m | 53m | **-9%** |
| es74x | 61m | 54m | **-12%** |
| es77x | 66m | 58m | **-13%** |
| es78x | 67m | 60m | **-11%** |
| es79x | 70m | 62m | **-12%** |
| es80x | 90m | 74m | **-18%** |
| es810x | 79m | 68m | **-14%** |
| es81x | 89m | 75m | **-16%** |
| es811x | 81m | 68m | **-16%** |
| es812x | 89m | 69m | **-22%** |
| es813x | 81m | 70m | **-14%** |
| es814x | 83m | 71m | **-14%** |
| es815x | 88m | 73m | **-17%** |
| es816x | 84m | 73m | **-14%** |
| es818x | 81m | 71m | **-13%** |
| es82x | 88m | 72m | **-18%** |
| es83x | 86m | 71m | **-17%** |
| es84x | 87m | 70m | **-19%** |
| es85x | 79m | 66m | **-17%** |
| es87x | 79m | 66m | **-16%** |
[...]

PR #1287, opened 2026-07-08
[ci] Fix BUILD_* legs on all PR builds: configuration-time quiet() log pollutes the printEsModules parse

sscarduzio:
**Changelog:** [ci] fix all-PR BUILD_[6-9]xx failures caused by a semantic conflict between #1263 and #1279.

## What broke

Every PR build now fails all four `BUILD_[6-9]xx` legs with:

```
Cannot locate tasks that match ':>>> integration-tests maxParallelForks=1 (1 ES/JVM; parallelism via -PshardCount):verifyRepackageBytecodeNewest'
```

First seen on #1269 (build 10700) — but it will hit **every** PR.

## Why

A semantic conflict between two individually-green PRs merged 7 minutes apart this morning:

- #1263 added `logger.quiet(">>> integration-tests maxParallelForks…")` inside the eager `test {}` block of `integration-tests/build.gradle`. It fires at **configuration time** and, being `quiet()`, **survives `--quiet`** (that was the deliberate point — keep the fork count visible).
- #1279 added `list_es_modules()` = parse the stdout of `./gradlew printEsModules --quiet` into module names.

Neither PR's CI ever saw the other's change, and develop pushes don't run the BUILD legs (only `AUDIT_BUILD_CHECK`) — so the combination was first exercised by the next PR merge build.

## Fix

1. `logger.quiet` → `logger.lifecycle`: still printed in normal IT runs (verified — it shows in a default-log-level gradle run), suppressed under `-q`, so machine-parsed output stays clean.
2. Defense in depth: `list_es_modules()` filters `printEsModules --quiet` output through `grep -E '^es[0-9]+x$'`, so any future configuration-time output can't re-poison the module list.

Verified locally: `./gradlew printEsModules -PesMajor=8 --quiet` emits exactly the 17 `es8xx` module names.

This PR's own BUILD legs are the end-to-end proof — they run `build_ror_plugins` against the merged tree containing the fix.

https://claude.ai/code/session_01GSyUyjGmV6wXtn3TPbxUbh

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command output handling so only the intended module names are returned, preventing extra build messages from affecting publishing workflows.

* **Chores**
  * Adjusted logging during test setup to better support quieter, more reliable automated output.

### DIFF
--- a/ci/publish-ror-plugins.sh
+++ b/ci/publish-ror-plugins.sh
@@ -20,9 +20,10 @@ cleanup_docker_and_build() {
 }
 
 # Emits one ES module name per line for the given generation (newest module first).
+# grep guards against configuration-time build-script output leaking into --quiet stdout.
 list_es_modules() {
   local es_major=$1
-  ./gradlew printEsModules "-PesMajor=$es_major" --quiet </dev/null
+  ./gradlew printEsModules "-PesMajor=$es_major" --quiet </dev/null | grep -E '^es[0-9]+x$'
 }
 
 # Emits two lines: the base ES version on line 1, all supported versions space-separated on line 2.
--- a/integration-tests/build.gradle
+++ b/integration-tests/build.gradle
@@ -133,8 +133,9 @@ test {
     // maxParallelForks=1 always — forks can't split the scalatest engine (gradle#8632), so a 2nd fork
     // gets no work. Parallelism = sharding (multiple -PshardCount invocations), not forks.
     maxParallelForks = 1
-    // Logged loud (quiet() survives -q) so the per-JVM fork count is never a mystery.
-    logger.quiet(">>> integration-tests maxParallelForks=${maxParallelForks} (1 ES/JVM; parallelism via -PshardCount)")
+    // lifecycle, NOT quiet(): this runs at configuration time, so quiet() pollutes the stdout that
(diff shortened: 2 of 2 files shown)


ci/publish-ror-plugins.sh
```diff
@@ -20,9 +20,10 @@ cleanup_docker_and_build() {
 }
 
 # Emits one ES module name per line for the given generation (newest module first).
+# grep guards against configuration-time build-script output leaking into --quiet stdout.
 list_es_modules() {
   local es_major=$1
-  ./gradlew printEsModules "-PesMajor=$es_major" --quiet </dev/null
+  ./gradlew printEsModules "-PesMajor=$es_major" --quiet </dev/null | grep -E '^es[0-9]+x$'
```
coutoPL, line 26: Why didn't we solve it where it's defined, rather than in bash?!

sscarduzio: Fair — the grep duplicated es-module naming knowledge that belongs to the task. Fixed where it's defined in #1290: `printEsModules` now writes `build/es-modules/es<major>x.txt` and `list_es_modules` reads the file; gradle stdout is never parsed at all, so no configuration-time logging can poison the list by construction. (The `logger.quiet`→`lifecycle` root-cause fix from this PR stands unchanged.)

PR #1288, opened 2026-07-09
[RORDEV-2112][ROR 1.68.0 patch] Fix ES nodes staying permanently unusable when ROR settings cannot be read at startup

coutoPL:
## The problem

A customer reported that after a cluster restart some ES nodes rejected every request, and kept
rejecting them until the node was restarted by hand. Two independent issues combined to cause it.

**1. ROR gave up after the first failed start attempt.**
When a node boots, ROR reads its settings from the `.readonlyrest` index. Early during a cluster
start that index is often not readable yet - the master node may not be elected, the shards may not
be allocated. ROR treated this like a fatal error: it marked itself as "not started" and never tried
again. The node stayed in that state forever, even though the settings became available seconds
later.

**2. ROR silently fell back to the settings file when the index was unreadable.**
ROR could not tell apart two very different situations: "the settings index has no settings in it"
and "the settings index cannot be read right now". In both cases it fell back to the local
`readonlyrest.yml` file. That is correct in the first case, but dangerous in the second - the file
may contain older or different rules than the ones the rest of the cluster is using, so a node could
end up enforcing the wrong access control settings without anybody noticing.

## The fix

**ROR now keeps trying to start.** Instead of one attempt, ROR retries until it succeeds. It waits 5
seconds before the first retry and gradually increases the delay up to 1 minute, so a node that boots
before its settings are available simply starts a little later, on its own, with no manual
intervention. Every failed attempt is logged and the node keeps reporting that ROR is not ready yet,
exactly as before.

**ROR no longer falls back to the file when the settings index is unreadable.** The two situations
above are now distinguished. If the settings genuinely are not in the index, the local file is used
as a fallback, as before. If the index exists but cannot be read, ROR reports a clear error
explaining that the file will *not* be used as a fallback - because it could differ from the settings
used by the rest of the cluster - and retries instead.


sscarduzio:
## Review — verdict: approach is correct; one fix requested before merge

The two-part design is right: (1) indefinite capped-backoff retry of the whole start sequence, (2) splitting "settings absent" (`DocumentNotFound`/`IndexNotFound` → file fallback) from "settings unreadable" (`DocumentUnreachable` → loud error + retry, **no** fallback). I traced the seams end-to-end; the security-relevant classification holds.

### Verified correct

- **Classification seam** (`EsIndexDocumentManager.onErrorRecover`): `IndexNotFound` is produced only from a genuine `IndexNotFoundException`; everything else (incl. `ClusterBlockException` before state recovery, `NoShardAvailableActionException` while shards are unassigned — the customer's exact boot scenarios) hits the catch-all → `DocumentUnreachable` → retry. Since ES blocks reads before cluster-state recovery, a not-yet-recovered node can't misreport "index missing", so the file fallback can't be reached spuriously.
- **Retry util** (`ScalaOps.retryUntilSuccessful`): Task-trampolined (stack-safe), delay capped at `maxDelay`, `onFailedAttempt` failures can't kill the loop, exceptions pass through by design and `startWithRetry` converts them to `Left` so they retry too. 5s→60s cap is a sane pace.
- **Resource lifecycle**: `EngineResources` released on `Left`, on thrown error, and via `doOnCancel`; `IndexLevelActionFilter.stop()` cancels the retry task, so node shutdown mid-retry doesn't spin forever. This closes the per-attempt HTTP-client/LDAP-pool leak that infinite retry would otherwise have introduced.
- **Cross-module consistency**: all 33 ES modules on this branch got the identical `startWithRetry` call-site change; no leftover `ror.start(...)` callers.
- **Tests** cover retry-until-success, exception-attempt, broken listener, no-retry-on-success, and both fallback-decision branches including the "will NOT be used as a fallback" message.
- Nice property worth keeping in the changelog: because each attempt re-reads settings, an operator can fix the index/file while the node retries and it recovers **without a restart**.

### Fix before merge

1. **`curl -fsSLk` in the 8 test-image Dockerfiles** — `-k` disables TLS verification for a JDK that replaces the ES-bundled one, with no checksum either. develop already solved this better: digest-pinned `FROM amazoncorretto:…@sha256:…` stages + BuildKit bind-mounts (no tarball layer, no network trust issue) — see develop's `es82x/Dockerfile`. Recommend backporting that pattern; at minimum drop `-k` (corretto.aws has a valid cert) and pin the sha256. Test-image scope only, but it's free to do right.

### Worth a note (not blocking)

2. **Behavior change: a corrupt/unparseable settings document now retries forever.** The JSON-parse `IllegalStateException` lands in the catch-all → `DocumentUnreachable` → infinite retry, where the old code fell back to the file. The new behavior is arguably safer (no silent divergence), but it's a *permanent* failure looping at 1/min — worth one line in the changelog so operators know the remedy is fixing/deleting the index doc, not restarting.
3. **Test-settings asymmetry**: `loadTestSettings()` still swallows *all* errors → `None`, so a node started while the test-settings doc is unreachable silently drops the cluster's impersonation settings — the same divergence class this PR fixes for main settings. Fine to keep out of scope; flagging for a follow-up.
4. **azure-pipelines.yml**: the disk-cleanup block is duplicated 6×; an in-repo step template would keep them in sync. (`target: host` is correct — the disk pressure is on the host.)
5. **Tiny race**: if cancellation lands while the `Left`-branch `release()` is running, `doOnCancel` can call `release()` a second time. Probably harmless (both shutdowns look idempotent), but an `AtomicBoolean` guard in `EngineResources.release()` would make it provably so.

sscarduzio:
## Addendum: broadened failure-scenario matrix — one more misclassification found

Follow-up to my review: I walked the full space of index-access failure modes through the new classification. One genuine hole, the rest hold up.

### The hole: a malformed settings document silently falls back to the file

`RawRorSettingsCodec` parses the YAML **at decode time**, so a doc that *exists* but doesn't decode (malformed YAML, truncated write, wrong JSON shape, schema written by a different ROR version during a rolling upgrade) produces `SettingsLoadingError.SettingsMalformed` — which lands in the `case _ =>` branch of `loadMainSettings()` and **silently falls back to `readonlyrest.yml`**:

```scala
.leftFlatMap {
  case SettingsLoadingError.SourceSpecificError(IndexSettingsSource.LoadingError.DocumentUnreachable) =>
    cannotReadExistingIndexSettings
  case _ =>                       // <- SettingsMalformed falls through here
    loadMainSettingsFromFile().leftMap(_.show)
}
```

This is exactly the divergence scenario the PR description warns about: the cluster's authoritative settings ARE in the index (an active shard served the doc), but a restarting node switches to its local file while non-restarted nodes keep enforcing the in-memory index settings — split-brain ACL. It's not a *regression* (the old code also fell back on any index error), but it's the PR's own new principle applied incompletely: "present but unreadable" should mean *loud error + retry*, and `SettingsMalformed` is "present but unreadable".

**Suggested fix:** route `SettingsMalformed` to the no-fallback branch with its own message ("settings found in index '…' but malformed: <cause>; will NOT fall back to file…"). The escape hatch stays intact: deleting the broken doc/index yields `DocumentNotFound`/`IndexNotFound` → legitimate file fallback. And thanks to the new retry loop, fixing the doc heals the node without a restart.

### Scenarios that hold up (verified against the `EsIndexDocumentManager` mapping, identical in all 33 modules)

| Scenario | Surfaces as | Classified | Outcome |
|---|---|---|---|
| Cluster state not recovered (boot) | `ClusterBlockException` | catch-all → `DocumentUnreachable` | retry ✓ (the PR's target case) |
| Index exists, shards unassigned (red) | `NoShardAvailableActionException` | catch-all → retry ✓ | if data loss is permanent, operator deletes the index → `IndexNotFoundException` → file fallback — worth one line in docs as the recovery path |
| Index genuinely absent | `IndexNotFoundException` (raised **locally** from cluster state; can't fire while state is unrecovered because the block wins first) | `IndexNotFound` → file fallback ✓ |
| Doc absent, index present | `isExists=false` from an **active** shard (authoritative) | `DocumentNotFound` → fallback ✓ |
| Index closed | `IndexClosedException` | catch-all → retry forever | acceptable; the error text won't say *why* — operator must spot the closed index |
| `_source` not valid JSON at all | `IllegalStateException` in the manager | catch-all → retry ✓ (consistent with treating it as unreadable — note the asymmetry with the decode-level case above) |
| Remote/wrapped exceptions (`RemoteTransportException` wrapping anything) | won't match the typed cases | catch-all → retry | fails **safe** — an unknown failure can never spuriously trigger the file fallback |
| Circuit breaker / threadpool rejection at boot | catch-all → retry ✓ |

### A design property worth stating (and preserving)

Only three things can ever reach the file fallback: local `IndexNotFoundException`, `ResourceNotFoundException`, and an authoritative `isExists=false` — every *unknown* failure fails toward retry. That's the right failure direction for an ACL: unknown ⇒ don't guess, retry loudly. Fixing the `SettingsMalformed` routing would make this property complete; a comment on the `onErrorRecover` block saying "new exception types must default to `DocumentUnreachable`" would keep future edits from eroding it.

sscarduzio:
## Addendum 2: logging audit of the retry / fallback decision points

Traced every operator-visible log line through the new decision tree. The outer loop is solid; four gaps at the decisions themselves.

### What's good

- Every failed start attempt logs `ROR starting failure: …` at **ERROR with the stacktrace** (`logAndSetStartingFailureState`), followed by `ReadonlyREST will try to start again in <delay>` at WARN — an operator tailing the log sees the node is down, why, and that it will self-heal.
- The no-fallback refusal (`cannotReadExistingIndexSettings`) logs at ERROR with the full "will NOT be used as a fallback, because…" rationale. Exactly the message that saves a support ticket.
- `EsIndexDocumentManager`'s catch-all logs the underlying exception at ERROR before classifying it `DocumentUnreachable`, so the root cause (cluster block, no shard, …) is never lost.

### Gaps

1. **The file-fallback decision is never logged explicitly** — the single most important decision in this PR. When `DocumentNotFound`/`IndexNotFound` → file, the log shows only the *sequence* ("Loading … from index … failed: Cannot found document" at INFO, then "Loading … from file …" at INFO); the operator must infer the causal link. One WARN would make it diagnosable at a glance: `"No settings found in index '.readonlyrest' — falling back to file '/…/readonlyrest.yml'. Index settings will take precedence once created."` Given the whole PR exists because this fallback silently did the wrong thing, the moment it fires deserves its own line.
2. **The inner `ConfigurableRetryStrategy` logs only at DEBUG** — both the per-attempt failures and the "failed permanently after N attempts" terminal line are invisible at default log level. The terminal line at least should be INFO/WARN: it's the event that *triggers* the fallback-or-refuse decision.
3. **Test-settings load failures are swallowed with zero trace** (`.recover { case _ => None }`) — a node that silently dropped the cluster's impersonation settings logs nothing at any level. Even a single WARN ("test settings could not be loaded from index — starting without them: <error>") would do.
4. **A throwing `onFailedAttempt` listener is silenced with no log** (`onErrorHandle(_ => ())` in `retryUntilSuccessful`). Surviving it is right (and tested 👍), but swallowing it invisibly means a broken failure-reporter can never be discovered. A `logger.debug`/`warn` inside that handler costs nothing.

### Polish (take or leave)

- The retry WARN has no attempt counter or elapsed time — `attempt 47, retrying in 1 minute (down for 43m)` turns a log grep into an incident timeline, and is cheap since `retryUntilSuccessful` already threads state through the recursion.
- On the `DocumentUnreachable` path each attempt currently emits ~3 ERROR entries (manager + `cannotReadExistingIndexSettings` + `logAndSetStartingFailureState`) — fine for visibility, just noting it's deliberate spam-per-attempt at up to 1/min, forever.

PR #1289, opened 2026-07-10
[RORDEV-2112] Fix ES nodes staying permanently unusable when ROR settings cannot be read at startup

coutoPL:
## The problem

A customer reported that after a cluster restart some ES nodes rejected every request, and kept
rejecting them until the node was restarted by hand. Two independent issues combined to cause it.

**1. ROR gave up after the first failed start attempt.**
When a node boots, ROR reads its settings from the `.readonlyrest` index. Early during a cluster
start that index is often not readable yet - the master node may not be elected, the shards may not
be allocated. ROR treated this like a fatal error: it marked itself as "not started" and never tried
again. The node stayed in that state forever, even though the settings became available seconds
later.

**2. ROR silently fell back to the settings file when the index was unreadable.**
ROR could not tell apart two very different situations: "the settings index has no settings in it"
and "the settings index cannot be read right now". In both cases it fell back to the local
`readonlyrest.yml` file. That is correct in the first case, but dangerous in the second - the file
may contain older or different rules than the ones the rest of the cluster is using, so a node could
end up enforcing the wrong access control settings without anybody noticing.

## The fix

**ROR now keeps trying to start.** Instead of one attempt, ROR retries until it succeeds. It waits 5
seconds before the first retry and gradually increases the delay up to 1 minute, so a node that boots
before its settings are available simply starts a little later, on its own, with no manual
intervention. Every failed attempt is logged and the node keeps reporting that ROR is not ready yet,
exactly as before.

**ROR no longer falls back to the file when the settings index is unreadable.** The two situations
above are now distinguished. If the settings genuinely are not in the index, the local file is used
as a fallback, as before. If the index exists but cannot be read, ROR reports a clear error
explaining that the file will *not* be used as a fallback - because it could differ from the settings
used by the rest of the cluster - and retries instead.


coutoPL:
@claude review please

coutoPL:
@claude review please

coutoPL:
review in this PR: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1288

PR #1292, opened 2026-07-10
[RORDEV-1228] Move CI from Azure Pipelines to GitHub Actions (Ubicloud Linux + GitHub-hosted Windows)

sscarduzio:
This PR moves our CI from Azure Pipelines to GitHub Actions. Linux jobs run on Ubicloud runners (4 CPU / 16 GB). Windows jobs run on GitHub's own `windows-2025` machines (free for public repos). Azure triggers are turned off in this same PR, but the Azure pipeline can still be run by hand as a fallback until we delete it.

The build scripts themselves are untouched — every job still calls `ci/run-pipeline.sh` with a `ROR_TASK`, exactly like on Azure. Only the orchestration around them was rewritten.

## The headline numbers

A full PR run now takes **~55 minutes**. On Azure it took **4h48**. Cost is about **$0.50 per PR run** (~$14/month).

| | Azure | Now | Why |
|---|---|---|---|
| One Linux test leg | 85–100 min (mostly queue) | **18–24 min** | each leg runs 4 test JVMs in parallel on its own VM |
| One Windows test leg | ~70–78 min | **28–40 min** | 3 parallel test JVMs, antivirus exclusions, faster ES installs |
| Whole run | 4h48 | **~55 min** | every job gets its own fresh VM — no shared queue |

## What's in the PR

- **`.github/workflows/ci.yml`** — all 14 Azure stages as GitHub jobs. Test matrices are computed per trigger: full 34 Linux / 33 Windows versions on develop/master, a 10 / 3 subset on PRs. Skipped versions never start a VM.
- **Test parallelism** (new, Azure ran everything serially):
  - Each Linux leg splits its suites across 4 JVMs, packed by measured duration so no JVM becomes the long pole (`integration-tests/suite-timings.json`).
  - A machine-wide gate (`HeavySuiteGate`) limits how many multi-node-cluster suites can boot at the same time. Without it, 16 GB machines run out of memory — we measured this, repeatedly.
  - Windows got the same treatment: 3 JVMs with separate port ranges and install dirs, since ES runs as native processes there (no Docker).
- **Leaner test ES nodes**: features no test ever uses are disabled per version — ML, watcher, built-in index templates, APM templates, and similar. Less memory, faster boots.
- **Fixes found along the way**: a secret that would have been printed to public logs by `ci/docker-hub-auth.sh`, a WireMock port clash between parallel Windows JVMs, an ILM test race, and a too-short wait in the audit tests.
- **Docs**: `ci/github/PORT.md` (how the port maps to Azure, and how the parallelism works), `ci/github/SECRETS.md` (all 17 secrets + 8 variables to set, with a helper script).

The current settings (4 JVMs, gate of 2, balanced packing) are the measured sweet spot for this machine size. We tried going further — 5 JVMs, gate of 3, smaller heaps, deleting unused ES modules from the image — each on an isolated one-leg test run. Every one of them hit a wall (out of memory, boot timeouts, or an ES-internal module dependency), so this is where the knobs stay.

## Validation

The full matrix (10 Linux + 3 Windows legs) ran green end to end, including the release-path job conditions (checked against a written scenario table for every trigger type). Two independent review passes over the whole diff found no correctness issues.

<details>
<summary><b>Azure → GitHub Actions mapping</b> (reviewer reference — the current-state architecture lives in <code>ci/README.md</code>)</summary>

| Azure stage | GH job | Notes |
|---|---|---|
| SUPERSEDE_GUARD | — | replaced by native `concurrency` (cancel-in-progress for PRs only; branch pushes queue, so a release run is never cancelled) |
| DISK_PROBE | `disk_probe` | manual `run_disk_probe` |
| ES_S3_UP | `es_s3_up` | `newes/*` branches; ordered before all check/test jobs |
| BUILD_TOOLCHAINS_IMAGE | `build_toolchains_image` | weekly cron + manual |
| TOOLCHAINS_VERIFY | `toolchains_verify` | |
| OPTIONAL_CHECKS (CVE) | `cve_check` | continue-on-error; monthly cache key |
| REQUIRED_CHECKS | `required_checks` | 4-way matrix |
| TEST (unit) | `unit_tests` | |
| TEST (Linux IT ×3 jobs) | `it_linux` | one dynamic matrix (full 34 / PR 10) |
| TEST (Windows IT ×3 jobs) | `it_windows` | one dynamic matrix (33 / 7 / 3) |
| TEST (Windows unit) | `unit_tests_windows` | manual |
| BUILD_ROR | `build_ror` | PR only; gates on Linux **and** Windows |
| DETERMINE_CI_TYPE → UPLOAD_PRE_ROR / RELEASE_ROR / PUBLISH_MVN | same names | `!cancelled()` + explicit `needs.<job>.result` (GH skips dependents of skipped jobs — this keeps `release_without_testing` working) |

All 5 Azure manual actions are ported: `run_all_tests_on_linux`, `run_all_tests_on_windows`, `build_toolchains_image`, `release_without_testing`, `run_disk_probe`.

Azure-feature translations: `$(System.AccessToken)` → `GITHUB_TOKEN`; `##vso[task.setvariable]` → `$GITHUB_OUTPUT`; secure file `secret.pgp` → base64 secret decoded in-step; `Cache@2` monthly CVE key → `actions/cache` + `date +%Y%m`; tag push via `permissions: contents: write` (no SSH deploy key).

Deliberately different from Azure: markdown-only changes anywhere skip CI (Azure's globs couldn't express that); the docker pre-clean/reap steps are dropped (runners are ephemeral VMs).
</details>

## Before merging

1. Set the repo secrets/variables (`ci/github/set-secrets.sh`, values map 1:1 from the Azure variable group).
2. Add an NVD API key to make `cve_check` green (it's the only red job).
3. After merge, keep Azure as a manual fallback for one release cycle, then delete it.

## Summary by CodeRabbit

* **New Features**
  * Added a GitHub Actions-based CI pipeline with sharded Linux/Windows integration testing, release/publish gating, auditing, and artifact reporting.
  * Introduced deterministic suite sharding utilities, suite timing support, and optional “heavy suite” concurrency gating.
* **Bug Fixes**
  * Improved integration-test resilience with async polling/retries for ILM step changes, settings reloads, audit sink recovery, and container startup.
  * Reduced cross-shard interference via Windows port/directory sharding, reused templates/ES unpacking, and better toolchain/worker reuse.
* **Documentation**
  * Added a migration guide covering workflow behavior, trigger mappings, and first-run checklist.


integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAuditingToolsSuite.scala
```diff
@@ -86,8 +86,11 @@ trait BaseAuditingToolsSuite
       }
     }
 
+  // 60s, not 15: audit entries flush to the audit sink asynchronously, and on sharded CI runners
```
coutoPL: sounds like PR comment, not long-living code comment - to remove

integration-tests/build.gradle
```diff
@@ -94,7 +118,24 @@ test {
                 exclude 'tech/beshu/ror/integration/suites/base/**'
             }.visit { fd ->
                 if (!fd.directory) {
-                    def fqcn = fd.path.replace('/', '.').replaceAll(/\.class$/, '')
+                    suites << fd.path.replace('/', '.').replaceAll(/\.class$/, '')
+                }
+            }
+            def matched = 0
```
coutoPL: **Can we move the shard-packing logic out of `build.gradle`?**

The sharding itself is fine — I just don't like where this code lives or how it reads.
We have ~45 lines of imperative Groovy inside `filter { }`: a mutable list built by a
file-tree `visit`, a `matched` counter bumped in two branches, a `long[] bins` array with
a hand-rolled loop to find the lightest bin. That's a real algorithm (LPT bin-packing,
hash-mod fallback, default weight, name tie-break) buried in a build script.

Two problems:

1. **It can't be tested.** Things worth asserting: every suite lands in exactly one shard,
   the same input always gives the same partition, a missing timings entry degrades
   gracefully. Today the only way to catch a bad pack is a 20-minute CI run.
2. **Its other half already lives elsewhere.** `ShardedGradlewTest` — which spawns the K
   child invocations — is a tested Java class in `build-base`. This is the same feature and
   belongs next to it.

Proposal — a pure function in `build-base`, no Gradle types:

```java
public final class SuiteSharder {
  public static List<String> suitesFor(
      List<String> allSuites, int shardCount, int shardIndex,
      Map<String, Long> timings /* empty => hash packing */);
}

build.gradle keeps only the Gradle I/O: walk the class tree, parse suite-timings.json,
call suitesFor(...), pass the result to includeTestsMatching, keep the zero-match guard
and the log line. ~45 lines become ~8. The important comments (why hash packing is safe on
memory, why balanced packing needs ROR_HEAVY_SUITE_PERMITS) move into the class javadoc,
next to the code they explain.

Leave the JSON parsing and file-tree walk in Gradle — keeping that I/O out is what makes
SuiteSharder testable with plain maps and lists.

File: integration-tests/suite-timings.json [hunk 3]
```diff
@@ -0,0 +1,67 @@
+{
+ "_comment": "Measured wall-clock seconds per suite (es94x, run 29167332908, incl. container boots). Regenerate from sharded-logs when suites change materially; unknown suites default to 60s in build.gradle.",
```
coutoPL: 1. How to regenerate it? Include it in the comment
2. Who is supposed to do it and when? How does the

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/XpackSecurityPlugin.scala
```diff
@@ -62,7 +62,8 @@ class XpackSecurityPlugin(esVersion: String, config: Config) extends Elasticsear
   override def updateEsConfigBuilder(builder: EsConfigBuilder): EsConfigBuilder = {
     builder
       .add("xpack.security.enabled: true")
-      .add("xpack.ml.enabled: true")
+      // xpack.ml.enabled: false comes from the base config (Elasticsearch.baseEsConfigBuilder) —
```
coutoPL: please remove this comment

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
```diff
@@ -154,12 +154,34 @@ abstract class EsContainer(
     try {
       super.start()
     } catch {
+      // Known docker classic-builder race: two sharded test JVMs concurrently building images that
+      // share parent layers can fail the export with "unknown parent image ID" (moby bug). It's
+      // transient — the parent exists once the sibling's build settles — and it cost a full-matrix
+      // run one red leg roughly every third run. One delayed retry re-resolves the image cleanly.
+      case ex: Throwable if isDockerParentImageRace(ex) =>
+        logger.warn("Docker parent-image build race detected - retrying container start once", ex)
```
coutoPL: put this code in the well-named private method to imrove readability

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/Wiremock.scala
```diff
@@ -27,11 +27,18 @@ object Wiremock {
   def create(mappings: List[String], portWhenRunningOnWindows: Int = 8080): WiremockContainer = {
     OsUtils.currentOs match {
       case CurrentOs.Windows =>
+        // Same per-shard port window as WindowsEsPortProvider: sharded Windows runs share one host,
+        // and two shards starting wiremock-backed suites concurrently raced for the same fixed port
+        // ("Failed to bind to /0.0.0.0:8080"). 8080/8081 + k*1000 never reaches 9200 + k*1000 (ES),
+        // so the windows stay disjoint. originalPort moves too — on Windows it is what gets
+        // substituted into readonlyrest.yml placeholders (RorSettingsAdjuster), and providePort is
+        // what test clients dial; all three must agree on the shifted port.
+        val shardedPort = portWhenRunningOnWindows + Integer.getInteger("ror.shard.index", 0) * 1000
```
coutoPL: We have "ror.shard.index" hardcoded in many places. Maybe we can introduce RorShard util and encapsulate all the ROR shard-related logic in it?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/HeavySuiteGate.scala
```diff
@@ -0,0 +1,97 @@
...
+ * How: N slot FILES under the root project dir, taken with OS file locks. Cross-process by
+ * construction (the shard workers are separate JVMs on one machine), and self-healing: if a
+ * worker crashes or is OOM-killed, its lock dies with the process and the slot frees itself.
+ *
+ * Deadlock-safe by design: ONE permit per SUITE, acquired before any of its clusters start and
+ * held until all are stopped — there is no hold-and-wait on a second permit.
+ *
+ * Opt-in: enabled only when ROR_HEAVY_SUITE_PERMITS is set (CI); local runs are unaffected.
+ */
+object HeavySuiteGate extends LazyLogging {
```
coutoPL: The code is too imperative and unreadable. Let's make it more FP-like.

coutoPL: **`HeavySuiteGate` — bad name.**                                                                                                                                               
                                                                                                                                                                                 
  There's nothing about suites in this class. `suiteName` is only used in two log lines.                                                                                         
  The rest is permits, slot files and OS locks — it's a plain cross-process semaphore.                                                                                           
                                                                                                                                                                                 
  Suggestion:                                                                                                                                                                    
  - rename to `FileLockSemaphore` (or `CrossProcessSemaphore`)                                                                                                                   
  - `suiteName` → `label`, it's just a log tag                                                                                                                                   
  - pass permits + slot dir as arguments instead of reading `ROR_HEAVY_SUITE_PERMITS`                                                                                            
    and `project.dir` inside the class                                                                                                                                           
                                                                                                                                                                                 
  `HeavySuiteGate` then becomes a couple of lines at the call site. Same code, no new                                                                                            
  abstraction — and the semaphore becomes testable, which today it isn't.

ci/github/PORT.md
coutoPL, line 1: do we need this file in the repo?

ci/github/SECRETS.md
coutoPL, line 1: do we need this file in the repo?

ci/github/set-secrets.sh
coutoPL, line 1: do we need this file in the repo?

coutoPL (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,681 @@
...
+    # Schedule runs are pinned to develop (Azure pinned the weekly rebuild to develop; GH schedules
+    # fire on the default branch — the ref guard protects us if the default branch ever changes).
+    if: >-
+      (github.event_name == 'schedule' && github.ref == 'refs/heads/develop') ||
+      (github.event_name == 'workflow_dispatch' && needs.setup.outputs.action == 'build_toolchains_image')
+    runs-on: ubicloud-standard-4
+    timeout-minutes: 240
+    steps:
+      - uses: actions/checkout@v4
+        with: { fetch-depth: 1 }
```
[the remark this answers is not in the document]
sscarduzio: Done in 80e056f — `persist-credentials: false` on all 13 non-pushing checkouts; `release_ror` keeps `true` for its tag push.

ci/github/set-secrets.sh
```diff
@@ -0,0 +1,59 @@
...
+# Register the CI secrets/variables needed by the GitHub Actions port (see SECRETS.md).
+# Fill in the values, then run:  ci/github/set-secrets.sh
+# Re-runnable (gh overwrites). Nothing here is committed with a real value.
+set -euo pipefail
+
+REPO="${REPO:-sscarduzio/elasticsearch-readonlyrest-plugin}"
+
+# --- helper: set only if a value is provided (skip blanks so you can do it in passes) ---
+set_secret() { [ -n "${2:-}" ] && gh secret   set "$1" -R "$REPO" --body "$2" && echo "secret  $1" || echo "skip    $1 (empty)"; }
+set_var()    { [ -n "${2:-}" ] && gh variable set "$1" -R "$REPO" --body "$2" && echo "var     $1" || echo "skip    $1 (empty)"; }
```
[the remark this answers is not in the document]
sscarduzio: `set-secrets.sh` was removed in 80e056f (one-time setup helper; the secrets/vars table now lives in PORT.md), so the failure/skip conflation is gone with it.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/HeavySuiteGate.scala
```diff
@@ -0,0 +1,97 @@
...
+          StandardOpenOption.WRITE
+        )
+        val lock = channel.tryLock()
+        if (lock != null) {
+          val waited = (System.currentTimeMillis() - start) / 1000
+          logger.info(s"[$suiteName] heavy-suite slot $i/$n acquired after ${waited}s")
+          acquired = new Slot(channel, lock)
+        } else {
+          channel.close()
+        }
```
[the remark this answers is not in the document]
sscarduzio: Fixed in 80e056f: `HeavySuiteGate` was replaced by `FileLockSemaphore`, whose `tryAcquireSlot` closes the channel on every branch, including a throwing `tryLock()`.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseAuditingToolsSuite.scala
sscarduzio: Done in 80e056f — rewritten as a timeless comment (why the window must be generous on a loaded machine), no run history.

integration-tests/build.gradle
sscarduzio: Done in 80e056f — as proposed: `SuiteSharder` in build-base (pure function, no Gradle types, next to `ShardedGradlewTest`) with 10 unit tests covering: disjoint cover in both modes and for k=1..8, determinism, input-order independence, sparse timings degrading to the 60s default, stale entries for deleted suites ignored, LPT bin spread bounded by the largest suite, invalid shard specs rejected. `build.gradle` keeps only the class-tree walk, the JSON parse and `includeTestsMatching`.

integration-tests/suite-timings.json
sscarduzio: Done in 80e056f — the file now embeds a `_readme`: what it is, advisory semantics (missing suites default to 60s and still run; stale entries are ignored — staleness affects balance, never coverage), the regeneration procedure (copy each root `<testsuite time=...>` from a green leg's `*-results` artifact), when to bother (only when shard durations drift apart), and ownership.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/XpackSecurityPlugin.scala
sscarduzio: Removed in 80e056f.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
sscarduzio: Done in 80e056f — extracted to `retryStartAfterDockerParentImageBuildRace(...)`, with `isDockerParentImageBuildRace(...)` as the guard.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/Wiremock.scala
sscarduzio: Done in 80e056f — `RorShard` (tests-utils misc) now owns `ror.shard.index`: `index`, `portOffset`, `shardedName`. `Wiremock`, `WindowsEsPortProvider` and `WindowsEsDirectoryManager` all read it from there; the only remaining literal is where build.gradle sets the property.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/HeavySuiteGate.scala
sscarduzio: Done in 80e056f — rewritten as `FileLockSemaphore`: `@tailrec` acquire loop, `Option`/`Try` instead of null checks and vars, `iterator.collectFirst` over the slots.

sscarduzio: Done in 80e056f — `FileLockSemaphore(permits, slotDir, slotFilePrefix)`, `suiteName` → `label`, and the env/project-dir wiring moved to the call site: `HeavySuiteGated` in support.scala is now the couple-of-lines glue you described. `HeavySuiteGate.scala` is deleted.

ci/github/PORT.md
sscarduzio: Kept — but it's now the *single* CI doc: stage map, per-trigger regression table, the sharding/gate architecture and its measured tuning limits, and (as of 80e056f) the secrets/variables inventory folded in from SECRETS.md. That's the operational knowledge someone debugging or re-creating this CI needs, and it fits nowhere else. If you'd rather have no doc at all, say so and I'll drop it.

ci/github/SECRETS.md
sscarduzio: Deleted in 80e056f; the durable part (secrets/variables table + the PGP base64 note) is folded into PORT.md.

ci/github/set-secrets.sh
sscarduzio: Deleted in 80e056f — one-time setup helper, and the real values live in the secret manager anyway. PORT.md documents the equivalent `gh secret set` / `gh variable set` usage.

integration-tests/suite-timings.json
```diff
@@ -0,0 +1,77 @@
+{
+  "_readme": [
+    "Measured wall time (seconds) per integration suite. Used by build-base's SuiteSharder for",
+    "duration-balanced shard packing when ROR_BALANCED_SHARDS=true (wired in integration-tests/",
+    "build.gradle; CI sets it in .github/workflows/ci.yml).",
+    "ADVISORY ONLY: suites missing here default to 60s and still run; entries for deleted suites",
+    "are ignored. Staleness affects shard balance, never correctness or coverage.",
+    "REGENERATE only when shard durations drift visibly apart in CI: download a green it_linux",
```
coutoPL: # 1. 
> download a green it_linux",
>    "leg's '*-results' artifact and copy the root <testsuite time=...> attribute of each"

No, it's not maintainable. Manual run it ok, but the drift should be auto-detected.
And maybe there should be optional job in the pipeline that will warn us where the drift is dectected, so someone could regenerate it.

#2.

I don't see how it could be regenerated. One liner, or sth like that needed.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteSharder.java
```diff
@@ -0,0 +1,110 @@
...
+   * Returns the subset of {@code allSuites} that shard {@code shardIndex} of {@code shardCount}
+   * must run.
+   *
+   * @param allSuites fully-qualified class names of every discovered suite
+   * @param shardCount total number of shards (K)
+   * @param shardIndex this shard's index, {@code 0 <= shardIndex < shardCount}
+   * @param timings measured suite wall times in seconds; empty map selects hash packing
+   */
+  public static List<String> suitesFor(
+      List<String> allSuites, int shardCount, int shardIndex, Map<String, Long> timings) {
```
coutoPL, line 66: IMO, suite-timings.json file parsing can be a part of this class. No need to keep it at the gradle's level. It's rather related to the SuiteSharder's logic

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/RemoteClusterAuditingToolsSuite.scala
```diff
@@ -166,24 +166,74 @@ class RemoteClusterAuditingToolsSuite
...
+  // Proves the audit pipeline recovered after a network re-enable: a probe event sent while the
+  // sink client still holds every node in dead-host backoff is dropped outright, so each attempt
+  // sends a FRESH probe request and briefly polls for its entry. Probe entries are throwaway —
+  // the caller truncates the sinks right after.
+  private def waitUntilAuditSinkRecovers(): Unit = {
+    val deadline = System.currentTimeMillis() + 180 * 1000L
+    val perProbeWaitMillis = 10 * 1000L
+
+    @scala.annotation.tailrec
+    def probeLanded(traceId: String, probeDeadline: Long): Boolean = {
```
coutoPL: Instead of using hard-to-understand metaphors, we could be stricter in naming. E.g. "entryVisibleInAllSinks"

ci/github/PORT.md
coutoPL: I doubt that in this form the document is needed. 
First of all - it's poorly named if it's going to describe how the current pipeline works.
If it's going to describe the process of the transition from azure pipeline to GH actions, it'd be better to place its content in the PR's description. 

From the perpective of the future reader, I would argue that we need to bother him about past trasction from one CI to another. The future reader would need to read about the architecture of the current state.

azure-pipelines.yml
coutoPL, line 1: why we didn't port all the things from this pipeline?

.github/workflows/ci.yml
coutoPL, line 1: 1. In the GH action graph view I see Matrix "it_linux" and entries: "it_linux (integration_ex...)" or in the "required_checks": "required checks(aud...", "required checks (co...)". The information that distinguishes the steps is truncated. It should be improved for the sake of human readability.

2. CVE stage failed with:

```
* What went wrong:
Execution failed for task ':audit:dependencyCheckAnalyze'.
> Analysis failed.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (powered by Develocity).
> Get more help at https://help.gradle.org./

BUILD FAILED in 1m 50s
```

3. It looks like e.g. "integration_es810x" step doesn't contain full logs. I would like to be able to browse them there.

4. all manual actions should be ported

coutoPL (changes requested):

ci/github/PORT.md
sscarduzio: Agreed and done in dfb3851: PORT.md is gone. `ci/README.md` now describes the pipeline **as it is** — jobs, triggers, the sharding/gate architecture, secrets, and the measured tuning limits — with zero transition history. The Azure→GH mapping moved to this PR's description (collapsed section), where the transition story belongs.

integration-tests/suite-timings.json
sscarduzio: Both points addressed in dfb3851:
1. **Auto-detection**: the es94x leg now runs `integration-tests:regenerateSuiteTimings` after the tests — it emits `::warning::` annotations for drifted/new suites and uploads the regenerated file as the `suite-timings-regenerated` artifact.
2. **Regeneration one-liner**: download that artifact and commit it. Locally: `./gradlew integration-tests:test -PesModule=es94x && ./gradlew integration-tests:regenerateSuiteTimings` (writes `build/suite-timings-regenerated.json`). The file's `_readme` documents exactly this.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteSharder.java
sscarduzio: Done in dfb3851 — `SuiteSharder.timingsFrom(File)` owns the JSON parsing (unit-tested incl. the file round-trip); build.gradle only hands it the `File`.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/audit/RemoteClusterAuditingToolsSuite.scala
sscarduzio: Renamed in dfb3851: `probeLanded` → `entryVisibleInAllSinks`.

azure-pipelines.yml
sscarduzio: Everything is ported — as of dfb3851 including the last manual action:
- all 14 stages have GH counterparts (mapping table now in the PR description),
- all 5 manual actions work via `workflow_dispatch` (`run_disk_probe` was the missing one, added in dfb3851),
- the only things without a literal counterpart are SUPERSEDE_GUARD (replaced by native `concurrency` with the same never-cancel-releases semantics) and Azure's docker pre-clean/reap steps (pointless on ephemeral per-job VMs).
The azure-pipelines.yml body is intentionally left intact so it stays manually runnable as a fallback until we delete it after a release cycle.

.github/workflows/ci.yml
sscarduzio: All four addressed in dfb3851:
1. **Truncated names**: matrix jobs are now named by their distinguishing value alone — `integration_es94x`, `win_es94x`, `audit_build_check`, `build_es9xx` — instead of the truncated `it_linux (integration_es…)` composite.
2. **CVE failure**: root cause is the missing `NVD_API_KEY` repo secret (NVD throttles keyless clients). The step now emits an explicit `::warning::` when the key is absent; the job stays `continue-on-error` so it never blocks. Green once the secret is set.
3. **Browsable logs**: shard stdout is still captured per-file (live interleaving of 4 JVMs would be unreadable), but each leg now prints every `shard-N.log` to the job console as collapsible groups after the run — browsable in the UI, and still uploaded as artifacts.
4. **Manual actions**: all 5 are now ported (`run_disk_probe` was the gap).

ci/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Jobs
+
+| Job | What it does | When |
+|---|---|---|
+| `setup` | computes branch flags + the IT matrices | always |
+| `toolchains_verify` | sanity-checks the toolchains image | always (fail-fast gate for tests) |
+| `required_checks` | audit build, cross-Scala compile, format, license | pushes + PRs |
+| `unit_tests` | `core:test` and friends | pushes + PRs |
+| `cve_check` | OWASP dependency-check (`continue-on-error`; needs `NVD_API_KEY`) | pushes + PRs |
```
coutoPL: CVE check in the old pipeline was in the Optional Checks. It's optional - when it's failing, there is a warning, but the whole stage passes. Can we move cve_check as a task inside the Optional Checks matrix? Do you confirm it works the same as I described?

ci/README.md
```diff
@@ -0,0 +1,84 @@
...
+directory contain the build logic; the workflow only orchestrates.
+
+## Jobs
+
+| Job | What it does | When |
+|---|---|---|
+| `setup` | computes branch flags + the IT matrices | always |
+| `toolchains_verify` | sanity-checks the toolchains image | always (fail-fast gate for tests) |
+| `required_checks` | audit build, cross-Scala compile, format, license | pushes + PRs |
+| `unit_tests` | `core:test` and friends | pushes + PRs |
```
coutoPL: Maybe `unit_tests_linux` (to be symmetric to `unit_tests_windows`)?

integration-tests/suite-timings.json
```diff
@@ -0,0 +1,77 @@
+{
+  "_readme": [
+    "Measured wall time (seconds) per integration suite. Used by build-base's SuiteSharder for",
+    "duration-balanced shard packing when ROR_BALANCED_SHARDS=true (see integration-tests/build.gradle).",
+    "ADVISORY ONLY: suites missing here default to 60s and still run; entries for deleted suites",
+    "are ignored. Staleness affects shard balance, never correctness or coverage.",
+    "DRIFT IS AUTO-DETECTED: the es94x CI leg runs integration-tests:regenerateSuiteTimings after",
```
coutoPL: Maybe we should use the `es90x` instead (the first ES 9 module)?

azure-pipelines.yml
coutoPL: Does it mean we can remove .azure-pipeline.yml and all other azure-pipeline-related files?

.github/workflows/ci.yml
coutoPL: Ad3. Each log should have a shard ID or name attached to it. Now I see sth like this:

```
2026-07-14T14:14:41.8153976Z     13:59:50.332 [-159017831] INFO  t.b.r.u.c.EsContainerWithRorAndXpackSecurity - STDOUT: {"@timestamp":"2026-07-14T13:59:50.329Z", "log.level": "INFO", "message":"[1921673982-2122610997] ROR main engine (id=c960f5dc9de0540967dcd136bb6479363125acf3) stopped!", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"scala-execution-context-global-91","log.logger":"tech.beshu.ror.boot.engines.MainSettingsBasedReloadableEngine","elasticsearch.cluster.uuid":"yigoAi68SpW0aTMHuypV0w","elasticsearch.node.id":"3IouEEy3Q7CQc4ilBhy2bw","elasticsearch.node.name":"ROR_SINGLE_1","elasticsearch.cluster.name":"ROR_SINGLE"}
```

Is `-159017831` the shard ID? Can we make it human-readable?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,737 @@
+# ReadonlyREST CI — ported from azure-pipelines.yml to GitHub Actions.
```
coutoPL: I don't think the comment is needed. We have ci/README.md

azure-pipelines.yml
sscarduzio: Everything IS ported — nothing was dropped. 2af180ea adds a stage-by-stage coverage table to ci/README.md ("Coverage: what happened to every Azure stage") so this can be verified at a glance: SUPERSEDE_GUARD → native concurrency, OPTIONAL_CHECKS → optional_checks, RELEASE_*_WITHOUT_TESTING → the release_without_testing dispatch input, secure-file secret.pgp → PGP_SECRET_KEY_B64, etc. This file keeps trigger:none/pr:none as a manually-runnable fallback for one release cycle after the switch, then it gets deleted.

.github/workflows/ci.yml
sscarduzio: 1. Fixed on the current branch: it_linux/it_windows/required_checks (and now optional_checks) set name: to the matrix value alone, so the graph shows "integration_es94x" / "win_es94x" / "compile_codebase_check" instead of the truncated "it_linux (integration_ex...)" composites. 2. The CVE failure was the missing NVD_API_KEY secret in this repo — the job is continue-on-error so it annotates but never blocks (details in the thread on ci/README.md).

ci/README.md
sscarduzio: Confirmed — continue-on-error on the job gives exactly the Azure Optional Checks semantics: a cve_check failure shows a red annotation on that job but the workflow (and the PR gate) passes. And done in 2af180ea: cve_check now lives inside an optional_checks matrix job mirroring required_checks (matrix has one member today; new optional checks just join the list).

ci/README.md
sscarduzio: Done in 2af180ea — renamed unit_tests_linux (all needs/result references updated).

integration-tests/suite-timings.json
sscarduzio: Agreed — done in 2af180ea. The drift-check leg, the _readme in suite-timings.json, and the build.gradle/README references now point at es90x (first ES 9 module — a reference that never moves when a newer es9Xx module lands). Note: the committed timings were measured on es94x, so the first es90x run may emit a one-time drift warning; committing its regenerated artifact re-baselines it.

.github/workflows/ci.yml
sscarduzio: Removed in 2af180ea — ci.yml now starts at name: CI; ci/README.md owns the architecture notes.

tests-utils/src/main/scala/tech/beshu/ror/utils/gradle/RorPluginGradleProject.scala
```diff
@@ -111,6 +125,28 @@ class RorPluginGradleProject(val moduleName: String) extends LazyLogging {
     s"${rootProjectProperties.getProperty("pluginName")}-${rootProjectProperties.getProperty("pluginVersion")}_es$getModuleESVersion.zip"
 
   private def runTask(task: String, extraArgs: List[String] = Nil): Unit = {
+    // Cross-process mutex: K sharded test JVMs (IT_PARALLELISM > 1) share this workspace, and their
+    // nested Tooling-API builds (prebuild + runtime container assembles) race Gradle's own project
+    // locks to death. Serialize every nested build across processes; an up-to-date nested build holds
+    // the lock only briefly. ponytail: one global lock — per-module locks if contention ever matters.
+    val lockFile = new JFile(RorPluginGradleProject.getRootProject, ".ror-nested-gradle.lock")
```
coutoPL, line 133: Why didn't you use the FileLockSemaphore? If it does not fit well; maybe it can be adapted and reused!

sscarduzio: Good timing — the inline lock block you commented on is already gone: 6dc61028 extracted it (plus the twin in WindowsEsDirectoryManager) into a shared `FileLocks.withExclusiveLock` next to FileLockSemaphore.

On why it is not FileLockSemaphore(permits = 1): the two shapes differ at the OS level. The semaphore's N-slot design *forces polling* — a process cannot block on "any of N files" at once, so waiters retry `tryLock` over the slots with a 3s sleep. A single named mutex can use the strictly better blocking `channel.lock()`: the kernel parks the waiter and wakes it the instant the holder releases, zero poll latency. This lock brackets every nested gradle build (taken many times per suite run), so up-to-3s of added latency per contended acquire from a 1-permit semaphore would be real waste.

So: FileLockSemaphore = N interchangeable slots (polling, inherent), FileLocks = one named mutex (blocking). They now live side by side in `misc` and 72437c17 adds cross-referencing scaladocs stating exactly this, so the next reader picks the right one. Happy to fold them into one file/object if you prefer that shape.

coutoPL (approved):

PR #1298, opened 2026-07-14
[RORDEV-2074] Fix repackaging building plugins from the wrong base Elasticsearch version

coutoPL:
## Problem                                                                                                                                     
An integration-test build failed with Elasticsearch refusing to start (container exited with code 1, first seen on the 7.4–7.6 build). The plugin installed into it had been repackaged from the wrong base version, so it didn't match the Elasticsearch version it was meant for and crashed on startup.

## Cause                                                                                                                                       
When a plugin for one Elasticsearch version is produced by repackaging another, the build always used the module's oldest supported version as the base. For modules covering a wide range, that built later versions from a base that no longer matched them.

## What this changes                                                                                                                           
- A module can now declare the base version each part of its range should be built from, so repackaging always uses a correct base close to the target instead of the oldest one. Applied to the ranges where this was wrong: 7.4–7.6, 7.11–7.13, 8.5–8.6, 9.2–9.3.                     
- Added a build-time safety check that verifies a repackaged plugin matches a fresh build for that version and fails the build with a clear error if it doesn't — so a wrong-base repackage can't silently reach CI or a release again.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bytecode compatibility verification by comparing normalized class behavior (with `.tasty` ignored) and falling back safely when needed.
  - Reduced stale/unsafe reuse by isolating per-version verification flows and stashing intermediate artifacts.
- **New Features**
  - Added configurable base Elasticsearch version sets for 7.4–7.6, 7.11–7.13, 8.5–8.6, and 9.2–9.3 to drive compatibility grouping.
- **Tests**
  - Expanded unit coverage for ES base/version selection and bytecode comparison scenarios.
- **Chores**
  - Refreshed Gradle dependency declarations and updated verification task wiring/order.


coutoPL:
@claude review please

sscarduzio (approved):
## Review

Good fix — the root cause (always using the oldest supported version as the repackaging base) is correctly addressed, and the build-time safety check ensures it can't regress silently.

**What I verified:**

1. **`EsVersions.java`** — `baseFor()` correctly picks the greatest declared base ≤ the requested version. Validation is thorough: all bases must be supported versions, must be oldest-first, and must start at the oldest supported version (so every version has a base at or below it). `groupNewest()` correctly identifies the farthest-from-base version in each sub-range for verification.

2. **`BytecodeComparison.java`** — The rewrite from raw SHA-256 to ASM textify (lightbend-labs/jardiff recipe) is a real improvement. It normalizes member ordering, strips Scala 3 TASTY attributes and pickle signatures, and catches actual bytecode differences (`invokevirtual` vs `invokeinterface` — the es74x `BytesReference` class→interface case) while ignoring irrelevant metadata. The fallback to raw bytes on parse failure is conservative (over-strict, never misses an unsafe repackage) — correct design choice.

3. **`readonlyrest.verify-repackage-bytecode.gradle`** — The `clean` + `--no-build-cache` + jar stashing pattern is well-reasoned. The comments explain exactly why each piece is load-bearing (Zinc incremental compiler reusing stale .class files, build cache poisoning, stashing outside build/ to survive the target's `clean`). Windows `gradlew.bat` handling is a nice addition.

4. **`RorPluginGradleProject.scala`** — Memoization via `assembledByModuleAndVersion` addresses the coderabbitai performance concern. The `synchronized` block is correct for the `mutable.Map` cache. Key space is bounded by (modules × versions) so unbounded growth isn't a concern.

5. **`gradle.properties` per-module overrides** — The `baseEsVersions` declarations for es74x, es711x, es85x, es92x are well-documented with the specific ES API changes that necessitate each split.

6. **Test coverage** — Comprehensive. BytecodeComparison tests cover .tasty ignoring, method asymmetry, different method bodies, invokevirtual vs invokeinterface, and member ordering normalization. EsVersions tests cover baseFor, groupNewest, baseline with group bases, and all validation error paths.

**Minor note:** IT tests are still running (14 jobs in progress), but since this PR only touches build infrastructure (not plugin code), the IT suite is unlikely to be affected. Will monitor for completion.

LGTM — approving.

PR #1301, opened 2026-07-21
[ci] Fix suite-timings drift check: wall time instead of junit exec time

sscarduzio:
The drift checker added in #1292 read the junit `time` attribute, which measures test execution only — but packing weights must include container boots (often the dominant part). Result: every run emitted ~40 bogus drift warnings (`CrossClusterCallsSuite: 338s -> 5s`), and committing the regenerated artifact would have wrecked balanced packing.

Fix: suites run serially within a shard, so `start(i+1) - start(i) = exec(i) + boot(i+1)`, giving `wall(i) = boot(i) + exec(i)` (first suite in a shard gets the median boot). Verified against the latest develop run's `*-results` artifact: regenerated walls now track the committed magnitudes, and drift warnings drop to 3 — all real speedups from the ES module trims.

Second commit-half: `suite-timings.json` refreshed from that artifact with the fixed formula (the designed update flow), so packing reflects post-trim reality.

https://claude.ai/code/session_01J7ZwM6WNAkeHKqSjxnqgPa

## Summary by CodeRabbit

- **Build / Chores**
  - Improved integration test timing estimation to account for suite startup overhead, yielding more accurate shard and rerun calculations.
  - Refreshed the stored suite timing results with the updated estimation method.
  - Disabled pre-commit hook setup when running in CI environments.
- **Tests**
  - Added comprehensive unit tests for wall-time estimation, including boot-time calculation, ordering, shard/rerun behavior, edge cases, and minimum/rounding handling.


integration-tests/build.gradle
```diff
@@ -201,13 +201,35 @@ tasks.register('prebuildEsImage', JavaExec) {
 //   ./gradlew integration-tests:test -PesModule=es90x && ./gradlew integration-tests:regenerateSuiteTimings
 tasks.register('regenerateSuiteTimings') {
     doLast {
-        def times = new TreeMap()
+        // WALL time per suite, not the junit `time` attribute: `time` is test execution only,
```
coutoPL: I see the logic is getting complicated. Can we move it to Java and test as we did in other places?

coutoPL (changes requested):

sscarduzio:
@coutoPL Done — the wall-time estimation moved to Java: `SuiteTimings.wallTimes` in build-base (next to `driftReport`, same pure-function split as `SuiteSharder`), with 8 unit tests covering boot-gap math, median fallback, cross-shard max, negative-gap clamp, and sub-second suites. The Gradle task now only parses the junit XMLs and does the JSON I/O. Also folded in coderabbit's precision point: gaps are computed in milliseconds as doubles (no truncation), rounded once at the end, and clamped to a minimum of 1s so no suite packs with zero weight. Rebased onto develop.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteTimings.java
```diff
@@ -38,8 +42,48 @@ public final class SuiteTimings {
...
+      List<Double> boots = new ArrayList<>();
+      for (int i = 1; i < suites.size(); i++) {
+        double gap =
+            Duration.between(suites.get(i - 1).start(), suites.get(i).start()).toMillis() / 1000.0;
+        boots.add(Math.max(0.0, gap - suites.get(i - 1).execSeconds()));
+      }
+      double medianBoot =
+          boots.isEmpty()
+              ? DEFAULT_BOOT_SECONDS
+              : boots.stream().sorted().toList().get(boots.size() / 2);
```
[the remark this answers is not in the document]
sscarduzio: Fixed in f810227ee: `median()` now averages the two middle values for even-sized lists, with a dedicated test (`evenNumberOfBootGapsUsesTrueMedian`, gaps 10+20 -> 15).

integration-tests/build.gradle
sscarduzio: Done in 0d0fbe63d: estimation logic moved to `SuiteTimings.wallTimes` in build-base (pure, 9 unit tests, same split as `SuiteSharder`); the task now only parses junit XMLs and writes JSON.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteTimings.java
```diff
@@ -38,8 +42,52 @@ public final class SuiteTimings {
...
+   * shards (or reruns) keeps its largest estimate; every value is clamped to at least 1s so no
+   * suite packs with zero weight.
+   */
+  public static Map<String, Long> wallTimes(Collection<List<SuiteRun>> shards) {
+    Map<String, Long> times = new TreeMap<>();
+    for (List<SuiteRun> shard : shards) {
+      List<SuiteRun> suites = new ArrayList<>(shard);
+      suites.sort(Comparator.comparing(SuiteRun::start));
+      List<Double> boots = new ArrayList<>();
+      for (int i = 1; i < suites.size(); i++) {
```
coutoPL: we could use java streams here instead (and in other places in this file too)

coutoPL (approved):
one minor comment

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteTimings.java
sscarduzio: Done in 8f634e57b: both indexed loops in `wallTimes` are now `IntStream.range(...)` pipelines, and `driftReport` builds its list via `entrySet().stream().sorted().flatMap(...)` (mutable accumulator gone, drift predicate extracted to `driftLine`). All 16 unit tests pass unchanged.

PR #1302, opened 2026-07-21
[ci] Delete the Azure pipeline

sscarduzio:
The "wait one release cycle" caution turned out unnecessary: the GitHub Actions pipeline is fully proven — full 34-leg matrix green twice on develop, and the pre-release path (`upload_pre_ror`, all 4 legs, real S3 credentials) already ran green on both develop merges. The only never-run pieces (`release_ror` tag push, `publish_mvn`) execute the same scripts Azure ran; only the reviewed orchestration is new. A dormant unverified Azure pipeline is a false safety net, so out it goes:

- `azure-pipelines.yml`, `ci/azure-templates/*`, `ci/stale-azure-pipeline-runs-canceler.sh` deleted
- `ci/docker-hub-auth.sh` simplified to the `::add-mask::` path only
- stale Azure comments in `ci/run-pipeline.sh` / `ci.yml` / `ci/README.md` updated

https://claude.ai/code/session_01J7ZwM6WNAkeHKqSjxnqgPa

## Summary by CodeRabbit

* **CI/CD**
  * Removed the legacy Azure Pipelines configuration and related templates.
  * Updated continuous integration documentation to reflect the migration to GitHub Actions.
  * Improved Docker registry authentication handling and credential masking in GitHub-based workflows.
  * Updated CI guidance for Maven publishing secrets and supported Linux module versions.
* **Maintenance**
  * Removed obsolete automation for cancelling stale Azure pipeline runs.


coutoPL (commented):
I'd wait with merging it until we release ROR 1.71.0 (the full release circe with GH Action pipeline). 

At the moment the Azure pipeline should be there and if needed it could be activated again. At the moment it should not be triggered on PR, push and any other way than manually. Do you confirm it works currently like that?

sscarduzio:
@coutoPL Confirmed: on develop, `azure-pipelines.yml` has `trigger: none` and `pr: none` — the Azure pipeline cannot be started by PRs, pushes, or schedules; only manually. Agreed on holding the merge until 1.71.0 has gone through the full GH-Actions release cycle; this PR just parks the deletion until then. (Rebased onto develop and applied the reviewer suggestion to mask the bare base64 credential in ci/docker-hub-auth.sh.)

coutoPL (approved):
LGTM.

But DO NOT merge it. We will merge it after releasing ROR 1.71.0

sscarduzio:
Rebased onto develop — it had gone conflicting, because develop rewrote the same CI files (the matrix variables in `ci.yml`, the docker-hub auth block).

Resolution: this PR's only `ci.yml` edit was deleting the comment "keep in sync with azure-pipelines.yml until Azure is deleted". Develop already rewrote that block and dropped the comment, so `ci.yml` now has **zero diff** against develop. The PR is again only what its title says: delete `azure-pipelines.yml`, the 4 azure templates, the stale-run canceler, and the 3 lines of prose that refer to them.

Verified: no file in the tree still references `azure-pipelines.yml`, `azure-templates/` or the canceler (the mapping table in `ci/README.md` is kept on purpose as history); `bash -n` and YAML parse are clean.

Still parked until 1.71.0 ships, per Simone.

sscarduzio:
Closing this for now — a fresh one will be raised after the release.

Context for whoever picks it up, so the reasoning is not lost:

This PR went `CONFLICTING` because the CI secret consolidation renamed every secret inside `azure-pipelines.yml` and the four templates under `ci/azure-templates/`, at review request. Two approved changes ended up pointing opposite ways at the same files.

**The state to be aware of when this is reopened:** the Azure DevOps variable group still holds the *old* secret names — `ROR_ARTIFACTS_STORE_*`, `ROR_LIBS_STORE_*`, `DOCKER_REGISTRY_*`, `ROR_ACTIVATION_KEY` and the rest. Doppler cannot write to Azure DevOps, so nothing renamed them and nothing will. The pipeline YAML now asks for `ROR_S3_*`, `DOCKER_HUB_RW_TOKEN`, `ROR_ENT_ACTIVATION_TOKEN` and `ROR_GH_TOKEN`, which that variable group does not define.

Azure expands an undefined non-secret variable to its own literal text, so a run today would not fail cleanly — every value would arrive as the eight-or-so character string `$(NAME)`. Only `ci/docker-hub-auth.sh` defends against that shape; everything else would take it as a real value and fail somewhere further down.

So the pipeline is not a working fallback in its current state. Whichever way this goes after the release:

- **Delete it** — no further work, and the renames become moot rather than wrong.
- **Keep it** — someone has to rename the entries in the Azure DevOps variable group to match, and then own keeping that group in step with the Doppler project `ror_ci` by hand, since there is no sync between them.

`ci/CI.md` says the pipeline lives one release cycle after the switch to GitHub Actions and is then deleted. Worth checking whether that cycle is over when this is revisited.

PR #1305, opened 2026-07-25
[RORDEV-2122][RORDEV-2143] Publish POMs alongside the ES jars in the libs store

mgoworko:
## Problem                                                                                                                                                                     
                                                                                                                                                                                 
  When Elastic releases an ES version before it lands on Maven Central, we mirror the jars                                                                                       
  ROR needs to the ROR libs store. The store served bare jars with no metadata, so Gradle                                                                                        
  resolved each one as a dependency-less leaf and none of ES's own transitive jars came                                                                                          
  along. `buildRorPlugin -PesVersion=9.5.0` failed at `jarHellCheck`:                                                                                                            
                                                                                                                                                                                 
      Could not find or load main class org.elasticsearch.jdk.JarHell  
      
 ## Fix                                                                                                                                                                         
                                                                                                                                                                                 
ror-tools:uploadArtifactsFromEsBinaries now publishes a POM next to each mirrored jar.                                                                                         
  The POM is built from the distribution itself: the artifact's classpath directory —                                                                                            
  lib for the server, its own module directory for a module — becomes its dependency                                                                                             
  list, at the versions the jar file names carry. Group ids are the one thing a jar                                                                                              
  cannot supply, so they are read from the release's verification-metadata.xml for                                                                                               
  third-party jars, and from Central for Elastic's own, whose groups move between                                                                                                
  releases. The libs store repository declares mavenPom() so Gradle reads the result.                                                                                            
                                                                                                                                                                                 
  Nothing changes for versions Central already has — Central stays declared first, and the                                                                                       
  store is only consulted for what Central cannot answer. No es*x module needs to declare                                                                                        
  anything extra.   

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added support for Elasticsearch 9.5.0.
  * Improved handling of Elasticsearch distributions, bundled libraries, Maven metadata, and mirrored artifacts.
  * Added clearer validation for Elasticsearch versions and compatibility settings.

* **Bug Fixes**
  * Improved artifact selection and dependency resolution.
  * Added safer repository metadata and POM processing.
  * Standardized artifact uploads and distribution handling across supported versions.

* **Tests**
  * Expanded coverage for distribution scanning, version validation, artifact mirroring, Maven processing, and repository errors.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


ror-tools/build.gradle
```diff
@@ -148,9 +157,12 @@ def registerLibsStoreUpload = { String taskName, String artifactName, String des
 
 registerLibsStoreUpload('uploadNetty4TransportFromEsBinaries', netty4TransportArtifactName, netty4TransportArtifactDestinationPath)
 registerLibsStoreUpload('uploadElasticsearchFromEsBinaries', elasticsearchArtifactName, elasticsearchArtifactDestinationPath)
+registerLibsStoreUpload('uploadElasticsearchRestClientFromEsBinaries', elasticsearchRestClientArtifactName, elasticsearchRestClientArtifactDestinationPath)
```
mgoworko: I tested it extensively on the evening, when ES 9.4.4, 9.3.8 and 8.19.19 were already available, but artifacts were not yet available in Maven Central. Both locally, and on Azure pipeline (on master branch) there were missing classes on classpath from those 3 modules. They were not pulled transitively from the `elasticsearch` dependency, because that dependency was pulled as a raw jar file from our S3.

Based on my tests, building ROR before ES artifacts are published works when:
- 5 jars from ES package are pushed to S3, not only netty4transport and main elasticsearch artifact
- we declare those dependencies explicitly
- we add also some lucene and org.apache.httpcomponents dependencies explicitly

This change allowed me to build and test ROR before any artifacts were published on Maven. Ultimately I did not push it to master when adding support for new ES version, because they were published to Maven the next day.

coutoPL: Could you please show me the logs from it? I haven't seen anything like this yet.

mgoworko: https://dev.azure.com/beshu-tech/ReadonlyREST%20for%20Elasticsearch/_build/results?buildId=10824&view=logs&j=a8f9bdaf-d411-50b3-cc2e-84fc3aff72f8&t=adcdd46c-b2f1-540d-f62c-bc6010a2fe9d

```
* What went wrong:
Execution failed for task ':es818x:compileScala'.
> Could not resolve all files for configuration ':es818x:compileClasspath'.
   > Could not find org.elasticsearch.client:elasticsearch-rest-client:8.19.19.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/elasticsearch/client/elasticsearch-rest-client/8.19.19/elasticsearch-rest-client-8.19.19.pom
       - file:/home/vsts_azpcontainer/.m2/repository/org/elasticsearch/client/elasticsearch-rest-client/8.19.19/elasticsearch-rest-client-8.19.19.pom
       - https://dgp.serve.beshu.tech/beshu/ror/libs/org/elasticsearch/client/elasticsearch-rest-client/8.19.19/elasticsearch-rest-client-8.19.19.jar
     Required by:
         project ':es818x'
```

- That log was after uploading the main `elasticsearch` jar to S3. The `elasticsearch`  dependency was correctly resolved from S3, but there was missing `elasticsearch-rest-client`. 
- After I added uploading `elasticsearch-rest-client` to S3, then there was no longer missing dependency error, but there were classes missing on classpath - from `elasticsearch-x-content` and `elasticsearch-core` modules.
- so I added those too to S3
- I also added explicit dependencies - those that are transitive dependencies of `elasticsearch` module, but were not pulled when the `elasticsearch` jar was itself pulled form our S3

Have you ever tested it in a situation, when ES artifacts were still completely missing from Maven Central? I tested it on a day, when new ES versions were released, and there were no jars available in Maven Central.

mgoworko: I can't access logs for some commits with incremental changes - adding those jars and dependencies one by one :( Azure CI returns errors and does not show them on a list of runs.

mgoworko: I think that we may pause it for now, and test it again when the new ES versions are released.

coutoPL: yeah, this is a good idea. Let's get back to this as soon as new es versions are released

coutoPL: I've checked it today, and I think I found a way to make it without changing build definitions. Will show you tomorrow

coutoPL: @mgoworko please take a look. The solution is bigger, but we improve only the upload task. No need to change the gradle modules setup.

coutoPL:
@claude review please

build-base/src/main/java/tech/beshu/ror/gradle/utils/EsDistribution.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * The ES distribution the build works with: where Elastic publishes it, where the build unpacks it, and -- once
+ * scanned -- an index of the jars it ships, taken from {@code lib} and every {@code modules/*} directory and
+ * keyed by artifact id.
+ *
+ * <p>A library can be shipped by several modules at different versions: ES 9.5.0 ships commons-codec 1.15 in
```
mgoworko, line 53: ES 9.5.0 - specific comment, perhaps to remove

build-base/src/test/java/tech/beshu/ror/gradle/utils/EsDistributionTest.java
```diff
@@ -0,0 +1,335 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+
+class EsDistributionTest {
+
+  private static final String ARCHIVE_NAME = "elasticsearch-9.5.0-linux-x86_64.tar.gz";
```
mgoworko, line 42: All the tests are ES 9.5.0-specific. I think that if that fixed version stays, then the version should be extracted to some global variable, so that it can be updated without refactoring multiple files.

ror-tools/build.gradle
```diff
@@ -123,75 +123,106 @@ tasks.withType(AbstractArchiveTask).configureEach {
...
+        'ROR_LIBS_STORE_PATH_PREFIX'      : System.getenv('ROR_LIBS_STORE_PATH_PREFIX') ?: 'ror/libs',
+]
+
+def targetEsVersion = esVersion.toString()
+def buildDirPath = { layout.buildDirectory.get().asFile.toPath() }
+def fileInBuildDir = { String name -> buildDirPath().resolve(name).toFile() }
+
+def mirrorPlanCache = [:]
+
+def mirrorPlan = {
```
mgoworko, line 156: It depends on the last (previous) ES version. It is a bit risky/fragile in my opinion. This new mechanism is great, but it would break the moment ES changes its dependencies or just adds one new jar. 

I would suggest adding a handle for uploading any jar by name. For example script that uploads any jar by name. I suggest that, because this change removed the very simple mechanism that existed before, where each jar was listed separately and it was easy to add a new one. Not it _should_  work automatically and better, but at the same time it is harder to _manually_ fix the missing jar if needed

ci/upload-es-artifacts.sh
```diff
@@ -4,6 +4,4 @@ set -xe
 
 echo ">>> ($0) UPLOADING ES ARTIFACTS ..."
 
-#./gradlew --stacktrace --info clean ror-tools:uploadArtifactsFromEsBinaries -PesVersion=9.4.4
-#./gradlew --stacktrace --info clean ror-tools:uploadArtifactsFromEsBinaries -PesVersion=9.3.8
-#./gradlew --stacktrace --info clean ror-tools:uploadArtifactsFromEsBinaries -PesVersion=8.19.19
+./gradlew --stacktrace --info clean ror-tools:uploadArtifactsFromEsBinaries -PesVersion=9.5.0
```
mgoworko, line 7: Using this opportunity I would suggest re-uploading 8.8.0 jars. They are currently missing in our s3 (it works if someone has them cached, otherwise it is impossible to import es88x module).

build-base/src/test/java/tech/beshu/ror/gradle/utils/EsDistributionTest.java
coutoPL: Because we use 9.5.0 just as an exmple here. We don't need to change it in the future.

ror-tools/build.gradle
coutoPL: The solution was improved, taking into consideration your comment

coutoPL:
@mgoworko please take a look

mgoworko (commented):
LGTM

PR #1309, opened 2026-08-05
[RORDEV-2146] Security fix: bump BouncyCastle FIPS, suppress netty HTTP/3 CVE

coutoPL:
## BouncyCastle FIPS — real, fixed by upgrade                                                                                                                        
                                                                                                                                                                       
  | artifact | from | to |                                                                                                                                             
  |---|---|---|                                                                                                                                                        
  | `bc-fips` | 2.1.2 | 2.1.3 |                                                                                                                                        
  | `bcpkix-fips` | 2.1.11 | 2.1.12 |                                                                                                                                  
  | `bctls-fips` | 2.1.23 | 2.1.24 |                                                                                                                                   
  | `bcutil-fips` | 2.1.6 | 2.1.7 |                                                                                                                                    
                                                                                                                                                                       
  * `bc-fips` 2.1.3 — CVE-2026-8763, CVE-2026-13506, CVE-2026-14682, CVE-2026-58059, CVE-2026-58060, CVE-2026-13586, CVE-2026-58063                                    
  * `bcpkix-fips` 2.1.12 — CVE-2026-12802, CVE-2026-59639                                                                                                              
                                                                                                                                                                       
  `bctls-fips`/`bcutil-fips` bumped because the new `bcpkix-fips`/`bctls-fips` need `bcutil-fips:[2.1.7,2.2.0)`, and CVE-2026-14682 also names `bctls-fips`.           
                                                                                                                                                                       
  ## Suppressed false positives                                                                                                                                                       
                                                                                                                                                                       
  * **Netty CVE-2026-56816** — affects `netty-codec-http3`, which only exists in Netty 4.2.x. We're on 4.1.136 (newest 4.1.x, the line ES bundles). NVD's CPE has no   
  lower bound, so it flags every 4.1.x jar.                                                                                                                            
  * **CVE-2026-63144, CVE-2026-63263** — flagged on `elasticsearch-rest-client`, but both are bugs in the ES server, which the client doesn't depend on. Fixed by      
  Elastic in ES 8.19.19 / 9.3.8 / 9.4.4.                                                                                                                               
                                                                                                                                                                       
  ## bc-fips 2.1.3 needs a new permission on older ES                                                                                                                  
                                                                                                                                                                       
  2.1.3 wipes DRBG buffers with a `Cleaner` instead of `finalize()`. `Cleaner.create()` starts a thread, which ES's `SecureSM` blocks without                          
  `ThreadPermission("modifyArbitraryThread")`. Without it the node won't start when `fips_mode` is on.                                                                 
                                                                                                                                                                       
  Two ES changes landed in different versions, so there are three cases:                                                                                               
                                                                                                                                                                       
  | ES | needs it? | allowed in `plugin-security.policy`? | how granted |                                                                                              
  |---|---|---|---|                                                                                                                                                    
  | ≤ 7.10 | yes | yes | policy file (9 modules) |                                                                                                                     
  | 7.11 – 7.17.7 | yes | **no** (`PolicyUtil` allowlist since 7.11) | `ror-tools patch` (es711x/714x/716x) |                                                          
  | ≥ 7.17.8 | no (`SecureSM` carve-out) | no | nothing |                                                                                                              
                                                                                                                                                                       
  In the middle case the permission is needed at runtime but rejected at install time, so it can only be added after install — the same patcher mechanism already used 
  for `getProperty.*`.                                                                                                                                                 
                                                                                                                                                                       
  `ModifyPolicyUtilClass` also had to extend `classPermissions`, not just `namedPermissions`: the policy parser can't resolve ES-owned permission classes, so they     
  arrive as `UnresolvedPermission`, which `namedPermissions` never matches.
  
  ## Two Windows patcher bugs (old bugs, exposed by this bump)                                                                                                         
                                                                                                                                                                       
  bc-fips 2.1.3 reads the security property `org.bouncycastle.asn1.max_cons_depth` when parsing PEM keys. ROR grants that via the patcher, but the grant never worked  
  on Windows, so PEM internode SSL nodes failed to start:                                                                                                              
                                                                                                                                                                       
  PEMException: problem parsing PRIVATE KEY: AccessControlException:                                                                                                   
  access denied ("java.security.SecurityPermission" "getProperty.org.bouncycastle.asn1.max_cons_depth")                                                                
                                                                                                                                                                       
  * `File.renameTo` fails on Windows if the target exists, and the result was ignored — the permission line was silently dropped. Now `Files.move(REPLACE_EXISTING)`.  
  * The ROR policy path was matched against `"/readonlyrest/plugin-security.policy"`; Windows uses `\`, so the ROR matcher never kicked in. Separators are now         
  normalised.                                                                                                                                                          
                                                                                                                                                                       
  Both predate this PR and were harmless until bc-fips started reading that property.


coutoPL:
@claude review please

mateuszkp96 (approved):
LGTM

PR #1307, opened 2026-08-01
[RORDEV-1636] Apply fields and filters rules for _terms_enum requests

mgoworko:
- type: security                                                                                                                                                                                                            
    components: [es]                                                                                                                                                                                                          
    text: "The _terms_enum API now applies the fields rule (restricting or obfuscating disallowed fields) and rejects requests guarded by a filter rule, since document-level security cannot be enforced on it."             

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **New Features**
  - Added access-controlled support for Elasticsearch `_terms_enum` requests across supported versions.
  - Applies index filtering and field-level security, including restrictions or field obfuscation where required.
  - Blocks requests when document-level security filters cannot be safely applied.

- **Tests**
  - Added unit and integration coverage for allowed, restricted, excluded, and obfuscated term fields.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #1313, opened 2026-08-09
[RORDEV-2151] De-duplicate the ES modules with one shared base per major

sscarduzio:
Jira: [RORDEV-2151](https://readonlyrest.atlassian.net/browse/RORDEV-2151)

_No changelog entry: no client-visible change. Every module compiles exactly the same sources as before._

Rebuilt on top of #1322, so this branch now contains **only the restructuring** — the drift fixes you asked to review separately are already on `develop`. One commit, no merge commits, no duplicated history.

## Why

The `es*` modules were whole-tree copies of each other: **5,507 source files, of which only ~534 were unique**. A single fix had to be hand-copied into up to 17 modules, and sometimes was missed — which is exactly what #1322 cleaned up.

## What

- **`es7-base`, `es8-base`, `es9-base`** hold the sources shared by every module of that major. `es67x` is the only ES 6 module, so it has no base.
- Each module compiles its major's base **plus** its own `src/main/scala`; a local file with the same relative path wins (an **override**). The base is derived from the module's ES major, so there is no per-module opt-in flag.
- **A file enters a base only when every module of that major has it** — otherwise the base would inject a file into a module that never had it.
- Completed majors take the **most common** shape of each file; the active major takes the **newest**, so a new module of that major needs the fewest overrides.
- One base per major rather than one global base (your review): a base that follows the newest ES version is never finished, because every new version pushes the previous shape down into every older module — including majors PR CI does not run. A completed major's base never changes again.
- The build stages the applicable base files into `build/es-base-src` with a `Sync` task. A direct source dir cannot work: `SourceDirectorySet` excludes apply to all source dirs at once, so excluding an overridden path would also drop the module's own copy.

**Folded in**, since it is the same "stop copying per module" idea:

- All 34 `es*/build.gradle` files become **one stub**. Every difference was a function of the module's ES version range — JDK 11/17/21, the JarHell class package move at 7.14, docker platforms, and the version-thresholded dependencies with their issue citations — so it is derived once in `plugin-common-conventions`.
- The 34 Dockerfiles had **7 distinct, deliberate contents**; they move to `/docker-image/dockerfiles`, and each module names its recipe with `rorDockerfile` in `gradle.properties`.

| major | before | after |
|---|---|---|
| es7x (12 modules) | 1,834 | **630** |
| es8x (17 modules) | 2,867 | **528** |
| es9x (4 modules) | 667 | **195** |
| es6x (1 module) | 139 | 139 |
| **total** | **5,507** | **1,492** |

Adding the next ES version is now: the fixed `build.gradle` stub + `gradle.properties` + `plugin-metadata/` + a handful of overrides — instead of a 178-file copy. A shared fix lands once per major instead of in up to 17 copies.

## Verification

- **Source-set equivalence**: every module tree was hashed before the migration; afterwards each module's effective set (base + own files, local wins) is **byte-identical to its pre-migration tree, for all 34 modules**. The compiler sees exactly the same input as before.
- `compileScala` **green for all 34 modules**.
- `es94x buildRorPluginZip` and `prepareDockerBuildContext` green, with the staged `Dockerfile` byte-identical to its recipe — so packaging, repackaging and the docker path are untouched.

## Docs

`CLAUDE.md` and the new `ES_MODULE_LAYOUT.md` cover the base rules, how to change a file, and how to add a new ES version or a new major.


coutoPL (changes requested):
The base idea is good. Sharing one copy of each file and staging it into the module                                                           
  before compilation is a clean mechanism, and it solves a real problem: today a single                                                         
  fix has to be copied by hand into 33 modules, and sometimes we miss some.                                                                     
                                                                                                                                                
  My concern is not the base itself. It is that there is only **one** base, and it always                                                       
  follows the newest ES version.                                                                                                                
                                                                                                                                                
  That means when we add a new ES version and a file changes shape, we put the new shape                                                        
  in the base and push the old shape down as overrides into the older modules. So adding                                                        
  one ES version writes into modules from ES 8.0 up to 9.4 — and we do not run most of                                                          
  those in PR tests. The effect is already visible in this PR: of 843 override files, 687                                                       
  are byte-identical copies. For example `ResolveIndexEsRequestContext` has 19 overrides                                                        
  with only 1 distinct content.                                                                                                                 
                                                                                                                                                
  **Proposal: one base per major version** — `es6-base`, `es7-base`, `es8-base`,                                                                
  `es9-base`. A module overrides only its own major's base.                                                                                     
                                                                                                                                                
  Why I think this is better:                                                                                                                   
                                                                                                                                                
  - Adding a new major (ES 10) = create `es10-base` from `es9-base`. Zero changes in old                                                        
    modules.                                                                                                                                    
  - Adding a new minor (ES 9.5) = only modules of that major can be touched. Today that is                                                      
    4, and 2 of them (es90x, es94x) run in PR integration tests.                                                                                
  - It is smaller: 697 files instead of 1077, for the same 21 modules.                                                                          
  - It also works for es7, which this PR leaves as full copies: 1829 files → 475.                                                               
  - A finished major never changes, so its base is frozen. A base that follows the newest                                                       
    version is never finished — it keeps reaching back into old modules forever.  

One detail: for a frozen major, pick the base as the most common shape, not the newest                                                        
  one. For es8, newest gives 849 files, most common gives 508. For the active major,                                                            
  newest is fine (es9: 191 vs 189).                                                                                                             
                                                                                                                                                
  I would prefer to re-cut the PR this way rather than merge and follow up later —                                                              
  otherwise we pay the migration twice.

sscarduzio:
You were right, and the effect is bigger than your numbers suggested. I re-cut the PR (`a9383f16c`) instead of merging and following up.

I verified your claim independently before changing anything: **798 of 981 override files (81%) were byte-identical copies of each other**, and your examples hold exactly — `ResolveIndexEsRequestContext` and `ForbiddenResponse` each had 19 overrides with 1 distinct content.

**Now: one base per major.** `es7-base`, `es8-base`, `es9-base`; `es67x` is the only ES 6 module so it keeps a full copy. Completed majors (6–8) take the most common shape, the active major (9) takes the newest — as you suggested. The base is derived from the module's ES major, so the per-module `useEsSharedSources` flag is gone entirely.

One rule I had to add for correctness: **a file enters a base only when every module of that major has it.** Otherwise the base would inject a file into a module that never had it — that constraint costs a little sharing but keeps each module's source set exactly what it was.

| major | before | after |
|---|---|---|
| es7x (12 modules) | 1,834 | **631** |
| es8x (17 modules) | 2,867 | **531** |
| es9x (4 modules) | 667 | **194** |
| es6x (1 module) | 139 | 139 |
| **total** | **5,507** | **1,495** |

This also picks up es70x–es717x, which the previous design left as full copies — your point that the per-major scheme works for es7 too.

Verification unchanged in method: every one of the 34 modules has an effective source set (base + own files, local wins) **byte-identical to before**, and `compileScala` is green for all 34 — the first time the ES 6/7 modules are exercised through the overlay.

coutoPL (changes requested):
can I asked you to extract the point 1 (drift fix) to separate PR. 
I would like to review it separatelly

sscarduzio:
Done — the drift fix (point 1) is extracted to #1322 as its own PR, same commit cherry-picked onto develop. Once #1322 merges, this PR's diff reduces to the overlay mechanism + migration + build-config derivation. I'll re-request review here after that lands.

sscarduzio:
Rebuilt on top of #1322 now that it is merged, so this PR contains **only the restructuring** — the drift fixes you asked to review separately are on `develop` already.

It is now **one commit**: no duplicated drift commit, no already-merged netty commits, no merge commits. A plain rebase could not produce that (the netty content reached `develop` as a squash, so replaying the originals conflicted on `suppressions_cve.xml`), so I rebuilt the branch from `develop` and re-ran the base computation.

Re-running it against your cleaned sources also **improved the result**: the `DeleteComponentTemplate` cleanup in #1322 made that file identical across more modules, so it now lives in the bases instead of being duplicated as overrides.

| | before | after |
|---|---|---|
| total source files | 5,507 | **1,492** (468 base + 1,024 module) |

Verification is unchanged in method: every one of the 34 modules has an effective source set byte-identical to its pre-migration tree, `compileScala` is green for all 34, and `es94x buildRorPluginZip` + `prepareDockerBuildContext` are green with the staged `Dockerfile` byte-identical to its recipe.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -149,9 +149,27 @@ tasks.register('prepareDockerBuildContext', Copy) {
     dependsOn prepareDockerFatLayer, prepareDockerThinLayer
     outputs.upToDateWhen { false }
 
-    from layout.projectDirectory.file("Dockerfile")
+    // The module names its image recipe via `rorDockerfile` in gradle.properties. The recipes live once in
+    // /docker-image/dockerfiles (7 variants; the differences are deliberate, e.g. the JDK-8287073 corretto
+    // stages exist only where an affected ES version is published). Staged into the context root under the
+    // name `Dockerfile`, so the build context is byte-identical to the old per-module copy.
+    def rorDockerfile = providers.provider {
+        def name = project.findProperty('rorDockerfile')
```
coutoPL: IMO, Dockerfile should stay where it was - in the es module folder. It should be treated as any other file - base defines it, but a module can overwrite it with its own version.

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
```diff
@@ -42,6 +42,57 @@ project.ext {
 
 group = 'org.elasticsearch.plugin'
 
+// Everything an es module's build.gradle used to declare is a pure function of the module's ES version
+// range (from its gradle.properties). Derive it here so a module file is just the plugins block.
+// A module can still override any of these ext values in its own build.gradle: the consumers read them
+// lazily, and the module body runs after this plugin.
+def esBase = project.ext.baselineEsVersion.tokenize('.')
```
coutoPL: I'm not sure about changes in this file. 
IMO, we should introduce one new convention:
base module for the major ES version defines common files, and a specific ES module can overwrite any file.

It should also apply to a `build.gradle` file. 

In the current approach, we have one source of truth for all modules. It's not good because:
1. it's another convention, specific for this kind of file
2. we will modify it, and breaking the old modules may happen more likely than in the requested approach above

coutoPL (changes requested):
I noticed two probematic things with the solution:
1. Compiler errors point at build/es-base-src/..
2. No IDE handling

Are we able to improve it?

sscarduzio:
Re-synced against develop — conflicts resolved, `MERGEABLE` again.

**Both of your requests are now satisfied:**
- The drift fix was extracted to #1322 (merged), so this PR is the overlay mechanism only.
- The branch is rebuilt on top of your squashed `[RORDEV-2151]` commit — I did not force-push over it.

**What the merge had to absorb:** #1307 (`_terms_enum` fields/filters) landed while this was in review, and it had to edit the same handful of files across **24 modules / 72 module files**. Rather than hand-resolve 46 conflicting paths, I regenerated the bases from develop's module trees using the documented rules (one base per major; a file enters a base only when every module of that major has it; completed majors take the most common shape, the active major the newest; a module keeps a copy only where it differs).

Worth noting as a check on your layout: **regeneration reproduced your existing base/override split exactly** — the only delta was the 28 files `_terms_enum` touches. After the merge its shared part lives once per major base instead of 24 times.

**Verification**
- Every module's effective source set (base minus overrides, plus its own files) is **byte-identical to origin/develop** — mechanical proof that the compiler sees exactly what it saw before.
- All 34 modules compile; representative modules per major (es714x, es717x, es80x, es818x, es90x, es94x) re-compiled with `--rerun-tasks`.
- Totals: 5531 module source files → 1031 module + 470 base.

Ready for another look when you have a moment.

coutoPL:
@sscarduzio did you address all my comments?

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
sscarduzio: Agreed, done. The Dockerfile is now an ordinary base file: `es{7,8,9}-base/Dockerfile` holds the major's recipe and a module that needs a different one simply keeps its own `Dockerfile`, which wins. The `rorDockerfile` property and `/docker-image/dockerfiles` are gone.

Overrides ended up being the modules that genuinely differ: es80x–es84x (the JDK-8287073 corretto stages) and, in ES 7, the four recipe variants that line has. Verified the staged build context is byte-identical to before for both paths — a base user (es94x, es85x) and an override user (es80x, es717x).

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
sscarduzio: You are right, and I have taken your version — the 51-line derived block is gone.

`build.gradle` now follows the same rule as everything else: `es{7,8,9}-base/module.gradle` holds the major's build config (JDK, JarHell class, dependencies), and a module's `build.gradle` is the plugins block, `apply from:` its base, then only what differs. es67x keeps a full script since it is the only ES 6 module and has no base.

Your point 2 is what convinced me: it is the same argument this branch already accepted when the single global source base was re-cut into one per major. A central derivation can break the frozen majors that PR CI does not run; a per-major base cannot reach outside its own major.

One wrinkle worth flagging: a module cannot un-declare a dependency its base declared, so three modules express their exclude at configuration level instead (es711x for elastic#68889, es92x/es94x for exponential-histogram). To prove that is not a behaviour change I compared the **resolved** compileClasspath, runtimeClasspath and distJars for those three before and after — byte-identical. For all 34 modules the declared config (dependencies, excludes, ext facts) is identical too, apart from those three lines, which the dump only records at dependency level.

While checking that I found something separate, which I did **not** change here: `exclude module: 'exponential-histogram'` matches nothing, because ES publishes the artifact as `elasticsearch-exponential-histogram`. I measured it — removing the exclude alters no resolved classpath, so that jar has been shipping all along. Left verbatim to keep this PR behaviour-neutral; worth a follow-up if the exclusion was meant to bite.

sscarduzio:
Both review points are implemented — replies on the threads.

The PR now has **one convention** instead of three: the major's base defines the common file, a module overrides any of it. That covers the sources, the `Dockerfile` and the module build script alike.

- `es{7,8,9}-base/module.gradle` — the major's build config; a module's `build.gradle` is the plugins block + `apply from:` + only its differences
- `es{7,8,9}-base/Dockerfile` — the major's recipe; a module keeps its own only when it differs
- deleted: `/docker-image/dockerfiles`, the `rorDockerfile` property in 34 files, and the 51-line version-derived block

`ES_MODULE_LAYOUT.md` and `CLAUDE.md` now describe the single rule.

**Verification**
- declared config (dependencies, excludes, ext facts) for all 34 modules: identical before/after, except three excludes that had to move from dependency level to configuration level (a module cannot un-declare what its base declared)
- for those three (es711x, es92x, es94x) the **resolved** compileClasspath, runtimeClasspath and distJars are byte-identical before/after
- all 34 modules compile
- staged Docker build contexts byte-identical for base users (es94x, es85x) and override users (es80x, es717x)

One incidental finding, deliberately not fixed here: the `exponential-histogram` exclude matches nothing (ES publishes it as `elasticsearch-exponential-histogram`), so that jar ships today. Measured — removing the exclude changes no resolved classpath. Flagged on the thread as a possible follow-up.

sscarduzio:
Re-synced onto develop after #1337 merged. **One commit again**, no merge commits.

The conflict looked alarming (a rebase hit 30 files) but was an artifact of replaying across this branch's own merge commits. The actual divergence was **one file**: this branch carried its own variant of the BouncyCastle CVE suppression, and develop now has the reviewed one from #1337.

So instead of re-running the migration, I rebuilt the tree directly and verified both directions:

- **vs the previous branch state:** the only difference is `suppressions_cve.xml` (develop's version wins) — nothing from the restructure was lost.
- **vs develop:** the diff is the restructure alone; `suppressions_cve.xml` does not appear in it.

The `ci/ci-lib.sh` hunk is part of the restructure — it repoints a comment at `preBuildDockerImageVersion`'s new home in `docker-image-conventions.gradle`.

The shape you reviewed is unchanged: one base per major, module overrides win, a file enters a base only when every module of that major has it.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
coutoPL, line 1: I noticed two problematic things with the solution:

1. Compiler errors point at build/es-base-src/..
2. No IDE handling

Are we able to improve it?

sscarduzio: Both improved.

**1. Compiler errors now point at the real files.** The Sync staging is gone entirely: per-file selection is impossible on a `SourceDirectorySet` (excludes are global across srcDirs — the original reason for staging), but it is perfectly possible on the **compile task's source**, so `compileScala` now receives `fileTree(es<major>-base) minus overrides` directly. No copies exist anywhere.

Verified with a deliberate syntax error in `es9-base/.../ResponseFieldsFiltering.scala`: the failure reports

```
es9-base/src/main/scala/tech/beshu/ror/es/ResponseFieldsFiltering.scala
```

— the real path. es94x/es818x/es717x compile clean, `formatCodeCheck` + `license` green (the `build/es-base-src` spotless exclusion is gone too, since there is nothing to exclude anymore).

**2. IDE:** the phantom `build/es-base-src` root no longer appears at all. Each `es<major>-base` is its own Gradle project, so IDEA shows it as a regular module — base sources are opened and edited there, and a per-ES-version error surfaces on that version's `compileScala`. One inherent limitation remains, shared by any shared-source scheme: the editor checks a base file against one classpath, not against all modules of the major at once — the per-module compile is where version-specific breakage shows. Documented in `ES_MODULE_LAYOUT.md`.

Still one commit; the source-set identity argument is unchanged (same selection predicate, different delivery).

sscarduzio: Both points handled, and I checked each one rather than reasoning about it.

**1. Compiler errors — already fixed, now proven.** The staged-copy design that produced `build/es-base-src/..` is gone; the applicable base files feed `compileScala.source` directly. To be sure the paths really are the source ones, I appended `val x: Int = "not an int"` to a base file and compiled `:es94x`:

```
[Error] .../es9-base/src/main/scala/tech/beshu/ror/es/RorRestChannel.scala:104:59: Found: ("not an int" : String)
```

The real base path, with line and column. Nothing points into `build/`.

**2. IDE handling — this was a genuine gap.** The IDE builds a module classpath from source sets, and the base sources reach a module through the compile *task* instead, so IntelliJ saw two broken things: `es9-base` had no ES on its classpath at all, and `es94x` could not resolve any base class its overrides touch. Two changes, both IDE-only:

`es<major>-base/build.gradle` now declares the same dependency set an `es<major>x` module gets, pinned to the oldest ES of that major — the version the modules compile the base against. Gradle still compiles nothing here.

`readonlyrest.plugin-common-conventions` adds the base as a PROVIDED module dependency in the IDEA model:

```groovy
idea.module {
    scopes.PROVIDED.plus += [
            configurations.detachedConfiguration(
                    dependencies.project(path: ":es${esMajorForBase}-base"))
    ]
}
```

A shared source *directory* cannot do this job — IntelliJ allows a directory in one module only, and this one belongs to every module of the major. A module dependency has no such limit.

What I verified:

- `es94x.iml` now contains `<orderEntry type="module" module-name="es9-base" scope="PROVIDED"/>`, so the IDE model carries it.
- `:es9-base:dependencies` resolves `org.elasticsearch:elasticsearch:9.0.0`; es8-base pins 8.0.0 and es7-base 7.0.0. Before the change the base projects had no ES dependency at all.
- No leak into the build: `es94x` `compileClasspath`, `runtimeClasspath` and `distJars` all contain zero references to `es9-base`.
- `:es94x:compileScala`, `:es818x:compileScala` and `:es717x:compileScala` are green after the change and after the rebase.

One limit I want to be straight about: I verified the Gradle-side IDEA model, not IntelliJ itself — I cannot run the IDE here. If you re-import and something still shows red, tell me what and I will chase it.

Also rebased onto current develop. That dropped the `bctls-fips` suppression this branch carried, because #1337 landed a broader one covering all four FIPS artifacts, so `suppressions_cve.xml` is untouched by this PR now.

sscarduzio:
CI note for the last force-push. 14 of the 15 red checks came from the Docker Hub unauthenticated pull rate limit, not from this branch:

- unit_tests_linux: the LDAP testcontainer image pull got `toomanyrequests` 24 times.
- all 11 it_linux legs: `warmDockerImageCache` failed on the same limit.
- the 3 e2e legs: the pre-build image could not pull `tianon/gosu` (HTTP 429).

develop fails with the same signature, and #1346 is the cure. This branch changes nothing for those legs.

cve_check was the one failure this branch caused. The new `es*-base` modules declare ES dependencies only so that the IDE can resolve symbols. dependency-check scanned them through `runtimeClasspath` and flagged the CVEs of the ES-provided jars, which ROR does not ship. The real es modules scan `distJars` and stay clean. The fix adds `dependencyCheck { skip = true }` to the three `es*-base` build.gradle files, with a comment that says why. I ran the three `dependencyCheckAnalyze` tasks locally: each one now reports "Skipping dependency-check-gradle". The fix is amended into the single commit.

sscarduzio:
Rebased onto develop once more, no code change: the branch needed the ci.yml from #1347 (renamed CI secrets), or the e2e legs fail at docker login.

sscarduzio:
@coderabbitai full review

coutoPL:
@sscarduzio, during the weekend we will probably do the release. This is a complicated change; I don't want to add to the develop branch before the release. 

So, it's low priority now. I'm going to do the next iteration of the review in the meantime. 

I just want to make you sure, I remember about it.

sscarduzio:
@coutoPL rebased onto develop, so this now carries your mirror PR (#1352) and ES 9.5.2 (#1353). It is `MERGEABLE` again, and CI is running.

Two conflicts, both resolved in favour of develop:

- `ci/ci-lib.sh` — my only change here was a comment, and your `gradle_property dockerImageNamespace` refactor made it obsolete. Your version stands.
- `es94x/gradle.properties` — mine had added a stray blank line. 9.5.2 is kept.

`readonlyrest.docker-image-conventions.gradle` auto-merged, and I read the result: the mirror and namespace code and the `rorDockerfile` base-override resolution sit in different parts of the file, so they do not interact. Of the 4,600 files this branch touches, the rebase changed 3, and all three took develop's content.

Nothing is open on my side — every thread has an answer, including the compiler-paths and IDE one, which I answered with a measured example. Over to you whenever you have a moment. 🙏

sscarduzio:
Rebased onto current develop (93a581581). Three PRs there touched files this branch hoists: #1356 (Sink→Output rename), #1359 (audit config rework), #1315 (lookup indices).

Resolution rule: develop's content is law (this branch is behavior-neutral), only placement is recomputed — base file only where every module of the major has the file, majority content → base, the rest stay module overrides. The new grouping matches the branch's previous override membership everywhere; #1359 added one new split (`RestClientAuditOutputService` now differs in es90x, which gained an override).

Verified mechanically: for all 34 modules and every Scala file, the effective source (override-or-base) is byte-identical to develop's per-module content — zero mismatches. Compile check on es67x/es70x/es710x/es717x/es80x/es85x/es818x/es90x/es94x: green.

sscarduzio:
CI note — `e2e_es94x` fails reproducibly on this branch, and I do not think it is the usual flake.

## Evidence

| Branch | Samples | `Reporting.cy.ts` renderer crash |
|---|---|---|
| this PR | 3 | **3** |
| `develop` (last 6 CI runs) | 6 | **0** |

The `develop` comparison is complete rather than sampled: a renderer crash fails the spec, so the 4 green runs cannot contain one, and I read the logs of both red ones — they failed on `Dev-tools.cy.ts` / `Sanity-check.cy.ts` and on a Docker Hub throttle, with zero crashes. `Reporting.cy.ts` passed 4/4 on `develop`.

The failure is not an assertion:

```
Renderer process crashed
We have failed the current spec but will continue running the next spec.
```

It always hits `Reporting.cy.ts`, always in the **admin** section, and it is getting earlier each run — 3 passing, then 2, then 1:

| Run | Result |
|---|---|
| 1 | 3 passing, 1 failing |
| 2 | 2 passing, 1 failing, 1 skipped |
| 3 | 1 passing, 1 failing, 2 skipped |

## What is ruled out

- **Docker Hub throttling** — `throttle=0` in all three runs.
- **A failing request before the crash** — the admin seeding calls return 201/200, same as `develop`.
- **A ROR error** — nothing in the ROR logs before the crash beyond the normal `[ROR|hiddenApps]` output.
- **The other e2e legs** — `es717x` and `es818x` pass.

Run 1 did log a real JS `TypeError: t.toUpperCase is not a function` from the dashboard bundle, which would fit a malformed response reaching the UI, but I have not tied it to the crash.

## What I do not know

The mechanism. This PR changes ES-module packaging and touches no Kibana code, so I have no explanation for how it crashes an Electron renderer. The moving crash point looks resource-driven rather than deterministic. Flagging it rather than guessing — this needs a look at the Cypress video (uploaded to the e2e reports bucket for build 33147993850) or a local repro before the PR is merged.

sscarduzio:
@coutoPL ready for the next pass whenever you have time.

The reason you parked this has cleared: you said on 08-20 you did not want a change this size on develop before the release, and 1.71.0 shipped on 08-28.

Where your three asks landed:

**Dockerfile placement.** Done as you asked. The Dockerfile is an ordinary base file now — `es7-base/Dockerfile`, `es8-base/Dockerfile`, `es9-base/Dockerfile` — and a module overrides it by simply having its own. The `rorDockerfile` property and the `/docker-image/dockerfiles` directory are gone. Overrides are limited to es80x–es84x and the four ES 7 recipe variants. I verified the staged build context is byte-identical for es94x, es85x, es80x and es717x.

**One convention, not a special case for this file type.** Also done your way. The 51-line derived block is gone; each major's build config lives in `es7-base/module.gradle`, `es8-base/module.gradle`, `es9-base/module.gradle`, and a module overrides by declaring its own. One wrinkle you should look at: es711x, es92x and es94x had to move their excludes from dependency level to configuration level, because a module cannot un-declare a dependency the base declared. Resolved classpaths verified byte-identical.

**Compiler errors pointing at `build/es-base-src`, and no IDE handling.** The sync staging is removed entirely — `compileScala` now takes `fileTree(es<major>-base) minus overrides` directly, so errors point at real source. Verified with a deliberate type error:

```
[Error] .../es9-base/src/main/scala/tech/beshu/ror/es/RorRestChannel.scala:104:59: Found: ("not an int" : String)
```

For the IDE half I added the base as a PROVIDED module dependency in the IDEA model. **This is the one thing I could not verify myself** — I checked the Gradle-side model, not IntelliJ. If you re-import and something still shows red, tell me what and I will chase it. That is the only open question I know of.

Ten force-pushes have landed since your last look, so it is worth re-reading rather than diffing against what you remember.

Two notes on the state of the PR:

- `e2e_es94x` went red on the 08-28 run, but it is a flake, not a regression: the Electron renderer crashed mid-spec (`Reporting.cy.ts`, the one test that ran passed first). This PR touches ES-module Gradle and source layout — the ES image was reused unchanged that run ("Sources unchanged … skipping build") and the specs come from the e2e repo's develop. This identical head passed full CI twice before, on 08-21 and 08-23. I have re-run the job; everything else is green.
- Incidental finding I deliberately did **not** fix, to keep this PR behaviour-neutral: `exclude module: 'exponential-histogram'` matches nothing, because ES publishes that artifact as `elasticsearch-exponential-histogram`. Left verbatim. Worth a follow-up ticket if the exclusion was meant to bite.

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
coutoPL: This part should be applied to base ES modules. Let's create another Gradle file with this part and apply it to all base-es modules

es7-base/build.gradle
coutoPL, line 1: In `readonlyrest.plugin-common-conventions.gradle` the base sources are given to the IDE like this:

```groovy
idea.module { scopes.PROVIDED.plus += [ configurations.detachedConfiguration(dependencies.project(path: ":es7-base")) ] }
```

This has no effect in IntelliJ. `idea.module.scopes` is only read by the `./gradlew idea` task that
generates `.iml` files. When IntelliJ imports the Gradle project itself (which is how we open it — you can
see it in `.idea/scala_compiler.xml`, the modules are `readonlyrest.es710x.main`,
`readonlyrest.es7-base.main`, i.e. one module per source set), dependencies come from the real
configurations, not from that block.

Result: `es710x.main` has no dependency on `es7-base`, so from
`es710x/…/IndexLevelActionFilter.scala` you cannot ctrl+click `AclAwareRequestFilter.EsContext` — it exists
only in `es7-base`. Symbols from `core` (`RequestIdAwareLogging`) and from the module's own overrides
(`EsEnvProvider`) navigate fine, because those have a real dependency / are in the same module. There are no
red marks because Scala 3 uses compiler-based highlighting, so the missing symbols simply don't get
underlined — but navigation is dead.

Fix: give the module a real dependency the IDE can see:

```groovy
dependencies {
    compileOnly project(path: ":es${esMajorForBase}-base")
}
```

The base compiles no bytecode (its compile tasks are disabled), so this jar is empty, and `compileOnly`
keeps it out of `runtimeClasspath`/`distJars` — the plugin zip is unchanged. Verified: `compileClasspath`
gains `project :es7-base`, `distJars` doesn't, `:es7-base:jar` is manifest-only, `:es710x:compileScala`
still green. Needs a Gradle re-import in the IDE.

es7-base/src/main/scala/tech/beshu/ror/es/handler/response/DocumentApiOps.scala
coutoPL, line 1: The base should hold the shape of the **oldest** ES version of the major, and every later module that is
incompatible should override it. Right now it holds the "most common" shape instead
(`ES_MODULE_LAYOUT.md` rule 2), which mixes shapes from different points in the major inside one base.

Example — `es7-base/…/response/DocumentApiOps.scala` calls the 10-argument
`GetResult(…, documentFields, metadataFields)`. That constructor only exists from ES 7.9 on; in 7.0.0
`GetResult` takes 9 arguments. So `es70x, es72x, es73x, es74x, es77x, es78x` all carry an override of that
file, and the base is the 7.9+ shape.

Two problems follow:

- `es7-base/build.gradle` pins the IDE classpath to the oldest ES 7 (`allSupportedEsVersions` is sorted
  ascending, so it resolves to 7.0.0). The base content doesn't match that classpath, so base files show as
  broken in the IDE — `DocumentApiOps` is red on the last argument. The comment there ("the oldest ES 7 we
  support, which is what the modules of this major compile the base against") is also not right: each module
  compiles the base against its own baseline (es79x → 7.9.0, es717x → 7.17.13).
- The base isn't one coherent shape: `IndexLevelActionFilter.scala` in the base is the 7.0–7.4 shape
  (overridden by es77x…es717x) while `DocumentApiOps.scala` is the 7.9+ shape, so no single ES version can
  ever resolve the whole base.

With the oldest shape as the rule, the base means one thing — the major's original shape — it never changes
again, and the pinned ES version matches it. Cost of the re-partition (the per-module compiled sources stay
identical, only the placement changes): ES 7 → 50 of 136 base files change content, overrides 498 → 727;
ES 8 → 35 of 168 change, overrides 359 → 724. Still well below the ~1900 full copies we had before the
de-duplication.

docs/architecture/ES_MODULE_LAYOUT.md
coutoPL, line 1: Maybe we can move this file to `docs/architeture`? Or something like that

CLAUDE.md
```diff
@@ -46,7 +46,8 @@ ReadonlyREST is an Elasticsearch security plugin providing access control, authe
 ## Module Architecture
 
 - **`core/`** — Shared security logic (Scala 3). Contains access control rules, settings parsing, API endpoints, field-level security, and boot/engine initialization. All ES modules depend on this.
```
coutoPL, line 48: Can we move this paragraph to `docs/architecture`? and link the new file here only?

CLAUDE.md
```diff
@@ -90,7 +91,7 @@ ES module entry point pattern: `es{version}x/src/main/scala/tech/beshu/ror/es/Re
 
 ## When Porting Changes Across ES Modules
 
-Many ES modules share near-identical code. When changing an ES-specific module, check whether the same change applies to other `es{version}x` modules. Divergence is intentional when ES internal APIs differ between versions — porting is not always 1:1.
+Shared code lives once per ES major in `es7-base/`, `es8-base/`, `es9-base/` — edit it there and the change applies to every module of that major (verify each still compiles; ES APIs differ between minors). Only touch a module's own `src/` when its ES API genuinely differs; that copy is an override and shadows the base file. When an override becomes identical to the base, delete it. A change that spans majors must be applied to each major's base. `es67x` has no base and is a full copy. See `ES_MODULE_LAYOUT.md`.
```
coutoPL: This could be moved to ES_MODULE_LAYOUT.md

CLAUDE.md
coutoPL: This is not needed at this level. Maybe a new file in the `docs/architecture`? We may develop the content later

es710x/build.gradle
coutoPL, line 1: make sure you ported this change: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1362

coutoPL (changes requested):

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
sscarduzio: Done — two new convention plugins in build-base:
- `readonlyrest.es-libs-repository` — this repositories block, applied by the es*x modules and by all three base projects, so the base IDE classpath resolves from the same places.
- `readonlyrest.es-base-conventions` — everything the three identical `es<N>-base/build.gradle` files shared. They now hold only the plugins block plus `apply from: module.gradle` (a precompiled script plugin cannot `apply from:` an external script — Gradle limitation).

es7-base/build.gradle
sscarduzio: Applied exactly as you specified: `compileOnly project(":es${esMajorForBase}-base")` in plugin-common-conventions; the `idea.module` block and the `idea` plugin are gone. Re-verified on this head: `:es717x:dependencies --configuration distJars` has no es7-base, `--configuration compileClasspath` has it, compileScala green. Thanks for the worked-out fix.

es7-base/src/main/scala/tech/beshu/ror/es/handler/response/DocumentApiOps.scala
sscarduzio: Re-cut as proposed (commit 55b79843f): es7-base and es8-base now hold the OLDEST module's shape. es70x and es80x have zero overrides — the base IS their shape. The Dockerfile follows the same rule.

Actual cost is lower than your estimate, because for many es7 files the most common shape already was the oldest: es7 — 50 of 136 base files changed, overrides at base paths now 523 (vs your 727); es8 — 37 of 169, now 719 (≈ your 724).

Verified: effective per-module content (override if present, else base) is byte-identical before and after the re-cut, asserted per module and path by the re-partition script; compileScala green for es70x, es717x, es80x, es818x, es94x.

The pin problem you found is gone by construction: the base project now pins its classpath to the version whose shape the base holds — oldest of the major for a completed major, newest for the active one (es9 keeps the newest shape). The wrong comment is replaced; rule 2 of the layout doc is updated.

docs/architecture/ES_MODULE_LAYOUT.md
sscarduzio: Moved: `docs/architecture/ES_MODULE_LAYOUT.md`. References updated.

CLAUDE.md
sscarduzio: Done — the two bullets are one-liners now; the detail lives in `docs/architecture/ES_MODULE_LAYOUT.md`.

CLAUDE.md
sscarduzio: Moved into the layout doc — its "How to change a file" section already covered most of this; the cross-major note is folded in. CLAUDE.md keeps a one-line pointer.

CLAUDE.md
sscarduzio: Moved to `docs/architecture/configuration-loading.md`, to grow later.

es710x/build.gradle
sscarduzio: Ported — the branch is rebased onto today's develop (through #1369, so also the 8.19.21 / 9.5.3 version bumps), and the `es710x/Dockerfile` override carries the `/tini` entrypoint from #1362.

sscarduzio:
@coutoPL all eight points from your 09-03 pass are in, as three commits on top of the rebased base commit, so you can review them step by step instead of re-reading the whole diff:

1. `55b79843f` — **oldest-shape re-cut** of es7-base and es8-base. es70x/es80x have zero overrides now. Effective per-module content verified byte-identical before/after (asserted per module and path).
2. `b7b56ae33` — **`readonlyrest.es-base-conventions`** (the three base build.gradle files are plugins-block-only now) + **`readonlyrest.es-libs-repository`** (applied by modules AND bases) + your **`compileOnly project(:es<major>-base)`** IDE fix, replacing the dead `idea.module` block.
3. `171f9ddb6` — **docs**: ES_MODULE_LAYOUT.md → `docs/architecture/` (rule 2 now says oldest shape), CLAUDE.md slimmed to pointers, configuration loading → `docs/architecture/configuration-loading.md`.

Also rebased onto develop: picks up 8.19.21/9.5.3 bumps and the es710x tini fix (#1362).

Verified in the CI toolchains container: compileScala green for es70x, es717x, es80x, es818x, es94x; `:es717x` distJars has no es7-base, compileClasspath has it.

Note: cve_check will stay red here until #1374 merges (three new ES-server CVEs false-positive-matched onto the rest-client jars — repo-wide, not this PR).

docs/architecture/ES_MODULE_LAYOUT.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Rules
+
+1. A file goes into the base only when **every** module of that major has it. If one module does not
+   have the file, the base would add it to that module.
+2. For a **completed** major (ES 7 and 8), the base holds the shape of the **oldest** ES version of
+   the major -- the major's original shape. That shape never changes again, and it is the one shape a
+   single ES version can resolve in full, so the base project pins its IDE classpath to exactly that
+   version. Every later module that is incompatible overrides the file.
+3. For the **active** major (ES 9), the base holds the **newest** shape, and the base project pins its
```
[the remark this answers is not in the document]
sscarduzio: Fixed in 558fcf019, your first option: the base now carries the es94x shape; es90x and es91x override with the old shape; es92x keeps its own. Effective per-module content is unchanged (the es94x delta was a trailing comma only), and rule 3 is now uniformly true — 13 of 13 non-identical es9-base files hold the newest shape.

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
sscarduzio: Done in the previous push — the block that sat at this line (the LibsStore `repositories` block) lives in `build-base/src/main/groovy/readonlyrest.es-libs-repository.gradle` now, applied by both `readonlyrest.plugin-common-conventions` and the new `readonlyrest.es-base-conventions`. Each `es{7,8,9}-base/build.gradle` is the same five-line shell that applies `readonlyrest.es-base-conventions`.

es7-base/build.gradle
sscarduzio: Taken verbatim, thanks for running it to ground. The `idea.module.scopes` block is gone; `readonlyrest.plugin-common-conventions.gradle` declares `compileOnly project(path: ":es${esMajorForBase}-base")`.

Re-verified today on the current head:
- `:es710x:dependencies --configuration compileClasspath` lists `project :es7-base`
- `runtimeClasspath` and `distJars` contain zero `es*-base` entries — the plugin zip is unchanged
- `:es7-base:jar` is a 261-byte, manifest-only jar
- es70x, es717x, es80x, es818x, es90x, es94x `compileScala` all green

es7-base/src/main/scala/tech/beshu/ror/es/handler/response/DocumentApiOps.scala
sscarduzio: Agreed — and it is now the rule for all three majors, with no exception.

es7 and es8 were re-cut to the oldest shape in the previous push: es70x and es80x override nothing, which is the structural test of the rule (the oldest module IS the base). The es9 base still held the newest shape under the old rule-3 "active major" carve-out; the new commit (2d6a6a08a) drops that carve-out. 13 es9-base files flip to the es90x shape, 26 override files appear on es91x/es92x/es94x holding the old base content, 17 now-redundant overrides go away, and `es90x/src` is gone entirely — a pure base consumer like es70x.

`ES_MODULE_LAYOUT.md` rules 2 and 3 are now one rule: every base holds the shape of the major's oldest ES version, later modules override. The base project's IDE pin is always `versionsOfMajor.first()`, so the pin and the base content agree by construction — the red mark you saw on `DocumentApiOps` goes away because the base now IS the 7.0 shape (9-arg `GetResult`), and the misleading "modules compile the base against the oldest ES" comment was already corrected to "each module compiles base sources against its own ES version".

Verification, not reasoning: a script computed the effective per-module source map (override if present, else base; sha256 per path) before and after — byte-identical for all 33 modules of majors 7/8/9, 5425 path entries. Oldest and newest module of each major compile green.

sscarduzio:
Rebased onto develop (7cfa1e61), no conflicts, no file of this branch changed.

The red `cve_check` was not this branch. The es8xx/es9xx scans flagged CVE-2026-72649, -78607, -78605 and -56143 (and -72656 on es811x–es818x) on `elasticsearch-rest-client`, and develop got the suppressions after this branch was cut (#1374, #1378). Verified in the CI toolchains image:
- `:es80x:dependencies --configuration distJars` is byte-identical on the old head, on develop and after the rebase: no `es*-base` project, no test-scope jars.
- `:es80x:dependencyCheckAnalyze` and `:es90x:dependencyCheckAnalyze` fail with the three rest-client CVEs on the old head and pass with 0 vulnerabilities after the rebase.

One slip while pushing: for about a minute the branch pointed at develop, which auto-closed the PR. Restored and reopened; the review threads and the content are intact.

sscarduzio:
@coutoPL this one is ready for another look. Your changes-requested is from before the last push: everything you raised is addressed, the branch is CLEAN against master, and CI is 36 pass / 0 fail.

sscarduzio:
@coutoPL all the points from your 09-03 review are addressed in commits `c6311a25e`–`9f43762bb` (09-09). Could you take another look? 10hexdev APPROVED after verifying the full byte-identity of all 33 modules.

PR #1315, opened 2026-08-09
[RORDEV-1563] Lookup indices support

mgoworko:
- type: security                                                                                                                                                                                                            
    components: [es]                                                                                                                                                                                                          
    text: "Fix authorization bypass allowing ESQL LOOKUP JOIN to read indices the user isn't granted"                                                                                                                         
                                                                                                                                                                                                                              
  ## What changed                                                                                                                                                                                                             
                                                                                                                                                                                                                              
  Added `EsqlIndexTable` (with a `From`/`LookupJoin` origin) so both `FROM` and `LOOKUP JOIN` tables are                                                                                                                      
  extracted, authorized, and narrowed/masked the same way — previously only `FROM` was.                                                                                                                                       
                                                                                                                                                                                                                              
  ## Why                                                                                                                                                                                                                      
                                                                                                                                                                                                                              
  Before this PR, `LOOKUP JOIN` targets were invisible to ROR's ACL: a user authorized for one index could                                                                                                                    
  join in data from any other index on the cluster, unmasked. This closes that bypass, with the same                                                                                                                          
  wildcard/alias narrowing `FROM` already had. 

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## New Features
- Added ESQL `LOOKUP JOIN` access control alongside standard `FROM` queries.
- Improved handling of authorized, narrowed, wildcard, aliased, and unauthorized index patterns.
- Added generic nonexistent-index masking for unauthorized or unavailable targets.
- Improved query rewriting while preserving table boundaries and independent source resolution.

## Bug Fixes
- Ensured lookup tables are analyzed and handled independently from regular query sources.

## Tests
- Added coverage for joins, wildcard narrowing, aliases, unauthorized targets, and index masking across supported Elasticsearch versions.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
+++ b/core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
@@ -31,7 +31,7 @@ sealed trait EsqlQueryRewriteResult
 
 object EsqlQueryRewriteResult {
   final case class Rewritten(newQuery: String) extends EsqlQueryRewriteResult
-  case object CannotRewriteQuery extends EsqlQueryRewriteResult
+  final case class CannotRewriteQuery(reason: String) extends EsqlQueryRewriteResult
 }
 
 sealed trait EsqlIndexTable {
@@ -51,8 +51,17 @@ object EsqlIndexTable {
   }
 
   object From {
+
     def parse(tableStringInQuery: String): Option[From] =
       requestedIndicesFrom(tableStringInQuery).map(From(tableStringInQuery, _))
+
+    private def requestedIndicesFrom(
+        tableStringInQuery: String
+    ): Option[NonEmptyList[RequestedIndex[ClusterIndexName]]] =
+      NonEmptyList.fromList(
+        tableStringInQuery.split(',').asSafeList.filter(_.nonEmpty).flatMap(RequestedIndex.fromString)
+      )
+
   }
 
   final case class LookupJoin(tableStringInQuery: String, index: IndexName.Full) extends EsqlIndexTable {
@@ -71,7 +80,7 @@ object EsqlIndexTable {
 
   }
 
-  final case class Replacement(table: EsqlIndexTable, newIndices: NonEmptyList[ClusterIndexName])
+  private[ror] final case class Replacement(table: EsqlIndexTable, newIndices: NonEmptyList[ClusterIndexName])
 
   def requestedIndicesOf(tables: NonEmptyList[EsqlIndexTable]): Set[RequestedIndex[ClusterIndexName]] =
     tables.toList.flatMap(_.requestedIndices.toList).toCovariantSet
@@ -84,37 +93,37 @@ object EsqlIndexTable {
     newQueryFrom(oldQuery, buildReplacements(tables, allowedIndices))
   }
 
-  def buildReplacements(
+  private[ror] def buildReplacements(
       tables: NonEmptyList[EsqlIndexTable],
       allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
   ): NonEmptyList[Replacement] = {
     val allowedIndexNames: Set[ClusterIndexName] = allowedIndices.includedOnly
     val fromTables = tables.collect { case table: From => table }
 
-    val lookupOnlyNames: Set[ClusterIndexName] =
+    val lookupOnlyIndexNames: Set[ClusterIndexName] =
       tables.collect { case table: LookupJoin => table.clusterIndexName }.toCovariantSet --
         fromTables.flatMap(_.requestedIndices.includedOnly).toCovariantSet
 
-    val matchedNamesByFromTable: Map[From, Set[ClusterIndexName]] =
+    val matchedIndexNamesByFromTable: Map[From, Set[ClusterIndexName]] =
       fromTables.map(table => (table, table.matcher.filter(allowedIndexNames))).toMap
 
     // an alias the ACL replaced with its backing index matches no FROM pattern, and nothing left in the
     // request says which table it came from
-    val unattributedNames: Set[ClusterIndexName] =
-      allowedIndexNames -- lookupOnlyNames -- matchedNamesByFromTable.values.flatten.toCovariantSet
+    val unattributedIndexNames: Set[ClusterIndexName] =
+      allowedIndexNames -- lookupOnlyIndexNames -- matchedIndexNamesByFromTable.values.flatten.toCovariantSet
 
-    def authorizedNamesFor(table: EsqlIndexTable): Option[NonEmptyList[ClusterIndexName]] = table match {
+    def authorizedIndexNamesFor(table: EsqlIndexTable): Option[NonEmptyList[ClusterIndexName]] = table match {
       case table: LookupJoin =>
         Option.when(allowedIndexNames.contains(table.clusterIndexName))(NonEmptyList.one(table.clusterIndexName))
       case table: From =>
-        NonEmptyList.fromList((matchedNamesByFromTable(table) ++ unattributedNames).toList)
+        NonEmptyList.fromList((matchedIndexNamesByFromTable(table) ++ unattributedIndexNames).toList)
     }
 
-    val authorizedNamesByTable: Map[EsqlIndexTable, Option[NonEmptyList[ClusterIndexName]]] =
-      tables.toList.map(table => (table, authorizedNamesFor(table))).toMap
+    val authorizedIndexNamesByTable: Map[EsqlIndexTable, Option[NonEmptyList[ClusterIndexName]]] =
+      tables.toList.map(table => (table, authorizedIndexNamesFor(table))).toMap
 
     val nonexistentIndexByTableText: Map[String, ClusterIndexName] =
-      authorizedNamesByTable
+      authorizedIndexNamesByTable
         .collect { case (table, None) => table.tableStringInQuery }
         .toList
         .distinct
@@ -122,31 +131,32 @@ object EsqlIndexTable {
         .toMap
 
     tables.map { table =>
-      val newIndices = authorizedNamesByTable(table)
+      val newIndices = authorizedIndexNamesByTable(table)
         .getOrElse(NonEmptyList.one(nonexistentIndexByTableText(table.tableStringInQuery)))
       Replacement(table, newIndices)
     }
   }
 
-  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): EsqlQueryRewriteResult = {
+  private[ror] def newQueryFrom(
+      oldQuery: String,
+      replacements: NonEmptyList[Replacement]
+  ): EsqlQueryRewriteResult = {
     val scan = QueryScan.of(oldQuery)
     replacements.toList
       .groupBy(_.table)
       .toList
       .traverse { case (table, tableReplacements) =>
         val spans = indexListSpansIn(scan, table)
-        Option.when(spans.size === tableReplacements.size)(spans.zip(tableReplacements))
+        Either.cond(
+          spans.size === tableReplacements.size,
+          spans.zip(tableReplacements),
+          s"the table [${table.show}] cannot be unambiguously located in the query text"
+        )
       }
       .map(_.flatten)
       .flatMap(spliceIndexLists(oldQuery, _))
-      .map(EsqlQueryRewriteResult.Rewritten.apply)
-      .getOrElse(EsqlQueryRewriteResult.CannotRewriteQuery)
+      .fold(EsqlQueryRewriteResult.CannotRewriteQuery.apply, EsqlQueryRewriteResult.Rewritten.apply)
   }
-
-  private def requestedIndicesFrom(tableStringInQuery: String): Option[NonEmptyList[RequestedIndex[ClusterIndexName]]] =
-    NonEmptyList.fromList(
-      tableStringInQuery.split(',').asSafeList.filter(_.nonEmpty).flatMap(RequestedIndex.fromString)
-    )
 
   private def indexListSpansIn(scan: QueryScan, table: EsqlIndexTable): List[(Int, Int)] = {
     val matcher = indexListPatternOf(table).matcher(scan.text)
@@ -284,15 +294,15 @@ object EsqlIndexTable {
 
   }
 
-  private def spliceIndexLists(query: String, spans: List[((Int, Int), Replacement)]): Option[String] = {
+  private def spliceIndexLists(query: String, spans: List[((Int, Int), Replacement)]): Either[String, String] = {
     val sortedSpans = spans.sortBy { case ((start, _), _) => start }
     val overlapping = sortedSpans.sliding(2).exists {
       case List(((_, previousEnd), _), ((start, _), _)) => previousEnd > start
       case _                                            => false
     }
-    if (overlapping) None
+    if (overlapping) Left("the index lists of the tables overlap in the query text")
     else
-      Some {
+      Right {
         sortedSpans.reverse.foldLeft(query) { case (currentQuery, ((start, end), replacement)) =>
           val newIndices = replacement.newIndices.toList.map(_.show).mkString(",")
           s"${currentQuery.substring(0, start)}$newIndices${currentQuery.substring(end)}"
--- a/core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
+++ b/core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
@@ -26,6 +26,13 @@ import tech.beshu.ror.utils.ScalaOps.*
 
 import java.util.regex.Pattern
 
+sealed trait EsqlQueryRewriteResult
+
+object EsqlQueryRewriteResult {
+  final case class Rewritten(newQuery: String) extends EsqlQueryRewriteResult
+  case object CannotRewriteQuery extends EsqlQueryRewriteResult
+}
+
 sealed trait EsqlIndexTable {
   def tableStringInQuery: String
   def requestedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
@@ -72,7 +79,7 @@ object EsqlIndexTable {
       oldQuery: String,
       tables: NonEmptyList[EsqlIndexTable],
       allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
-  ): Option[String] = {
+  ): EsqlQueryRewriteResult = {
     newQueryFrom(oldQuery, buildReplacements(tables, allowedIndices))
   }
 
@@ -102,14 +109,12 @@ object EsqlIndexTable {
     NonEmptyList.fromListUnsafe(fromReplacements ++ lookupReplacements)
   }
 
-  /** `None` when the rewrite cannot be applied. Callers must reject the request rather than skip it - the
-    * query was allowed only on the strength of the narrowed set, so unrewritten it runs against the
-    * user's original indices.
-    */
-  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): Option[String] = {
+  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): EsqlQueryRewriteResult = {
     replacements.toList
       .traverse(replacement => indexListSpanIn(oldQuery, replacement.table).map((_, replacement)))
       .flatMap(spliceIndexLists(oldQuery, _))
+      .map(EsqlQueryRewriteResult.Rewritten.apply)
+      .getOrElse(EsqlQueryRewriteResult.CannotRewriteQuery)
   }
 
   private def requestedIndicesFrom(tableStringInQuery: String): Option[NonEmptyList[RequestedIndex[ClusterIndexName]]] =
--- a/core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
+++ b/core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
@@ -24,7 +24,7 @@ import tech.beshu.ror.implicits.*
 import tech.beshu.ror.syntax.*
 import tech.beshu.ror.utils.ScalaOps.*
 
-import java.util.regex.{Matcher, Pattern}
+import java.util.regex.Pattern
 
 sealed trait EsqlIndexTable {
   def tableStringInQuery: String
@@ -72,7 +72,7 @@ object EsqlIndexTable {
       oldQuery: String,
       tables: NonEmptyList[EsqlIndexTable],
       allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
-  ): String = {
+  ): Option[String] = {
     newQueryFrom(oldQuery, buildReplacements(tables, allowedIndices))
   }
 
@@ -100,21 +100,22 @@ object EsqlIndexTable {
     NonEmptyList.fromListUnsafe(fromReplacements ++ lookupReplacements)
   }
 
-  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): String = {
-    replacements.toList.foldLeft(oldQuery) { case (currentQuery, replacement) =>
-      val (beforeFrom, afterFrom) = currentQuery.splitBy("FROM")
-      afterFrom match {
-        case None =>
-          replaceTableNameInQueryPart(currentQuery, replacement.table.tableStringInQuery, replacement.newIndices)
-        case Some(tablesPart) =>
-          s"${beforeFrom}FROM ${replaceTableNameInQueryPart(tablesPart, replacement.table.tableStringInQuery, replacement.newIndices)}"
-      }
-    }
+  /**
+   * `None` whenever a table ES reported cannot be located in the query, or two located index lists overlap.
+   *
+   * The request has already been allowed on the strength of the narrowed index set, so a rewrite that cannot
+   * be performed is not one that can be skipped - forwarding the query untouched would run it against the
+   * user's original, unauthorized indices. Callers turn `None` into a forbidden response.
+   */
+  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): Option[String] = {
+    replacements.toList
+      .traverse(replacement => indexListSpanIn(oldQuery, replacement.table).map((_, replacement)))
+      .flatMap(spliceIndexLists(oldQuery, _))
   }
 
-  // The returned map is threaded into `buildFromReplacements`: if the same literal text is masked in
-  // both a FROM and a LOOKUP JOIN, `newQueryFrom`'s single per-table substitution needs both to
-  // resolve to the same nonexistent name.
+  // The returned map is threaded into `buildFromReplacements` so that a literal masked in both a FROM and
+  // a LOOKUP JOIN reads as the same nonexistent index in the rewritten query, the way it read as the same
+  // index in the original one.
   private def buildLookupReplacements(
       tables: List[LookupJoin],
       allowedIndices: Set[ClusterIndexName]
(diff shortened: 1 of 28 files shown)


coutoPL:
@mgoworko, is this ready to review? I don't know why @sscarduzio requested it on your behalf

sscarduzio:
@coutoPL @mgoworko sorry — that review request came from my automation, which saw a green PR with no reviewer and assumed it was waiting. It was not its call to make: the author decides when a PR is ready for review. I withdrew the request; over to you, @mgoworko, whenever it is ready.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  final case class Replacement(table: EsqlIndexTable, newIndices: NonEmptyList[String])
+
+  private implicit val stringMatchable: PatternsMatcher.Matchable[String] =
+    PatternsMatcher.Matchable.caseSensitiveStringMatchable
+
+  def buildReplacements(
+      tables: NonEmptyList[EsqlIndexTable],
+      narrowedFlat: Set[String]
```
coutoPL: It's hard to tell (without reading the implementation) what this is. 
Maybe we should:
1. think about better name of this param
2. maybe the type too

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,124 @@
...
+  ): NonEmptyList[Replacement] = {
+    val (lookupTables, fromTables) = tables.toList.partition(_.origin == Origin.LookupJoin)
```
coutoPL: to catch the potential new Origin implementations, maybe we should do it like this:

```scala
val (lookupTables, fromTables) = tables.toList.partitionMap { table =>                                                                                                                                                               
    table.origin match {                                                                                                                                                                                                               
      case Origin.LookupJoin => Left(table)                                                                                                                                                                                            
      case Origin.From       => Right(table)                                                                                                                                                                                           
    }                                                                                                                                                                                                                                  
  }  
```

WDYT?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    // A name only counts as "lookup-only" if no FROM table itself asked for it - otherwise the same
+    // literal index legitimately requested by both clauses would be wrongly stripped from FROM's own
+    // replacement.
+    val lookupOnlyNames: Set[String] =
```
coutoPL: Let's split the processing of lookup and from tables - let's have two dedicated private methods for them

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,124 @@
...
+      lookupTables.flatMap(_.indices.toList).toCovariantSet -- fromTables.flatMap(_.indices.toList).toCovariantSet
+
+    val nonexistentIndexByOriginal = scala.collection.mutable.Map.empty[String, String]
```
coutoPL: Do we really need the mutable map here? 
I'm not sure if the performance matters here. If not, IMO, it'd be better to stick with the immutable one. The code probably will be more readable.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,124 @@
...
+    // `table.indices` is the table's own *literal query text* (e.g. an alias like "bookshop"), which
+    // can differ from what actually ends up in `narrowedFlat`: ROR's ACL resolves an alias to its
+    // underlying concrete index name(s) whenever the alias isn't itself a literal allowed pattern, so a
+    // plain PatternsMatcher re-match of the query text against narrowedFlat would wrongly find no
+    // candidates and mask a fully-authorized request. `narrowedFlat -- lookupOnlyNames` recovers those
+    // alias-resolved names; the PatternsMatcher term on top still attributes any name a wildcard FROM
+    // genuinely matches (even one primarily requested via a LOOKUP JOIN, e.g. book_* matching
+    // book_prices), which lookupOnlyNames alone would otherwise exclude.
+    val fromReplacements = fromTables.map { table =>
+      val patternMatched = PatternsMatcher.create(table.indices.toList).filter(narrowedFlat)
```
coutoPL: it looks like the PatternsMatcher can be a lazy val in the EsqlIndexTable class, right?

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseXpackApiSuite.scala
```diff
@@ -1277,6 +1282,59 @@ trait BaseXpackApiSuite
...
       }
     }
+    "LOOKUP JOIN is used" should {
+      "be allowed" when {
+        "the LOOKUP JOIN target is authorized" excludeES (allEs6x, allEs7x, allEs8xBelowEs818x) in {
+          val result = dev7EsqlManager.execute(
+            """FROM book_catalog | LOOKUP JOIN book_prices ON book_id | SORT book_id | LIMIT 100"""
+          )
+          result should have statusCode 200
+          result.columnNames should contain allOf ("book_id", "title", "discount_price")
```
coutoPL: I'm not sure about the "allOf" here. IMO, it would be better to use "only" as in the tests (somewhere) above. It applies to this test, and all other tests too.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/base/BaseXpackApiSuite.scala
```diff
@@ -1426,6 +1487,30 @@ object BaseXpackApiSuite {
     indexManager.createAliasOf("bookstore", "bookshop").force()
   }
 
+  private def configureBookPrices(documentManager: DocumentManager, indexManager: IndexManager): Unit = {
+    documentManager.createDocAndAssert(
+      "book_catalog",
```
coutoPL: Named params?

coutoPL (changes requested):
Additional things:
1. changelog entry: use improvement instead of new
2. In Jira, there was a request to check if this is not a security issue - it means if the code before this change didn't allow for an information leak. Did you check it? If there was a leak, we should add a "security" changelog entry too (or instead)
3. Let's generate a PR description with information about what was changed and why (the "how" part is the code)

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Done, switched to partitionMap

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Done, buildLookupReplacements and buildFromReplacements.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: changed to immutable map

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Done, matcher is now a lazy val

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Renamed to authorizedIndices, type changed to `NonEmptyList[RequestedIndex[ClusterIndexName]]` instead of Set[String] and is stringified only when it becomes necessary

mgoworko (commented):
answered

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,162 @@
...
+import tech.beshu.ror.utils.ScalaOps.*
+
+import java.util.regex.{Matcher, Pattern}
+
+private[es] given esqlIndexTableStringMatchable: PatternsMatcher.Matchable[String] =
+  PatternsMatcher.Matchable.caseSensitiveStringMatchable
+
+final case class EsqlIndexTable(
+    tableStringInQuery: String,
+    indices: NonEmptyList[String],
```
coutoPL: At the `EsqlRequestHelper`'s level, we used a string as an index name (it was used like this locally - within the limited scope), but if we moved the logic to the core, we should type it better, IMO. WDYT about using `IndexName` hierarchy here? Or `IndexName.Full` if we are sure there are no wildcards here.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,162 @@
...
+    origin: EsqlIndexTable.Origin
+) {
+  lazy val matcher: PatternsMatcher[String] = PatternsMatcher.create(indices.toList)
+}
+
+object EsqlIndexTable {
+
+  sealed trait Origin extends EnumEntry
```
coutoPL: Do we need enumeratum here?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,162 @@
...
+  ): (Map[String, String], String) = {
+    nonexistentIndexByOriginal.get(originalTableText) match {
+      case Some(existing) => (nonexistentIndexByOriginal, existing)
+      case None           =>
+        val generated = ClusterIndexName.Local.randomNonexistentIndex().stringify
+        (nonexistentIndexByOriginal.updated(originalTableText, generated), generated)
+    }
+  }
+
+  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): String = {
```
coutoPL: A public method should be declared before the private ones

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,162 @@
...
+  sealed trait Origin extends EnumEntry
+
+  object Origin extends Enum[Origin] {
+    case object From extends Origin
+    case object LookupJoin extends Origin
+
+    override val values: IndexedSeq[Origin] = findValues
+  }
+
+  final case class Replacement(table: EsqlIndexTable, newIndices: NonEmptyList[String])
```
coutoPL: WDYT about `IndexName`/`IndexName.Full` instead of `String` here?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,162 @@
...
+import cats.implicits.*
+import enumeratum.{Enum, EnumEntry}
+import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, RequestedIndex}
+import tech.beshu.ror.accesscontrol.matchers.PatternsMatcher
+import tech.beshu.ror.syntax.*
+import tech.beshu.ror.utils.ScalaOps.*
+
+import java.util.regex.{Matcher, Pattern}
+
+private[es] given esqlIndexTableStringMatchable: PatternsMatcher.Matchable[String] =
```
coutoPL: Do we have to have it defined here?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,162 @@
...
+    case object LookupJoin extends Origin
+
+    override val values: IndexedSeq[Origin] = findValues
+  }
+
+  final case class Replacement(table: EsqlIndexTable, newIndices: NonEmptyList[String])
+
+  def buildReplacements(
+      tables: NonEmptyList[EsqlIndexTable],
+      authorizedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
```
coutoPL: In our domain, we name it `allowedIndices`

coutoPL (changes requested):
I stopped the review, because I think we should work a little bit more on proper types in the Esql domain. Until now, we assumed that indices related to ESQL table are strings. IMO, we should improve it because the logic of this part is becoming complicated. First of all we should decide if index related to the ESQL table is defined by Full name or maybe can be a Wildcard. When we model it properly, we will be more sure that the logic in the EsqlndexTable class is right.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Done, additionally split by clause 
- FROM keeps NonEmptyList[RequestedIndex[ClusterIndexName]] (I can use wildcards, exclusions, comma lists and cluster:index)
- LOOKUP JOIN accepts only a single local concrete name or alias there, so it's IndexName.Full

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: This trait is now removed. EsqlIndexTable is now a sealed trait with `From` and `LookupJoin` subtypes.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Done, still more general type `ClusterIndexName` rather than `IndexName.Full`
(because a FROM replacement is not necessarily IndexName.Full, but whatever the ACL resolved)

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: No, removed. It only existed, because the indices were strings

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: done, renamed across the change

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+        .toMap
+
+    tables.map { table =>
+      val newIndices = authorizedNamesByTable(table)
+        .getOrElse(NonEmptyList.one(nonexistentIndexByTableText(table.tableStringInQuery)))
+      Replacement(table, newIndices)
+    }
+  }
+
+  def newQueryFrom(oldQuery: String, replacements: NonEmptyList[Replacement]): EsqlQueryRewriteResult = {
```
coutoPL: private?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+
+  def newQueryFrom(
+      oldQuery: String,
+      tables: NonEmptyList[EsqlIndexTable],
+      allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
+  ): EsqlQueryRewriteResult = {
+    newQueryFrom(oldQuery, buildReplacements(tables, allowedIndices))
+  }
+
+  def buildReplacements(
```
coutoPL: private?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+  object LookupJoin {
+
+    def parse(tableStringInQuery: String): Option[LookupJoin] =
+      ClusterIndexName.fromString(tableStringInQuery).collect { case ClusterIndexName.Local(index: IndexName.Full) =>
+        LookupJoin(tableStringInQuery, index)
+      }
+
+  }
+
+  final case class Replacement(table: EsqlIndexTable, newIndices: NonEmptyList[ClusterIndexName])
```
coutoPL: private?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  def buildReplacements(
+      tables: NonEmptyList[EsqlIndexTable],
+      allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
+  ): NonEmptyList[Replacement] = {
+    val allowedIndexNames: Set[ClusterIndexName] = allowedIndices.includedOnly
+    val fromTables = tables.collect { case table: From => table }
+
+    val lookupOnlyNames: Set[ClusterIndexName] =
```
coutoPL: lookupOnlyIndexNames?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+      tables.collect { case table: LookupJoin => table.clusterIndexName }.toCovariantSet --
+        fromTables.flatMap(_.requestedIndices.includedOnly).toCovariantSet
+
+    val matchedNamesByFromTable: Map[From, Set[ClusterIndexName]] =
+      fromTables.map(table => (table, table.matcher.filter(allowedIndexNames))).toMap
+
+    // an alias the ACL replaced with its backing index matches no FROM pattern, and nothing left in the
+    // request says which table it came from
+    val unattributedNames: Set[ClusterIndexName] =
```
coutoPL: "...IndexNames"?
let's be explicit. Here and below too

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+    * literals and backquoted identifiers, so a source command spelled out inside one cannot be rewritten in
+    * place of the real one. Blanking keeps offsets, so spans found in [[text]] index the original query too.
+    */
+  private final class QueryScan(val text: String, literal: Array[Boolean]) {
+    def spansLiteral(from: Int, until: Int): Boolean = (from until until).exists(literal(_))
+  }
+
+  private object QueryScan {
+
+    def of(query: String): QueryScan = {
```
coutoPL, line 201: Probably to be rewritten in a more declarative way in the follow-up PR.
For now, it's ok, assuming it's well tested.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+      .map(EsqlQueryRewriteResult.Rewritten.apply)
+      .getOrElse(EsqlQueryRewriteResult.CannotRewriteQuery)
+  }
+
+  private def requestedIndicesFrom(tableStringInQuery: String): Option[NonEmptyList[RequestedIndex[ClusterIndexName]]] =
+    NonEmptyList.fromList(
+      tableStringInQuery.split(',').asSafeList.filter(_.nonEmpty).flatMap(RequestedIndex.fromString)
+    )
+
+  private def indexListSpansIn(scan: QueryScan, table: EsqlIndexTable): List[(Int, Int)] = {
```
coutoPL, line 161: Too low level. Hard to grasp. To be "civilized" in the follow-up PR

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+        val spans = indexListSpansIn(scan, table)
+        Option.when(spans.size === tableReplacements.size)(spans.zip(tableReplacements))
+      }
+      .map(_.flatten)
+      .flatMap(spliceIndexLists(oldQuery, _))
+      .map(EsqlQueryRewriteResult.Rewritten.apply)
+      .getOrElse(EsqlQueryRewriteResult.CannotRewriteQuery)
+  }
+
+  private def requestedIndicesFrom(tableStringInQuery: String): Option[NonEmptyList[RequestedIndex[ClusterIndexName]]] =
```
coutoPL: We should move the private method above or below. 
I read `newQueryFrom` and see this method as next, but `indexListSpansIn` is the one I'm interested in next.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+        if (query.charAt(index) != '`') index += 1
+        else if (query.startsWith("``", index)) index += 2
+        else { index += 1; closed = true }
+      }
+      math.min(index, query.length)
+    }
+
+  }
+
+  private def spliceIndexLists(query: String, spans: List[((Int, Int), Replacement)]): Option[String] = {
```
coutoPL: the same comment as [here](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1315/changes#r3847209131)

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+      .groupBy(_.table)
+      .toList
+      .traverse { case (table, tableReplacements) =>
+        val spans = indexListSpansIn(scan, table)
+        Option.when(spans.size === tableReplacements.size)(spans.zip(tableReplacements))
+      }
+      .map(_.flatten)
+      .flatMap(spliceIndexLists(oldQuery, _))
+      .map(EsqlQueryRewriteResult.Rewritten.apply)
+      .getOrElse(EsqlQueryRewriteResult.CannotRewriteQuery)
```
coutoPL: Do we need a human-readable reason in it?

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
...
+    val keywords = table match {
+      case _: From       => "FROM|TS|METRICS"
+      case _: LookupJoin => "LOOKUP\\s+JOIN"
+    }
+    val optionalQuote = "(?:\"\"\"|\")?"
+    val indexList = table.tableStringInQuery
+      .split(',')
+      .map(index => s"$optionalQuote${Pattern.quote(index)}$optionalQuote")
+      .mkString("\\s*,\\s*")
+    Pattern.compile(
```
coutoPL, line 185: We compile a pattern each time - this won't offer great performance for sure. To be optimized in the follow-up PR.

core/src/test/scala/tech/beshu/ror/unit/es/EsqlIndexTableTest.scala
```diff
@@ -0,0 +1,508 @@
...
+        val replacements = EsqlIndexTable.buildReplacements(
+          NonEmptyList.of(fromTable, lookupTable),
+          allowedIndices = authorizedIndicesOf("shared_idx")
+        )
+
+        newIndicesOf(replacements, fromTable) shouldBe NonEmptyList.one("shared_idx")
+        newIndicesOf(replacements, lookupTable) shouldBe NonEmptyList.one("shared_idx")
+      }
+
+    "resolve a literal name shared by a forbidden FROM and a forbidden LOOKUP JOIN to the same " +
```
coutoPL: why "+"? Is it a formatter?  here and below

es90x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/EsqlIndicesEsRequestContext.scala
```diff
@@ -71,35 +72,47 @@ class EsqlIndicesEsRequestContext private (
...
-        request
-      case Right(r @ EsqlRequestClassification.IndicesRelated(tables)) =>
-        val filteredIndicesStrings = filteredIndices.stringify.toCovariantSet
-        if (filteredIndicesStrings != r.indices) {
-          EsqlRequestHelper.modifyIndicesOf(request, tables, filteredIndicesStrings)
-        } else {
-          request
+      case Right(r @ EsqlRequestClassification.IndicesRelated(tables))
+          if filteredRequestedIndices.toList.toCovariantSet != r.requestedIndices =>
+        EsqlRequestHelper.modifyIndicesOf(request, tables, filteredRequestedIndices) match {
```
coutoPL: let's create a new private method for this

es818x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
```diff
@@ -40,10 +42,16 @@ class EsqlRequestHelper(esVersion: EsVersion) {
 
   def modifyIndicesOf(
```
coutoPL, line 47: I have some problems with this method in its current shape. 
It modifies the request but returns the `Rewritten` which can be misleading in the place where it's called ("should I set the new query, or is it already set and I can ignore the result?")

es818x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
```diff
@@ -131,11 +123,18 @@ class EsqlRequestHelper(esVersion: EsVersion) {
...
-      statement.map { s =>
-        NonEmptyList.fromList(indicesFrom(s)) match {
-          case Some(indices) => new IndicesRelatedStatement(statement, indices)
-          case None          => OtherCommand(statement)
-        }
+      statement.flatMap(statementWithIndices)
+    }
+
+    private def statementWithIndices(statement: Any): Either[ClassificationError, Statement] = {
+      Try(indicesFrom(statement)) match {
```
coutoPL, line 134: it looks like some problem was fixed here, right?
What about porting the fix to older (untouched in this PR) modules?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
```diff
@@ -0,0 +1,303 @@
+/*
+ *    This file is part of ReadonlyREST.
+ *
+ *    ReadonlyREST is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    ReadonlyREST is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
```
mgoworko: a

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Yes, I already crafted the second version of the PR to be much more strongly typed and safe. It uses types to express the operations. But the second version is also much larger and rewrites much of the logic. The solution in this PR is simple and it works. But it does not fix the underlying issue of working on raw strings.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Yes, I agree. The second version of the feature already has `IndexListLocator` util and expresses it in type `LocatedIndexList`. I did not port it here on purpose - too much things to change/fix.

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: added human-readable reason

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Yes, already done in util `IndexListReplacer` in the second version of the PR

core/src/main/scala/tech/beshu/ror/es/EsqlIndexTable.scala
mgoworko: Second versions of the PR does not use this regex pattern at all and does not compile patterns on each request. It is a side effect of operating on raw strings

core/src/test/scala/tech/beshu/ror/unit/es/EsqlIndexTableTest.scala
mgoworko: Sorry, I didn't notice it, I think the agent tried this approach with + to bypass the test name being too long. FIxed.

es818x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
mgoworko: It now returns IndicesModified (nothing left for the caller to do) or CannotModifyIndices with reason

es818x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
mgoworko: Ported to 8.11-8.16.

es813x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/EsqlIndicesEsRequestContext.scala
```diff
@@ -71,35 +71,49 @@ class EsqlIndicesEsRequestContext private (
...
-      case Right(r @ EsqlRequestClassification.IndicesRelated(tables)) =>
-        val filteredIndicesStrings = filteredIndices.stringify.toCovariantSet
-        if (filteredIndicesStrings != r.indices) {
-          EsqlRequestHelper.modifyIndicesOf(request, tables, filteredIndicesStrings)
-        } else {
-          request
-        }
+      case Right(r @ EsqlRequestClassification.IndicesRelated(tables))
+          if filteredRequestedIndices.stringify.toCovariantSet != r.indices =>
+        requestWithIndicesNarrowedTo(request, tables, filteredRequestedIndices, filter, fieldLevelSecurity)
```
coutoPL: Shouldn't it be named `updateRequest...`?

coutoPL (approved):
LGTM

coutoPL:
@claude review please

mgoworko:
@claude  re-review please

PR #1304, opened 2026-07-21
[RORDEV-2122][RORDEV-2124][RORDEV-2126] ES 9.4.4, 9.3.8, 8.19.19 support

mgoworko:
## Summary by CodeRabbit

* **New Features**
  * Added support for Elasticsearch 8.19.19, 9.3.8, and 9.4.4.
  * Expanded Elasticsearch artifact packaging to include additional client and core components.

* **Bug Fixes**
  * Improved compatibility across supported Elasticsearch distributions through updated dependency handling.

* **Chores**
  * Extended release checkout timeouts to reduce failures during lengthy release operations.
  * Disabled selected artifact upload commands while release packaging is updated.

PR #1314, opened 2026-08-09
[RORDEV-2150] Bump netty to 4.1.137.Final (CVE-2026-59903)

sscarduzio:
### Changelog entry

```yaml
  - type: security
    components: [es]
    text: "[CVE-2026-59903](https://nvd.nist.gov/vuln/detail/CVE-2026-59903)"
```

Suggested body for the release notes:

> This release addresses a cache-poisoning and information-disclosure vulnerability in Netty's HTTP codec, a dependency used by ReadonlyREST. A remote attacker could exploit incorrect handling of HTTP messages to poison a shared cache or read responses intended for another client. The vulnerability is fixed by updating Netty to 4.1.137.Final.

---

Jira: [RORDEV-2150](https://readonlyrest.atlassian.net/browse/RORDEV-2150)

The netty 4.1.137.Final release (2026-08-06) fixes CVE-2026-59903, a cache-poisoning and information-disclosure issue in `netty-codec-http`. Our OWASP `cve_check` fails on every branch until this lands, including PRs that do not touch dependencies (first seen on #1313).

Patch-level bump of the nine netty pins in `core/build.gradle`. The `netty#15053` SslHandshakeReadTrigger workaround is unaffected by a patch release.

This PR also suppresses four Elasticsearch **server** CVEs (CVE-2026-56144, CVE-2026-56145, CVE-2026-63136, CVE-2026-63140) that the OWASP scanner CPE-matches to the `elasticsearch-rest-client` jars we ship. Per NVD, these affect the Elasticsearch server's ingest-simulation and search paths, not the REST client, and the ES server jar is excluded from `distJars` — only the client ships. The entries are appended to the existing `elasticsearch-rest-client` suppression block, whose note already documents this rationale for ~30 earlier CVEs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1316, opened 2026-08-09
[RORDEV-2152] Dead-wood batch: delete dead scripts/plugins/build plumbing, drop jitpack, CI hygiene

sscarduzio:
Jira: [RORDEV-2152](https://readonlyrest.atlassian.net/browse/RORDEV-2152)

_No changelog entry: this change has no client-visible effect. The shipped plugin is unchanged (source sets, declared dependencies and build outputs are all verified identical)._

---

Implements the "proceed now" tier of the simplification audit (items 2, 5-commit-1, 6-PR-A, 7a/b/c), with every amendment from the adversarial challenge applied. Four commits, ~2,100 lines deleted, zero behavior change intended.

## 1. Dead scripts, plugins and build plumbing
Travis-era S3 helpers, Python-2 `genHash.py` (imports `crypt`, removed in Python 3.13), `RunCommandCombiner`, the `ldapserver` task + `LDAPServer.java` (task points at a package the class left in 2019 — broken ~7 years; `LdapContainer` stays), two 2023 playground dirs with checked-in test `.p12`s, the bmuschko + checksum plugins (taskless since #1279; both `docker { registryCredentials }` no-op blocks removed atomically with the plugin id), the orphan `wagon` configuration, and ror-managed-libs' dead-on-arrival `compile.extendsFrom`.

**Kept on purpose** (challenge findings): `ci/upload-es-artifacts.sh` + the `es_s3_up` job — a live toggle last used in #1304 for release-day Maven Central lag — and `ci/probe-host-disk.sh` (live `disk_probe` consumer).

## 2. Dead scalaCompilerPlugin machinery
The configuration fed macro paradise to Scala 2.12 (2020); empty since #896 (2023), unrevivable since Scala 3 (#1010). Deleted in core/tests-utils/integration-tests — each module's real scalac flag list is unchanged and `-Xfatal-warnings` makes that self-verifying. Also deleted: core's no-op `parameters -= ["-target:8"]` and the inert `compileScala { options.compilerArgs }` blocks in audit/ror-tools/ror-tools-core (Java options on a Scala task — proven inert by audit's 2.11/2.12 crossbuild never failing on Scala-3-only flags). core's sibling `all*.exclude` rules are untouched.

## 3. jitpack.io removed from the supply chain
Added 2017 for a testcontainers snapshot; need died the same year. Every `com.github.*` coordinate verified on Maven Central (HTTP 200 per pom): better-files 3.9.2, caffeine 3.2.3, scala-csv 2.0.0, wiremock-jre8 2.35.2, scopt 4.1.0.

## 4. CI hygiene
- **Toolchains tag → YAML anchor** (11 literals → 1). Deliberately *not* a repo variable — the tag is coupled to the commit's baked dependency cache, and branch-atomic Gradle upgrades / bisects must keep working. Drift guard kept as a belt.
- **actionstrings_gen.yml**: checkout v4; `|| true` off the push (failed cross-repo pushes went green and docs silently stopped updating) but kept on the commit; `concurrency` group; paths filter + `workflow_dispatch`.
- **`defaultEsModule` derived via `EsModuleFinder`** instead of hardcoded `es94x` — the resolver the old comment was waiting for landed in #1279; the default now auto-flips per new es module. CI unaffected (`run-pipeline.sh` always passes `-PesModule`).

## Verification
- `./gradlew help` — every project configures.
- `compileScala` green for core, tests-utils, ror-tools, ror-tools-core, audit + `compileTestScala` for integration-tests (exercises the EsModuleFinder wiring at configuration time).
- Both workflow YAMLs parse; the anchor resolves to the same image string in all 10 jobs.
- All five com.github poms probed on repo1.maven.org (200).

## Pre-merge checklist (from the challenge's Group D)
- [ ] Ping support on Slack: the Elastic-Cloud POC recipe moves to git history (`docker-envs/misc-clusters/ror-with-elastic-cloud-poc`).
- [ ] @coutoPL sign-off on the two CI-file changes (toolchains anchor, actionstrings_gen) — recent rulings in this area are yours.

## Summary by CodeRabbit

- **CI/CD Improvements**
  - Improved workflow triggering and run coordination, including manual runs.
  - Updated build infrastructure components for more reliable automation.
  - Push failures are now reported instead of silently ignored.
  - Standardized build environments across verification, testing, release, and publishing.

- **Maintenance**
  - Simplified build and Scala compilation configuration.
  - Updated the Netty library version.
  - Added security advisories for known dependency vulnerabilities.

- **Removed**
  - Retired legacy manual performance-test and proof-of-concept environments, scripts, configurations, and utilities.


.github/workflows/actionstrings_gen.yml
coutoPL, line 1: Image

It looks like the action name doesn't follow the convention (see the sceenshot)

BTW, do we use the pages-build-deployment?

.github/workflows/actionstrings_gen.yml
```diff
@@ -2,31 +2,41 @@ on:
...
+concurrency:
+  group: actionstrings
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout ror es
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
```
coutoPL, line 21: maybe we should pin the specific version?

.github/workflows/actionstrings_gen.yml
```diff
@@ -2,31 +2,41 @@ on:
...
           mv docs main/readonlyrest-docs
-          cd main 
+          cd main
           git config --global advice.detachedHead false
           bash -x ci/actionstrings/run.sh
           cd readonlyrest-docs
           git config user.name github-actions
           git config user.email github-actions@github.com
           git add .
+          # `|| true` on commit only: an empty diff is a normal outcome. A FAILED PUSH IS NOT --
```
coutoPL: unnecessary/wrong comment related to the change, not the final code

audit/build.gradle
```diff
@@ -68,21 +68,6 @@ test {
     }
 }
 
-compileScala {
```
coutoPL, line 71: @mgoworko do you confirm we can remove it?

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -51,13 +51,6 @@ def buildxPushArgs = { List<String> tags ->
     cmd + ['--output', 'type=image,push=true,rewrite-timestamp=true', '.']
 }
 
-docker {
```
coutoPL, line 54: what about local development?

docker-envs/misc-clusters/performance-manual-tests/certs/elastic-certificates.p12
coutoPL, line 1: IMO, we can remove all `docker-envs`. They are not used these days. ROR Sandbox took their role.

integration-tests/build.gradle
```diff
@@ -56,15 +41,17 @@ sourceSets {
     }
 }
 
-// No -PesModule → newest supported module. Hardcoded until buildSrc's EsModuleResolver lands
-// and can derive it automatically.
-def defaultEsModule = 'es94x'
+// No -PesModule → the newest supported module, derived from the es* modules' supportedEsVersions
+// (same resolver RorTaskFinder uses). Auto-flips when a newer es module is added.
+def defaultEsModule = tech.beshu.ror.gradle.utils.EsModuleFinder
```
coutoPL, line 48: what about `findTheNewestSupportedEsVersion` method?

ror-tools-core/build.gradle
```diff
@@ -54,21 +53,6 @@ java {
     }
 }
 
-compileScala {
```
coutoPL, line 57: @mgoworko and here?

ror-tools/build.gradle
```diff
@@ -83,21 +82,6 @@ java {
     }
 }
 
-compileScala {
```
coutoPL, line 86: @mgoworko and here?

suppressions_cve.xml
```diff
@@ -69,6 +69,10 @@ False positive: CVE-2021-34364 is for "Refined GitHub" (browser extension), not
         <cve>CVE-2026-56148</cve>
         <cve>CVE-2026-56149</cve>
         <cve>CVE-2026-63144</cve>
+        <cve>CVE-2026-56144</cve>
```
coutoPL, line 72: this should be done in separate PR

coutoPL (changes requested):

.github/workflows/actionstrings_gen.yml
sscarduzio: Fixed — added `name: ROR ES Action Strings`. It was the only workflow without a `name:` field, which is why it showed as a path.

On pages-build-deployment: the repo has no GitHub Pages site configured (`GET /repos/.../pages` returns 404), so that workflow is not ours and nothing here depends on it.

.github/workflows/actionstrings_gen.yml
sscarduzio: The repo convention today is major tags — `ci.yml` uses `@v4` for checkout, cache, setup-java, upload-artifact and setup-gradle — so `@v4` here matches. Happy to pin everything to commit SHAs, but I would rather do that repo-wide in its own PR than leave one workflow inconsistent with the rest.

.github/workflows/actionstrings_gen.yml
sscarduzio: Good catch, fixed. The comment now describes only the current behaviour: `|| true` stays on the commit (an empty diff is normal), and the push has no `|| true`, so a failed push fails the job.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
sscarduzio: Local development is unaffected. That block was already a no-op: it configured the `com.bmuschko.docker-remote-api` extension, whose tasks were removed in #1279, and it is deleted here together with the plugin id. `localRorDockerImage` runs `buildx build --load` without a push, so it needs no credentials. Pushes (`pushRorDockerImage`, `pushRorPreBuildDockerImage`) authenticate with `docker login` in CI, which is what already happens today.

integration-tests/build.gradle
sscarduzio: Switched to it, thanks. It returns the newest ES *version*, so it now pairs with `findEsModuleFor(...)` to get the module. Verified the result: `newestEsVersion=9.5.0` -> `defaultEsModule=es94x`, i.e. the same value as the old hardcoded constant.

suppressions_cve.xml
sscarduzio: They are in a separate PR — #1314, together with the netty bump that the same CVE scan requires. They only appear in this diff because I merged that branch into this one so CI could pass here. They disappear from this PR once #1314 merges.

On the CodeRabbit comment above: its own NVD lookups confirm these are Elasticsearch *server* CVEs that do not affect `elasticsearch-rest-client`, which is exactly why they are suppressed. The suppression is appended to the existing `elasticsearch-rest-client` block whose note already states that rationale for ~30 earlier CVEs. Upgrading ES is not an option here: each es* module targets a fixed ES version, and the ES server jar is excluded from `distJars` anyway — only the rest client ships.

docker-envs/misc-clusters/performance-manual-tests/certs/elastic-certificates.p12
sscarduzio: Agreed in principle, but it needs one decision first, so I have left `docker-envs` alone in this PR apart from the two dirs already deleted here.

Evidence it is unused: the only reference anywhere in the repo is `development.md:55`, and it points at `docker-envs/build-ror-in-docker/builds` — a path that does not exist.

The catch: the plan for replacing `eshome` (which cannot run ES >= 8.6) was to point macOS/Windows developers at `docker-envs/base-ror-docker`, because `ror-tools:downloadEsBinaries` is hardcoded to linux-x86_64. If all of `docker-envs` goes, that replacement needs a different fallback — ROR Sandbox may well be the answer, but I did not want to assume it. Shall I delete the whole `docker-envs` tree in the eshome PR, where both halves can be decided together?

audit/build.gradle
sscarduzio: For @mgoworko, the evidence that these three blocks are inert: they call `options.compilerArgs`, which is the **Java** compiler option list on the ScalaCompile task, so scalac never receives these flags. The proof is that this list contains Scala-3-only flags while the `audit` module still cross-compiles for Scala 2.11 and 2.12 — if the flags reached scalac, those builds would fail.

The flags that do apply in these modules are the `scalaCompileOptions.additionalParameters` lines, which are untouched.

sscarduzio:
Thanks for the review — pushed 6dd97cb39 with the three code fixes (workflow `name:`, the comment rewrite, and `findTheNewestSupportedEsVersion` + `findEsModuleFor` for the default ES module, verified to resolve to es94x as before). Replied in each thread for the questions.

Two things still need someone else:
- `@mgoworko` to confirm the three inert `compileScala { options.compilerArgs }` blocks (audit, ror-tools, ror-tools-core) — rationale in that thread.
- Your call on deleting the whole `docker-envs` tree; I suggested folding it into the eshome PR so it and the macOS/Windows fallback are decided together.

.github/workflows/actionstrings_gen.yml
coutoPL: sure

docker-envs/misc-clusters/performance-manual-tests/certs/elastic-certificates.p12
coutoPL: ok, agreed

coutoPL (approved):
LGTM!

Before merge, let's @mgoworko confirm the 3 outstanding comments.

audit/build.gradle
mgoworko: Yes, I think all those sections can be removed. I've checked and they indeed seem unused. 

Those sections have existed since before the Scala 2.13->Scala 3 upgrade and were not cleaned then. There is even an old task in the backlog for cleaning and aligning the Scala compiler options (RORDEV-1469)

sscarduzio: Thanks @mgoworko — that matches what I found. For the record, the mechanism: those blocks call `options.compilerArgs`, which is the **Java** compiler option list on a ScalaCompile task, so scalac never saw the flags. The proof is that the list contains Scala-3-only flags while `audit` still cross-compiles for Scala 2.11/2.12 — had they reached scalac, those builds would have failed.

The flags that do apply in these modules are the `scalaCompileOptions.additionalParameters` lines, which stay untouched. If you have the backlog task number, we can close it against this PR.

PR #1317, opened 2026-08-10
[RORDEV-2153] Action strings sync: read the ES versions from the build

sscarduzio:
Jira: [RORDEV-2153](https://readonlyrest.atlassian.net/browse/RORDEV-2153)

_No changelog entry: CI-only change, the shipped plugin is unchanged._

> ⚠️ Stacked on #1316 (it also touches `actionstrings_gen.yml`). Review/merge that one first; this diff is the two commits on top.

## The bug

`ci/actionstrings/run.sh` extracted ES versions with `awk -F= '{print $2}'` over the whole of each module's `gradle.properties`. That worked when the file held a single `esVersion=X.Y.Z` line. It now holds a multi-line `supportedEsVersions` list, so the script extracts **a backslash** instead of a version:

```
$ awk -F= '{print $2}' es67x/gradle.properties
\
```

It reached production. `beshu-tech/readonlyrest-docs/actionstrings/` contains a 0-byte file literally named **`action_strings_es\.txt`**. Nobody noticed because the workflow ended in `git push || true`.

## The fix

**The build is the single source of truth.** New `printAllSupportedEsVersions` task prints every supported ES version from each module's `supportedEsVersions`; `run.sh` asks for that instead of parsing properties itself. It returns **256 versions** where the old parser returned `\`.

Three guards so this class of failure cannot be silent again:
1. `run.sh` rejects any token that is not an ES version, and fails if the task returns nothing.
2. `fetchIfNecessary.sh` writes to a temp file and publishes **only a non-empty result**. Previously a failed extraction committed an *empty* list, and the "file exists" check then treated that version as done forever — that is exactly how the junk file became permanent.
3. `fetchIfNecessary.sh` used `return` outside a function (an error in an executed script); now `exit`.

Also: the workflow gets a JDK (run.sh calls Gradle now), and the `ror-release` skill's step 2 — which told the reader to edit `ci/supported-es-versions/es{N}x.txt`, **a directory that no longer exists**, and called it the thing "CI actually keys off" — is corrected to point at `supportedEsVersions`.

## Verification

Sandbox tests with a stubbed `gradlew` (no network, no commits):

| Case | Result |
|---|---|
| Task returns `\` (the exact old output) | rejected, exit 1, **no file created** |
| Task returns nothing | fails loudly, exit 1 |
| ES tag does not exist (`v0.0.0`) | exit 1, **no empty file published** |
| Action-string file already exists | skipped, exit 0 |

Plus `./gradlew printAllSupportedEsVersions` → 256 lines, **0** not matching `^\d+\.\d+\.\d+(-\w+)?$`.

## Cleanup — done

The junk file `actionstrings/action_strings_es\.txt` has been **deleted** from `beshu-tech/readonlyrest-docs` (blob SHA was `e69de29b…`, git's hash for an empty file). 278 valid action-string files remain.

It cannot come back: with the guards in this PR, a non-version token never reaches the fetch step, and an empty extraction is never published.

## Summary by CodeRabbit

* **New Features**
  * Added a Gradle task to display all supported Elasticsearch versions in order.
  * Supported Elasticsearch versions are now consistently derived across builds, CI, and documentation.

* **Bug Fixes**
  * Improved action-string generation reliability with safer extraction, cleanup, and error reporting.
  * Generation now continues across individual fetch failures and clearly reports unsuccessful versions.

* **Chores**
  * Added manual workflow execution and improved workflow triggering and run management.
  * Updated build automation to use Java 17 and newer checkout tooling.

### DIFF
--- a/.github/workflows/actionstrings_gen.yml
+++ b/.github/workflows/actionstrings_gen.yml
@@ -29,6 +29,13 @@ jobs:
           repository: beshu-tech/readonlyrest-docs
           token: ${{ secrets.GH_PAT }}
 
+      # run.sh asks the build for the supported ES versions, so this job needs a JDK.
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+
       - run: |
           mv docs main/readonlyrest-docs
           cd main
--- a/.github/workflows/actionstrings_gen.yml
+++ b/.github/workflows/actionstrings_gen.yml
@@ -51,8 +51,12 @@ jobs:
           set -e
           cd readonlyrest-docs
           git add .
-          # `|| true` on the commit only: an empty diff is a normal outcome. The push has no
-          # `|| true`, so a failed push to the docs repo fails this job.
-          git commit -m "commit new action strings" || true
+          # Commit only when there is something staged. `git commit || true` would also swallow a REAL
+          # failure (a hook, an index lock) and report success while publishing nothing.
+          if git diff --cached --quiet; then
+            echo "no new action strings to commit"
+          else
+            git commit -m "commit new action strings"
+          fi
           git push
           exit $generated
--- a/.github/workflows/actionstrings_gen.yml
+++ b/.github/workflows/actionstrings_gen.yml
@@ -40,12 +40,19 @@ jobs:
           mv docs main/readonlyrest-docs
           cd main
           git config --global advice.detachedHead false
+          # Set the commit identity BEFORE generating, and in the docs repo — that is the repo we commit to.
+          git -C readonlyrest-docs config user.name github-actions
+          git -C readonlyrest-docs config user.email github-actions@github.com
+          # Generate first and remember the outcome, then publish whatever was generated: one ES version
+          # that fails must not discard the versions that succeeded. The job still fails afterwards.
+          set +e
           bash -x ci/actionstrings/run.sh
+          generated=$?
+          set -e
           cd readonlyrest-docs
-          git config user.name github-actions
-          git config user.email github-actions@github.com
           git add .
           # `|| true` on the commit only: an empty diff is a normal outcome. The push has no
           # `|| true`, so a failed push to the docs repo fails this job.
           git commit -m "commit new action strings" || true
           git push
+          exit $generated
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,23 @@ tasks.register("publishEsRorPreBuildDockerImage", RorTaskFinder) {
 // Usage: ./gradlew printEsModules -PesMajor=8 --quiet
 tasks.register("printEsModules", PrintEsModulesTask)
 
+// Prints every ES version ROR supports, one per line, oldest first. The versions come from each es module's
+// `supportedEsVersions` (gradle.properties), so scripts that need the full list cannot drift from what CI
+// actually builds and tests. Invoke with --quiet to suppress Gradle's own output.
+// Usage: ./gradlew printAllSupportedEsVersions --quiet
+tasks.register("printAllSupportedEsVersions") {
+    group = 'help'
+    description = 'Prints every supported ES version (one per line, oldest first)'
+    outputs.upToDateWhen { false }
+    doLast {
+        tech.beshu.ror.gradle.utils.EsModuleFinder
+                .sortedEsModules(rootProject, tech.beshu.ror.gradle.utils.EsModuleFinder.newestEsVersionComparator())
+                .collectMany { tech.beshu.ror.gradle.utils.EsVersions.of(it).all }
+                .unique()
+                .each { println it }
+    }
+}
+
 group = 'tech.beshu.ror'
 version = pluginVersion
 
--- a/build.gradle
+++ b/build.gradle
@@ -70,11 +70,7 @@ tasks.register("printAllSupportedEsVersions") {
     description = 'Prints every supported ES version (one per line, oldest first)'
     outputs.upToDateWhen { false }
     doLast {
-        tech.beshu.ror.gradle.utils.EsModuleFinder
-                .sortedEsModules(rootProject, tech.beshu.ror.gradle.utils.EsModuleFinder.newestEsVersionComparator())
-                .collectMany { tech.beshu.ror.gradle.utils.EsVersions.of(it).all }
-                .unique()
-                .each { println it }
+        tech.beshu.ror.gradle.utils.EsModuleFinder.allSupportedEsVersions(rootProject).each { println it }
     }
 }
 
--- a/ci/actionstrings/fetchIfNecessary.sh
+++ b/ci/actionstrings/fetchIfNecessary.sh
@@ -1,18 +1,42 @@
+#!/bin/bash
+set -euo pipefail
+
+# Generates the action-string list for ONE ES version and commits it to the readonlyrest-docs checkout,
+# unless that version's file already exists. Called by run.sh.
+
 ES_VERSION=$1
 
 DEST_DIR="readonlyrest-docs/actionstrings"
-mkdir -p $DEST_DIR || echo "$DEST_DIR already present."
+mkdir -p "$DEST_DIR"
 
 FILENAME="$DEST_DIR/action_strings_es$ES_VERSION.txt"
-if test -f $FILENAME; then
+if test -f "$FILENAME"; then
   echo "$FILENAME exists."
-  return 0
+  exit 0
 fi
 
 echo "processing ES action extractions for: $FILENAME"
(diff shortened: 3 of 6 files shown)


ci/actionstrings/fetchIfNecessary.sh
```diff
@@ -1,18 +1,42 @@
...
+  exit 1
+fi
+
+mv "$TMP_FILE" "$FILENAME"
+trap - EXIT
+cat "$FILENAME"
+
+git add "$FILENAME"
+git commit "$FILENAME" -m "adding actions list for ES $ES_VERSION"
+git push
```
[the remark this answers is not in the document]
sscarduzio: Correct, and worse than described — fixed in `8e1f7568f`.

I verified it locally with a nested-repo sandbox:

- `git add <path inside a nested repo>` **stages nothing and returns 0** (git ignores paths inside an embedded repository)
- `git commit <that path>` fails with `did not match any file(s) known to git` and **returns 1**

So the per-file git block never worked; the workflow's own trailing commit has always been what published. And because this branch adds `set -euo pipefail`, that failing commit would have aborted the script for **every newly generated version** — a regression I introduced. Good catch.

Fix taken slightly differently from the suggestion: rather than repairing the block with `git -C`, I removed it. The workflow already publishes everything in **one** commit in the docs repo, so per-file commits are redundant. The workflow now also:
- sets the commit identity **before** generation, in the docs repo (your second point), and
- keeps run.sh's exit code, publishes what was generated, then exits with that code — so one failed ES version cannot discard the versions that succeeded, while the job still fails.

ci/actionstrings/fetchIfNecessary.sh
```diff
@@ -1,18 +1,42 @@
...
+if ! ci/actionstrings/fetch.sh "v$ES_VERSION" > "$TMP_FILE"; then
+  echo "extraction failed for ES $ES_VERSION"
+  exit 1
+fi
+if [ ! -s "$TMP_FILE" ]; then
+  echo "no action strings found for ES $ES_VERSION (is the tag v$ES_VERSION published?)"
+  exit 1
+fi
+
+mv "$TMP_FILE" "$FILENAME"
```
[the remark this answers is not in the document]
sscarduzio: Keeping `mktemp` in `TMPDIR` deliberately. `mv` across filesystems copies rather than renames, which is correct here — the file is small and the operation still succeeds.

The reason not to create it in the destination is the opposite risk: the workflow publishes with `git add .` in the docs repo, so a temp file left behind by a killed script would be **committed**. `TMPDIR` keeps that impossible.

.github/workflows/actionstrings_gen.yml
```diff
@@ -1,32 +1,58 @@
...
           bash -x ci/actionstrings/run.sh
+          generated=$?
+          set -e
           cd readonlyrest-docs
-          git config user.name github-actions
-          git config user.email github-actions@github.com
           git add .
+          # `|| true` on the commit only: an empty diff is a normal outcome. The push has no
+          # `|| true`, so a failed push to the docs repo fails this job.
           git commit -m "commit new action strings" || true
```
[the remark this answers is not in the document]
sscarduzio: Right — fixed. `|| true` was covering the empty-diff case, but it also swallowed a real failure (a hook, an index lock), which would report success while publishing nothing.

The workflow now commits only when something is staged:

```bash
git add .
if git diff --cached --quiet; then
  echo "no new action strings to commit"
else
  git commit -m "commit new action strings"
fi
git push
```

So an empty diff is a normal, silent outcome, and a failing commit fails the job.

audit/build.gradle
```diff
@@ -68,21 +68,6 @@ test {
     }
 }
 
-compileScala {
```
coutoPL: How is this related to the PR?!
I found this code in another PR. Don't mix different things in every PR context!

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -51,13 +51,6 @@ def buildxPushArgs = { List<String> tags ->
     cmd + ['--output', 'type=image,push=true,rewrite-timestamp=true', '.']
 }
 
-docker {
```
coutoPL: the same here

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
```diff
@@ -92,7 +92,6 @@ java {
 }
 
 configurations {
-    wagon
```
coutoPL: and here

core/build.gradle
```diff
@@ -80,9 +79,9 @@ dependencies {
     // ES 9.1+ (auto-read disabled, elastic/elasticsearch#127817) affects the TLS-over-HTTP read flow. 4.1.136
     // (netty#15053) required the SslHandshakeReadTrigger workaround used by es{version}x SSLNetty4HttpServerTransport
     // (tech.beshu.ror.utils.SslHandshakeReadTrigger) — re-verify that path before/after changing this version.
-    api group: 'io.netty',                          name: 'netty-codec-http2',                   version: '4.1.136.Final'
-    api group: 'io.netty',                          name: 'netty-handler',                       version: '4.1.136.Final'
-    api group: 'io.netty',                          name: 'netty-transport',                     version: '4.1.136.Final'
+    api group: 'io.netty',                          name: 'netty-codec-http2',                   version: '4.1.137.Final'
```
coutoPL, line 82: !!!

coutoPL (changes requested):
rebase to develop and get back to me with ready to rereview PR

audit/build.gradle
sscarduzio: You are right, and this was my mistake — not a stray change. I stacked this branch on the parent PR to keep CI green, but left it targeting `develop`, so GitHub showed the parent's files as if they were mine.

Now that the parent is merged, I rebased this branch onto the new `develop`. The diff is **5 files** instead of 49, and only the action-strings work remains. Sorry for the noise.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
sscarduzio: You are right, and this was my mistake — not a stray change. I stacked this branch on the parent PR to keep CI green, but left it targeting `develop`, so GitHub showed the parent's files as if they were mine.

Now that the parent is merged, I rebased this branch onto the new `develop`. The diff is **5 files** instead of 49, and only the action-strings work remains. Sorry for the noise.

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
sscarduzio: You are right, and this was my mistake — not a stray change. I stacked this branch on the parent PR to keep CI green, but left it targeting `develop`, so GitHub showed the parent's files as if they were mine.

Now that the parent is merged, I rebased this branch onto the new `develop`. The diff is **5 files** instead of 49, and only the action-strings work remains. Sorry for the noise.

core/build.gradle
sscarduzio: You are right, and this was my mistake — not a stray change. I stacked this branch on the parent PR to keep CI green, but left it targeting `develop`, so GitHub showed the parent's files as if they were mine.

Now that the parent is merged, I rebased this branch onto the new `develop`. The diff is **5 files** instead of 49, and only the action-strings work remains. Sorry for the noise.

build.gradle
```diff
@@ -61,6 +61,23 @@ tasks.register("publishEsRorPreBuildDockerImage", RorTaskFinder) {
...
+// Prints every ES version ROR supports, one per line, oldest first. The versions come from each es module's
+// `supportedEsVersions` (gradle.properties), so scripts that need the full list cannot drift from what CI
+// actually builds and tests. Invoke with --quiet to suppress Gradle's own output.
+// Usage: ./gradlew printAllSupportedEsVersions --quiet
+tasks.register("printAllSupportedEsVersions") {
+    group = 'help'
+    description = 'Prints every supported ES version (one per line, oldest first)'
+    outputs.upToDateWhen { false }
+    doLast {
+        tech.beshu.ror.gradle.utils.EsModuleFinder
```
coutoPL: Let's have the proper function in the EsModuleFinder util

coutoPL (changes requested):

build.gradle
sscarduzio: Done — `EsModuleFinder.allSupportedEsVersions(rootProject)` is now the util function, and the Gradle task is just:

```groovy
tech.beshu.ror.gradle.utils.EsModuleFinder.allSupportedEsVersions(rootProject).each { println it }
```

Verified unchanged output: 256 versions, none malformed.

coutoPL (approved):
LGTM

PR #1318, opened 2026-08-10
[RORDEV-2154] Check the PR conventions automatically

sscarduzio:
Jira: [RORDEV-2154](https://readonlyrest.atlassian.net/browse/RORDEV-2154)

_No changelog entry: CI-only change, the shipped plugin is unchanged._

## Why

Your review of #1314 was, in full: *"Every code change that has visible/meaningful effect for clients should have a jira and the description should have the changelog style entry added at the top."* That is a regex and a file check costing a review round-trip. This makes CI do it, so review attention goes to design only.

## What

- **`ci/check-pr-conventions.sh`** — the logic lives in a script, so it runs locally *before* opening a PR: `ci/check-pr-conventions.sh 1234`
- **`.github/workflows/pr-conventions.yml`** — a thin wrapper, on every non-draft PR (open / edit / label / push).

**Check 1 — Jira key in the title:** `[RORDEV-1234] Short summary`.

**Check 2 — changelog entry**, required when the PR changes production sources **or a dependency declaration**. Dependencies count because a changed dependency ships in the plugin zip and is the most common changelog entry we write — a CVE fix. Only `api`/`implementation`/`compileOnly`/… lines match, so touching a `plugins {}` block or compiler flags does not ask for a changelog.

Both documented forms are accepted — the YAML of `readonlyrest-docs/changelog/<version>.yaml` **and** the emoji phrase from the `ror-dev-process` skill (`🚨**Security Fix** (ES) …`).

**Escape hatches:** the `no-jira` and `no-changelog` labels, for what a script cannot judge (refactors, build changes, tests). Both labels are created. A reviewer should never have to ask for these again.

## Verification — run against the four live PRs

| PR | Result |
|---|---|
| #1314 (netty CVE) | **passes** on its changelog entry — and correctly *required* one, via the dependency rule |
| #1313, #1316, #1317 | correctly **asked** for a changelog until labelled `no-changelog` (they are refactors) |

Title regex, table-tested: `[RORDEV-2154] x` passes; bare titles, `[ci] …`, `[RORDEV-]`, a key with no summary, and a key not at the start are all rejected.

## Also

Corrects a stale line in the `ror-dev-process` skill that named `ci/supported-es-versions/es{N}x.txt` — **a directory that no longer exists** — as the thing gating CI build/test. (#1317 fixes the same claim in the `ror-release` skill.)

## Note for the maintainer

Whether to make `conventions` a **required** check is a repo-settings decision — I have not touched branch protection.

## Summary by CodeRabbit

* **New Features**
  * Added automated convention checks for non-draft pull requests.
  * Pull request titles are validated for the required issue-key format, with an approved bypass option.
  * Production and dependency changes are checked for required changelog entries, with a bypass option.
  * Checks continue running to provide consolidated validation results.

* **Documentation**
  * Updated guidance for supported ES versions and pull request convention validation.


.github/workflows/pr-conventions.yml
```diff
@@ -0,0 +1,31 @@
+name: PR Conventions
+
+# Checks the conventions a reviewer used to check by hand: the Jira key in the title, and a changelog
+# entry when the change can reach a client. The logic lives in ci/check-pr-conventions.sh so it can be
+# run locally before opening the pull request.
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize, labeled, unlabeled, ready_for_review]
```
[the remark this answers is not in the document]
sscarduzio: Fair for a policy gate — hardened in `7c04a81b5`.

The job now checks out the **base branch's** copy of the script (`ref: base.sha`, sparse-checkout of just that one file); everything else it needs comes from the API, so no pull-request code runs. A PR can no longer edit the rule it is judged by.

There is a `not on the base branch yet` guard, which is only ever true for this pull request — the one introducing the check.

ci/check-pr-conventions.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# --- 2. Changelog entry for production changes --------------------------------------------------
+production_changed=$(grep -cE "$PRODUCTION_PATHS" <<<"$files" || true)
+dependencies_changed=$(grep -cE "$DEPENDENCY_LINE" <<<"$dependency_diff" || true)
+if [ "$production_changed" -eq 0 ] && [ "$dependencies_changed" -eq 0 ]; then
+  pass "changelog: not needed (no production sources or dependencies changed)"
+elif has_label no-changelog; then
+  pass "changelog: skipped ('no-changelog' label)"
+elif { grep -q 'type:' <<<"$body" && grep -q 'components:' <<<"$body" && grep -q 'text:' <<<"$body"; } \
+  || grep -qE '\*\*(Security Fix|New|Enhancement|Fix)\*\*[[:space:]]*\((ES|KBN)\)' <<<"$body"; then
```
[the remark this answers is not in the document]
sscarduzio: Right, fixed in `7c04a81b5`. It now requires a complete entry: a `type:` line with a real type (`security | new | fix | enhancement`), followed within three lines by `components:` and `text:`.

Tested against your decoy:

| description | before | now |
|---|---|---|
| real YAML entry | accept | accept |
| three stray unrelated lines (`type: whatever` … `components: unrelated` … `text: nope`) | accept | **reject** |
| bare `- type: fix` with nothing after | accept | **reject** |
| `- type: banana` + valid keys | accept | **reject** |

All five open PRs give the same verdicts as before.

core/build.gradle
```diff
@@ -80,9 +80,9 @@ dependencies {
     // ES 9.1+ (auto-read disabled, elastic/elasticsearch#127817) affects the TLS-over-HTTP read flow. 4.1.136
     // (netty#15053) required the SslHandshakeReadTrigger workaround used by es{version}x SSLNetty4HttpServerTransport
     // (tech.beshu.ror.utils.SslHandshakeReadTrigger) — re-verify that path before/after changing this version.
-    api group: 'io.netty',                          name: 'netty-codec-http2',                   version: '4.1.136.Final'
-    api group: 'io.netty',                          name: 'netty-handler',                       version: '4.1.136.Final'
-    api group: 'io.netty',                          name: 'netty-transport',                     version: '4.1.136.Final'
+    api group: 'io.netty',                          name: 'netty-codec-http2',                   version: '4.1.137.Final'
+    api group: 'io.netty',                          name: 'netty-handler',                       version: '4.1.137.Final'
+    api group: 'io.netty',                          name: 'netty-transport',                     version: '4.1.137.Final'
```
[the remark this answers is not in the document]
sscarduzio: Reasonable flag — `core/build.gradle` says in its own comment to re-verify that path on a netty change, so here is the evidence rather than an assurance.

That TLS path **is** exercised by the PR matrix and is green with 4.1.137:

- `SSLNetty4HttpServerTransport` (which wires `SslHandshakeReadTrigger`) lives in `es9-base`, so `es90x` and `es94x` — both in the PR integration matrix — compile and run against it.
- The suites include `XpackClusterWithRorNodesAndInternodeSslKeystoreSuite`, `XpackClusterWithRorNodesAndInternodeSslPemSuite` and `FipsSslSuite`; sharding runs the full suite set on every leg.
- Those legs passed green on four separate PRs carrying the bump (#1313, #1314, #1316, #1317).

As you note, 4.1.137 is still above the 4.1.136 boundary, so the workaround remains required and remains wired — the bump does not cross the boundary in either direction. `es91x`/`es92x` are outside the PR subset but are covered by the full 34-leg matrix on the merge to develop.

ci/check-pr-conventions.sh
```diff
@@ -0,0 +1,85 @@
...
+  pass "changelog: skipped ('no-changelog' label)"
+# Both documented forms count: the YAML entry of readonlyrest-docs/changelog/<version>.yaml, and the
+# emoji phrase of the `ror-dev-process` skill (which is how the YAML renders in changelog.md).
+# The YAML form asks for a complete entry — a `type:` line with a real type, then `components:` and
+# `text:` within the following lines — so that three unrelated lines elsewhere in the description
+# cannot pass the check by accident.
+elif awk '
+      /^[[:space:]]*-?[[:space:]]*type:[[:space:]]*(security|new|fix|enhancement)[[:space:]]*$/ { found=NR }
+      found && NR>found && NR<=found+3 && /components:/ { c=1 }
+      found && NR>found && NR<=found+3 && /text:/       { t=1 }
```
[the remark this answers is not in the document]
sscarduzio: Confirmed and fixed — I reproduced it first:

```
  - type: fix
    components: [es]
  - type: new
    text: "other"
```
Two incomplete entries were accepted, because the flags survived across entries. A new `type:` line now resets them, and `text:` must have a value.

| description | before | now |
|---|---|---|
| real single entry | accept | accept |
| two half-entries (your case) | accept | **reject** |
| `text:` with an empty value | accept | **reject** |
| second entry complete, first not | accept | accept (correct — one complete entry exists) |

The five open PRs are unchanged.

.github/workflows/pr-conventions.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Check PR conventions
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          # Only true for the pull request that introduces the check.
+          if [ ! -x ci/check-pr-conventions.sh ]; then
+            echo "::notice::ci/check-pr-conventions.sh is not on the base branch yet; skipping"
+            exit 0
```
coutoPL: Why not `exit 1`?

core/build.gradle
```diff
@@ -80,9 +80,9 @@ dependencies {
     // ES 9.1+ (auto-read disabled, elastic/elasticsearch#127817) affects the TLS-over-HTTP read flow. 4.1.136
     // (netty#15053) required the SslHandshakeReadTrigger workaround used by es{version}x SSLNetty4HttpServerTransport
     // (tech.beshu.ror.utils.SslHandshakeReadTrigger) — re-verify that path before/after changing this version.
-    api group: 'io.netty',                          name: 'netty-codec-http2',                   version: '4.1.136.Final'
-    api group: 'io.netty',                          name: 'netty-handler',                       version: '4.1.136.Final'
-    api group: 'io.netty',                          name: 'netty-transport',                     version: '4.1.136.Final'
+    api group: 'io.netty',                          name: 'netty-codec-http2',                   version: '4.1.137.Final'
```
coutoPL, line 82: not related to the PR!

suppressions_cve.xml
```diff
@@ -69,6 +69,10 @@ False positive: CVE-2021-34364 is for "Refined GitHub" (browser extension), not
         <cve>CVE-2026-56148</cve>
         <cve>CVE-2026-56149</cve>
         <cve>CVE-2026-63144</cve>
+        <cve>CVE-2026-56144</cve>
```
coutoPL, line 72: not related to the PR!

ci/check-pr-conventions.sh
coutoPL, line 1: IMO, each convention should be a separate script.
The main script should go through all the convention scripts (in a predefined order) and run them one by one. Then it should check for errors and pass if there are no errors. If there are errors, it should fail and report all errors.

coutoPL (changes requested):

core/build.gradle
sscarduzio: You are right, and this was my mistake — not a stray change. This branch carried the netty CVE branch so `cve_check` could pass, but it targeted `develop`, so those files showed up as mine.

The netty PR is merged now, so I rebased onto `develop`: the diff is **3 files** and contains only the conventions check.

suppressions_cve.xml
sscarduzio: You are right, and this was my mistake — not a stray change. This branch carried the netty CVE branch so `cve_check` could pass, but it targeted `develop`, so those files showed up as mine.

The netty PR is merged now, so I rebased onto `develop`: the diff is **3 files** and contains only the conventions check.

.github/workflows/pr-conventions.yml
sscarduzio: Fair — a policy check that silently passes is a bad smell. Removed the skip in `4dfa4924d`.

The reason it existed: the check runs the **base branch's** copy of the scripts, so a PR cannot edit the rule that judges it — and on the PR that *introduces* the checks, the base branch has none. `exit 1` would have made that PR unmergeable; `exit 0` was silent.

It now does neither. If the base branch has no checks yet, it runs the PR's own copy and says so loudly:

```
::warning::the checks are not on the base branch yet, so this run uses the pull request's own copy
```

So the check always actually runs, and after this merges the fallback can never trigger again.

ci/check-pr-conventions.sh
sscarduzio: Done in `4dfa4924d` — you are right that this scales better, and it fits where we are heading: adding a convention is now adding a file, which is exactly the shape an agent can extend.

```
ci/pr-conventions/
  run.sh                          # fetches the PR data once, runs checks/*.sh in name order, aggregates
  checks/10-title-jira-key.sh
  checks/20-changelog-entry.sh
```

`run.sh` keeps going after a failure, so one run reports everything that is wrong instead of only the first thing.

The restructure also caught a real bug, which is a good argument for your suggestion: passing the file list through the environment broke on **#1313**, whose 4,590 changed files exceed the 128 KB limit for a single environment variable — every later command then failed to start with `Argument list too long`. The runner now passes the big inputs as files. Verified: identical results on all five open PRs, #1313 included.

ci/pr-conventions/checks/10-title-jira-key.sh
```diff
@@ -0,0 +1,13 @@
+# The title must carry the Jira key: "[RORDEV-1234] Short summary".
+# Escape hatch: the `no-jira` label, for a change that does not need a ticket.
+#
+# Sourced by run.sh; uses PR_TITLE and the has_label / pass / fail helpers.
+
+if has_label no-jira; then
+  pass "title: skipped ('no-jira' label)"
+elif [[ $PR_TITLE =~ ^\[RORDEV-[0-9]+\][[:space:]]+.+ ]]; then
```
coutoPL: Maybe it would be better if all logic were in the check scripts? 
E.g., here is the logic of title extraction. In some cases it won't be 100% optimal, but clear and readable.

coutoPL (changes requested):

ci/pr-conventions/checks/10-title-jira-key.sh
sscarduzio: Agreed, done. The runner now only fetches the pull request once and formats results; each check reads `PR_JSON_FILE` and does its own extraction, so the logic of a convention is visible in the file that owns it:

```bash
title=$(jq -r ".title" "$PR_JSON_FILE")
labels=$(jq -r ".labels[].name" "$PR_JSON_FILE")
```

The shared `has_label` helper is gone for the same reason. It stays one API call, so the readability costs nothing.

ci/pr-conventions/run.sh
```diff
@@ -0,0 +1,58 @@
...
+# start with "Argument list too long". (PR #1313 changed 4,590 files.)
+WORK_DIR=$(mktemp -d)
+trap 'rm -rf "$WORK_DIR"' EXIT
+export PR_FILES_FILE="$WORK_DIR/files"
+export PR_BUILD_DIFF_FILE="$WORK_DIR/build-diff"
+gh api "repos/$REPO/pulls/$PR/files" --paginate --jq '.[].filename' > "$PR_FILES_FILE"
+# Only the build files' diffs: a check that cares about dependencies needs the changed lines, not just
+# the file names.
+gh api "repos/$REPO/pulls/$PR/files" --paginate \
+  --jq '.[] | select(.filename | test("(^|/)build\\.gradle$")) | .patch // ""' > "$PR_BUILD_DIFF_FILE"
```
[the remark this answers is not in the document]
sscarduzio: Real, and demonstrable on this repo — fixed.

```
#1313 changed_files (API):   4590
#1313 files actually listed: 3000
```

Rather than fetch base and head git objects (a full checkout for a check that runs on every PR), the check now compares `.changed_files` against the number of lines it got and **fails closed** when they differ:

```
note: the changed-file list is truncated (3000 of 4590); assuming production code changed
```

Failing open would have silently passed exactly the largest changes — the ones most likely to need a changelog entry. The `no-changelog` label remains the way out, which is what #1313 uses.

ci/pr-conventions/checks/20-changelog-entry.sh
```diff
@@ -0,0 +1,46 @@
...
+# them.
+elif awk '
+      /^[[:space:]]*-?[[:space:]]*type:[[:space:]]*(security|new|fix|enhancement)[[:space:]]*$/ {
+        found=NR; c=0; t=0; next
+      }
+      found && NR<=found+3 && /components:/ { c=1 }
+      found && NR<=found+3 && /text:[[:space:]]*[^[:space:]]/ { t=1 }
+      c && t { complete=1 }
+      END { exit !complete }' <<<"$PR_BODY" \
+  || grep -qE '\*\*(Security Fix|New|Enhancement|Fix)\*\*[[:space:]]*\((ES|KBN)\)' <<<"$PR_BODY"; then
```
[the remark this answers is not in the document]
sscarduzio: Correct, fixed as proposed. Every `type:` line now resets the state and only a supported type opens an entry.

Your counter-example, tested:

```
  - type: fix
  - type: docs
    components: [es]
    text: "x"
```
before: accepted — now: **rejected**. A genuine complete entry is still accepted.

.github/workflows/pr-conventions.yml
```diff
@@ -0,0 +1,53 @@
+name: PR Conventions
+
+# Checks the conventions a reviewer used to check by hand: the Jira key in the title, and a changelog
```
coutoPL, line 3: I wonder why we need a separate workflow? Why cannot it be a part of our CI workflow?

sscarduzio: Good question — I tried to keep it in `ci.yml` first, and the trigger types make it impossible without a large cost. Three reasons, all concrete:

**1. The check must run on events `ci.yml` deliberately excludes.** `ci.yml` is `types: [opened, synchronize, reopened, ready_for_review]`. This check has to re-run on **`edited`** (you fix the title, or paste the changelog entry into the description) and on **`labeled` / `unlabeled`** (you add `no-jira` / `no-changelog`). Without those, you fix the title and the check stays red until you push a commit — and the label escape hatch never works at all.

**2. Adding those types to `ci.yml` is expensive.** It would re-run the whole matrix — 16 jobs / 27 checks, including the integration legs on the 16 GB runners — on **every description edit and every label toggle**. On PRs where we add a label or fix a typo a few times, that is a hundred-plus extra job runs each, on the runners we just spent RORDEV-2156 trying to relieve.

**3. `ci.yml` has `paths-ignore: docs/**, .claude/**, **/*.md`.** A docs-only PR does not trigger it at all — but it still needs a Jira key in the title. As a `ci.yml` job the check would silently not run on exactly those PRs.

There is also a practical difference in shape: this job needs no container, no toolchains image and no JDK; it finishes in ~8 seconds and has its own `concurrency` group so an edit cancels the previous run. Sharing `ci.yml`'s concurrency group would make a label toggle cancel an in-flight test matrix.

The only consolidation that keeps the behaviour is making it a reusable workflow called from `ci.yml`, which still leaves two files — so it buys nothing.

Happy to reopen this if you would still rather have it inside `ci.yml`; I just did not want to trade the label escape hatch or the runner budget for one fewer file.

coutoPL (approved):
LGTM

sscarduzio:
Ran a self-review on this PR and applied four fixes (`8730b2fc4`):

| # | Finding | Fix |
|---|---|---|
| 1 | `run.sh` paginated the files endpoint **twice** for the same data — ~17s per crawl on #1313, and double the rate limit | one crawl, both views derived locally; runner on #1313 now **16s total** instead of ~34s just fetching |
| 2 | The truncation guard used `wc -l`, which counts *newlines* — a one-file PR whose output lacked a trailing newline would look truncated and be asked for a changelog it doesn't need | count with `awk END{print NR}` |
| 3 | A check run directly failed with `jq: Could not open file` instead of saying what it needs | each check now declares its contract: `: "${PR_JSON_FILE:?...}"` |
| 4 | Bare `3` governing how far `components:`/`text:` may sit from `type:` | named `window` variable with a comment |

Verified: all four open PRs give unchanged results, and every changelog-matcher case still behaves (genuine entry accepted; incomplete-then-unsupported, two half-entries, empty `text:`, and three stray lines all rejected).

One finding I deliberately did **not** fix here: the second `actions/checkout` (the fallback for when the base branch has no checks yet) is only needed by *this* PR — after it merges it runs on every PR and is never used. Removing it now would stop this PR's own check from running, so it's filed as [RORDEV-2158](https://readonlyrest.atlassian.net/browse/RORDEV-2158) to do immediately after merge.

PR #1322, opened 2026-08-11
[RORDEV-2151] Fix copy-drift between ES modules

sscarduzio:
Extracted from #1313 on @coutoPL's request, for separate review. Same commit, cherry-picked onto develop — once this merges, the #1313 diff shrinks to the overlay mechanism and migration only.

Three drift cases, found by hashing every file across all 34 es* modules and flagging paths whose version sequence across modules is non-contiguous:

| Where | Fix |
|---|---|
| `es87x` EsIndexDocumentManager | Remove two redundant imports (the class extends same-package `IndexDocumentManager`; `warnEx` comes from `RequestIdAwareLogging`). File becomes identical to its neighbors. |
| `es87x`/`es88x`/`es89x` DeleteComposableIndexTemplateEsRequestContext | Remove the dead runtime branch for ES ≤ 7.12 carried over from the 7.x lineage. Always false on ES 8.x — behavior unchanged. Files now match es80x..es85x. |
| `es90x` AclAwareRequestFilter, IndexLevelActionFilter, NodeClientBasedAuditSinkService | Align with the es91x..es94x lineage: `EsContext` carries `EsVersion`; audit sink uses `BulkProcessor2` (present in ES 9.0). es90x was forked before these landed in es818x and never caught up. **The `BulkProcessor2` switch is the only behavioral change** — same max retries, ES-managed scheduling instead of the deprecated `BulkProcessor`. |

Deliberately not "fixed": the object-style `EsqlRequestHelper` in single-minor modules is intentional divergence (the class-style `EsVersion`-switch variant exists only where one module spans ESQL-incompatible minors), and the `QueryType` anomaly is a false positive (a `@nowarn` legitimately dropped in 8.x).

Verified: compileScala green for es87x–es90x; all four modules passed their integration legs on #1313's runs (27/27 × 3 CI rounds).

## Summary by CodeRabbit

* **Compatibility**
  * Improved component and composable index template deletion across supported Elasticsearch versions by consistently handling multiple template names.
* **Audit Logging**
  * Updated audit event delivery with improved bulk processing and streamlined retry and scheduling behavior.
* **Bug Fixes**
  * Improved request processing so Elasticsearch version information is handled consistently across operations.
  * Removed outdated compatibility behavior that could affect template deletion requests.

### DIFF
--- a/es89x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/templates/DeleteComposableIndexTemplateEsRequestContext.scala
+++ b/es89x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/templates/DeleteComposableIndexTemplateEsRequestContext.scala
@@ -18,7 +18,6 @@ package tech.beshu.ror.es.handler.request.context.types.templates
 
 import cats.data.NonEmptyList
 import cats.implicits.*
-import org.elasticsearch.Version
 import org.elasticsearch.action.admin.indices.template.delete.{
   DeleteComposableIndexTemplateAction,
   DeleteIndexTemplateRequest
@@ -71,40 +70,11 @@ class DeleteComposableIndexTemplateEsRequestContext(
   extension (request: DeleteComposableIndexTemplateAction.Request)
 
     private def getNames: List[TemplateNamePattern] = {
-      if (isEsNewerThan712) getNamesForEsPost12
-      else getNamesForEsPre13
-    }
-
-    private def updateNames(names: NonEmptyList[TemplateNamePattern]): Unit = {
-      if (isEsNewerThan712) updateNamesForEsPost12(names)
-      else updateNamesForEsPre13(names)
-    }
-
-    private def isEsNewerThan712 = {
-      Version.CURRENT.after(Version.fromString("7.12.1"))
-    }
-
-    private def getNamesForEsPre13 = {
-      val name: String = on(request).call("name").get[String]
-      Option(name).toList.flatMap(TemplateNamePattern.fromString)
-    }
-
-    private def getNamesForEsPost12 = {
       val names: Array[String] = on(request).call("names").get[Array[String]]
       names.asSafeList.flatMap(TemplateNamePattern.fromString)
     }
 
-    private def updateNamesForEsPre13(names: NonEmptyList[TemplateNamePattern]) = {
-      names.tail match {
-        case Nil =>
-        case _   =>
-          logger.warn(s"""[${id.show}] Filtered result contains more than one template pattern. First was taken.
-               | The whole set of patterns [${names.show}]""".oneLiner)
-      }
-      on(request).call("name", names.head.value.value)
-    }
-
-    private def updateNamesForEsPost12(names: NonEmptyList[TemplateNamePattern]): Unit = {
+    private def updateNames(names: NonEmptyList[TemplateNamePattern]): Unit = {
       on(request).set("names", names.toList.map(_.value.value).toArray)
     }
 
--- a/es90x/src/main/scala/tech/beshu/ror/es/services/NodeClientBasedAuditSinkService.scala
+++ b/es90x/src/main/scala/tech/beshu/ror/es/services/NodeClientBasedAuditSinkService.scala
@@ -17,13 +17,13 @@
 package tech.beshu.ror.es.services
 
 import cats.data.NonEmptyList
-import org.elasticsearch.action.bulk.{BulkProcessor, BulkRequest, BulkResponse}
+import org.elasticsearch.action.bulk.{BulkProcessor2, BulkRequest, BulkResponse}
 import org.elasticsearch.action.index.IndexRequest
 import org.elasticsearch.action.{ActionListener, DocWriteRequest}
 import org.elasticsearch.client.internal.node.NodeClient
-import org.elasticsearch.common.BackoffPolicy
 import org.elasticsearch.common.unit.{ByteSizeUnit, ByteSizeValue}
 import org.elasticsearch.core.TimeValue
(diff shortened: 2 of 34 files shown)


sscarduzio:
Both red checks here are **not from this PR**:

- **\`cve_check\`** — a new CVE landed today: [CVE-2026-64607](https://nvd.nist.gov/vuln/detail/CVE-2026-64607) in `httpclient5-5.6.2` (affected range 5.0-alpha ≤ 5.6.2). It fails on every branch. Fixed in #1324 (bump to 5.6.4); once that merges this goes green.
- **`unit_tests_linux`** — one test, `LdapServicesSettingsTests.scala:2330` "when some of LDAP services are unavailable". It is environment-sensitive rather than broken by this change:

```
expected: "There was a problem with ldap1 LDAP connection to: ldaps://ssl-ldap2.foo.com:836"
actual:   "There was a problem with resolving ldap1 LDAP hosts: ldaps://ssl-ldap2.foo.com:836"
```

The test points at fake hostnames and asserts the *connection* failure message, but on this runner DNS did not resolve them, so the code produced the *resolution* failure message instead. Whether it passes depends on how the runner resolves `*.foo.com`. This PR only touches es87x/es88x/es89x/es90x sources — it cannot reach `core`.

I will rerun it once #1324 is in.

es90x/src/main/scala/tech/beshu/ror/es/services/NodeClientBasedAuditSinkService.scala
```diff
@@ -39,20 +39,23 @@ import tech.beshu.ror.utils.RequestIdAwareLogging
...
-    BulkProcessor
-      .builder(BulkRequestHandler, new AuditSinkBulkProcessorListener, "ror-audit-bulk-processor")
+    BulkProcessor2
+      .builder(BulkRequestHandler, new AuditSinkBulkProcessorListener, threadPool)
       .setBulkActions(AUDIT_SINK_MAX_ITEMS)
       .setBulkSize(ByteSizeValue.of(AUDIT_SINK_MAX_KB, ByteSizeUnit.KB))
       .setFlushInterval(TimeValue.timeValueSeconds(AUDIT_SINK_MAX_SECONDS))
-      .setConcurrentRequests(1)
-      .setBackoffPolicy(BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(100), AUDIT_SINK_MAX_RETRIES))
+      .setMaxNumberOfRetries(AUDIT_SINK_MAX_RETRIES)
```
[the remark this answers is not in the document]
sscarduzio: Worth checking, but this PR does not introduce that behaviour — it **removes a difference**.

`es91x`, `es92x` and `es94x` on `develop` already use exactly this configuration today:

```
BulkProcessor2 ... .setMaxNumberOfRetries(AUDIT_SINK_MAX_RETRIES)
```

`es90x` was the odd one out: it was forked before `BulkProcessor2` landed in the 9.x lineage and never caught up, so it kept the deprecated `BulkProcessor` with `BackoffPolicy.exponentialBackoff`. This change aligns it with its three siblings, which is the entire point of the drift PR — one ES major should not have two different audit-sink implementations.

So if immediate `429` retries are a real concern, it is a concern for the whole ES 9.x line that is already shipping, not something introduced here. That deserves its own ticket against the shared behaviour rather than a fix applied to one module, which would recreate the drift we are removing.

es89x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/templates/DeleteComposableIndexTemplateEsRequestContext.scala
```diff
@@ -71,40 +70,11 @@ class DeleteComposableIndexTemplateEsRequestContext(
   extension (request: DeleteComposableIndexTemplateAction.Request)
 
     private def getNames: List[TemplateNamePattern] = {
-      if (isEsNewerThan712) getNamesForEsPost12
```
coutoPL, line 74: This kind of code was (probably) ported by mistake from one of the `es7*x` modules. But AFAIS, it probably makes sense for one module only. Please check the codebase and leave the code only in the module/modules where is make sense. Remove from other places.

coutoPL (changes requested):

es89x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/templates/DeleteComposableIndexTemplateEsRequestContext.scala
sscarduzio: Checked, and you were right — it makes sense in exactly one module.

The branch switches on `Version.CURRENT.after("7.12.1")`, so a module needs it only if its supported range spans that boundary. Of the 19 modules carrying it, only one does:

| module | supported range | needs the branch |
|---|---|---|
| **es711x** | 7.11.2 – 7.13.4 | **yes — kept** |
| es714x | 7.14.0 – 7.15.2 | no |
| es716x | 7.16.0 – 7.17.12 | no |
| es717x | 7.17.13 – 7.17.29 | no |
| es80x … es815x | 8.x | no |

Your instruction to check the codebase also caught something my first pass missed: **the same dead branch exists in a second file**, `DeleteComponentTemplateEsRequestContext`, next to `DeleteComposableIndexTemplateEsRequestContext`. Both are now cleaned in those 18 modules — 27 files, −837 lines.

The result is the shape `es816x` and newer already have, so this closes the gap rather than inventing anything. Each module keeps its own body text; I did not copy code between modules. `compileScala` is green for all 19, es711x included.

coutoPL (approved):
LGTM

PR #1323, opened 2026-08-11
[RORDEV-2157] Give the ES test containers a memory limit

sscarduzio:
Jira: [RORDEV-2157](https://readonlyrest.atlassian.net/browse/RORDEV-2157)

_No changelog entry: test infrastructure only._

Follow-up to #1321 (RORDEV-2156), which capped the orchestrator JVM. This bounds the other half.

## The problem

The ES test containers pin their heap (`-Xms512m -Xmx512m`) but have **no cgroup memory limit**, so their total footprint is unbounded. When a leg runs out of memory, the **host** OOM killer picks the victim — which is why today's failures surfaced as an unrelated Gradle daemon dying (`hs_err_pid`, "daemon disappeared") or a test worker being cancelled, while the container that actually took the memory kept running.

With a limit, Docker kills the container that went over its budget, and the failure names itself instead of costing someone a log-read to reach "not my code".

## Choosing the number — measured, not guessed

I ran ES 8.18 with the exact heap these tests pin:

| Metric | Value |
|---|---|
| Container RSS, idle | **1.092 GiB** |
| Container RSS, under indexing + search load | **1.105 GiB** |
| JVM heap used / max | 319 / 512 MiB |
| Non-heap | 197 MiB |

A full heap plus the ROR plugin lands around ~1.5 GiB, so **2 GiB** is the limit. Verified by starting a node with `--memory=2g`:

```
ES booted and serving under the 2 GiB cap (after ~90s)
usage under cap: 1.082GiB / 2GiB (54.09%)
cgroup limit applied: 2147483648 bytes
OOMKilled: false  Status: running
```

54% utilisation — comfortable headroom, while still catching a node that balloons to 4–6 GiB.

`ROR_ES_CONTAINER_MEMORY_MB` overrides it without a code change, for a suite that genuinely needs more.

## Scope

Docker path only (`EsContainer.scala`); the Windows legs run ES as native processes and are untouched. `tests-utils:compileScala` is green.

## Honest limitation

This does **not** reduce the memory a leg uses — it bounds each container and makes failures deterministic and attributable. The actual headroom recovery came from #1321 (~4 GB per leg). If a container does start getting OOM-killed at 2 GiB, that is now visible information rather than a mystery crash elsewhere, and the env var is the immediate relief valve.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Linux-based Elasticsearch containers by applying a configurable memory limit.
  * Uses a 2048 MB default limit when no valid custom value is provided.
  * Supports customizing the container memory limit through the `ROR_ES_CONTAINER_MEMORY_MB` setting.

### DIFF
--- a/tests-utils/build.gradle
+++ b/tests-utils/build.gradle
@@ -54,11 +54,11 @@ dependencies {
     api group: 'io.monix',                       name: 'monix_3',                           version: '3.4.1'
     api group: 'com.lihaoyi',                    name: 'os-lib_3',                          version: '0.11.6'
     api group: 'eu.timepit',                     name: 'refined_3',                         version: '0.11.3'
-    implementation group: 'org.typelevel',        name: 'squants_3',                         version: '1.8.3'
     api group: 'org.scala-lang',                 name: 'scala3-library_3',                  version: '3.3.7'
     api group: 'org.scala-lang.modules'  ,       name: 'scala-parallel-collections_3',      version: '1.2.0'
     api group: 'com.typesafe.scala-logging',     name: 'scala-logging_3',                   version: '3.9.5'
     api group: 'org.scalatest',                  name: 'scalatest_3',                       version: '3.2.19'
+    api group: 'org.typelevel',                  name: 'squants_3',                         version: '1.8.3'
     api group: 'com.dimafeng',                   name: 'testcontainers-scala-scalatest_3',  version: '0.44.0'
     api group: 'org.testcontainers',             name: 'testcontainers',                    version: "2.0.3"
     api group: 'eu.rekawek.toxiproxy',           name: 'toxiproxy-java',                    version: "2.1.11"
--- a/tests-utils/build.gradle
+++ b/tests-utils/build.gradle
@@ -54,6 +54,7 @@ dependencies {
     api group: 'io.monix',                       name: 'monix_3',                           version: '3.4.1'
     api group: 'com.lihaoyi',                    name: 'os-lib_3',                          version: '0.11.6'
     api group: 'eu.timepit',                     name: 'refined_3',                         version: '0.11.3'
+    implementation group: 'org.typelevel',        name: 'squants_3',                         version: '1.8.3'
     api group: 'org.scala-lang',                 name: 'scala3-library_3',                  version: '3.3.7'
     api group: 'org.scala-lang.modules'  ,       name: 'scala-parallel-collections_3',      version: '1.2.0'
     api group: 'com.typesafe.scala-logging',     name: 'scala-logging_3',                   version: '3.9.5'
--- a/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
+++ b/tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
@@ -23,6 +23,7 @@ import org.apache.http.message.BasicHeader
 import org.testcontainers.containers.GenericContainer
 import org.testcontainers.containers.output.{OutputFrame, Slf4jLogConsumer}
 import org.testcontainers.images.builder.ImageFromDockerfile
+import squants.information.{Information, Mebibytes}
 import tech.beshu.ror.utils.containers.ElasticsearchNodeWaitingStrategy.AwaitingReadyStrategy
 import tech.beshu.ror.utils.containers.EsContainer.Credentials.{BasicAuth, Header, None, Token}
 import tech.beshu.ror.utils.containers.EsContainer.{Credentials, EsContainerImplementation}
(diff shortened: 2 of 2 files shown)


tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
```diff
@@ -54,6 +54,21 @@ abstract class EsContainer(
...
+  //
+  // Measured on ES 8.18 with the 512m heap these tests pin: ~1.09 GiB resident idle and ~1.10 GiB under
+  // indexing and search load (heap 319/512 MiB, non-heap 197 MiB). 2 GiB leaves room for a full heap plus
+  // the ROR plugin -- a node boots and serves at ~54% of it -- while still catching a runaway node.
+  // Override with ROR_ES_CONTAINER_MEMORY_MB when a suite genuinely needs more.
+  private val esContainerMemoryLimitBytes: Long =
+    Option(System.getenv("ROR_ES_CONTAINER_MEMORY_MB"))
+      .flatMap(_.toLongOption)
+      .filter(_ > 0)
+      .getOrElse(2048L) * 1024L * 1024L
```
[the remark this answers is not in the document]
sscarduzio: Real bug, fixed. The bound is now applied to the megabyte value **before** multiplying, with an out-of-range value falling back to the default:

| `ROR_ES_CONTAINER_MEMORY_MB` | before | after |
|---|---|---|
| 2048 | 2147483648 | 2147483648 |
| 8796093022207 (max safe) | 9223372036853727232 | 9223372036853727232 |
| 8796093022208 | **-9223372036854775808** | 2147483648 (default) |
| 99999999999999 | overflowed | 2147483648 (default) |

Worth stressing the failure mode you identified: a negative byte count does not merely ignore the override — Docker rejects it, so a typo would have failed **every** ES container in the run.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
```diff
@@ -54,6 +54,21 @@ abstract class EsContainer(
...
+
```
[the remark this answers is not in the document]
sscarduzio: Comment narrowed, and you are right on both counts.

Every Docker profile pins the same `-Xms512m -Xmx512m` — that is what makes the measurement meaningful, since the heap is fixed and only off-heap varies. Windows runs ES as native processes with a 400m heap and no cgroup limit at all, so this does not apply there. The comment now names ES 8.18 as the measured version and scopes the claim to the Docker path.

I did not re-measure every ES version deliberately: the heap is identical across them, the 2 GiB limit sits ~85% above the measured 1.10 GiB, and `ROR_ES_CONTAINER_MEMORY_MB` is the relief valve if some version needs more, without a code change. If a container is ever OOM-killed at 2 GiB, that is now attributable information instead of a mystery crash in an unrelated job — which is the point of the PR.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
```diff
@@ -54,6 +54,26 @@ abstract class EsContainer(
...
+  // Each ES container gets a memory budget. Without one the container is unbounded, so when a leg runs
+  // out of memory the HOST OOM killer chooses the victim -- which is why those failures surfaced as an
+  // unrelated Gradle daemon or test worker dying while the container that took the memory kept running.
+  // With a limit, Docker kills the container that went over and the failure names itself.
+  //
+  // Measured on ES 8.18 with the 512m heap every Docker profile pins: ~1.09 GiB resident idle and ~1.10 GiB
+  // under indexing and search load (heap 319/512 MiB, non-heap 197 MiB). 2 GiB leaves room for a full heap
+  // plus the ROR plugin -- a node boots and serves at ~54% of it -- while still catching a runaway node.
+  // Override with ROR_ES_CONTAINER_MEMORY_MB when a suite genuinely needs more.
+  private val esContainerMemoryLimitBytes: Long = {
```
coutoPL: you can use `squants.information.Bytes` instead of Long here (and then you can change the val name because the `Bytes` suffix won't be needed)

coutoPL (changes requested):

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/EsContainer.scala
sscarduzio: Done — `esContainerMemoryLimit: Information`, built with `Mebibytes` (the override is given in megabytes) and converted to bytes only where Docker needs a raw number. Name loses the suffix as you suggested.

Two things worth flagging, since they change the cost of the suggestion slightly:

**1. squants was not on tests-utils' classpath.** It is declared in `core`, and tests-utils does not depend on core, so this adds the dependency. I scoped it `implementation` rather than `api` so it does not leak onto the classpath of everything that consumes tests-utils (most of its other deps are `api`, which is why the test classpath is already large).

**2. It removed a bug rather than just renaming one.** The previous version needed an explicit upper bound, because `mb * 1024 * 1024` on a `Long` overflows to a **negative** byte count that Docker rejects — failing every container in the run. `Information` holds a `Double`, and JVM double→long conversion *saturates* instead of wrapping, so that guard is no longer reachable and I dropped it:

| `ROR_ES_CONTAINER_MEMORY_MB` | before | now |
|---|---|---|
| 8796093022208 | **-9223372036854775808** | 9223372036854775807 |
| 99999999999999 | overflowed | 9223372036854775807 |

So the type change is doing real work here, not just cosmetics. Thanks.

tests-utils/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     api group: 'io.monix',                       name: 'monix_3',                           version: '3.4.1'
     api group: 'com.lihaoyi',                    name: 'os-lib_3',                          version: '0.11.6'
     api group: 'eu.timepit',                     name: 'refined_3',                         version: '0.11.3'
+    implementation group: 'org.typelevel',        name: 'squants_3',                         version: '1.8.3'
```
coutoPL: 1. Why implementation, not api? if the implementation is better, let's change the other libs, otherwise - follow convention
2. The deps should be ordered by name

coutoPL (changes requested):

tests-utils/build.gradle
sscarduzio: Both applied — it is `api` now, and sits in name order between `scalatest_3` and `testcontainers-scala-scalatest_3`.

On the first point, my reasoning was that `api` on tests-utils widens the classpath every consumer inherits, so a new dependency used by exactly one private val seemed a good candidate for `implementation`. But you are right that a lone exception in an otherwise all-`api` block is worse than the wider classpath: the next person has to work out why this one line differs. If narrowing is worth doing it is a change for the whole file, not a special case smuggled in with an unrelated PR.

Separately, the `integration_es818x` failure on this PR is **exit code 137** — SIGKILL from the host OOM killer, with no test failure and no container reported as OOM-killed. That is the ambient memory pressure this PR is meant to make attributable, not something the cap causes: a limit bounds a container, it does not increase what it uses. Worth noting that once this is in, that class of failure names the container instead of killing an unrelated process.

coutoPL (approved):
LGTM

PR #1330, opened 2026-08-12
[RORDEV-1229] Mirror ES libs in a manual workflow, not in CI

coutoPL:
## Why

When Elastic releases an ES version, Maven Central does not have its jars yet. We mirror them into
our libs store, and every plugin build resolves ES dependencies from there.

Until now that mirroring worked like this: uncomment a line in `ci/upload-es-artifacts.sh`, push a
`newes/*` branch to fire the `es_s3_up` CI job, then re-comment the line before merging.

Two problems with it:

- If you forgot to uncomment the line, the job was **green and uploaded nothing**. The mistake showed
  up much later as an es*x module failing to resolve `org.elasticsearch:elasticsearch:X.Y.Z`.
- Something we do a few times a year ran on every push to such a branch, and five CI jobs had a
  `needs` on a job that almost always did nothing.

## What changed

**Mirroring is now its own manual workflow.** `.github/workflows/mirror-es-libs.yml` takes the ES
versions as an input (`9.5.1 9.4.5`), so there is nothing to uncomment and nothing to revert. Which
versions were mirrored, and when, is now visible in the workflow run history. `ci/upload-es-artifacts.sh`
reads them from `ES_VERSIONS_TO_UPLOAD`, validates the whole list before uploading anything (a typo
must not leave half a mirror in the store), and can also be run by hand.

**The libs store address lives in one place.** The build that *reads* the store (the Maven repository
in `plugin-common-conventions`) and the task that *writes* it (`ror-tools:uploadArtifactsFromEsBinaries`)
each had their own copy of the endpoint, bucket and prefix defaults. They can drift, and they did. Both
now take the address from the new `LibsStore` class (`build-base`), which has unit tests.

**`es_s3_up` and the `newes/*` trigger are removed** from `ci.yml`, together with the `needs` on it in
the five jobs that waited for it.

## Also in here

- `ci-lib.sh`: `upload_using_aws_s3_uploader` gained a sibling, `..._to_key`, for uploads where the S3
  key is a path in a mirrored tree rather than a file name. Store resolution stays in one function.
- `ci-lib.sh`: `CI_DIR` now uses `BASH_SOURCE` instead of `$0`, which was wrong whenever the lib was
  sourced rather than run.
- `s3-uploader.sh`: dropped `curl -k`, so TLS certificates are verified again.
- `run-pipeline.sh`: ~35 near-identical `if [[ $ROR_TASK == "integration_esNNx" ]]` blocks (and the
  build/upload/release ones) collapsed into two pattern matches. Adding an ES module no longer needs an
  edit here.
- `ci/README.md` and the `ror-release` skill describe the new flow.

## How to use it

Supporting a new ES version, step 2 is now: run **Mirror ES Libs** with `es_versions` set, *before*
the branch adding that version goes through CI.

## Testing

Build-level change only, no product code touched. `LibsStoreTest` covers the store defaults and the
read/write address agreement.

## Summary by CodeRabbit

* **New Features**
  * Added a manually triggered workflow for mirroring selected Elasticsearch versions to the libraries store.
  * Version mirroring accepts space- or comma-separated lists, validates all versions, and reports uploaded versions.
  * Added centralized library-store configuration with environment-based defaults and normalized repository paths.
  * Added support for uploading artifacts using exact storage keys.

* **Bug Fixes**
  * TLS certificate verification is now enabled for library uploads.
  * CI task detection now supports Elasticsearch generations 6–9 more consistently.

* **Documentation**
  * Updated CI guidance for library mirroring, storage ownership, credentials, and workflow usage.

### DIFF
--- a/.claude/skills/ror-release/SKILL.md
+++ b/.claude/skills/ror-release/SKILL.md
@@ -33,7 +33,7 @@ Source: `beshu-tech/readonlyrest-internal` (`versioning.md`, `releasing.md`), re
 Verified against recent PRs (e.g. #1257). Files touched:
 
 1. `es{NN}x/gradle.properties` — add the version to `supportedEsVersions` in the matching module. **This is the single source of truth**: CI build/test, the default ES module, and the docs action-string sync all derive from it (`EsModuleFinder` / `printAllSupportedEsVersions`). There is no separate version list to update.
-2. `ci/upload-es-artifacts.sh` — add a *commented-out* line for the new version (the existing entries are all commented, kept as a historical record). To actually upload raw ES jars to S3: uncomment the line, push to a `newes/*` branch (the `ES_S3_UP` pipeline stage fires on that branch name), then re-comment before merging. This is a separate manual step — it is NOT what gates CI build/test (that's `supportedEsVersions`, step 1).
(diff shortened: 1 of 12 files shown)


coutoPL:
@claude review please

mgoworko (approved):
LGTM

PR #1331, opened 2026-08-12
[RORDEV-2168] Add CI memory telemetry and cut container RSS to stop host-OOM kills

sscarduzio:
## Why

The integration legs sporadically die with `exit code 137`. Deep RCA result:

- The SIGKILL comes from the **kernel OOM killer**, not from a JVM. No leg log contains a `java.lang.OutOfMemoryError`, a heap dump, or an `ExitOnOutOfMemoryError` trigger. The host runs dry; no single process hits its own cap.
- The worst case holds more containers than the memory budget assumed. The per-JVM **singleton ES never stops**, so each of the 4 shard JVMs keeps one ES resident under every later suite. Two gated heavy suites can add 4+3 nodes. Total: **~11 concurrent ES containers** (~12GB at the measured ~1.1GiB each) plus 9 JVMs on a 16GB runner.
- One multi-container suite **bypassed the heavy-suite gate**: `RemoteClusterAuditingToolsSuite` boots 2 extra ES nodes + 2 toxiproxy containers via `BaseSingleNodeEsClusterTest`, which carries no gate.
- All dependency containers run **uncapped**. WireMock's 2016 image runs JDK 8 with no `-Xmx`; that JDK is not cgroup-aware, so its default max heap is 1/4 of the **host** RAM (4GB).
- The test workers were the **only JVMs with unbounded metaspace** in the leg.

## What (all measured, `docker stats`, ES 8.18.0)

| Change | Effect |
|---|---|
| CI sets `ROR_ES_JDWP=false`; the JDWP agent stays out of ES containers | **−94MB per container**, ~−1GB at the 11-container worst case. Local runs keep the debugger (default on). The image tag hashes `ES_JAVA_OPTS`, so no stale-image reuse. |
| WireMock CMD gets `-Xmx256m` | Bounds a 4GB default to 256m. Verified: image boots, `__admin` answers, RSS 87MB. |
| Worker `-XX:MaxMetaspaceSize=512m` | A runaway now dies as an attributable `OutOfMemoryError`, not as a host OOM SIGKILL. Same cap every other JVM has. |
| `RemoteClusterAuditingToolsSuite` mixes in `HeavySuiteGated` | The 2-permit gate now covers all multi-container suites. |
[...]

PR #1333, opened 2026-08-13
[RORDEV-2170] Pre-pull the WireMock base image before the parallel workers

sscarduzio:
## Why

Integration legs fail intermittently in the WireMock suites (`DuplicatedResponseHeadersIssueSuite`, `AdminApiAuthMockSuite`):

```
org.testcontainers.containers.ContainerLaunchException: Container startup failed
  Caused by: ContainerFetchException: Can't get Docker image
    Caused by: DockerClientException: Could not build image: unknown blob
```

Seen today on develop (run 31677291593, 4 legs) and on #1302 (4 legs). Both were green on a plain rerun, so it is a race, not a broken image.

**The race.** Each WireMock suite builds its own image (its mappings differ), all `FROM rodolpheche/wiremock:2.5.1`. Up to 4 shard JVMs do that concurrently against one Docker daemon. When the base image is not yet in the runner cache, those builds pull it at the same time and one of them reads a half-materialised layer — `unknown blob`.

This is the same failure mode the ES image already guards against. `prebuildEsImage` exists precisely because *"at shardCount>=3 the worker JVMs otherwise build this identical image concurrently, and the Docker build layers fail intermittently"*. The dependency base images never got that treatment.

## What

`PrebuildSingletonEsImage` — already run once, before the shards spawn — now also pulls the base images the workers build `FROM`. One line of list, so adding the next dependency base is trivial.

The pull goes through testcontainers' `RemoteDockerImage`, not `docker pull`, so the configured pull policy and image-name substitutor keep applying.

The WireMock base image name moves to `WireMockContainer.BASE_IMAGE`, so the puller and the Dockerfile builder cannot drift apart.

## Verification

Not deduced — measured:

1. Removed `rodolpheche/wiremock:2.5.1` from the local daemon and confirmed it was gone.
2. Ran the new pull path on that empty cache: `PULLED_OK=rodolpheche/wiremock:2.5.1`.
3. Confirmed the image is present afterwards, which is exactly the state the shards need.

Also green locally: `tests-utils:compileScala`, `spotlessCheck`, `license`.

This removes the window; it does not change any test's behaviour.

JIRA: [RORDEV-2170](https://readonlyrest.atlassian.net/browse/RORDEV-2170)

## Summary by CodeRabbit

* **Performance**
  * Improved Docker image preparation by warming required image layers in advance, reducing startup time for integration-test environments.
* **Bug Fixes**
  * Enhanced Docker preparation logging and failure reporting.
  * Added Windows-aware handling to skip unsupported Docker warm-up operations.
* **Maintenance**
  * Streamlined image warming for parallel test runs.
  * Centralized the WireMock base-image configuration for more consistent container setup.


sscarduzio:
Windows legs caught a real gap in the first push: Windows runs ES as a native process and has **no Docker environment at all**, so `RemoteDockerImage.get()` threw `Could not find a valid Docker environment` and failed the prebuild step before it ever reached the ES part.

Fixed by guarding the pull with the repo's existing `OsUtils.ignoreOnWindows` idiom.

Verified both branches, by forcing `os.name` and watching the local daemon:
- `os.name=Windows 10` → pull skipped, image still absent afterwards.
- `os.name=Linux` → pull runs, image present afterwards.

Starting from an empty cache in both cases.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/PrebuildSingletonEsImage.scala
```diff
@@ -39,18 +44,31 @@ import com.typesafe.scalalogging.StrictLogging
 object PrebuildSingletonEsImage extends StrictLogging {
```
coutoPL: IMO, it would be nice to change the name of the object. Now, it's not only prebuilding a singleton ES image.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/PrebuildSingletonEsImage.scala
```diff
@@ -39,18 +44,31 @@ import com.typesafe.scalalogging.StrictLogging
...
 
   def main(args: Array[String]): Unit = {
-    logger.info("Pre-building the singleton ES image (once, before parallel test workers)...")
+    logger.info("Warming the Docker image cache (once, before parallel test workers)...")
     try {
+      pullBaseImages()
       // Touching `singleton` triggers its construction + start(), which builds the image + its layers.
       val _ = SingletonEsContainerWithRorSecurity.singleton.nodes.head
```
coutoPL: it would be nice to put the code below in own private, well-named method. Like the pullBaseImages one.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/WarmDockerImageCache.scala
```diff
@@ -39,18 +44,31 @@ import com.typesafe.scalalogging.StrictLogging
...
+      pullBaseImages()
       // Touching `singleton` triggers its construction + start(), which builds the image + its layers.
       val _ = SingletonEsContainerWithRorSecurity.singleton.nodes.head
       logger.info("Singleton ES image pre-build complete; layers cached for the test workers to rebuild from.")
       SingletonEsContainerWithRorSecurity.singleton.stop()
       sys.exit(0)
     } catch {
       case ex: Throwable =>
-        logger.error("Pre-build of the singleton ES image failed", ex)
+        logger.error("Warming the Docker image cache failed", ex)
```
coutoPL, line 52: The message should be generic because we do two things now

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/WarmDockerImageCache.scala
```diff
@@ -39,18 +44,31 @@ import com.typesafe.scalalogging.StrictLogging
 object PrebuildSingletonEsImage extends StrictLogging {
 
   def main(args: Array[String]): Unit = {
-    logger.info("Pre-building the singleton ES image (once, before parallel test workers)...")
+    logger.info("Warming the Docker image cache (once, before parallel test workers)...")
```
coutoPL, line 44: the same here

coutoPL (changes requested):

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/PrebuildSingletonEsImage.scala
sscarduzio: Renamed to `WarmDockerImageCache` (file, object, and the gradle task `prebuildEsImage` -> `warmDockerImageCache`). No stale reference is left in the repo.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/PrebuildSingletonEsImage.scala
sscarduzio: Extracted to `prebuildSingletonEsImage()`, next to `pullBaseImages()`. `main` is now just the two calls.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/WarmDockerImageCache.scala
sscarduzio: Both messages are generic now: it starts with "Warming the Docker image cache..." and ends with "Docker image cache warmed; the test workers can reuse the cached layers." The per-step detail moved into the step methods.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/WarmDockerImageCache.scala
sscarduzio: Same — covered by the rename and the generic messages.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/WireMockContainer.java
```diff
@@ -56,7 +57,7 @@ public static WireMockContainer create(String... mappings) {
         new WireMockContainer(
             dockerfile.withDockerfileFromBuilder(
                 builder -> {
-                  DockerfileBuilder b = builder.from("rodolpheche/wiremock:2.5.1");
+                  DockerfileBuilder b = builder.from(BASE_IMAGE);
```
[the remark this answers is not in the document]
sscarduzio: Good catch, and confirmed against the library source: `ImageFromDockerfile` pre-fetches its Dockerfile dependencies with `.withImageNameSubstitutor(ImageNameSubstitutor.noop())` (ImageFromDockerfile.java:200-202 in 2.0.3), while my pull applied the configured substitutor. With a substitutor configured, the warm pull and the build would have used different names and the warm-up would silently do nothing. The pull now uses `noop()` too, so both sides resolve identically.

integration-tests/build.gradle
```diff
@@ -161,21 +161,21 @@ dependencies {
     testImplementation  group: 'org.scalatest',                 name: 'scalatest_3',                version: '3.2.19'
 }
 
-// Builds + stably tags the singleton ES image ONCE before parallel workers; without it concurrent
-// identical builds fail intermittently at shardCount>=3. Pre-building gives every worker a cache hit.
-tasks.register('prebuildEsImage', JavaExec) {
+// Pulls the base images and builds the singleton ES image ONCE before parallel workers; without it
```
coutoPL: You describe what the warmDockerImageCache does internally at this level. It's not necessary. The contract (based on the name) is clear. If someone needs to know what it does internally, it shhould take a look at the proper class.

sscarduzio:
Pushed a fix for the CI failure my own rename caused, and a note on how I let it through.

**What broke:** `ShardedGradlewTest` passes `-x integration-tests:prebuildEsImage` to each child gradle invocation, so the shards do not redo the prebuild. Renaming the task left that string behind, and every shard died with:

```
Cannot locate excluded tasks that match 'integration-tests:prebuildEsImage'
as task 'prebuildEsImage' not found in project ':integration-tests'
```

**Why I missed it:** when I searched for references to rename, I filtered by `--include=*.gradle,*.scala,*.sh,*.yml,*.md` — and the reference is a magic string in a **`.java`** file. My own filter hid it. I re-ran the search with no file-type filter; there is now no occurrence of the old name anywhere in the repo.

**Verified with a negative control**, not just a green run:
- old name: `./gradlew integration-tests:testClasses -x integration-tests:prebuildEsImage --dry-run` reproduces CI's exact error.
- new name: same command with `warmDockerImageCache` resolves and succeeds.

So the check actually fails when it should, which is what makes the pass meaningful.

coutoPL (changes requested):

integration-tests/build.gradle
sscarduzio: Agreed — cut to `// Runs ONCE before the parallel workers; see WarmDockerImageCache.` The why lives in the class scaladoc, which is the single place it can stay correct.

integration-tests/build.gradle
```diff
@@ -161,8 +161,7 @@ dependencies {
     testImplementation  group: 'org.scalatest',                 name: 'scalatest_3',                version: '3.2.19'
 }
 
-// Pulls the base images and builds the singleton ES image ONCE before parallel workers; without it
-// the concurrent pulls/builds fail intermittently at shardCount>=3. See WarmDockerImageCache.
+// Runs ONCE before the parallel workers; see WarmDockerImageCache.
```
coutoPL: now you tell how it's used! this is bad too.

If any comment here is needed it should say, that its job to to warm docker image cache and the details are in the class. 

On the other hand, this is obvious

coutoPL (changes requested):

integration-tests/build.gradle
sscarduzio: Removed it. The task name is the contract.

coutoPL (approved):
LGTM

sscarduzio:
The `integration_es94x` failure here is **not this PR** — every test passed. Step 7 (`integration_es94x`) succeeded; step 10 (`Memory telemetry report`) failed. That is the diagnostics I added in #1331: when Ryuk reaps a container between `docker ps` and `docker inspect`, inspect returns 1, the `while` loop propagates it, and the step runs under `bash -e`.

Fixed separately in #1334, kept off this PR so the pre-pull change stays reviewable on its own.

Worth recording what the telemetry captured on this leg, since it is the first real data since #1331 landed — the host went down to **836 MB available** (7 of 128 samples below 1 GB), with **7 ES containers resident at once**:

```
avail_mb=836 swapfree_mb=1258 psi_some_avg10=1.46
7 x ES containers: 726M, 734M, 948M, 1.02G, 1.01G, 934M, 1.14G  (~6.5 GB)
12 x JVMs: 1053M, 486M, 389M, 389M, 329M, 288M, 284M, 270M, 266M, 247M, 72M, 69M  (~4.1 GB)
```

So the ~11-container worst case in the #1331 analysis is not theoretical: 7 concurrent ES containers happen in a normal leg. No kernel OOM fired this time, but the margin was under 1 GB.

PR #1335, opened 2026-08-15
Drop org.reflections and two more jars from the plugin bundle

sscarduzio:
```yaml
- type: enhancement
  components: [es]
  text: "Smaller plugin ZIP: unused libraries removed from the bundle"
```

From the simplification audit (item 3), with the amendments the adversarial review added. No Jira key — I cannot create tickets; happy to have one assigned and the title updated (`no-jira` label applied so the conventions check passes).

## Why

`org.reflections` 0.9.11 (2017) was carried for two call sites, both of which are "walk the class hierarchy and collect fields" — a few lines of JDK reflection:

- `ReflecUtils.setIndices` used `getAllFields` for the String/String[] fields it rewrites.
- `BulkShardEsRequestContext` used `getAllFields` + `withName("index")` to replace the `Index` inside `ShardId`.

It also dragged `javassist` in, and it was the only reason `guava` reached the classpath at all.

## The one subtle part

`ReflecUtils.fieldsNamed` matches on the **name only**, deliberately. The BulkShard call writes an `org.elasticsearch.index.Index`; routing it through `setIndices` (which only ever matches `String`/`String[]`) would have silently written nothing and left the index rewrite — an access-control operation — undone. That trap is what the adversarial review flagged, and `ReflecUtilsTests` now pins the behaviour: name-only matching, superclass traversal, and `setIndices` leaving other types alone.

## Also in this PR

- **`guava` is now a direct dependency of `core`** — ~16 production files import `com.google.common`, but it only arrived transitively via reflections. Same version (`32.0.1-jre`, the floor that fixed CVE-2023-2976); the constraint that used to carry it is removed, so the version is declared in one place.
- **`scala-parallel-collections` dropped** — no `.par` usage anywhere in `core`.

## Verification

Built the es94x plugin zip before and after and diffed the jar sets — exactly three jars leave, nothing else moves:

```
- javassist-3.21.0-GA.jar
- reflections-0.9.11.jar
- scala-parallel-collections_3.jar
```

96 jars → 93, and `guava` still ships. Plus: `core` compiles (`-Xfatal-warnings` on), all 6 new `ReflecUtilsTests` cases pass, and es67x/es94x compile — the two ends of the module range that share the patched `BulkShardEsRequestContext`.

The same file is patched in all 34 modules here because they are still full copies; once #1313 lands this becomes 4 files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bulk shard request handling across supported Elasticsearch versions, including updates to shard index information.
  - Preserved existing error handling and request-processing behavior while improving compatibility with inherited and private fields.

- **Tests**
  - Added broader validation for field discovery, inherited fields, array and scalar values, incompatible fields, and empty updates.


coutoPL:
@sscarduzio what about this one?

suppressions_cve.xml
```diff
@@ -208,4 +208,21 @@ False positive: CVE-2021-34364 is for "Refined GitHub" (browser extension), not
...
+   CPE matcher maps the artifact version 2.1.24 onto the generic bouncy_castle_for_java CPE.
+   Two further signs it is a version-number coincidence: the flagged module is the TLS/JSSE
+   provider, while BKS/UBER keystores live in the provider jar, and none of the sibling FIPS
+   artifacts we ship (bc-fips 2.1.3, bcpkix-fips 2.1.12, bcutil-fips 2.1.7) is flagged -- only the
+   one whose version happens to land in the matched range.
+   Scoped to this one CVE and this one artifact so a real future finding is not masked. There is no
+   upgrade to take: all four FIPS pins are already the newest published versions.
+   ]]></notes>
+        <packageUrl regex="true">^pkg:maven/org\.bouncycastle/bctls-fips@.*$</packageUrl>
+        <cve>CVE-2026-12185</cve>
```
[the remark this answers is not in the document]
sscarduzio: Moot now: this PR no longer touches `suppressions_cve.xml`. The suppression it carried came from #1336, which is superseded by the already-merged #1337 — that one covers all four FIPS artifacts, so the narrower `bctls-fips` entry, and this regex question with it, is gone from the branch.

sscarduzio:
Rebased onto develop. Two notes. First, the branch now uses the ci.yml from #1347 — the old one read secret names that do not exist any more, so every rerun failed at docker login. Second, the branch carried its own suppression for CVE-2026-12185 on bctls-fips; develop got a broader one for all four FIPS artifacts in #1337, so I dropped the branch copy and suppressions_cve.xml is now identical to develop. The one red e2e leg before the rebase was a Cypress Electron renderer crash (memory pressure on the runner), not this PR — no missing-class errors appear in the log.

sscarduzio:
@coderabbitai review

core/src/main/scala/tech/beshu/ror/utils/ReflecUtils.java
coutoPL, line 1: This class should be considered deprecated. In this codebase, we use `joor` when we deal with reflection. 
WDYT about getting rid of this class and use `joor` in all the places where the ReflecUtils are used today? 

If it's necessary, we can provide (in the utils package) a joor extension (using a Scala extension) if we are not able to replace the methods from ReflecUtils with a simple joor-provided method.

core/build.gradle
```diff
@@ -83,11 +83,12 @@ dependencies {
     api group: 'io.netty',                          name: 'netty-handler',                       version: '4.1.137.Final'
     api group: 'io.netty',                          name: 'netty-transport',                     version: '4.1.137.Final'
     api group: 'eu.timepit',                        name: 'refined_3',                           version: '0.11.3'
-    api group: 'org.reflections',                   name: 'reflections',                         version: '0.9.11'
     api group: 'org.mozilla',                       name: 'rhino',                               version: '1.8.1'
     api group: 'com.github.ben-manes.caffeine',     name: 'caffeine',                            version: '3.2.3'
+    // Direct dependency: ~16 production files import com.google.common. It used to arrive as a
```
coutoPL: remove the unecessary comment

coutoPL (changes requested):

core/build.gradle
sscarduzio: Trimmed in `80cb0a624`. The importer headcount was the part that rots, so it is gone; one line stays for the version floor (`32.0.1-jre` is the CVE-2023-2976 fix), since that is the fact a future bump needs to know. Say the word if you want the whole comment gone.

core/src/main/scala/tech/beshu/ror/utils/ReflecUtils.java
sscarduzio: Agreed on the direction — `joor` should be the one reflection API here, and `ReflecUtils` is legacy. I would rather not fold it into this PR, and I checked the ground first rather than argue from taste.

**Size.** `ReflecUtils` has 66 call sites in 42 files: 41 `setIndices` and 25 `extractStringArrayFromPrivateMethod`, spread across every `esXXXx` module. That is a bigger diff than this PR, and it lands in exactly the request-handling paths that the IT matrix covers per ES major.

**The part that is not mechanical.** Both entry points run inside `AccessController.doPrivileged`, and the plugin ships `plugin-metadata/plugin-security.policy` granting `ReflectPermission "suppressAccessChecks"` and `RuntimePermission "accessDeclaredMembers"`. joor 0.9.15 calls `setAccessible` with **no** `doPrivileged` anywhere in the jar — I checked the bytecode, not the docs. So a straight swap moves the reflection out of the plugin's privileged block, and where ES frames sit below us on the stack that is an `AccessControlException` at runtime, on a security manager, in production — which the unit tests (no SM) would not catch. Your "joor extension in the utils package" is the right shape for this, and it needs to wrap the privileged call.

**Scope.** This PR removes a JAR from the shipped bundle. `ReflecUtils` stays either way — dropping `org.reflections` only replaced its one use of `getAllFields` with a local hierarchy walk (joor's `Reflect` walks `getSuperclass`/`getDeclaredFields` too, so it is a fine replacement later).

Happy to take the migration as the next piece of work: a `joorPrivileged` extension in `utils` that keeps the `doPrivileged` wrapper, then the 66 call sites converted per ES major so the IT matrix bisects a break. Want me to open a RORDEV ticket for it, or would you rather scope it yourself?

coutoPL (changes requested):
> 🧐Enhancement (ES) three unused libraries removed from the plugin bundle

This should be changed. It should be:
1) in the yaml format
2) should be created from the clients' perspective - they are not interested in how many libraries we removed, but maybe they may be interested in the fact the the size of the plugin was reduced (nevertheless, I'm still not sure if we should put in the changelog)

core/src/main/scala/tech/beshu/ror/utils/ReflecUtils.java
coutoPL: ok, let's do this in the follow-up PR

sscarduzio:
Changelog entry reworked as requested. It is now the YAML form at the top of the description, and the text speaks to the client: "Smaller plugin ZIP: unused libraries removed from the bundle". No counts, no library names. If you decide it does not belong in the changelog at all, say so and I will replace it with the `no-changelog` label — your call.

Also re-ran the four red `build_es*` legs. They failed when the runner pulled the `ror-ci-toolchains` job container: the runner logged in with the RO credentials, and Docker Hub still answered "unauthenticated pull rate limit" (run 32351052650, job 96400182992, 10:51 UTC). Same symptom as the open point from RORDEV-2183: the login succeeds and the pull still counts as anonymous.

coutoPL (approved):
LGTM

PR #1340, opened 2026-08-17
[RORDEV-2179] Strip unused ES modules from the test images

sscarduzio:
## Why

`xpack.*.enabled: false` stops the services, but every module's jars are still scanned and class-loaded at boot. The test ES containers pay for 86 modules while the integration suites exercise a fraction of them.

Measured on ES 8.18.0, identical config, side by side:

| image | RSS after boot+load | boot to HTTP |
|---|---|---|
| official | 1.074 GiB | 50 s |
| 48 modules stripped | **829 MiB (−25%)** | 48 s |

At the 7–8 concurrent ES containers the telemetry records at peak, that is **~2 GB returned exactly where the host has been running within ~0.7 GB of the OOM killer**.

## What

One flag-gated `RUN rm -rf` step in the test image builder, listing 48 modules no suite uses.

- **The list is closure-verified, not guessed**: extracted every module's `extended.plugins` from the plugin descriptors and kept full dependency chains. Two non-obvious keeps found that way: the **server-cli launcher hard-requires the `apm` module directory** ("Installation is corrupt" otherwise), and **`x-pack-esql` declares a dependency on `x-pack-ml`** (which needs `x-pack-autoscaling`) — so ml stays despite being the fattest module on disk.
- Kept because suites use them: sql/esql (+`x-pack-ql`, `x-pack-esql-core`, async), ilm/slm, data-streams, reindex, painless, mustache, security, kibana, analytics, fs-repository snapshots.
- **One list serves every ES version**: `rm -rf` ignores names a version does not ship. The layer sits early in the image, so all suite images of a version share it.
- **Opt-out, not opt-in** (`ROR_ES_SLIM_MODULES=false`): CI and local runs boot the same image, so the module set can never cause "passes locally, fails in CI".

## Verification

- Local: stripped 8.18 boots green, serves index/search/cat, RSS and boot numbers above.
- Dependency edges were discovered the hard way first (boot failures naming each missing dep), then the whole graph extracted and the closure checked in one pass; the wired Scala list is byte-identical to the boot-verified probe list (diffed, 48/48).
- The real oracle is this PR's ready matrix: 10 ES versions run their full integration suites against stripped images. A version-specific dependency difference fails loudly at boot with the module's name.

`tests-utils:compileScala`, `formatCodeCheck`, `license` green locally.

JIRA: [RORDEV-2179](https://readonlyrest.atlassian.net/browse/RORDEV-2179)

## Summary by CodeRabbit

* **New Features**
  * Elasticsearch Docker images are now slimmed by default by removing unused modules.
  * Set `ROR_ES_SLIM_MODULES` to anything other than `true` to retain all modules.
* **Documentation**
  * Added guidance for running Docker-based integration tests, including version and module selection, commands, environment settings, sharding, resource limits, debugging, and CI options.
  * Centralized integration test execution and configuration details in the integration test documentation.


sscarduzio:
The matrix did exactly its job: es717x + all 8.x/9.x legs failed at boot with

```
unknown setting [cluster.deprecation_indexing.enabled]
```

**Cause:** our own test config disables subsystems whose modules this PR strips. With the registering module gone, its setting key becomes unknown and ES refuses to start. The version split matched perfectly — that entry is only added ≥7.16, so 6.7–7.10 passed.

**Fix:** the four disable-entries whose owners are stripped (`ingest.geoip.downloader.enabled`, `xpack.monitoring.enabled`, `xpack.apm_data.enabled`, `cluster.deprecation_indexing.enabled`) now apply only when slim mode is off — stripping supersedes disabling.

`xpack.watcher.enabled: false` stays unconditional on purpose: I boot-tested the stripped image with that setting present, and it starts — the key is registered by x-pack-core, not the watcher module. The kept entries (`ml`, `slm`, `ilm`, `stack.templates`) all belong to modules the slim image keeps.

sscarduzio:
Second matrix catch, second fix: `integration_es70x` failed with 60× `invalid_index_name_exception: Invalid index name [_rollup]` — `BaseXpackApiSuite` **creates rollup jobs**, so `x-pack-rollup` was wrongly on the strip list (my usage census missed it because the endpoint lives behind `XpackApiManager`, not a literal `_rollup` string in the suite). With the module stripped, the `_rollup` route disappears and ES parses the path as an index name.

`x-pack-rollup` is back on the keep list (47 stripped now), keep-list comment updated. Only es70x tripped because the newer legs' rollup tests are version-gated, so this was exactly the cross-version coverage the matrix exists for.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -28,6 +28,67 @@ import tech.beshu.ror.utils.misc.{JDK, Version}
 
 object Elasticsearch {
 
+  // Opt-out (ROR_ES_SLIM_MODULES=false) rather than opt-in: CI and local runs then boot the SAME
+  // image, so "passes locally, fails in CI" cannot come from the module set.
+  private[images] def slimModulesEnabled: Boolean =
+    Option(System.getenv("ROR_ES_SLIM_MODULES")).forall(_.equalsIgnoreCase("true"))
```
coutoPL, line 32: it's another env we use to set up tests somehow. Let's have all of the envs documented somewhere. Is there a Readme file dedicated for integration tests? Or another file that the information fits well?

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -28,6 +28,67 @@ import tech.beshu.ror.utils.misc.{JDK, Version}
 
 object Elasticsearch {
 
+  // Opt-out (ROR_ES_SLIM_MODULES=false) rather than opt-in: CI and local runs then boot the SAME
```
coutoPL: Unnecessary comment

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -28,6 +28,67 @@ import tech.beshu.ror.utils.misc.{JDK, Version}
...
+  // image, so "passes locally, fails in CI" cannot come from the module set.
+  private[images] def slimModulesEnabled: Boolean =
+    Option(System.getenv("ROR_ES_SLIM_MODULES")).forall(_.equalsIgnoreCase("true"))
+
+  // ES modules no integration suite exercises. Kept (with their `extended.plugins` dependency
```
coutoPL: I have doubts about this comment too.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -28,6 +28,67 @@ import tech.beshu.ror.utils.misc.{JDK, Version}
...
+  // closures): sql/esql (esql needs x-pack-ml, ml needs x-pack-autoscaling), async-search, ilm/slm,
+  // data-streams, reindex, painless/mustache, security, kibana, analytics, aggregations, mappers,
+  // percolator, rollup (BaseXpackApiSuite creates rollup jobs), snapshots over the fs repository,
+  // and `apm` (the server-cli launcher requires the
+  // directory to exist). `rm -rf` ignores names a given ES version does not ship.
+  private[images] val modulesUnusedByTests: List[String] = List(
```
coutoPL: maybe `modulesToRemoveInSlimModuleMode`?
or sth like that

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -183,6 +244,13 @@ class Elasticsearch(val esVersion: String, val config: Config, val plugins: Seq[
       // Package tar is required by the RorToolsAppSuite, and the ES >= 9.x is based on
       // Red Hat Universal Base Image 9 Minimal, which does not contain it.
       .runWhen(Version.greaterOrEqualThan(esVersion, 9, 0, 0), "microdnf install -y tar")
+      // Modules no suite uses still load classes at boot; stripping them cuts a test container from
```
coutoPL: This code is self-documenting. The comment is useless here.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -183,6 +244,13 @@ class Elasticsearch(val esVersion: String, val config: Config, val plugins: Seq[
...
+      // 1.07GiB to 829MiB RSS (measured, ES 8.18). `rm -rf` ignores names absent in a version, so one
+      // list serves all majors. Early in the layer order: shared by every suite image of a version.
+      .runWhen(
```
coutoPL: IMO, we should have an extension that will gather all the "slim modules"-related settings. 
It should look like this:
`.enableSlimModules(when = Elasticsearch.slimModulesEnabled)`

coutoPL (changes requested):

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Added `integration-tests/README.md` with a table of every knob — env vars (`ROR_ES_SLIM_MODULES`, `ROR_ES_JDWP`, `ROR_ES_CONTAINER_MEMORY_MB`, `ROR_HEAVY_SUITE_PERMITS`, `ROR_BALANCED_SHARDS`, `IT_PARALLELISM`, `IT_ORCHESTRATOR_JVMARGS`) and gradle properties (`esModule`, `esVersion`, `shardCount`/`shardIndex`, `itTestHeap`). The code comments now point there instead of re-explaining.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Removed.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Trimmed to one line pointing at the README, where the keep/strip rationale now lives.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Renamed to `modulesToRemoveInSlimMode`.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Removed.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Done: `.enableSlimModules(when = Elasticsearch.slimModulesEnabled, esDir = config.esDir)` — an extension on `DockerImageDescription` that owns the strip step and the module list. The four elasticsearch.yml entries guarded by `!slimModulesEnabled` stay where they are: they belong to the config builder chain, not the image chain, so the extension could not gather them without crossing builders.

sscarduzio:
Third matrix catch, and this one only the **full 34-leg** run could find (I dispatched it on the stacked branch #1341, since a feature-based PR gets no CI — see that PR).

es85x/87x/88x/89x/816x failed at `warmDockerImageCache`, all with the ROR plugin install returning exit 1 inside the slim image. Reproduced locally, and it is **not ROR-specific** — on a slim 8.7.1 image even `elasticsearch-plugin install analysis-icu` fails:

```
Exception in thread "main" java.lang.IllegalArgumentException:
  Missing plugin [x-pack-aggregate-metric], dependency of [x-pack-rollup]
```

**Cause:** exactly the version-specific dependency difference this list has to survive. In 8.18/9.x `x-pack-rollup` declares only `x-pack-core`; in **8.5 and 8.7 it also declares `x-pack-aggregate-metric`**. When I kept rollup (needed by `BaseXpackApiSuite` on es70x), its *older* dependency stayed stripped — and the plugin installer, not just the server, validates the module graph.

**Fix, done systematically rather than edge-by-edge:** I extracted the module dependency graph from **8 ES versions** (6.7.2, 7.10.2, 7.17.13, 8.5.3, 8.7.1, 8.16.0, 8.18.0, 9.0.8) and checked the keep-set closure against every one. Exactly one broken edge existed — `x-pack-aggregate-metric` — and after keeping it the closure is **clean across all versions**. 47 modules stripped now.

Verified on the precise failing case: `elasticsearch-plugin install analysis-icu` on the fixed slim 8.7.1 image → `Installed analysis-icu`.

Gates green; still one commit.

integration-tests/README.md
```diff
@@ -0,0 +1,34 @@
+# Integration tests
+
+Docker-based suites (TestContainers). Run one ES module's suites with:
+
+```bash
+./gradlew integration-tests:test -PesModule=es818x
```
coutoPL: It's an example. We should use `<es_module>` or something like that instead of `es818x`. And explain what the `es_module` can be.

integration-tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+See `CLAUDE.md` for the execution model (serial suites per JVM, sharding for parallelism).
```
coutoPL: No! It should be the other way around. The CLAUDE.md should reference this file.

integration-tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Knobs
+
+All are optional; defaults reproduce CI behaviour except where noted.
```
coutoPL: All? The ones below. You should be stricter. It's misleading now.

integration-tests/README.md
coutoPL: I see it's described below together with all other options. Nevertheless, this one should be refactored.

integration-tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Environment variables
+
+| Variable | Default | Meaning |
+|---|---|---|
+| `ROR_ES_SLIM_MODULES` | `true` | Strip the ES modules no suite uses from the test images (measured: −25% container RSS, slightly faster boot). Same default locally and in CI, so the module set can never cause "passes locally, fails in CI". Set `false` to test against a stock ES image, e.g. when experimenting with a feature whose module the slim image drops. |
```
coutoPL: "measured: −25% container RSS, slightly faster boot" 

This is a useless comment. Related rather to the PR description. Hard to maintain. Easily becoming out-of-date.

integration-tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+| `ROR_ES_JDWP` | `true` | Adds a JDWP agent to every ES test container (attach a debugger on port 8000). CI sets `false`: nothing attaches there and the agent costs ~90MB RSS per container. |
+| `ROR_ES_CONTAINER_MEMORY_MB` | `2048` | Hard memory limit per ES container. |
+| `ROR_HEAVY_SUITE_PERMITS` | off | Machine-wide cap on concurrently running multi-container ("heavy") suites, enforced across shard JVMs via file locks. CI sets `2` on 16GB runners. |
```
coutoPL: I'm not sure about saying what CI sets in this file. A description of why CI sets, e.g. 2, should be written in the place where the value is set.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -26,8 +26,69 @@ import tech.beshu.ror.utils.containers.images.Elasticsearch.Plugin.{PluginInstal
 import tech.beshu.ror.utils.containers.windows.WindowsEsDirectoryManager
 import tech.beshu.ror.utils.misc.{JDK, Version}
 
+extension (image: DockerImageDescription)
```
coutoPL: IMO, for the sake of readability, we should move the extension below (after the code it's used in)

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -26,8 +26,69 @@ import tech.beshu.ror.utils.containers.images.Elasticsearch.Plugin.{PluginInstal
...
+
+  private[images] def enableSlimModules(when: Boolean, esDir: Path): DockerImageDescription =
+    image.runWhen(
```
coutoPL: all the code related to the enabled slim modules should be moved in here. E.g.
```
.addWhen(
        Version.greaterOrEqualThan(esVersion, 8, 15, 0) && !Elasticsearch.slimModulesEnabled,
        entry = "xpack.apm_data.enabled: false"
      )
```

We only add this entry when the slim modules are enabled, so it'd be good to gather code like this in one place.

coutoPL (changes requested):

integration-tests/README.md
sscarduzio: Reversed. `CLAUDE.md` now points here ("Knobs ... are documented in integration-tests/README.md"), and the README states the execution model itself instead of deferring to CLAUDE.md.

integration-tests/README.md
sscarduzio: You are right — "All are optional" was a claim about things not on the page. Now: "The variables and properties below are optional."

integration-tests/README.md
sscarduzio: Refactored: the intro no longer half-documents `-PesModule` in prose while the table documents it properly. It is now just the command you run, and the table below is the single description of every option.

integration-tests/README.md
sscarduzio: Removed. Benchmark numbers belong in the PR description, not in a file nobody will update when the numbers move.

integration-tests/README.md
sscarduzio: Dropped every "CI sets X" from the README — the value and its justification live together at the place that sets it (`ci.yml`), so they cannot drift apart. The README now documents only what each knob means and its default.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Moved. Both extensions now sit at the bottom of the file, after the code that uses them.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Done — all slim-mode code is in one place now. The four `!slimModulesEnabled` entries scattered through the config chain are gone, replaced by a single `.disableSubsystemsOfModulesKeptOutsideSlimMode(esVersion)` extension that holds them together with the version guards; the image side keeps `.enableSlimModules(esDir)`, and both sit next to each other under one comment explaining why the two halves are opposites (a setting whose module was stripped is an *unknown* setting, so ES will not boot).

Verified the refactor is behaviour-preserving: every emitted key (`ingest.geoip.downloader`, `xpack.monitoring`, `xpack.apm_data`, `cluster.deprecation_indexing`, plus the untouched `ml`/`watcher`/`stack.templates`/`slm`/`destructive_requires_name`) appears exactly as before, with the same version guards.

integration-tests/README.md
sscarduzio: Reversed: `CLAUDE.md` now points at `integration-tests/README.md` for the knobs, and the README no longer refers back to it.

integration-tests/README.md
sscarduzio: Tightened — it now says the variables and properties in the tables below are optional, instead of claiming anything about "all".

integration-tests/README.md
sscarduzio: Removed. Agreed it belongs in the PR description, not in a file that has to stay true.

integration-tests/README.md
sscarduzio: Removed all "CI sets X" statements from the README. The reasoning for a specific CI value stays where the value is set (`ci.yml` / `ci/run-pipeline.sh`), which is the only place it can stay accurate.

integration-tests/README.md
sscarduzio: Refactored along with the rest of that section.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Moved: both extensions now sit at the bottom of the file, after the code that uses them.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: Done — and it turned out better than a simple move. All four of those `!slimModulesEnabled` entries are now gathered in `disableSubsystemsOfModulesKeptOutsideSlimMode(esVersion)`, one call in the config chain.

Two extensions rather than one, because the two halves live on different builders: `enableSlimModules` on the image chain removes the module directories, and this one on the `EsConfigBuilder` chain skips the disable-entries for modules that are no longer installed (a setting owned by a stripped module is an UNKNOWN key and ES refuses to boot). They sit next to each other with a single comment explaining the pairing, so the whole slim-mode story is in one place to read.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -299,9 +353,8 @@ class Elasticsearch(val esVersion: String, val config: Config, val plugins: Seq[
         entry = s"cluster.initial_master_nodes: ${config.masterNodes.toList.mkString(",")}",
         orElseEntry = "node.master: true"
       )
-      .addWhen(Version.greaterOrEqualThan(esVersion, 7, 14, 0), entry = "ingest.geoip.downloader.enabled: false")
+      .disableSubsystemsOfModulesKeptOutsideSlimMode(esVersion)
```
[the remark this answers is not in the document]
sscarduzio: Correct, and thanks — this is a real defect, and it predates the refactor: the guards I added to fix the "unknown setting" boot failure keyed off the **global flag**, while only `toOfficialEsImageBasedDockerImageDescription` actually strips modules. The apt/Ubuntu image (and the native Windows path) keep every module, so with the flag on by default they were losing the GeoIP, monitoring, APM and deprecation-indexing settings for no reason.

Fixed by making the question "does THIS installation strip modules?" instead of "is slim mode on?":

```scala
def modulesAreStripped: Boolean =
  Elasticsearch.slimModulesEnabled && (config.esInstallationType match {
    case EsInstallationType.EsDockerImage                  => true
    case EsInstallationType.UbuntuDockerImageWithEsFromApt => false
    case EsInstallationType.NativeWindowsProcess           => false
  })
```

Both halves now consume it — `.enableSlimModules(esDir, when = config.modulesAreStripped)` and `.disableSubsystemsOfStrippedModules(esVersion, whenModulesArePresent = !config.modulesAreStripped)` — so the image and the config can no longer disagree about what is installed. It sits beside `esDir`/`esConfigDir`, which already switch on the installation type, and the exhaustive match means a new installation type has to state its answer.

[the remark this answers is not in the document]
sscarduzio: Correct, and a real defect — thank you. `enableSlimModules` runs only in `toOfficialEsImageBasedDockerImageDescription`, but `esConfigFile` is shared by all three installation types, so the apt/Ubuntu image and the native Windows process kept every module **and** lost the four disable-entries. Those subsystems (geoip downloader, monitoring, apm-data, deprecation indexing) would have quietly come back on for exactly the installs that still have them.

Fixed at the root rather than by threading a flag through: the decision is now a property of the installation, `config.modulesAreStripped`, with an exhaustive match so a future `EsInstallationType` cannot silently default to the wrong answer:

```scala
def modulesAreStripped: Boolean =
  Elasticsearch.slimModulesEnabled && (config.esInstallationType match {
    case EsInstallationType.EsDockerImage                  => true
    case EsInstallationType.UbuntuDockerImageWithEsFromApt => false
    case EsInstallationType.NativeWindowsProcess           => false
  })
```

Both sides read it: the image chain strips only when it is true, and the config chain keeps the disable-entries whenever the modules are still present.

sscarduzio:
The red legs here are **not this PR** — they are Docker Hub rate limiting, and the root cause is worth fixing centrally.

```
toomanyrequests: You have reached your unauthenticated pull rate limit
```

**Why "unauthenticated", when the job logs `Docker Hub authenticated pulls ENABLED`:** `ci/docker-hub-auth.sh` exports `DOCKER_AUTH_CONFIG`, which is a **testcontainers** convention. It never runs `docker login`, so anything going through the plain `docker` CLI is anonymous. `docker_image_exists()` in `ci-lib.sh` is exactly that:

```bash
docker_image_exists() { docker manifest inspect "$1" >/dev/null 2>&1; }
```

The e2e jobs poll it in a loop, per ELK version, waiting for `beshultd/kibana-readonlyrest-dev:<tag>` — so each e2e run spends anonymous quota, and once the per-IP limit is hit the *integration* legs start failing too (they pull the WireMock/ES base images).

@coutoPL — this is in the new e2e path, so flagging rather than changing it under you. A `docker login` in `docker-hub-auth.sh` (guarded by the same secret check) would authenticate the CLI as well as testcontainers, and cost nothing when the secrets are absent.

My share of the blame: I dispatched the full 34-leg matrix twice yesterday to prove the stacked #1341 combination, which will have consumed a good part of the window. I have stopped doing that.

This PR's own code is unaffected — its diff is 3 files (`CLAUDE.md`, `integration-tests/README.md`, `Elasticsearch.scala`), and the e2e stack builds its ES from the `docker-image-conventions` path, not from the testcontainers image builder this PR touches.

integration-tests/README.md
sscarduzio: Now a placeholder, with the explanation next to it:

```bash
./gradlew integration-tests:test -PesModule=<es_module>
```

> An `<es_module>` is one of the `esNNx` directories in the repository root — `es67x`, `es717x`, `es818x`, `es94x` and so on. Each one supports a range of ES versions, and the suites run against the newest version in that range unless `-PesVersion` says otherwise. To list the modules of a given ES major:
> ```bash
> ./gradlew printEsModules -PesMajor=8 --quiet
> ```
> Omit `-PesModule` to run against the module that publishes the newest supported ES version.

I ran that command before documenting it — it prints the 17 modules of major 8, `es818x` down to `es80x`.

Apologies for the duplicate replies across this review: I had two copies of my PR watcher running and each answered every thread. Both are stopped.

sscarduzio:
Refining the diagnosis — the failure mode is worse than "the CLI is anonymous", and it now affects **every** open PR (#1313 and #1340 both have 13 red legs).

**`ci/docker-hub-auth.sh` reports success without ever verifying it.** It checks only that the two variables are non-empty, then prints:

```
[TEST] Docker Hub authenticated pulls ENABLED (user '***')
```

Nothing validates the credentials. So an **expired or revoked `DOCKER_HUB_RO_TOKEN` degrades silently to anonymous pulls** while the log still says ENABLED — which is exactly the contradiction in these runs: the job claims authenticated, and Docker Hub answers `unauthenticated pull rate limit`.

Two independent gaps, then:

1. **The `docker` CLI is never authenticated.** `DOCKER_AUTH_CONFIG` is a testcontainers convention; `docker_image_exists()` (`docker manifest inspect`) and any script-level `docker pull` are anonymous. The e2e jobs poll that in a loop, per ELK version.
2. **Credentials are never validated**, so failure is invisible.

A `docker login` inside the existing secret-guarded block fixes both: it authenticates the CLI, testcontainers falls back to `~/.docker/config.json`, and a bad credential fails **loudly** at that point instead of silently halving throughput.

@coutoPL @sscarduzio — worth checking whether `DOCKER_HUB_RO_TOKEN` is still valid, since that would explain why this started biting now. I have not touched the script: it is shared CI infrastructure and the token question is yours. Happy to open the `docker login` PR on request.

coutoPL:
ref: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1346

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
```diff
@@ -432,3 +487,28 @@ final case class EsJavaOptsBuilder(options: Seq[String]) {
...
+
+  private[images] def enableSlimModules(esDir: Path, when: Boolean): DockerImageDescription =
+    image.runWhen(
+      when,
+      s"cd ${esDir.toString()}/modules && rm -rf ${Elasticsearch.modulesToRemoveInSlimMode.mkString(" ")}"
+    )
+
+extension (config: EsConfigBuilder)
+
+  private[images] def disableSubsystemsOfStrippedModules(
```
coutoPL: do we have to have two separate methods? I mean `enableSlimModules` and `disableSubsystemsOfStrippedModules`? Cannot we have one that do all of that things?

CLAUDE.md
```diff
@@ -49,7 +49,7 @@ ReadonlyREST is an Elasticsearch security plugin providing access control, authe
...
+- **`integration-tests/`** — Docker-based integration tests using TestContainers. Suites run **serially within a worker JVM** (they share one mutable singleton ES, guarded by an acquire/release latch), so `maxParallelForks = 1` always (forks-as-parallelism is a confirmed dead end — Gradle's JUnit-Platform engine runs the scalatest Launcher once per worker, so `maxParallelForks`/`forkEvery` can't split it; gradle/gradle#8632). Parallelism instead comes from **suite SHARDING**: K separate `integration-tests:test` invocations (`-PshardCount=K -PshardIndex=i`), each a fresh JVM with its own singleton ES, over a disjoint name-hash partition of the suites. `IT_PARALLELISM` (default 1) drives it in CI; at K=1 everything runs in one JVM. Knobs (env vars, gradle properties) are documented in `integration-tests/README.md`.
```
coutoPL: IMO, we have too much info at this level. We have all of them in the integration-tests/README.md, so maybe the reference is enough? WDYT?

sscarduzio: Agreed — trimmed. That bullet now says what the module is, that suites run serially per worker JVM with parallelism from sharding, and points at `integration-tests/README.md` for the execution model, the env vars and the gradle properties.

Nothing was lost by cutting it: the non-obvious part (why forks cannot replace sharding — the scalatest engine runs the JUnit-Platform Launcher once per worker, gradle#8632) already sits next to `maxParallelForks = 1` in `integration-tests/build.gradle`, which is where someone tempted to change it will actually be looking.

tests-utils/src/main/scala/tech/beshu/ror/utils/containers/images/Elasticsearch.scala
sscarduzio: I tried, and one method is not possible — the two act on **different builders, in different chains, at different moments**:

```scala
enableSlimModules(esDir, when):        DockerImageDescription => DockerImageDescription   // line 248, image chain
disableSubsystemsOfStrippedModules(…): EsConfigBuilder        => EsConfigBuilder          // line 365, elasticsearch.yml chain
```

One removes directories from the image; the other decides which lines the config file must NOT contain. There is no single value they could both thread through, and folding either into the other would mean the image chain reaching into config generation (or vice versa).

What I *can* do, if the goal is one concept in one place rather than literally one method: wrap both in a `SlimModules` object so there is a single named home for them. The cost is losing the fluent `.enableSlimModules(...)` form the rest of this file uses. As it stands they are adjacent, share one comment explaining the pairing, and both derive from the single predicate `config.modulesAreStripped` — so the decision exists once even though it is applied twice. Tell me if you would rather have the object and I will change it.

sscarduzio: Yes — and the honest answer turned out to be better than merging them. The second method existed only because stripping four modules invalidated four settings we set. So I asked whether those four are worth stripping at all, and measured it (ES 8.18, same config, side by side):

| image | RSS |
|---|---|
| strips 46 modules (incl. the four) | 1008 MiB |
| strips 42, **keeps** `ingest-geoip`, `x-pack-monitoring`, `x-pack-apm-data`, `x-pack-deprecation` | **997 MiB** |

They cost nothing — the difference is noise, in the wrong direction. So they stay, and:

* `disableSubsystemsOfStrippedModules` is **deleted**; the four settings go back to plain version guards that know nothing about slim mode
* `modulesAreStripped` and the installation-type coupling are **deleted** too — `enableSlimModules` is only ever called from the official-image builder, so that check was redundant
* one extension remains, exactly as you asked

It also removes the bug class CodeRabbit found on this PR: with no coupling between image and config, the apt/Ubuntu path cannot disagree with the image any more — the defect is now impossible rather than fixed.

Verified on the original failure case: `elasticsearch-plugin install` works on the new image, and 8.7.1 boots clean **with** `cluster.deprecation_indexing.enabled` and `ingest.geoip.downloader.enabled` set — the exact "unknown setting" failure that started this.

CLAUDE.md
sscarduzio: Agreed — trimmed to one line: `Docker-based integration tests using TestContainers. Execution model, knobs and sharding: integration-tests/README.md`. The detail now lives in the README only, so the two cannot drift.

sscarduzio:
Rebased onto develop, no code change. The old branch had the ci.yml from before #1347, which reads secret names that do not exist any more. Because of that, each rerun failed at once with "Must provide --username with --password-stdin". The two red e2e legs before the rebase were not this PR: es94x died when Kibana hit its Node heap limit at startup, and es818x failed one known-flaky Tenancy short-URL test.

sscarduzio:
@coderabbitai review

integration-tests/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Knobs
+
+The variables and properties below are optional.
+
+### Environment variables
+
+| Variable | Default | Meaning |
+|---|---|---|
+| `ROR_ES_SLIM_MODULES` | `true` | Strip the ES modules no suite uses from the test images. Set `false` to run against a stock ES image, e.g. when adding a suite for a feature whose module the slim image drops. |
```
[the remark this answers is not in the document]
sscarduzio: Valid, and fixed. Checked against the code first: `enableSlimModules` is called once, inside `toOfficialEsImageBasedDockerImageDescription`. The apt/Ubuntu builder does not call it, and the native Windows process is not a docker image at all. So "the test images" did read as all of them, and only one is stripped.

The row now says:

> Strip the ES modules no suite uses from the test image. This applies to the official Elasticsearch Docker image only — the Ubuntu image with ES from apt, and the native Windows process, keep every module.

coutoPL (approved):
LGTM

PR #1341, opened 2026-08-17
[RORDEV-2180] Release each shard's singleton ES at the phase boundary

sscarduzio:
Releases each shard's shared ES as soon as the suites that need it are finished, instead of letting it idle until the shard ends.

## The problem

A suite that uses the shared ES keeps it alive until its JVM exits. A shard that ran those suites first and then went on to the heavy own-cluster suites was therefore still holding an idle ES underneath them — at exactly the point in the run where the host has least memory to spare.

## The change

A shard now runs its suites as two `Test` tasks, so the JVM boundary releases the ES:

1. `integration-tests:sharedEsSuitesTest` — the suites that share the ES. The ES dies with this task's JVM.
2. `integration-tests:test` — the remaining suites, in a fresh JVM that never starts a shared ES.

`test.mustRunAfter 'sharedEsSuitesTest'` declares the order, because naming both tasks on the command line does not guarantee it. A shard whose partition holds no suite for one step disables that step, so a shard runs one task or two. Unsharded, `test` alone still runs every suite, exactly as before.

Splitting the suites needs to know which ones use the shared ES **at configuration time**, before any test classpath exists. `SuiteClassGraph` answers that by READING class files with ASM rather than loading them: it inverts the hierarchy into a direct-subtype map, then sweeps down from the marker type once. Loading would resolve against a runtime classpath that does not reliably exist yet, because a project dependency's jar is built later.

## Files

| File | Why |
|---|---|
| `integration-tests/build.gradle` | the two-task split, the partition, disjoint per-shard report dirs |
| `build-base/.../SuiteClassGraph.java` | which suites use the shared ES, read from class files |
| `build-base/.../SuiteClassGraphTest.java` | direct, transitive, unrelated, and parent-outside-the-tree |
| `build-base/.../ShardedGradlewTest.java` | each shard child runs both tasks, in order |
| `integration-tests/README.md` | the execution model, where `CLAUDE.md` points |
| `tests-utils/.../Elasticsearch.scala` | shared-ES lifecycle |

## Verification

- `SuiteClassGraphTest` 2/2 green. Negative control: dropping the interface edge from `declaredSupertypesOf` fails 1 of the 2 cases, so they do exercise the traversal rather than passing vacuously.
- The 4-shard partition of es94x, as CI actually ran it (run 32348377952, balanced packing), is **14 / 11 / 11 / 12** shared-ES suites against **5 / 5 / 5 / 4** own-cluster — 19+16+16+16 = 67 suites, a disjoint cover. Both tasks ran in every shard.

  Earlier revisions of this description quoted these numbers twice from local dry-runs, and got them wrong both times (14/11/11/12 from a stale tree, then 16/12/11/9). The figures above are read from the CI log of this branch, which is the only measurement that describes what the change does in production.
- `javac -Xlint:all -Werror` clean; `google-java-format` with the build's own settings reports no violation.

## Review history

Thanks to @coutoPL for three rounds that each improved this:
- the task name and its comment now say *what* it does and *why* it exists, at the registration where a reader first looks — `singletonPhaseTest` named an implementation detail;
- `SuiteClassGraph` inverts the map and sweeps once, instead of walking up from every candidate with a memo;
- the `singleton` vocabulary is gone from `build.gradle`, which had outlived the rename.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sharded integration-test execution by running shared-environment suites before isolated-cluster suites.
  - Prevented empty test filters from unintentionally running all suites.
  - Preserved required Elasticsearch modules for non-Docker installations.
  - Improved compatibility across supported Elasticsearch installation types.

- **Tests**
  - Added coverage for discovering direct and transitive test-suite relationships.

- **Documentation**
  - Documented the two-phase sharded test execution process and suite ordering.


sscarduzio:
Heads-up on a CI gap this PR exposed, worth knowing beyond this PR:

`.github/workflows/ci.yml` limits the PR trigger to `branches: [master, develop, 'epic/**', '**/epic/**']`. A PR whose **base is a feature branch gets no CI at all** — this one sat with 3 trivial checks and zero integration legs, silently. Any future stacked PR has the same blind spot.

Rather than retarget this PR (which would show #1340's diff mixed in) or change the trigger list unilaterally, I dispatched the workflow directly on this branch — it contains both changes, so this is the combined proof:

**https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runs/32036409544** — the FULL 34-leg Linux matrix (`run_all_tests_on_linux`), i.e. broader coverage than a PR run's 10-leg ready set: slim images + phase-split singleton release, every ES version from 6.7 to 9.4.

@coutoPL — worth deciding separately whether `pull_request.branches` should include `feature/**` so stacked PRs get CI automatically. I did not change it here; it affects every PR.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
```diff
@@ -0,0 +1,148 @@
...
+ * Answers "which of these classes extend/mix in the given marker type?" by PARSING class files
+ * (constant pool: super class + interfaces), never by loading them. Loading would resolve the full
+ * hierarchy against the runtime classpath, which does not reliably exist at Gradle configuration
+ * time (a project dependency's jar only materializes at execution).
+ *
+ * <p>Sound for markers defined IN the scanned tree: a path to the marker can only run through
+ * classes of the same source set (a dependency of the tree cannot extend a type defined in it), so
+ * an edge leaving the scanned tree can be pruned.
+ */
+public final class SuiteClassGraph {
```
coutoPL, line 46: Let's create the class in a less imperative way.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+      }
+    }
+    return parents;
+  }
+
+  private record ParsedClass(String name, List<String> parents) {}
+
+  /** Minimal class-file reader: constant pool, then this/super/interfaces. Nothing else is read. */
+  private static ParsedClass parse(DataInputStream in) throws IOException {
```
coutoPL: OMG :D Let's use `asm` instead

integration-tests/build.gradle
```diff
@@ -48,24 +48,45 @@ def newestEsVersion = tech.beshu.ror.gradle.utils.EsModuleFinder
 def defaultEsModule = tech.beshu.ror.gradle.utils.EsModuleFinder
         .findEsModuleFor(rootProject, newestEsVersion).orElseThrow().name
 
-test {
-    systemProperty "project.dir", rootProject.projectDir
+// Suites that mix in SingletonPluginTestSupport share the per-JVM singleton ES, which lives until
```
coutoPL: In this comment, I see things that were already described. We don't have to repeat things over and over.

integration-tests/build.gradle
```diff
@@ -48,24 +48,45 @@ def newestEsVersion = tech.beshu.ror.gradle.utils.EsModuleFinder
...
+// JVM exit. A shard therefore used to hold its singleton UNDER every own-cluster suite it ran. The
+// shard now runs as two SEQUENTIAL Test tasks: `singletonPhaseTest` (the singleton suites — the
+// singleton dies WITH that JVM) and then `test` (the rest — a fresh JVM that never lazily starts a
+// singleton). The JVM boundary is the release mechanism: no stop hook, no restart path, and the
+// singleton lifecycle itself is untouched. Phase split applies only to sharded runs; a plain
+// `integration-tests:test` behaves exactly as before.
+def partitionBySingletonUsage = { org.gradle.api.tasks.testing.Test task, List<String> suiteNames ->
+    // Hierarchy walk over PARSED class files (build-base's SuiteClassGraph, pure + unit-tested), not
```
coutoPL: Do we need this comment?

integration-tests/build.gradle
```diff
@@ -129,26 +167,33 @@ test {
...
-    maxHeapSize = project.findProperty('itTestHeap') ?: '512m'
+    task.maxHeapSize = project.findProperty('itTestHeap') ?: '512m'
     // bounded like every other JVM in the CI leg, so a runaway fails in-process, not via the host OOM killer
-    jvmArgs '-XX:MaxMetaspaceSize=512m'
+    task.jvmArgs '-XX:MaxMetaspaceSize=512m'
 }
 
+test { configureIntegrationTest(it, false) }
+tasks.register('singletonPhaseTest', Test) { configureIntegrationTest(it, true) }
+// Declarative ordering (CLI order alone is not a contract): in a shard child both tasks are
```
coutoPL: review all comments added in the PR and decide if they bing additional context, or are just a noise

integration-tests/build.gradle
```diff
@@ -129,26 +167,33 @@ test {
...
     // live heap is ~50MB, so 512m is ~10x headroom. Overridable via -PitTestHeap for outlier suites.
-    maxHeapSize = project.findProperty('itTestHeap') ?: '512m'
+    task.maxHeapSize = project.findProperty('itTestHeap') ?: '512m'
     // bounded like every other JVM in the CI leg, so a runaway fails in-process, not via the host OOM killer
-    jvmArgs '-XX:MaxMetaspaceSize=512m'
+    task.jvmArgs '-XX:MaxMetaspaceSize=512m'
 }
 
+test { configureIntegrationTest(it, false) }
+tasks.register('singletonPhaseTest', Test) { configureIntegrationTest(it, true) }
```
coutoPL: yes, and in the place where I would like to read what the `singletonPhaseTest` is, there is no comment. 

This is the place where I expect the comment that describes what the task does. `What`, not `How` - to be clear. 

We have this in many places (wrong ones), but not in the most important one.

coutoPL (changes requested):

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: You are right, and it was worse than a style problem: **asm-tree is already a `build-base` dependency**, so I hand-rolled a constant-pool parser next to a library that was on the classpath the whole time.

Replaced with ASM: `ClassReader.accept(node, SKIP_CODE | SKIP_DEBUG | SKIP_FRAMES)` and then `node.superName` + `node.interfaces`. The ~50 lines of tag-by-tag parsing (including the long/double two-slot rule I had to special-case) are gone.

The existing unit tests pass **unchanged** — 2/2, covering direct mixin, transitive inheritance through a base class, an unrelated class, and a parent outside the scanned tree — which is what makes me confident the swap is behaviour-preserving rather than merely compiling.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Done as part of the ASM rewrite: the traversal is now stream-based (`filter`/`anyMatch`/`collect`) instead of nested loops with mutable accumulators, and the only remaining mutable state is the memo map that makes the reachability walk linear.

integration-tests/build.gradle
sscarduzio: Removed — it repeated what the block below already states.

integration-tests/build.gradle
sscarduzio: No. Removed.

integration-tests/build.gradle
sscarduzio: Went through every comment this PR adds and cut the ones that restate the code or the PR description. What is left is the non-obvious "why": that loading classes is impossible at configuration time, that an empty filter would run every suite, and that the ordering must be declared rather than implied by CLI argument order.

integration-tests/build.gradle
sscarduzio: Fair, and it was the wrong way round: plenty of commentary on mechanics, none where a reader first looks. The task registration now carries the *what* — a shard runs its suites in two phases so its singleton ES is released as soon as the suites needing it are finished, instead of idling until the shard ends; `singletonPhaseTest` runs the singleton suites, `test` runs the rest in a fresh JVM, and unsharded `test` alone still runs everything.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Agreed, and embarrassing in hindsight — `asm-tree` is **already a dependency of build-base**, so I hand-wrote a constant-pool reader the project could already do. Gone: it now reads the header with `ClassReader` into a `ClassNode` (`SKIP_CODE | SKIP_DEBUG | SKIP_FRAMES`) and takes `superName` + `interfaces`. ~60 lines of parsing deleted.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Rewritten in the same pass: the class is now a small declarative walk — `classesReaching` filters candidates through a memoised `reaches`, and the parent map is built by streaming the class files. No manual byte handling left.

integration-tests/build.gradle
sscarduzio: Removed. That block restated what the class doc and the task comment already say.

integration-tests/build.gradle
sscarduzio: Removed — it repeated `SuiteClassGraph`'s own javadoc.

integration-tests/build.gradle
sscarduzio: Went through every comment this PR adds and cut **18 lines**. What is left states only things the code cannot: that a registered `Test` task does not inherit `testClassesDirs`/`classpath` from the java plugin, that an *enabled* Test task with an empty filter silently runs every suite (twice, at both guard sites), and that task ordering has to be declared because passing both names on the command line does not guarantee it.

integration-tests/build.gradle
sscarduzio: You are right, and this was the real gap — the explanation was everywhere except where a reader looks for it. The task registration now carries it, and says *what*, not *how*:

```groovy
// A shard runs its suites in two phases, so its singleton ES is released as soon as the suites that
// need it are done, instead of idling until the shard ends.
//
// `singletonPhaseTest` runs the suites that share the per-JVM singleton ES; the singleton dies with
// this task's JVM.
// `test` then runs the remaining suites (each boots its own cluster) in a fresh JVM, which never
// starts a singleton.
//
// Unsharded, `test` alone runs everything, exactly as before.
tasks.register('singletonPhaseTest', Test) { ... }
```

I also moved the `singletonPhaseTest` registration **above** `test`, so the file reads in execution order.

Verified the ASM rewrite changes nothing: `SuiteClassGraphTest` 2/2 green, and the 4-shard dry-run partition is identical to before — 19=14+5, 16=11+5, 16=11+5, 16=12+4.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Took another pass. The ASM rewrite removed the hand-rolled parsing, but `parentsByClassIn` was still a loop with a mutable accumulator — the last imperative piece. It is now one stream:

```java
private static Map<String, List<String>> parentsByClassIn(Collection<Path> classDirs) {
  // flatMap closes each Files.walk stream when the outer stream closes, so one try suffices.
  try (Stream<Path> classFiles =
      classDirs.stream()
          .filter(Files::isDirectory)
          .flatMap(SuiteClassGraph::walk)
          .filter(file -> file.getFileName().toString().endsWith(".class"))) {
    return classFiles
        .map(SuiteClassGraph::read)
        .collect(Collectors.toMap(
            node -> binaryName(node.name),
            SuiteClassGraph::directParentsOf,
            (first, duplicate) -> first));
  }
}
```

Two things this fixes beyond style. The duplicate binary name across two class dirs used to resolve by silent overwrite; the merge function makes the tie explicit. And `classesReaching` no longer declares `throws IOException` — the class was already throwing `UncheckedIOException` from `read`, so the checked-exception plumbing was inconsistent, and it is now unchecked throughout.

The one mutable structure left is the `memo` map in `reaches`. That is the memoisation itself: without it the walk is exponential on a deep hierarchy, so I kept it.

Verified rather than assumed: `SuiteClassGraphTest` is 2/2 green, and I confirmed it can fail for the right reason — emptying the class-file stream makes `findsDirectAndTransitiveUsersAndIgnoresTheRest` fail, so the pass is real and not a vacuous match.

Apologies for the duplicate replies on this review: two copies of my PR watcher were running and each answered every thread. Both are stopped.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public final class SuiteClassGraph {
+
+  private SuiteClassGraph() {}
+
+  /** The subset of {@code candidates} (binary names) that reach {@code marker} via extends/implements. */
+  public static Set<String> classesReaching(
+      String marker, Collection<Path> classDirs, Collection<String> candidates) {
+    Map<String, List<String>> parents = parentsByClassIn(classDirs);
+    Map<String, Boolean> memo = new HashMap<>();
```
coutoPL: instead of walking up from every candidate and memoizing "does this reach the marker?", build the map the other way round (parent -> children) and do one downward BFS from the marker. Whatever that closure contains is
  the answer; the query becomes a set-membership filter

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    memo.put(name, false); // also breaks a cycle, which only a malformed tree could contain
+    boolean found =
+        parents.getOrDefault(name, List.of()).stream()
+            .anyMatch(parent -> reaches(parent, marker, parents, memo));
+    memo.put(name, found);
+    return found;
+  }
+
+  private static Map<String, List<String>> parentsByClassIn(Collection<Path> classDirs) {
```
coutoPL: I'm not sure about the naming domain picked in this class's implementations. We use Child/Parent naming, but we are looking for files with classes that extend/implement some class/interfaces. It's misleading and not easy to follow.

integration-tests/build.gradle
coutoPL: Still, the comments don't clearly tell what the task does. 

Take the reader's perspective. The reader doesn't have to know all the details of the solution, but after reading this comment, they should understand why it was added. 

Moreover, the `singletonPhaseTest` doesn't look like the best choice for the name. Now, the name doesn't tell much, nor does the comment.

coutoPL (changes requested):

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Better algorithm — adopted. The map is now built inverted (`supertype -> subtypes`) and one downward sweep from the marker collects the whole closure; the query is then a set-membership filter.

It removes the recursion, the memo map and its cycle guard, and it does the work once for the whole tree instead of once per candidate. The tests pass unchanged, including the transitive case, which is what tells me the traversal is equivalent.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Agreed — `Child`/`Parent` describes a tree, not `extends`/`implements`. The vocabulary is now the one the language uses:

- `subtypesOf(supertype, classDirs, candidates)` — the entry point
- `subtypesBySupertypeIn(classDirs)` — the inverted map
- `declaredSupertypesOf(node)` — `superName` plus `interfaces`
- `allSubtypesOf(...)` — the downward sweep

The test names followed (`findsDirectAndTransitiveSubtypesAndIgnoresTheRest`).

integration-tests/build.gradle
sscarduzio: Both fair. Renamed the task to **`sharedEsSuitesTest`** — it says what it runs, whereas `singletonPhaseTest` named an implementation detail ("phase") and left the reader to guess.

The comment now leads with why it exists rather than how it works:

> a suite that uses the shared ES keeps it alive until its JVM exits, so a shard that ran those suites first and then went on to the heavy own-cluster suites was still holding an idle ES underneath them. Splitting the shard into two Test tasks makes the JVM boundary release it […] Less memory held at the point in the run where the host needs it most.

Verified after the rename that the shard partition is unchanged: 14/11/11/12 across the four shards of es94x, the same numbers as before.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Correction to my own reply above: the partition numbers I quoted were wrong.

I said "14/11/11/12 across the four shards of es94x, the same numbers as before". Those came from a measurement taken two days ago, before develop moved, and I presented them as a fresh check. The real shared-ES counts on this branch are **16/12/11/9**.

The conclusion still holds, and I have now established it properly rather than by re-reading an old note. I ran the same command on both commits against the same tree — `c56aa2944` (walk up from each candidate, with a memo) and `a74416a33` (invert the map, one sweep down from the supertype):

```
shard 0/4  6 of 22 suites (16 shared / 6 own-cluster)
shard 1/4  4 of 16 suites (12 shared / 4 own-cluster)
shard 2/4  6 of 17 suites (11 shared / 6 own-cluster)
shard 3/4  3 of 12 suites ( 9 shared / 3 own-cluster)
```

Byte-identical between the two algorithms. 67 suites, 48 shared and 19 own-cluster, and every shard lands the same way. That is the behaviour-preservation evidence; "the unit tests still pass" was not enough on its own for an algorithm swap.

I also confirmed the unit test can still fail for the right reason: inverting the map the wrong way round (keying by the subtype instead of the declared supertype) makes `findsDirectAndTransitiveSubtypesAndIgnoresTheRest` fail. So the 2/2 pass is meaningful and not vacuous.

One more thing I checked, because a rename of this task broke every shard once before: `singletonPhaseTest` appears nowhere in the tree, and the new name reaches `ShardedGradlewTest.java:68`, which is the file that holds the task name as a plain string and is invisible to a Java refactor.

sscarduzio:
@coutoPL All three open threads are addressed as of `a74416a3`, so this is ready for another look:

- **Comment where the reader looks** — the task registration now carries the *what*, and the task is renamed `singletonPhaseTest` → `sharedEsSuitesTest`, since the old name described a phase (an implementation detail) rather than what it runs.
- **Downward BFS instead of memoized upward walks** — the map is built inverted and one sweep from the marker collects the closure; the recursion, memo map and cycle guard are gone.
- **extends/implements vocabulary** — `subtypesOf` / `subtypesBySupertypeIn` / `declaredSupertypesOf` / `allSubtypesOf` replace the `Child`/`Parent` tree language, test names included.

I verified each of these is actually on the branch rather than just described in a reply — the old names return no hits.

One correction still standing from my side: I earlier quoted shard partitions of 14/11/11/12 as a fresh measurement when they came from a stale run. The real shared-ES counts on this branch are 16/12/11/9. The conclusion is unchanged, but the number I gave you was wrong.

sscarduzio:
Rebased onto the rebased base branch, no code change. The base branch needed the ci.yml from #1347 (renamed CI secrets), so this branch moved with it.

sscarduzio:
@coderabbitai full review

CLAUDE.md
```diff
@@ -49,7 +49,7 @@ ReadonlyREST is an Elasticsearch security plugin providing access control, authe
...
 - **`ror-shadowed-libs/`** — Shaded dependencies (auto-relocated to `tech.beshu.ror` prefix via Shadow plugin to avoid classpath conflicts with ES internals).
-- **`integration-tests/`** — Docker-based integration tests using TestContainers. Execution model, knobs and sharding: `integration-tests/README.md`.
+- **`integration-tests/`** — Docker-based integration tests using TestContainers. Suites run **serially within a worker JVM** (they share one mutable singleton ES, guarded by an acquire/release latch), so `maxParallelForks = 1` always (forks-as-parallelism is a confirmed dead end — Gradle's JUnit-Platform engine runs the scalatest Launcher once per worker, so `maxParallelForks`/`forkEvery` can't split it; gradle/gradle#8632). Parallelism instead comes from **suite SHARDING**: K separate `integration-tests:test` invocations (`-PshardCount=K -PshardIndex=i`), each a fresh JVM with its own singleton ES, over a disjoint name-hash partition of the suites. `IT_PARALLELISM` (default 1) drives it in CI; at K=1 everything runs in one JVM. Knobs (env vars, gradle properties) are documented in `integration-tests/README.md`.
```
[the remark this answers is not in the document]
sscarduzio: Correct, and the cause was worse than a stale sentence: this branch **restored** that long bullet. RORDEV-2179 had already replaced it with a one-line pointer to `integration-tests/README.md`, at coutoPL's request, and my commit here carried the pre-trim copy forward and undid that.

So the fix is not to edit the bullet. It is to drop the hunk — `CLAUDE.md` keeps the one-liner — and to put the description in the README it points at, which is the file that now owns the execution model:

> Each shard runs its share in two steps, so it frees the shared ES as soon as the suites that need it are done, instead of holding it idle until the shard ends:
> 1. `integration-tests:sharedEsSuitesTest` — the suites that use the shared ES. The ES stops when this task's JVM exits.
> 2. `integration-tests:test` — the remaining suites, in a fresh JVM. Each one starts its own cluster, and this JVM never starts a shared ES.
>
> `shardedTest` starts both steps for every shard, in that order. Without sharding, `test` alone still runs every suite.

Checked against the code rather than the suggestion: `ShardedGradlewTest` adds `integration-tests:sharedEsSuitesTest` then `integration-tests:test` (lines 68-69), and `test.mustRunAfter 'sharedEsSuitesTest'` declares the order, because passing both names on the command line does not guarantee it.

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
```diff
@@ -0,0 +1,120 @@
...
+        if (reached.add(subtype)) {
+          pending.push(subtype);
+        }
+      }
+    }
+    return reached;
+  }
+
+  /** The hierarchy inverted: each type mapped to the types that declare it as a supertype. */
+  private static Map<String, List<String>> subtypesBySupertypeIn(Collection<Path> classDirs)
```
coutoPL: WDYT about doing it like this?
```java
  private static Map<String, List<String>> subtypesBySupertypeIn(Collection<Path> classDirs) {
    return classDirs.stream()
        .filter(Files::isDirectory)
        .flatMap(SuiteClassGraph::classFilesIn)
        .map(SuiteClassGraph::read)
        .flatMap(node ->
            declaredSupertypesOf(node).stream()
                .map(declared -> Map.entry(declared, binaryName(node.name)))
        )
        .collect(
            Collectors.groupingBy(
                Map.Entry::getKey, Collectors.mapping(Map.Entry::getValue, Collectors.toList())));
  }

  private static Stream<Path> classFilesIn(Path dir) {
    try {
      return Files.walk(dir).filter(file -> file.getFileName().toString().endsWith(".class"));
    } catch (IOException e) {
      throw new UncheckedIOException("cannot walk " + dir, e);
    }
  }
```

coutoPL: maybe a better name should be `directSubtypesBySupertypeIn`?

integration-tests/build.gradle
```diff
@@ -48,24 +48,34 @@ def newestEsVersion = tech.beshu.ror.gradle.utils.EsModuleFinder
 def defaultEsModule = tech.beshu.ror.gradle.utils.EsModuleFinder
         .findEsModuleFor(rootProject, newestEsVersion).orElseThrow().name
 
-test {
-    systemProperty "project.dir", rootProject.projectDir
+def partitionBySingletonUsage = { org.gradle.api.tasks.testing.Test task, List<String> suiteNames ->
+    def dirs = task.testClassesDirs.files.collect { it.toPath() }
+    def singleton = tech.beshu.ror.buildbase.SuiteClassGraph.subtypesOf(
```
coutoPL: `singleton`? the name could be better

coutoPL (changes requested):
and please update the PR's description - it seems to be out-of-date now

build-base/src/main/java/tech/beshu/ror/buildbase/SuiteClassGraph.java
sscarduzio: Adopted as written. The pipeline replaces the nested loop and `computeIfAbsent`, and `classFilesIn` keeps the `Files.walk` resource handling in one place — `flatMap` closes each mapped stream once its contents are consumed, so nothing leaks.

It also removes the checked `IOException` from `subtypesOf`, since `classFilesIn` wraps it.

Verified rather than eyeballed, because a traversal rewrite can pass a test vacuously:
- `javac -Xlint:all -Werror` — clean
- `SuiteClassGraphTest` — 2/2, before and after
- negative control: dropping the interface edge from `declaredSupertypesOf` fails 1 of the 2 cases, so they do exercise the graph

sscarduzio: Renamed to `directSubtypesBySupertypeIn`. It is the accurate name: the map holds only the DIRECT edges, and the transitive answer comes from `allSubtypesOf` sweeping it. The javadoc now says so too, so the pairing is not left implicit.

integration-tests/build.gradle
sscarduzio: Right — that vocabulary had outlived the `singletonPhaseTest` -> `sharedEsSuitesTest` rename, and `singleton` said which mechanism it is rather than which suites it holds.

That file now reads: `partitionBySharedEsUsage`, `sharedEsSuites` / `ownClusterSuites`, and `sharedEsPhase` for the flag. The log line says "shared-ES / own-cluster" and the per-shard report dir suffix is `-shared-es`.

`SingletonPluginTestSupport` keeps its name — it is the Scala marker type and renaming it is a separate change.

integration-tests/README.md
```diff
@@ -20,6 +20,17 @@ Omit `-PesModule` to run against the module that publishes the newest supported
...
 
+Each shard runs its share in two steps, so it frees the shared ES as soon as the suites that need
+it are done, instead of holding it idle until the shard ends:
+
+1. `integration-tests:sharedEsSuitesTest` — the suites that use the shared ES. The ES stops when
+   this task's JVM exits.
+2. `integration-tests:test` — the remaining suites, in a fresh JVM. Each one starts its own
+   cluster, and this JVM never starts a shared ES.
+
+`shardedTest` starts both steps for every shard, in that order. Without sharding, `test` alone
```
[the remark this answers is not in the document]
sscarduzio: Correct, and I confirmed it in the code before changing the text: `if (phaseSuites.isEmpty()) { task.enabled = false }` in `integration-tests/build.gradle`. A shard whose partition holds no suite for one step really does run one task, not two.

The README now says each shard runs its share in "up to two steps", and states the rule plainly: "A shard whose partition holds no suite for one of the steps disables that step, so a shard runs one task or two."

coutoPL (approved):
LGTM

PR #1342, opened 2026-08-17
[RORDEV-2181] Stop the audit tests depending on wall-clock timing

sscarduzio:
## Why

`integration_es78x` on develop failed with the ECS-serializer cases in `AuditSettingsTests` comparing a whole JSON document that contains `event.duration`:

```
"duration" : 5001000000   (expected 5000000000)
```

`AuditResponseContext.duration` is `Instant.now() - requestContext.timestamp` — **two separate clock reads**. The value is exactly 5000ms only when nothing elapses between them, so under CI load it becomes 5001 and the whole-document comparison fails.

Worth noting what does *not* fix it: freezing the mock's `timestamp` in a `val`. `duration` is computed later, when the response context is built, so a nonzero gap remains — it would only narrow the window, not close it.

## What

`event.duration` is normalised to `"IGNORED"` on **both** sides of the comparison, exactly as `@timestamp` already is (same `circeJsonWithIgnoredTimestamp` helper). The test keeps asserting the whole document; it just stops asserting a wall-clock measurement it cannot control.

## Verification

A/B against an injected 3ms delay placed **between** the two clock reads (my first attempt put the delay before both, which cannot widen the gap — it proved nothing, and I only caught that because the control passed when it should have failed):

| | control (no fix) | with fix |
|---|---|---|
| same injected drift | **3 failures** — `5003000000` vs `5000000000` | **0 failures** |

Without drift: 104/104 green.

JIRA: [RORDEV-2181](https://readonlyrest.atlassian.net/browse/RORDEV-2181)

## Summary by CodeRabbit

* **Tests**
  * Improved audit JSON test comparisons by normalizing event durations alongside timestamps.
  * Added handling for expected results that include nested duration values.

### DIFF
--- a/core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
+++ b/core/src/test/scala/tech/beshu/ror/unit/acl/factory/AuditSettingsTests.scala
@@ -2687,6 +2687,7 @@ class AuditSettingsTests extends AnyWordSpec with Inside {
   private def circeJsonWithIgnoredTimestamp(json: JSONObject): Option[Json] = {
     json
       .withTimestampValue("IGNORED")
+      .withEventDurationValue("IGNORED")
       .circeJsonE
       .toOption
   }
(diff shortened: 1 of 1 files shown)


coutoPL (changes requested):
Why you added this weird changelog entry, when there is no changelog for this change?!
Moreover, there is nothing like "core" component in the changelog.

sscarduzio (commented):
You are right on both counts, and it was sloppy of me.

The changelog block is **gone**. Writing `text: "no-changelog"` inside a changelog entry was nonsense — if a change needs no entry, it needs no block. The `no-changelog` label is on the PR, which is what the conventions check actually reads. This is test-only, so no entry is correct.

On the component: you are right that `core` does not exist — the vocabulary is `[es]`, which I confirmed against the merged PRs and the example in `20-changelog-entry.sh`. I invented `core`.

One consequence worth flagging rather than leaving for you to find: **#1338 was merged with the same invalid `components: [core]`**. If release notes are harvested from PR descriptions, that entry is wrong. I can edit that merged PR's description to `[es]` — say the word and I will, or correct it yourself if the harvest has already happened.

coutoPL (approved):
LGTM

PR #1343, opened 2026-08-17
[RORDEV-2182] Add a CI flake ledger

sscarduzio:
We have triaged CI reds **one at a time all week**. This makes the repetition visible, so we can tell a pattern from a one-off before spending an hour on it.

`ci/flake-ledger.sh` ranks CI failures by how often they recur. Read-only, runs locally, no workflow change.

```bash
ci/flake-ledger.sh --runs 25                 # fast, job level
ci/flake-ledger.sh --branch develop --runs 10 --with-tests   # + junit artifacts (slow, GBs)
```

## Two layers, because our failures live in both

**Job level** — a leg that dies *before any test runs*: image build, container start, host OOM, runner provisioning. **Every red we chased this month was of this kind.**

**Test level** — assertion/initialization failures, from the `TEST-*.xml` each leg already uploads.

I built the junit-only version first and it reported **0 failures across 8 runs and 1,387 tests** — not a bug, but proof the design was wrong: those runs were green, and the failures we actually care about never produce a test XML at all. The job-level layer exists because of that finding.

Two further traps the header documents: reruns **overwrite** artifacts, so a leg rerun to green leaves no test-level trace (job level survives it); and green runs contribute nothing, so mining only successful develop runs is empty by construction.

## What it says about us today

First real run, last 25 runs, unfiltered:

```
    5/21  FLAKE   cve_check                 failing step: cve_check x5
    2/22  FLAKE   integration_es94x
    2/22  FLAKE   integration_es816x
    1/22  FLAKE   integration_es90x / es818x / es810x / es80x / es717x / es70x
    1/5   FLAKE   integration_es89x / es88x / es87x / es85x
```

Which matches what we found the hard way, and adds one thing: **`cve_check` is the single most frequent red in the repo** — 5 of 21 runs, all from NVD throttling keyless clients rather than any vulnerability. An `NVD_API_KEY` repo secret would remove the noisiest signal we have.

Nothing is BROKEN (always-failing); everything is intermittent, which is the useful distinction.

JIRA: [RORDEV-2182](https://readonlyrest.atlassian.net/browse/RORDEV-2182)

## Summary by CodeRabbit

* **New Features**
  * Added a manually triggered CI reliability report for reviewing recent workflow failures.
  * Supports filtering results by branch and selecting the number of CI runs to analyze.
  * Optionally analyzes downloaded test reports to identify recurring test failures.
  * Reports failed jobs, failing steps, flaky behavior, and unavailable or malformed test results.
  * Publishes results directly in the workflow summary for easier review.

### DIFF
--- a/.github/workflows/flake-ledger.yml
+++ b/.github/workflows/flake-ledger.yml
@@ -1,0 +1,56 @@
+name: CI Flake Ledger
+
+# Ranks CI failures by how often they REPEAT, so triage can target a pattern instead of a single red.
+# Two layers: job level (a leg dying before any test runs — image build, container start, host OOM,
+# runner provisioning, registry throttling) and, optionally, test level from the junit artifacts.
+#
+# Manual-only: it reads CI history, changes nothing, and is meant to be run when deciding whether a
+# red is worth chasing. The same script runs locally — see ci/utils/flake-ledger.sh.
+on:
+  workflow_dispatch:
+    inputs:
+      runs:
+        description: 'How many recent ci.yml runs to inspect'
+        type: number
+        default: 20
+        required: true
+      branch:
+        description: 'Branch to scan (empty = all branches)'
+        type: string
+        required: false
+      with_tests:
+        description: 'Also download junit artifacts for the test-level ledger (slow, GBs)'
+        type: boolean
+        default: false
+        required: false
+
+permissions:
+  contents: read
+  actions: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ledger:
+    name: Ledger
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with: { persist-credentials: false }
+
+      - name: Build the ledger
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # Inputs via env, never `${{ }}` in the script body: that is a shell-injection vector.
+          INPUT_BRANCH: ${{ inputs.branch }}
+          INPUT_RUNS: ${{ inputs.runs }}
+          INPUT_WITH_TESTS: ${{ inputs.with_tests }}
+        run: |
+          args=(--runs "$INPUT_RUNS")
+          [ -n "$INPUT_BRANCH" ] && args+=(--branch "$INPUT_BRANCH")
+          [ "$INPUT_WITH_TESTS" = "true" ] && args+=(--with-tests)
+          # Also into the job summary, so the result is readable without opening the log.
+          ci/utils/flake-ledger.sh "${args[@]}" | tee -a "$GITHUB_STEP_SUMMARY"
--- a/ci/utils/flake-ledger.sh
+++ b/ci/utils/flake-ledger.sh
@@ -1,0 +1,168 @@
+#!/bin/bash
+# Flake ledger: which CI failures REPEAT, so triage targets patterns instead of single reds.
+#
+# Two layers, because our failures live in both:
+#   JOB level  — a leg that dies before any test runs (image build, container start, host OOM,
+#                runner provisioning, registry throttling). A junit-only ledger reports NOTHING for
+#                these, and in Aug 2026 they were every red we chased.
+#   TEST level — assertion/initialization failures, from the `TEST-*.xml` each leg uploads.
+#
+# A job/test failing in some runs and passing in others is a FLAKE; one that always fails is BROKEN.
+# Reruns overwrite artifacts, so a leg rerun to green leaves no test-level trace; job level survives.
+#
+# Usage: ci/utils/flake-ledger.sh [--branch B] [--runs N] [--with-tests]
+set -uo pipefail
+
+for tool in gh python3; do
+  command -v "$tool" >/dev/null 2>&1 || { echo "error: '$tool' is required but not on PATH"; exit 2; }
+done
+gh auth status >/dev/null 2>&1 || { echo "error: 'gh' is not authenticated (run: gh auth login)"; exit 2; }
+# `gh run download --pattern` landed in 2.11.0.
+gh_version=$(gh --version | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1)
+if [ -n "$gh_version" ] && [ "$(printf '%s\n2.11.0\n' "$gh_version" | sort -V | head -1)" != "2.11.0" ]; then
+  echo "error: gh $gh_version is too old; 'gh run download --pattern' needs 2.11.0+"; exit 2
+fi
+
+BRANCH=""
+RUNS=20
+WITH_TESTS=false
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --branch)
+      [ -n "${2-}" ] || { echo "error: --branch requires a value"; exit 2; }
+      BRANCH="$2"; shift 2 ;;
+    --runs)
+      case "${2-}" in
+        ''|0|*[!0-9]*) echo "error: --runs requires a positive integer"; exit 2 ;;
+      esac
+      RUNS="$2"; shift 2 ;;
+    --with-tests) WITH_TESTS=true; shift ;;
+    -h|--help) sed -n '2,13p' "$0"; exit 0 ;;
+    *) echo "error: unknown argument '$1'"; exit 2 ;;
+  esac
+done
+
+list_args=(--workflow ci.yml --limit "$RUNS" --json databaseId)
+[ -n "$BRANCH" ] && list_args+=(--branch "$BRANCH")
+
+echo ">>> scanning the last $RUNS ci.yml runs${BRANCH:+ of $BRANCH}"
+run_ids=$(gh run list "${list_args[@]}" -q '.[].databaseId') || { echo "error: cannot list runs"; exit 1; }
+[ -z "$run_ids" ] && { echo "no runs found"; exit 1; }
+
+JOBS_TSV=$(mktemp)
+WORK=$(mktemp -d)
+trap 'rm -rf "$JOBS_TSV" "$WORK"' EXIT
+
+inspected=0
+skipped_runs=""
+for r in $run_ids; do
+  # Every terminal conclusion that is not success counts as a failure: a leg that times out or whose
+  # runner never starts is just as much a red as an assertion failure, and reporting only "failure"
+  # would silently drop them from the ledger.
+  if rows=$(gh run view "$r" --json jobs,conclusion -q \
+      'if (.jobs | length) == 0 then [ "'"$r"'", "(workflow startup)", (.conclusion // "running"), "" ] | @tsv
+       else .jobs[] | [ "'"$r"'", .name, (.conclusion // "running"), ([.steps[]? | select(.conclusion=="failure") | .name] | join("+")) ] | @tsv end' 2>/dev/null) \
+      && [ -n "$rows" ]; then
+    echo "$rows" >> "$JOBS_TSV"
+    inspected=$((inspected + 1))
+  else
+    skipped_runs="$skipped_runs $r"
+  fi
+done
+
+downloaded=0
+download_failures=""
+if [ "$WITH_TESTS" = true ]; then
+  echo ">>> downloading junit artifacts (slow)"
+  for r in $run_ids; do
+    mkdir -p "$WORK/$r"
+    if gh run download "$r" -D "$WORK/$r" -p "it-*-results" >/dev/null 2>&1; then
+      downloaded=$((downloaded + 1))
+    else
+      download_failures="$download_failures $r"
+    fi
+  done
+fi
+
+WITH_TESTS=$WITH_TESTS INSPECTED=$inspected TOTAL=$(echo "$run_ids" | wc -w) \
+SKIPPED="$skipped_runs" DOWNLOADED=$downloaded DL_FAILED="$download_failures" \
+python3 - "$JOBS_TSV" "$WORK" <<'PY'
+import sys, os, glob, collections, xml.etree.ElementTree as ET
+
+jobs_tsv, work = sys.argv[1], sys.argv[2]
+inspected, total = int(os.environ["INSPECTED"]), int(os.environ["TOTAL"])
+skipped = os.environ["SKIPPED"].split()
+with_tests = os.environ["WITH_TESTS"] == "true"
+
+runs_seen, runs_failed = collections.defaultdict(set), collections.defaultdict(set)
+steps = collections.defaultdict(collections.Counter)
+for line in open(jobs_tsv):
+    row = line.rstrip("\n").split("\t")
+    if len(row) < 3:
+        continue
+    run, job, concl = row[0], row[1], row[2]
+    step = row[3] if len(row) > 3 else ""
+    # A run cancelled before its matrix expanded reports the literal template as the job name.
+    if "${{" in job:
+        continue
+    # Only real outcomes go in the denominator. Counting a cancelled or unfinished run there would
+    # make a job that fails whenever it completes read as a FLAKE, understating the failure rate.
+    if concl in ("running", "cancelled", ""):
+        continue
+    runs_seen[job].add(run)
+    # "cancelled" is not a flake: a push superseding an in-flight run cancels it by design.
+    if concl not in ("success", "skipped", "neutral"):
+        runs_failed[job].add(run)
+        steps[job][step or f"(no step; job {concl})"] += 1
+
+print(f"\n{'='*74}\nJOB-LEVEL LEDGER — {inspected} of {total} runs inspected\n{'='*74}")
+if skipped:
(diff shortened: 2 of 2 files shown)


ci/flake-ledger.sh
coutoPL, line 1: IMO, we should create a `ci/utils` folder and put in there. Then we could try to clean up the `ci` folder a little bit. We have a few scripts like this one. They could be moved there.

coutoPL, line 1: let's create a new, manual workflow for this thing. Like we did in case of `Mirror ES Libs` workflow

ci/flake-ledger.sh
```diff
@@ -0,0 +1,106 @@
...
+WITH_TESTS=false
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --branch) BRANCH="$2"; shift 2 ;;
+    --runs) RUNS="$2"; shift 2 ;;
+    --with-tests) WITH_TESTS=true; shift ;;
+    *) echo "unknown arg: $1"; exit 2 ;;
+  esac
+done
+
```
coutoPL: shouldn't we check if `gh` and `python3` are available?

coutoPL (changes requested):

ci/flake-ledger.sh
sscarduzio: Moved to `ci/utils/flake-ledger.sh`. Agreed the `ci/` root has accumulated; I did not move the other scripts in this PR because `run-pipeline.sh`, `ci-lib.sh` and `docker-hub-auth.sh` are referenced from workflows and from each other, so relocating them is its own change with its own blast radius. Happy to do that as a follow-up if you want the tidy-up now.

sscarduzio: Added `.github/workflows/flake-ledger.yml`, modelled on Mirror ES Libs: `workflow_dispatch` only, `contents: read` + `actions: read`, inputs for branch / how many runs / whether to pull the junit artifacts. It writes to `$GITHUB_STEP_SUMMARY` as well as the log, so the ranking is readable without opening the job. The same script still runs locally.

ci/flake-ledger.sh
sscarduzio: Added: the script checks `gh` and `python3` are on PATH, and that `gh` is authenticated, before doing anything. Verified — with `gh` removed from PATH it now exits with `error: 'gh' is required but not on PATH` instead of failing obscurely later.

ci/flake-ledger.sh
```diff
@@ -0,0 +1,106 @@
...
+RUNS=20
+WITH_TESTS=false
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --branch) BRANCH="$2"; shift 2 ;;
+    --runs) RUNS="$2"; shift 2 ;;
+    --with-tests) WITH_TESTS=true; shift ;;
+    *) echo "unknown arg: $1"; exit 2 ;;
+  esac
+done
```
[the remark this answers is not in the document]
sscarduzio: Fixed as suggested. Verified each path: `--runs` with no value, `--runs 0`, `--runs abc` and `--branch` with no value all exit 2 with a specific message, and an unknown flag is rejected rather than ignored.

ci/flake-ledger.sh
```diff
@@ -0,0 +1,106 @@
...
+[ -z "$run_ids" ] && { echo "no runs found"; exit 1; }
+
+JOBS_TSV=$(mktemp); trap 'rm -f "$JOBS_TSV"' EXIT
+total_runs=0
+for r in $run_ids; do
+  total_runs=$((total_runs+1))
+  # One row per job: run, job name, conclusion, and the step that failed (if any).
+  gh run view "$r" --json jobs -q \
+    '.jobs[] | [ "'"$r"'", .name, (.conclusion // "running"), ([.steps[]? | select(.conclusion=="failure") | .name] | join("+")) ] | @tsv' \
+    2>/dev/null >> "$JOBS_TSV"
```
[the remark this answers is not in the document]
sscarduzio: This was the most valuable comment on the PR — a ledger that quietly under-reports is worse than none, because the whole purpose is to decide whether a red is worth chasing.

Now: unreadable runs are counted and named, the header says **"N of M runs inspected"**, artifact-download failures and unparsable junit files each print an explicit `INCOMPLETE:` line. The very first real run proved the point — it printed `INCOMPLETE: 1 run(s) could not be read: 32114352995` rather than silently ranking 24 runs as if they were 25.

ci/flake-ledger.sh
```diff
@@ -0,0 +1,106 @@
...
+runs_failed = collections.defaultdict(set) # job -> runs it failed in
+steps = collections.defaultdict(collections.Counter)
+for row in rows:
+    if len(row) < 3: continue
+    run, job, concl = row[0], row[1], row[2]
+    step = row[3] if len(row) > 3 else ""
+    runs_seen[job].add(run)
+    if concl == "failure":
+        runs_failed[job].add(run)
+        steps[job][step or "(no step recorded)"] += 1
```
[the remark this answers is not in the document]
sscarduzio: Right, and this was a genuine hole: only `conclusion == "failure"` counted, so a leg that **timed out** or whose runner never started was invisible — exactly the 120-minute-timeout and provisioning failures we have hit.

Now every terminal conclusion that is not a success counts. Two deliberate exceptions, which the code states: `cancelled` (a push superseding an in-flight run cancels it by design — not a flake) and an empty conclusion (the job has not finished). I also had to drop jobs whose name still contains `\${{ matrix... }}`: a run cancelled before its matrix expands reports the literal template as a job name, and those were showing up as BROKEN 1/1.

ci/flake-ledger.sh
```diff
@@ -0,0 +1,106 @@
...
+    rel = os.path.relpath(f, root); run = rel.split(os.sep)[0]
+    leg = next((p for p in rel.split(os.sep) if p.startswith("it-")), "?").replace("it-", "").replace("-results", "")
+    runs.add(run)
+    try: suite = ET.parse(f).getroot()
+    except ET.ParseError: continue
+    for case in suite.iter("testcase"):
+        name = f'{case.get("classname","?").split(".")[-1]}.{case.get("name","?")}'
+        failed = case.find("failure") is not None or case.find("error") is not None
+        seen[name][run] = seen[name].get(run, False) or failed
+        if failed: legs[name].add(leg)
```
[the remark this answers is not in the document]
sscarduzio: Correct — fixed. The full `classname` is kept, so `api.TestClient.test_connect` and `db.TestClient.test_connect` stay distinct. Merging them would have invented a FLAKE from two independent results, which is the exact failure mode this tool exists to avoid.

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/checkout@v4
+        with: { persist-credentials: false }
+
+      - name: Build the ledger
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          args=(--runs "${{ inputs.runs }}")
+          [ -n "${{ inputs.branch }}" ] && args+=(--branch "${{ inputs.branch }}")
+          [ "${{ inputs.with_tests }}" = "true" ] && args+=(--with-tests)
```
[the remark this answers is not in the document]
sscarduzio: Real vulnerability, and mine — fixed. The inputs now reach bash through `env:` (`INPUT_BRANCH`, `INPUT_RUNS`, `INPUT_WITH_TESTS`) instead of `\${{ }}` expanding into the script text before bash parses it. As written, a branch input of `x" $(id) "` would have executed on the runner with `GH_TOKEN` in scope. The comment in the workflow now says exactly that, so nobody reintroduces it.

ci/utils/flake-ledger.sh
```diff
@@ -0,0 +1,155 @@
...
+    step = row[3] if len(row) > 3 else ""
+    # A run cancelled before its matrix expanded reports the literal template as the job name.
+    if "${{" in job:
+        continue
+    runs_seen[job].add(run)
+    # "cancelled" is not a flake: a push superseding an in-flight run cancels it by design. An empty
+    # conclusion means the job has not finished, which is not an outcome either.
+    if concl not in ("success", "skipped", "running", "neutral", "cancelled", ""):
+        runs_failed[job].add(run)
+        steps[job][step or f"(no step; job {concl})"] += 1
```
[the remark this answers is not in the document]
sscarduzio: Correct, and it was a bug I created when I excluded `cancelled` from the numerator without excluding it from the denominator — so a job that fails whenever it completes could read as a FLAKE.

Runs without a real outcome (`cancelled`, `running`, empty) are now skipped entirely rather than counted as appearances. The effect is visible in the same data: `e2e_es94x` went from `3/7` to `3/6` and `unit_tests_linux` from `2/24` to `2/21`.

ci/flake-ledger.sh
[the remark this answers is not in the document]
sscarduzio: Also fixed. `gh run view` now requests `jobs,conclusion`, and a run with zero job rows is recorded as a synthetic `(workflow startup)` entry carrying the run's own conclusion — so a workflow that dies before creating any job is ranked instead of being filed as unreadable.

Measurable on the same window: the header went from `24 of 25 runs inspected` with `INCOMPLETE: 32114352995` to **`25 of 25`**, and that run is precisely the no-job-rows case.

ci/utils/flake-ledger.sh
```diff
@@ -0,0 +1,155 @@
...
+  fi
+done
+
+downloaded=0
+download_failures=""
+if [ "$WITH_TESTS" = true ]; then
+  echo ">>> downloading junit artifacts (slow)"
+  for r in $run_ids; do
+    mkdir -p "$WORK/$r"
+    if gh run download "$r" -D "$WORK/$r" -p "it-*-results" >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
sscarduzio: Added a version gate: the preflight parses `gh --version` and refuses to run below 2.11.0, naming `gh run download --pattern` as the reason. Checked against the local `gh 2.97.0`, which passes.

ci/utils/flake-ledger.sh
```diff
@@ -0,0 +1,155 @@
...
+    --branch)
+      [ -n "${2-}" ] || { echo "error: --branch requires a value"; exit 2; }
+      BRANCH="$2"; shift 2 ;;
+    --runs)
+      case "${2-}" in
+        ''|0|*[!0-9]*) echo "error: --runs requires a positive integer"; exit 2 ;;
+      esac
+      RUNS="$2"; shift 2 ;;
+    --with-tests) WITH_TESTS=true; shift ;;
+    -h|--help) sed -n '2,12p' "$0"; exit 0 ;;
```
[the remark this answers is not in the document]
sscarduzio: Fixed — though the suggested `2,13` was right and my first attempt at `2,14` pulled in `set -uo pipefail`. It now ends on the Usage line.

ci/utils/flake-ledger.sh
```diff
@@ -0,0 +1,164 @@
...
+    except ET.ParseError:
+        bad_xml.append(rel)   # a leg killed mid-write leaves a truncated file
+        continue
+    for case in suite.iter("testcase"):
+        # Full classname: two same-named tests in different packages are different tests, and
+        # merging them would invent a FLAKE out of two independent results.
+        name = f'{case.get("classname", "?")}.{case.get("name", "?")}'
+        failed = case.find("failure") is not None or case.find("error") is not None
+        seen[name][run] = seen[name].get(run, False) or failed
+        if failed:
```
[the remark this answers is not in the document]
sscarduzio: Fixed — and it is the same bug as the cancelled-run one at job level: a non-outcome inflating the denominator. A `<skipped/>` case is now dropped before `seen` is updated, so a test that fails whenever it actually runs reads as BROKEN rather than FLAKE. The comment says so, next to the job-level equivalent.

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,57 @@
+name: Flake Ledger
```
coutoPL: Maybe "Flake Tests Ledger"?

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,57 @@
...
+# Manual-only: it reads CI history, changes nothing, and is meant to be run when deciding whether a
+# red is worth chasing. The same script runs locally — see ci/utils/flake-ledger.sh.
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to scan (empty = all branches)'
+        type: string
+        required: false
+      runs:
```
coutoPL: the required params should be described above the optional ones

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,57 @@
...
+        description: 'How many recent ci.yml runs to inspect'
+        type: string
```
coutoPL: String or int?

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,57 @@
...
+  run:
+    shell: bash
+
+jobs:
+  ledger:
+    name: Ledger
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
```
coutoPL: pin the action version

coutoPL (changes requested):

.github/workflows/flake-ledger.yml
sscarduzio: Reordered: `runs` (required) now comes first, then the optional `branch` and `with_tests`.

.github/workflows/flake-ledger.yml
sscarduzio: int — and thank you for asking, because I had assumed `workflow_dispatch` only accepted `string`/`boolean`/`choice`/`environment`. I checked the docs: **`number` is allowed**. Changed to `type: number` with `default: 20`, so the dispatch form validates it instead of the script being the first thing to complain.

.github/workflows/flake-ledger.yml
sscarduzio: Pinned to the commit SHA: `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2`, verified against the real `v4.2.2` tag via the API rather than copied from memory. A tag can be moved; a SHA cannot.

Worth flagging: the other 20 usages of `actions/checkout@v4` in this repo (plus `upload-artifact`, `setup-java`, `cache`, `claude-code-action`) are all tag-pinned. If you want SHA pinning as a rule, that is a separate sweep — I did not change them here to keep this PR to its subject.

.github/workflows/flake-ledger.yml
sscarduzio: Renamed to **`CI Flake Ledger`** rather than `Flake Tests Ledger`, because the ledger ranks failing CI **jobs** as well as tests — and the job layer is the one that matters most for us: every red we chased this month (image build, container start, host OOM, runner provisioning, registry throttling) kills the leg before a single test runs, so a tests-only reading would miss them. `CI` keeps your added specificity without narrowing it to tests. Say the word if you still prefer `Flake Tests Ledger` and I will change it.

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,58 @@
...
+  run:
+    shell: bash
+
+jobs:
+  ledger:
+    name: Ledger
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      # Pinned to the commit the v4 tag pointed at, so a retag cannot change what runs here.
```
coutoPL: unecessary comment

.github/workflows/flake-ledger.yml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with: { persist-credentials: false }
+
+      - name: Build the ledger
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # Through env, NOT `${{ }}` inside the script: a template expands into the script TEXT
```
coutoPL: the same here

coutoPL (changes requested):

.github/workflows/flake-ledger.yml
sscarduzio: Removed — the trailing `# v4.2.2` already says what the SHA is.

.github/workflows/flake-ledger.yml
sscarduzio: Compressed to one line rather than removed, and I would like your call on whether even that should go:

```yaml
# Inputs via env, never `${{ }}` in the script body: that is a shell-injection vector.
```

My hesitation is that this exact code **was** the vulnerable form earlier in this PR — `${{ inputs.branch }}` expanded into the script text, so `x" $(id) "` would have executed on the runner with `GH_TOKEN` in scope. The `env:` indirection looks like pointless ceremony to anyone who does not know that, which makes it the kind of thing a later cleanup removes. One line seemed a fair price for that.

If you still read it as noise, say so and I will drop it.

coutoPL (approved):
LGTM

PR #1345, opened 2026-08-18
Make the Docker Hub credential check real, not a variable check

sscarduzio:
## The symptom

`it_linux_*` legs fail like this:

```
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500:
{"message":"toomanyrequests: You have reached your unauthenticated pull rate limit."}
tc.osixia/openldap:1.5.0 - Retrying pull for image: osixia/openldap:1.5.0 (118s remaining)
```

`LdapIntegrationSuite`, `ImpersonationSuite` and `AdminApiAuthMockSuite` then fail with `initializationError`, and the leg dies after ~25 minutes.

Earlier in the **same log**:

```
[TEST] Docker Hub authenticated pulls ENABLED (user '***')
```

## Why both are true

`ci/docker-hub-auth.sh` checked that `DOCKER_HUB_USER` and `DOCKER_HUB_RO_TOKEN` are non-empty, exported `DOCKER_AUTH_CONFIG`, and printed ENABLED. It never asked the registry whether the credentials are accepted. A revoked, expired or mistyped token reads exactly like a working one.

So the message reports what we *configured*, not what the registry *does*, and the real answer arrives 25 minutes later as a rate limit we were supposed to be exempt from.

## The change

Log the daemon in too, and let the login be the check.

- The credentials now also cover pulls this script cannot reach through the environment.
- A bad credential fails in the first seconds of the leg, with a message that names the cause.

## One thing worth knowing

A sourced script **cannot** stop its caller. `set -e` does not abort on a failing `source` — I checked on this bash, with a bare `return 1` and with `return 1 2>/dev/null || exit 1`, and the caller continued in both cases. My first attempt did exactly that, printed a stern message, and then let the leg run anyway.

Every call site therefore guards explicitly, and the script returns 1 for that guard to test:

```bash
source ci/docker-hub-auth.sh || exit 1
```

Four call sites: two in `ci.yml` (`unit_tests_linux`, `it_linux`), plus `azure-pipelines.yml` and `ci/azure-templates/integration-test-steps.yml`.

## Verification

With a stub `docker` on `PATH`, all four paths, each observed rather than assumed:

| case | result | leg rc |
|---|---|---|
| no credentials | DISABLED, runs anonymously | 0 |
| credentials, login accepted | ENABLED (login verified) | 0 |
| credentials, login rejected | named error, leg aborts | 1 |
| credentials, no docker CLI | warns, still exports the config | 0 |

The rejected case is the negative control: before the call-site guard it printed the error and returned 0, which is how I found that `set -e` was not enough.

`bash -n` on the script and a YAML parse on all three workflow files are clean.

## What this does not answer

It does not tell us *why* the current token is not authenticating. This PR makes CI say so on the next run instead of hiding it. If the login fails, the token needs rotating; if it succeeds and pulls are still anonymous, the problem is in how testcontainers 2.0.3 resolves `DOCKER_AUTH_CONFIG` and I will chase that next.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker Hub authentication validation during automated tests.
  * CI now fails clearly when configured credentials are rejected, preventing tests from continuing in an unauthenticated state.
  * Authentication failures now provide guidance for resolving credential issues.
  * Environments without configured credentials continue without authentication; unavailable Docker installations are handled with a warning.

* **Documentation**
  * Updated CI authentication guidance to reflect the new failure behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,7 +288,7 @@ jobs:
           DOCKER_HUB_USER: ${{ secrets.DOCKER_HUB_USER }}
           DOCKER_HUB_RO_TOKEN: ${{ secrets.DOCKER_HUB_RO_TOKEN }}
         run: |
-          source ci/docker-hub-auth.sh
+          source ci/docker-hub-auth.sh || exit 1
           ci/run-pipeline.sh
       - name: Publish test results
         if: failure()
@@ -351,7 +351,7 @@ jobs:
           DOCKER_HUB_USER: ${{ secrets.DOCKER_HUB_USER }}
           DOCKER_HUB_RO_TOKEN: ${{ secrets.DOCKER_HUB_RO_TOKEN }}
         run: |
-          source ci/docker-hub-auth.sh
+          source ci/docker-hub-auth.sh || exit 1
           ci/run-pipeline.sh
       - name: Publish test results
         # always(): green-run XMLs feed suite-timings.json regeneration (duration-balanced sharding)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -337,7 +337,7 @@ stages:
             persistCredentials: true
           - script: |
               # core_tests starts testcontainers too (LDAP/WireMock) — same Docker Hub rate limit as IT.
-              source ci/docker-hub-auth.sh
+              source ci/docker-hub-auth.sh || exit 1
               echo "[TEST] executing ROR_TASK = $ROR_TASK"
               ci/run-pipeline.sh
             env:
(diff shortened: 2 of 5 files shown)


sscarduzio:
Superseded by #1346 — closing in its favour.

@coutoPL's PR is the larger and better fix. Mine addressed one of the four anonymous-pull sources; his covers all of them, including the two I never reached:

- the job `container:` image, pulled by the runner before step 1, which no script can authenticate — **22 anonymous pulls in a plain PR run, 46 on the full matrix**, before a test starts;
- `build_toolchains_image` logging in *after* `docker build`, so the five `FROM` pulls stayed anonymous.

He also settled a fact I had checked only halfway. I confirmed testcontainers 2.0.3 still reads `DOCKER_AUTH_CONFIG` (I unpacked the jar), and concluded the variable was fine. The missing half is that **Docker CLI 27 ignores it** — so every `docker pull`, `manifest inspect` and `compose` call stayed anonymous while the log said authentication had succeeded.

One thing from here that his version already does better. I found that a sourced script cannot stop its caller — `set -e` does not abort on a failing `source`, with either `return 1` or `return 1 || exit 1`; I verified both and the caller continued. My answer was to guard all four call sites with `|| exit 1`. His script exits the caller shell instead, which needs no guard **and** holds for steps that never set errexit. That is the stronger fix.

Nothing here is worth porting. I have left one review note on #1346 about the two Azure call sites.

PR #1347, opened 2026-08-19
Use one S3 credential set and consistent CI secret names

sscarduzio:
Part 1 of 3. Companions: `sscarduzio/readonlyrest_kbn` and `beshu-tech/readonlyrest-e2e-tests`, same branch name. **All three must merge together** — the Doppler project `ror_ci` feeds every repo from one config.

## What changes

The CI secrets had grown three spellings of the same credential. This collapses them and renames the rest to say what they are. `ror_ci/prd` already holds the new names **alongside** the old ones with identical values, so this PR is safe to merge on its own schedule; the old names get removed only after all three land and one full cycle is green.

| Was | Is |
|---|---|
| `ROR_{ARTIFACTS,LIBS,E2E_REPORTS}_STORE_{ACCESS_KEY_ID,ACCESS_KEY_SECRET,BUCKET,REGION,ENDPOINT_URL}` (15 names) | `ROR_S3_{ACCESS_KEY_ID,SECRET_ACCESS_KEY,BUCKET,REGION,ENDPOINT_URL}` (5) |
| `ROR_*_STORE_PATH_PREFIX` (3) | `ROR_S3_PATH_{ARTIFACTS,LIBS,E2E_REPORTS}` (3) |
| `DOCKER_REGISTRY_USER` / `_PASSWORD` | `DOCKER_HUB_USER` / `DOCKER_HUB_RW_TOKEN` |
| `GH_PAT`, `KBN_REPO_GH_TOKEN` | `ROR_GH_TOKEN` |
| `ROR_ACTIVATION_KEY` (secret) | `ROR_ENT_ACTIVATION_TOKEN` (secret) |

`ci-lib.sh` no longer resolves a credential family per store. One credential set serves every store, so the store name selects a path and nothing else.

## The claim this removes was false

`ci/CI.md` said each store had its own credentials *"because the gateway authorizes each prefix separately"*. Signed S3 requests say otherwise — the artifacts key writes all three prefixes:

| Prefix | artifacts key | the separate reports key |
|---|---|---|
| `ror/builds/` | 200 | 200 |
| `ror/libs/` | 200 | **403** |
| `ror/e2e_reports/` | 200 | **403** |

The doc is corrected rather than deleted, so the next person does not re-derive it.

## What is deliberately untouched

**`ROR_ACTIVATION_KEY` stays as the env var name.** Only the *secret* is renamed. `readonlyrest_kbn`'s `proxy/core/license/LicenseService.ts` reads `process.env.ROR_ACTIVATION_KEY`, it is documented for customers, and `docker-envs/**` sets it. Renaming it would break every customer deployment. Same rule for `DOCKER_AUTH_CONFIG`, `GH_TOKEN` and `GITHUB_TOKEN`.

**`azure-pipelines.yml` and `ci/azure-templates/` are not updated.** They hold 114 of the references, the pipeline is `trigger: none` / `pr: none`, and Doppler does not feed Azure DevOps variable groups. Half-updating it would be worse than leaving it. `ci/CI.md:138` says it lives one release cycle after the switch and is then deleted — that call is worth making now, separately.

## Verification done

- Strict duplicate-key YAML parse on every workflow (the ARTIFACTS and LIBS families shared an `env:` block, so a naive rename would have collided).
- `bash -n` on every changed script.
- `LibsStore.java` and `LibsStoreTest.java` agree on the env-var name set.
- Signed PUT to each of the three prefixes with the retained key: all 200.

**Not run locally: the Gradle test.** There is no JDK on the machine this was prepared on, so `LibsStoreTest` needs CI to confirm.

## Summary by CodeRabbit

- **Release Infrastructure**
  - Standardized artifact, library, and report publishing with shared storage configuration.
  - Improved Docker Hub authentication across build, test, and release workflows.
  - Updated automated checks and end-to-end workflows with consistent credentials and storage paths.
  - Preserved reliable uploads across supported platforms and release stages.

- **Documentation**
  - Updated CI setup guidance for simplified credentials and revised publishing variables.


coutoPL (approved):
LGTM

sscarduzio:
@coutoPL — added the Azure pipeline as you asked, in `4d1a3e3`. It was excluded from the first pass on the grounds that it is `trigger: none` and Doppler cannot feed Azure DevOps variable groups; carrying the renames through keeps it readable rather than making it runnable.

Azure passes every secret through two hops, so all three layers are renamed:

```
VAR_ROR_S3_BUCKET: $(ROR_S3_BUCKET)      # variable group -> step env
export ROR_S3_BUCKET=$VAR_ROR_S3_BUCKET  # step env -> what the scripts read
```

Same treatment in PowerShell for the Windows legs, and the ARTIFACTS/LIBS families collapse to one credential set plus a per-store path, matching the GitHub workflows.

One thing this cannot do from here: **the entries in the Azure DevOps variable group still carry the old names.** If anyone runs this pipeline before they are renamed, every value resolves to the literal string `$(NAME)` rather than empty. `ci/docker-hub-auth.sh` is the only script that defends against that form.

Re-approve when you have a moment — the four Azure files are the only change since your approval.

PR #1348, opened 2026-08-19
Use one S3 credential set and consistent CI secret names (master)

sscarduzio:
Mirrors the same change already merged to `develop`. Opened separately because `master` and `develop` have diverged substantially — replaying the develop commit onto master conflicts in every changed file, so this is applied to master's own versions rather than cherry-picked.

## What changes

| Was | Is |
|---|---|
| per-store S3 credential families | one set: `ROR_S3_{ACCESS_KEY_ID,SECRET_ACCESS_KEY,BUCKET,REGION,ENDPOINT_URL}` |
| `ROR_*_STORE_PATH_PREFIX` | `ROR_S3_PATH_{ARTIFACTS,LIBS,E2E_REPORTS}` |
| `DOCKER_REGISTRY_USER` / `_PASSWORD` | `DOCKER_HUB_USER` / `DOCKER_HUB_RW_TOKEN` |
| cross-repo GitHub PATs | `ROR_GH_TOKEN` |
| `ROR_ACTIVATION_KEY` (secret) | `ROR_ENT_ACTIVATION_TOKEN` (secret) |

Doppler project `ror_ci` holds both the old and the new names with identical values, so master and develop can carry this independently and neither branch breaks while the other waits.

## Deliberately untouched

**`ROR_ACTIVATION_KEY` remains the environment variable name.** Only the secret is renamed. The shipped plugin reads `process.env.ROR_ACTIVATION_KEY`, customers set it, and the Docker environments pass it through — renaming it would break customer deployments. Same rule for `DOCKER_AUTH_CONFIG`, `GH_TOKEN` and `GITHUB_TOKEN`.

## Verification

- Strict duplicate-key YAML parse on every changed workflow — the ARTIFACTS and LIBS families shared an `env:` block, so a naive rename collides.
- `bash -n` on every changed script.
- Runtime-constructed names were fixed by hand, not by pattern: `ROR_${STORE}_STORE_*` is invisible to a grep, so the store resolution in `ci-lib.sh` and the Cypress uploaders was rewritten to look up a path only.


coutoPL (approved):
LGTM

PR #1350, opened 2026-08-19
[RORDEV-2183] Adopt the new pre-build wait from the e2e repo

coutoPL:
This PR comes out of #1346, which now carries the Docker Hub authentication only. The rewrite of the wait is beshu-tech/readonlyrest-e2e-tests#111. This PR is the ROR ES side of it.

## The old wait polled the registry

`wait_for_prebuild_image` ran `docker manifest inspect` every 60 seconds until the ROR KBN image appeared. Each `e2e_tests` job ran its own copy of that loop, with a timeout that grows with the number of versions in the matrix.

Three jobs, 90 minutes, one pull each minute: **up to 270 pulls for one pipeline run**, against 200 in 6 hours. This repository was the largest consumer of the limit.

## The new wait follows the run

A job now waits for the ROR KBN **run**, through the GitHub API. It asks the registry only after that run succeeds. One job spends 1 pull. A ROR KBN build that fails stops all the jobs at once, and costs no pulls.

| | Before | After |
|---|---|---|
| One `e2e_tests` job, waiting | up to 90 pulls | 1 |
| One pipeline run, three jobs | up to 270 | 3 |
| A ROR KBN build that fails | up to 90 pulls for each job | 0 pulls, 1 API request |

## What changes here

- `ci/e2e-tests-lib.sh` calls `wait_for_kbn_prebuild_images`.
- An empty `ROR_KBN_PREBUILD_RUN_ID` is now an error, and not a warning. The wait follows the run, and `e2e_prepare` fails if it cannot identify the run that it started. An empty value in a test job thus shows that the wiring in `ci.yml` is broken, and the job says so.
- `.github/workflows/publish-pre-builds.yml` gets `run-name: ROR ES pre-build ${{ inputs.tag }}`. This line is for the other side. The e2e repository and the ROR KBN repository dispatch **our** pre-build, and a dispatch is told nothing about the run that it starts. Without a title, all runs of the workflow look the same, and two dispatches a few seconds apart cannot be told apart. A job then follows a run that is not its own. The tag holds a run id and an attempt, so the title is unique.
- A job needs `ROR_GH_TOKEN` to read the runs of the ROR KBN repository. If GitHub refuses that read, the job now stops in seconds. The job also names the permission that the token needs. Before, the job waited for the full timeout, and the answer could not change.
- `ci/CI.md` records what this repository owes the shared wait: pass the run down, log in first, give the job a token that can read the runs, and keep the title. For the rest it points at the contract in the shared library, so this file cannot drift from the code.

`ci.yml` needs no change. It already passes `kbn_run_id`, `kbn_run_url`, `kbn_wait_timeout_seconds` and `ROR_GH_TOKEN` to each job.

## Merge order

This PR is one part of one change. beshu-tech/readonlyrest-e2e-tests#111, sscarduzio/elasticsearch-readonlyrest-plugin#1350 and sscarduzio/readonlyrest_kbn#1009 must merge together, in one sitting, when the three are approved and green. No one of them is safe alone:

| Merged alone | What breaks |
|---|---|
| the e2e PR | each dispatch in the three repositories. The `develop` of the plugin repositories has no `run-name:` line, so no run can be found (code 4) |
| a plugin PR | the e2e jobs on the `develop` of that repository. They call the plural names, which the e2e `develop` does not have (`command not found`, 127) |

Merge the two plugin PRs first, and the e2e PR immediately after. That order keeps the e2e repository correct at each moment, and only plugin CI is open to a fault in the gap. A run that starts in the gap fails, and it passes when you start it again. The ROR ES repository has scheduled runs, so look for one in the middle.

The three authentication PRs are different. beshu-tech/readonlyrest-e2e-tests#108, sscarduzio/elasticsearch-readonlyrest-plugin#1346 and sscarduzio/readonlyrest_kbn#1006 need nothing else, and each one can merge alone at any time.

Delete the two transitional aliases in `ci/prebuild-images-lib.sh` after the three merge. Nothing calls the singular names then.

## Tests

The wait was tested against a stub of `docker` and `gh` that counts the calls: the run succeeds and the image is there (1 pull), the image arrives late (3 pulls), the image never arrives (3 pulls, then a named error), the registry cannot be queried (a different named error), the run fails (0 pulls, 1 API request), and the run never finishes (0 pulls, and a timeout that names the run).

## Summary by CodeRabbit

- **New Features**
  - Pre-build workflow runs now display the selected dispatch tag, or the Elasticsearch version when no tag is provided.
  - Local end-to-end test runs now show the environment details needed for setup.

- **Bug Fixes**
  - End-to-end tests validate pre-build run information before starting and wait for image preparation to complete, reducing confusing failures and unnecessary work.

- **Documentation**
  - Clarified pre-build workflow integration, run tracking, fallback run titles, required workflow naming, authentication, and permissions.

### DIFF
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -1,5 +1,11 @@
 name: ROR ES Publish Pre-Builds
 
+# Two repos dispatch this workflow, and a dispatch is told nothing about the run it creates. They
+# find their run by this title, so the tag has to stay in it: the tag holds a run id and an
+# attempt, which makes the title unique. The e2e repo builds the same string in
+# ci/prebuild-images-lib.sh, and the two must match.
+run-name: ROR ES pre-build ${{ inputs.tag }}
+
 # Manual-only: no push/pull_request triggers (Azure's `trigger: none` / `pr: none`).
 on:
   workflow_dispatch:
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -3,8 +3,8 @@ name: ROR ES Publish Pre-Builds
 # Two repos dispatch this workflow, and a dispatch is told nothing about the run it creates. They
 # find their run by this title, so the tag has to stay in it: the tag holds a run id and an
 # attempt, which makes the title unique. The e2e repo builds the same string in
-# ci/prebuild-images-lib.sh, and the two must match. The tag is optional: a hand-made dispatch
-# without one falls back to the versions, so the title never ends in a dangling space.
+# ci/prebuild-images-lib.sh, and the two must match. The tag is optional. If a dispatch sends no
+# tag, the title shows the versions. Then no title stops after the word "pre-build".
 run-name: ROR ES pre-build ${{ inputs.tag || inputs.es_versions }}
 
 # Manual-only: no push/pull_request triggers (Azure's `trigger: none` / `pr: none`).
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -3,8 +3,9 @@ name: ROR ES Publish Pre-Builds
 # Two repos dispatch this workflow, and a dispatch is told nothing about the run it creates. They
 # find their run by this title, so the tag has to stay in it: the tag holds a run id and an
 # attempt, which makes the title unique. The e2e repo builds the same string in
-# ci/prebuild-images-lib.sh, and the two must match.
-run-name: ROR ES pre-build ${{ inputs.tag }}
+# ci/prebuild-images-lib.sh, and the two must match. The tag is optional: a hand-made dispatch
+# without one falls back to the versions, so the title never ends in a dangling space.
+run-name: ROR ES pre-build ${{ inputs.tag || inputs.es_versions }}
 
 # Manual-only: no push/pull_request triggers (Azure's `trigger: none` / `pr: none`).
 on:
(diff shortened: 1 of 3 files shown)


sscarduzio (approved):
Approve. I compared the contract with the code on the e2e branch, not with the descriptions:

- The function name and its signature match e2e #111 (`wait_for_kbn_prebuild_images <versions> <run tag>`).
- The run title that `prebuild-images-lib.sh` builds is the same string as the `run-name` added here, character for character.
- The old wait on e2e `develop` does not read the title, so this PR cannot break it. The merge order you describe is safe.
- `ci.yml` already logs in to Docker before the wait. It passes the run id, the URL, the scaled timeout and `ROR_GH_TOKEN` to every leg.
- The early run-id check makes a mis-wired leg fail in seconds, before the Gradle build. Good.

The three e2e legs on this PR are red, but not because of this change. The KBN pre-build run (sscarduzio/readonlyrest_kbn run 32274394766) hit the Docker Hub unauthenticated rate limit on `tianon/gosu` and failed after 4 buildx attempts. Your new wait found that failed run and stopped fast with exit 6 and a clear message. That is the new code at work, and it is one more reason to land #1346 and its KBN-side equivalent.

Two small points:

1. The PR description says that `ci.yml` passes `KBN_REPO_GH_TOKEN` to each job. It does not. It passes `ROR_GH_TOKEN` (ci.yml lines 588 and 634), and the wait in the shared lib reads `ROR_GH_TOKEN` too. `KBN_REPO_GH_TOKEN` appears only in the contract comment of the e2e lib. CI.md in this PR has the correct name. Please correct the description, so that nobody looks for a secret that does not exist.

2. The `tag` input of `publish-pre-builds.yml` is optional. A manual dispatch without a tag makes a run with the title "ROR ES pre-build " and nothing after it. The search in the e2e lib is not affected, because it only matches when the caller gives a tag. If you want tidy titles, `run-name: ROR ES pre-build ${{ inputs.tag || inputs.es_versions }}` keeps the exact string when a tag is set. Feel free to ignore.

The contract text in the e2e lib has the same token-name mismatch as the description. I put that note on beshu-tech/readonlyrest-e2e-tests#111, because the text lives there.

PR #1352, opened 2026-08-20
[RORDEV-2183] Pull Docker Hub images through a mirror

coutoPL:
## Why

Docker Hub rejects our pulls with a bare `429 Too Many Requests`. It limits pulls per IP address and
ignores the account when it does, so the login we added earlier cannot prevent it. A CI runner takes
an ephemeral address from the Ubicloud pool and shares it with other tenants. When a neighbour fills
the bucket, our build fails. Docker Hub also counts every pull against one account quota of 200 per
6 hours, and one CI run makes about 60 pulls.

This change pulls `docker.io` images through `mirror.gcr.io`, which answers from a different host.
It serves `docker.io` only. It cannot serve `docker.elastic.co`, and it cannot accept a push.

## Where the mirror is applied

`ci/docker-hub-auth.sh` becomes `ci/configure-docker.sh`. It still authenticates, and it now also
points Docker at the mirror. Three clients pull images, and each one needs its own setting:

| Client | How |
|---|---|
| docker daemon | `registry-mirrors` in `/etc/docker/daemon.json` |
| buildx and BuildKit | `build-base/buildkitd.toml`, through `docker buildx create --config` |
| testcontainers | `DockerHubMirror.applyTo(...)`, called at each image name |

BuildKit does not read `daemon.json`, so the first two are separate. Only a job without `container:`
gets the daemon half: a job that runs inside the toolchains image cannot restart the host daemon.
The script detects that and skips it. The other two halves work everywhere.

The mirrored buildx builder has its own name, `ror_kbn_builder_mirrored`. Only our own builder
carries the config file, so the build must not reuse whatever builder the runner already has.

The test suite names its mirrored images one by one. testcontainers can prefix names by itself,
through `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX`, but it prefixes every name without a registry host.
A locally built name looks the same as a Docker Hub name, so it rewrote our own `ror-it-es:<hash>`
image and then failed to pull it. `DockerHubMirror` therefore reads our own variable, and each call
site says what it wants mirrored.

## Where the fallback happens

When the mirror cannot answer, the pull falls back to Docker Hub. The daemon does this by itself.
BuildKit does it too: this was measured on v0.32, both with a mirror that does not resolve and with
one that answers 404. It is not a promise for every failure, because moby/buildkit#1972 is still
open, so `retry_with_backoff` in `ci/ci-lib.sh` covers the rest. It wraps the two Gradle steps that
push a Docker image. A green run pays nothing for it. The first 15s delay comes only after a push
has already failed, and there the choice is 15s or a dead pipeline.

`ROR_DOCKER_HUB_MIRROR=false` switches the mirror off for a job.

## What stays on Docker Hub

- Every push. A pull-through cache is read-only.
- `docker manifest inspect` and `docker buildx imagetools`. They read a tag we pushed a moment
  before, and a cache can answer stale. Both run in the CLI, which reads neither mirror setting.
- `e2e_tests`. It is the only job that switches the mirror off, for the same reason.
- The `container:` image. The runner pulls it before step 1 starts, so no step can set a mirror
  first. That is about 11 pulls per CI run. Only `ghcr.io` can remove them, and that is not here.

## Other changes

- `DnsServerContainer` pulls `coredns/coredns:1.13.2` instead of building an image from a
  Dockerfile. It copies the Corefile and the zone file into the container, so the two
  `coredns-image/` resource folders go away. One pull replaces one local build.
- The `beshultd` namespace moves into `gradle.properties`, as `dockerImageNamespace`. A new
  `gradle_property` helper is the only reader of that file from the CI scripts. A second parser
  drifts.
- `DOCKER_AUTH_REQUIRED` is removed. When a job supplies credentials and they fail, the script now
  always stops that job. It never falls back to anonymous pulls, because such a fallback becomes a
  rate-limit failure in a different job one hour later.

## Summary by CodeRabbit

- **Improvements**
  - Docker image pulls can use a configurable mirror across CI builds, BuildKit, Docker daemon operations, and container-based tests.
  - Mirror configuration supports safe opt-out behavior and preserves existing Docker settings.
  - Transient registry failures during image publishing are retried automatically with increasing delays.
  - Docker authentication behavior is more reliable across pull requests and publishing workflows.
  - Container-based DNS and service tests now use mirrored images more consistently.

- **Documentation**
  - Added guidance on mirror configuration, limitations, quotas, supported clients, authentication, and opt-out behavior.


tests-utils/src/main/scala/tech/beshu/ror/utils/containers/DockerHubMirror.scala
```diff
@@ -0,0 +1,74 @@
...
+  private val propertyName = "hub.image.name.prefix"
+
+  /** The image name to pull. It is unchanged when no mirror is configured, and when the name already carries a registry
+    * host. The mirror serves docker.io only, so `docker.elastic.co/elasticsearch/elasticsearch` and
+    * `ghcr.io/shopify/toxiproxy` must pass through untouched.
+    */
+  def applyTo(imageName: String): String = {
+    val prefix = configuredPrefix
+    if (prefix.isEmpty || hasRegistryHost(imageName) || imageName.startsWith(prefix)) imageName
+    else prefix + imageName
```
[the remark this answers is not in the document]
coutoPL: This is empirically wrong. Its web research says mirror.gcr.io is a passive cache that 404s on a miss rather than fetching from Docker Hub, so a direct mirror.gcr.io/... reference would break for anything uncached.

ci/ci-lib.sh
```diff
@@ -6,9 +6,54 @@
...
+# Two failures look the same from outside, and both pass:
+#   1. Docker Hub answers a bare 429. It limits pulls per IP address and ignores the account, so a
+#      login does not prevent it, and waiting is the only answer.
+#   2. The mirror is unavailable. BuildKit falls back to Docker Hub for most mirror failures, but
+#      not for all of them (moby/buildkit#1972). See build-base/buildkitd.toml.
+#
+# The first delay is 30s. That is short against the minutes an abuse limit lasts, but three attempts
+# cover a short one and add nothing to a run that succeeds at once.
+retry_with_backoff() {
+  local attempts=${ROR_RETRY_ATTEMPTS:-3}
```
sscarduzio, line 48: OK the robustness, but if we regularly pay one or more 30+s retry tax, we should absolutely find a solution, not tolerable.

coutoPL: The retry only wraps the two Docker Hub pushes. Pulls go through the mirror this PR adds, and a green run pays nothing because the sleep runs only after a push has already failed. Each retry prints a log, so if it starts firing regularly, we see it in the log and fix the cause. AFAIR, the 429 at this level was not a problem.

coutoPL:
@claude review please

ci/configure-docker.sh
```diff
@@ -0,0 +1,325 @@
...
+_ror_docker_auth_failed() {
+  echo "[CI] Docker authentication FAILED. Cause: $1" >&2
+  [ -n "${GITHUB_ACTIONS:-}" ] && echo "::error::Docker authentication failed: $1"
+  return 1
+}
+
+_ror_docker_auth() {
+  local user token auth
+
+  if ! _ror_docker_auth_isset "${DOCKER_REGISTRY_USER:-}" ||
```
sscarduzio, line 183: **An unresolved Azure macro counts as supplied credentials, and the job stops.**

This PR corrects `azure-pipelines.yml` to map `DOCKER_REGISTRY_USER: $(DOCKER_HUB_USER)`. Azure leaves an undefined variable as the literal text `$(DOCKER_HUB_USER)`, which is not empty. `_ror_docker_auth_isset` accepts it, the script goes to the login path, `docker login` fails, and `_ror_docker_auth_failed` stops the job. `DOCKER_AUTH_REQUIRED: "false"` cannot save it, by design.

On `develop` the variable names were wrong, so the same build saw no credentials and continued with anonymous pulls. So an Azure build with no variable group linked (a pull request from a fork) now fails where it degraded before.

A guard that rejects a value of the form `$(...)` restores the old behaviour.

ci/configure-docker.sh
```diff
@@ -0,0 +1,325 @@
...
+  local backup="" merged
+
+  if sudo -n test -f "$config"; then
+    # The runner may keep its own settings here, so add the mirror to the file instead of replacing
+    # the file. jq does the merge, and without jq the script leaves the file alone.
+    if ! command -v jq >/dev/null 2>&1; then
+      echo "[CI] /etc/docker/daemon.json exists and jq is missing, so the daemon keeps its own settings."
+      return 0
+    fi
+    merged=$(sudo -n cat "$config" |
```
sscarduzio: **An empty `daemon.json` gives a blank file and a false success message.**

`jq` prints nothing and exits 0 for empty input. I measured this. If `/etc/docker/daemon.json` exists but is empty or holds only whitespace, `merged` becomes an empty string, the `||` guard does not fire, `tee` writes a blank file, and the daemon restarts without complaint. The backup is then deleted and line 127 prints `The docker daemon now mirrors Docker Hub.`

No mirror is configured, every daemon pull goes to Docker Hub, and the log says the opposite. That is the silent state this script exists to remove.

Test that `merged` is not empty and is valid JSON before you write it.

ci/ci-lib.sh
```diff
@@ -86,7 +133,8 @@ publish_ror_es_prebuild_plugin() {
   if [ "$FORCE_REBUILD_NORM" != "true" ] && docker_image_exists "${ES_DEV_IMAGE_REPO}:${SOURCE_TAG}"; then
     echo ">>> Sources unchanged (image for this commit already published), skipping build"
   else
-    if ! ./gradlew publishEsRorPreBuildDockerImage "-PesVersion=$ES_VERSION" </dev/null; then
+    # Retried because this build pulls base images, and a registry can answer 429.
+    if ! retry_with_backoff ./gradlew publishEsRorPreBuildDockerImage "-PesVersion=$ES_VERSION" </dev/null; then
```
sscarduzio: **The retry repeats deterministic failures.**

This sharpens @sscarduzio's point on the retry tax. The answer given covers the green run, which is correct. The cost is on the red run.

`retry_with_backoff` cannot tell a 429 from a compile error, a wrong `-PesVersion`, or a Dockerfile error. This task and `:module:pushRorDockerImage` both build multi-arch images before they push, so any deterministic failure now costs two more full builds and 45s of sleep before it reports the same result.

Retry only on a registry error (429 or 5xx in the task output), and a broken build fails at once.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -53,34 +54,53 @@ def buildxPushArgs = { List<String> tags ->
...
+        if (!currentBuilderIsEnough) {
             def inspectResult = execOps.exec {
                 executable = 'docker'
-                args = ['buildx', 'inspect', 'ror_kbn_builder']
+                args = ['buildx', 'inspect', multiplatformBuilderName]
                 ignoreExitValue = true
                 errorOutput = new ByteArrayOutputStream()
             }
             if (inspectResult.exitValue != 0) {
+                def createArgs = ['buildx', 'create', '--platform=linux/amd64,linux/arm64', "--name=${multiplatformBuilderName}".toString()]
```
sscarduzio, line 94: **`--config` applies only when the builder is created.**

The flag is inside `if (inspectResult.exitValue != 0)`. On a host that keeps docker state between runs (a self-hosted agent, a developer machine), `ror_kbn_builder_mirrored` survives. A later change to `build-base/buildkitd.toml` — a different mirror host, or one more registry — is then ignored in silence: buildx reuses the builder made with the old config, and the log still says the mirror is on.

The name-carries-the-config rule above separates a mirrored builder from an unmirrored one, but it does not separate one config version from the next. A hash of the config file in the builder name, or `buildx rm` before create, closes it.

ci/configure-docker.sh
```diff
@@ -0,0 +1,325 @@
...
+#    the login is necessary. A later CLI reads the variable and gives it priority over the login.
+#    That priority is safe here, because both halves use the same credentials. It is not safe if
+#    you add a second login with different credentials. Use this script instead.
+
+ROR_DOCKER_HUB_MIRROR_HOST="mirror.gcr.io"
+
+# The daemon runs on the host. A container job reaches it through the mounted socket and cannot
+# restart it, so this is best effort: it stops at the first thing it cannot do. The daemon also
+# falls back to Docker Hub by itself, so a job without a daemon mirror still works.
+_ror_docker_mirror_daemon() {
```
sscarduzio: **No pull request or push build runs this block.**

`_ror_docker_mirror_daemon` returns at the systemd test inside any container job, and `sudo` is never installed in the toolchains image. Of the jobs that source this script, only `build_toolchains_image` (schedule on develop, or dispatch) and `publish-pre-builds` / `publish` (dispatch) run without `container:`, and `e2e_tests` turns the mirror off.

So the code that rewrites `/etc/docker/daemon.json` and restarts the daemon — the largest blast radius in this PR — executes in no ordinary CI run, this PR's own green run included. The empty-file defect above, and any later regression here, ships green and appears first during a scheduled toolchains rebuild.

One dispatch run of `publish-pre-builds` before merge would exercise it.

ci/ci-lib.sh
```diff
@@ -5,10 +5,59 @@
 # lib is sourced rather than run (`source ci/ci-lib.sh && reap_ci_job_containers` resolved it to ".").
 CI_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 
+# Reads one key from gradle.properties, which holds the build's own values. The file sits beside this
+# one, so the caller's working directory does not matter.
+#
+# An absent key fails. An empty value would name a wrong image or a wrong version, and nothing later
+# would show the cause. Every reader of the file goes through here: a second parser drifts.
+gradle_property() {
```
sscarduzio, line 16: **`gradle_property` truncates a value that holds `=`.**

`awk -F= '$1==k {print $2}'` returns only the text between the first and the second `=`. The three keys read today hold no `=`, so nothing breaks now. The trap sits in a new shared helper that the description calls the only reader of that file, and `gradle.properties` is a file where a `-Dkey=value` form is normal. A duplicated key also gives a two-line value.

`sub(/^[^=]*=/, "")` on the last match reads the whole value.

sscarduzio (commented):
Thorough pass over the final state. The design is sound and the reasoning in the comments is unusually good. Six points below, most severe first.

**Two open threads I checked myself:**

1. CodeRabbit says `mirror.gcr.io` is a passive cache that 404s on a miss, so a direct `mirror.gcr.io/...` reference breaks for anything uncached. **@coutoPL is right and CodeRabbit is wrong.** I measured it: a first-ever manifest request returns 200, so the mirror fetches on a miss. My own first test showed a 404 only because I sent a narrow `Accept` header, which hides a manifest list. All four rewritten images (`coredns/coredns:1.13.2`, `osixia/openldap:1.5.0`, `rodolpheche/wiremock:2.5.1`, `ubuntu:24.04`) resolve on the mirror.
2. CodeRabbit says `ci/publish-ror-plugins.sh` shadows `retry_with_backoff` with a local copy. That is stale — there is one definition, in `ci/ci-lib.sh`.

Also confirmed fixed in this branch: the Azure credential variable names (`DOCKER_REGISTRY_USER` / `DOCKER_REGISTRY_PASSWORD`), the retry-configuration validation, and the `daemon.json` merge and restore.

ci/ci-lib.sh
```diff
@@ -5,10 +5,132 @@
...
+  done
+}
+
+# A --retry-if function for a command that pushes or pulls a Docker image. True when the output
+# holds a registry error, which another attempt can clear. A compile error, a wrong -PesVersion and
+# a broken Dockerfile give the same result every time, and both commands we retry build a
+# multi-arch image before they push, so a repeat of such a failure costs two more full builds.
+is_docker_registry_error() {
+  grep -Eqi \
+    'toomanyrequests|429 Too Many Requests|received unexpected HTTP status: 4?5[0-9][0-9]|unexpected status: 5[0-9][0-9]|50[0234] (Internal Server Error|Bad Gateway|Service Unavailable|Gateway Time-?out)|TLS handshake timeout|i/o timeout|connection reset by peer|unexpected EOF|net/http: request canceled' \
```
sscarduzio: Nit: `4?5[0-9][0-9]` matches `4500`-`4599`, which is not a status code. If `[45][0-9][0-9]` was meant, a repeat of a 401, a 403 or a 404 would be wrong — those give the same answer every time. So the safe correction is to drop the `4?` and keep 5xx.

429 is already covered by the two terms before it.

ci/ci-lib.sh
```diff
@@ -5,10 +5,132 @@
...
+  log=""
+  if [ -n "$retry_if" ]; then
+    log=$(mktemp) || return 2
+  fi
+
+  while true; do
+    if [ -n "$log" ]; then
+      # tee keeps the output on the console. PIPESTATUS holds the status of the command itself, not
+      # the status of tee.
+      "$@" 2>&1 | tee "$log"
```
sscarduzio: Nit, for later: this pipeline is a bare command, so a caller with `errexit` that does not put the call in a condition would stop here instead of repeating. Both call sites use `if ! retry_with_backoff ...`, which turns errexit off for the whole function, so nothing is wrong today.

`if ! { "$@" 2>&1 | tee "$log"; }; then status=${PIPESTATUS[0]}; fi` removes the trap for whoever writes the third call site.

sscarduzio (commented):
Re-reviewed at `1a36b11`. All six points are fixed. I checked each fix rather than the commit message.

| Point | Fix | Checked |
|---|---|---|
| Azure macro read as a credential | `ci/azure-templates/docker-credentials.sh` drops a `$(...)` literal before `configure-docker.sh` reads it | The fork case now sees no credentials and pulls anonymously, which is what `DOCKER_AUTH_REQUIRED=false` asks for |
| Blank `daemon.json` | The daemon half is gone | The defect cannot occur |
| Retry repeats a broken build | `--retry-if is_docker_registry_error` | A compile error now fails on the first attempt; `PIPESTATUS[0]` carries the real status, and every exit path removes the temporary log |
| `--config` counts once | The builder name carries a hash of the file, and a stale builder is dropped | `BuildxBuildersTest` covers it |
| `gradle_property` truncation | Reads everything after the first `=`, and keeps the last line | A `-Dkey=value` value now survives |
| Daemon half runs in no PR build | Removed, replaced by the `MIRROR` build argument | `build_toolchains_image` is now the one job that needs it, and it gets it |

**Dropping the daemon half also removed a hazard I had not yet posted.** The header invites a developer to source this file by hand, the mirror defaults to ON, and nothing guarded the daemon path with `GITHUB_ACTIONS`. On a Linux box with passwordless sudo it would have written `/etc/docker/daemon.json` and restarted the local daemon, which stops the containers the developer is running. I measured the three guards on my own machine and all three passed. The build-argument approach has none of that.

**Measured on the new toolchains path:** all five names answer on the mirror — `library/eclipse-temurin:{8,11,17,21}-jdk` and `library/docker:27.3.1-cli` — and `library/x` resolves the same with the prefix and without it. Together with the digest-pinned Docker Hub `FROM` lines in `es*x/Dockerfile`, a mirror cannot substitute content anywhere in this PR.

**One open item** from my second pass: `DockerHubMirror.applyTo` still has no test, while `BuildxBuildersTest` now shows the pattern. The same shape would cover the registry-host rule, the `library/` prefix and the idempotence guard. A follow-up is fine.

Two nits inline, and two more here:

- `ci/configure-docker.sh` lost its executable bit (100755 → 100644), while the new `ci/azure-templates/docker-credentials.sh` has it. Both are only ever sourced, so nothing breaks either way — but the two files should agree.
- `staleMirroredBuilders` matches `ror_kbn_builder_mirrored_`, so a builder named `ror_kbn_builder_mirrored` by the earlier revision of this branch is never dropped. This touches only a host that ran that revision.

CI is green on this head. LGTM.

sscarduzio (approved):
Approved at `1a36b11`.

All six review points are fixed, and I checked each fix against the code. CI is green on this head and the branch is CLEAN.

Two nits stand in the thread above, and neither blocks: the `4?` in the registry-error regex, and the bare pipeline that a future third caller could trip on. The one follow-up is a test for `DockerHubMirror.applyTo`, in the shape of the new `BuildxBuildersTest`.

Nice call deleting the daemon half instead of patching it. It removed the defect, the coverage gap and a developer-machine hazard in one move.

sscarduzio:
The three red e2e legs are the 429 this PR fixes — in the one job that opts out of the mirror. 🙂

e2e finds no prebuild for this commit, so it builds one, and BuildKit resolves `docker.io/tianon/gosu:1.17@sha256:5591f0…` with the mirror off. Docker Hub answers `429 Too Many Requests`. The new retry did its job: it recognised the 429 and tried again at 15s and 30s. The runner is self-hosted, so the address is static and the limit outlasted all three attempts.

I think `ROR_DOCKER_HUB_MIRROR: 'false'` on `e2e_tests` can go now. Its reason went away when you deleted the daemon mirror:

- The ROR ES and KBN images come from compose, which has no mirror any more, so nothing can answer stale for them.
- What the flag still switches off is BuildKit base-image resolution and the 4 test-suite names.
- `tianon/gosu` is digest-pinned. A digest-pinned pull cannot be stale — it matches or it fails.

If you want to keep a guard, leave the flag `true` and blank only `ROR_DOCKER_HUB_MIRROR_PREFIX`.

A rerun will probably go green, but it would hide this rather than close it.

coutoPL:
@sscarduzio, the issue you are talking about is in the ROR KBN repo. I will port these changes to the project. Now, I retried it and will merge as soon as I see all green jobs.

PR #1354, opened 2026-08-23
[RORDEV-1563] Lookup indices - refactoring

mgoworko:
## Changelog entries

```yaml
- type: security
  components: [es]
  text: "When ReadonlyREST could not parse an ES|QL query, it sent the query to Elasticsearch unchanged. ReadonlyREST now forbids such a query when the ACL limits the user's indices."

- type: security
  components: [es]
  text: "Fix an authorization bypass in ES|QL PROMQL queries (ES 9.4+): a query with no `index` parameter read every `metrics-*` index. ReadonlyREST now writes the allowed indices into such a query."
```

## Summary by CodeRabbit

* **New Features**
  * Added parser-aware ES|QL index filtering and query rewriting.
  * Added support for `LOOKUP JOIN`, source commands, nested queries, comments, metadata clauses, remote selectors, and PROMQL.
  * Added validation that rewritten queries reference only authorized indices.

* **Bug Fixes**
  * Requests are rejected when index access cannot be safely determined.
  * Anonymous query parameters and malformed index lists are handled safely.
  * Unauthorized or empty index results are safely masked.
  * Preserved interruption results when handling nonexistent indices.

* **Tests**
  * Expanded coverage for ES|QL authorization, rewriting, errors, and lookup joins.


core/src/main/scala/tech/beshu/ror/es/esql/IndexListLocator.scala
```diff
@@ -0,0 +1,156 @@
...
+  private val sourceCommandIndexList: Regex =
+    """(?is)^\s*(?:FROM|TS)\s+(.*?[^\s,])(?:\s+METADATA\b.*|\s*\[\s*METADATA\b.*|\s*)$""".r
+
+  private val promqlCommand: Regex = """(?is)^\s*PROMQL\b.*""".r
+
+  private val comment: Regex = """(?s)/\*.*?(?:\*/|$)|//[^\n]*""".r
+
+  private val queryParameter: Regex = """^\s*\?\??[A-Za-z_0-9]*\s*$""".r
+
+  def locatedIn(
```
mgoworko, line 49: Uses the line/column that the ES parser reports, instead of using raw string operations

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,205 @@
...
+      read.indexList.split(',').asSafeList.filter(_.nonEmpty).flatMap(RequestedIndex.fromString)
+    )
+
+}
+
+/** A rewritten query, together with what ES has to read out of it for the rewrite to have done its job. */
+final case class ReplacedQuery(query: Query, intendedReads: List[IndexListRead]) {
+
+  /** Held to what ES reads back out of the rewrite - the only thing saying which indices it will really run against. */
+  def checkedAgainst(esReads: List[IndexListRead]): Either[Query.Rejection, Query] = {
```
mgoworko: re-parses the rewritten query and compares what ES reads to what ROR intended

core/src/main/scala/tech/beshu/ror/es/esql/IndexListLocator.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Everything between the command keyword and a `METADATA` clause, in either the current or the ES 8.x form. The
+   * list has to end on a character that could close an index name, or `FROM a, metadata` would read its last index
+   * as the start of the clause.
+   */
+  private val sourceCommandIndexList: Regex =
+    """(?is)^\s*(?:FROM|TS)\s+(.*?[^\s,])(?:\s+METADATA\b.*|\s*\[\s*METADATA\b.*|\s*)$""".r
+
+  private val promqlCommand: Regex = """(?is)^\s*PROMQL\b.*""".r
```
mgoworko, line 41: ESQL syntax is handles explicitly, instead of raw string operations. In the previous approaches (both on current develop branch, and in the first abandoned approach to RORDEV-1636) the index pattern was replaced as a string operation. Not detected index patterns were just silently allowed. This new approach is more resilient.

mgoworko (commented):
I've abandoned the previous, minimalistic approach to RORDEV-1636. The previous approach was raw-string-based. The new approach is much larger change, but introduces strong typing of ESQL parsing and operations, and additional verifications.

coutoPL:
> https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1354#pullrequestreview-5003035229

But we do it in two stages. The first stage - the one from the previous PR. The second one - better design with types. WDYT?

coutoPL (commented):
please let me know when it's my time to take a look at this

mgoworko:
@coutoPL it is ready for the new review

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/EsqlLookupJoinSuite.scala
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  private lazy val catalogOnlyEsqlManager = new EsqlApiManager(basicAuthClient("catalog", "test"), esVersionUsed)
+  private lazy val pricesOnlyEsqlManager = new EsqlApiManager(basicAuthClient("prices", "test"), esVersionUsed)
+  private lazy val bothIndicesEsqlManager = new EsqlApiManager(basicAuthClient("both", "test"), esVersionUsed)
+  private lazy val bookWildcardEsqlManager = new EsqlApiManager(basicAuthClient("books", "test"), esVersionUsed)
+  private lazy val allJoinTargetsEsqlManager = new EsqlApiManager(basicAuthClient("joins", "test"), esVersionUsed)
+  private lazy val adminEsqlManager = new EsqlApiManager(basicAuthClient("admin", "container"), esVersionUsed)
+
+  "An ESQL LOOKUP JOIN request" should {
```
coutoPL: I see we test the lookup join in two places. I'm not sure about it.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/EsqlLookupJoinSuite.scala
```diff
@@ -0,0 +1,283 @@
...
+import org.scalatest.wordspec.AnyWordSpec
+import tech.beshu.ror.integration.suites.base.support.BaseSingleNodeEsClusterTest
+import tech.beshu.ror.integration.utils.{ESVersionSupportForAnyWordSpecLike, SingletonPluginTestSupport}
+import tech.beshu.ror.utils.TestUjson.ujson
+import tech.beshu.ror.utils.TestUjson.ujson.{Null, Num, Str}
+import tech.beshu.ror.utils.elasticsearch.{DocumentManager, EsqlApiManager, IndexManager}
+import tech.beshu.ror.utils.httpclient.RestClient
+import tech.beshu.ror.utils.misc.{CustomScalaTestMatchers, Version}
+
+/** Own cluster, not the shared xpack one: `book_*` is caught by the `*` its wildcard cases assert on. */
```
coutoPL: The BaseXpackApiSuite is huge - this is true. Maybe it's high time to split it. 

In the past, I had an idea to have a suite per API type. E.g., a suite for ESQL. Can we do that? 
And because of our magic with X-Pack in patches, I see a value in running the same suite with enabled and disabled X-Pack security.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,224 @@
...
+import tech.beshu.ror.implicits.*
+import tech.beshu.ror.syntax.*
+import tech.beshu.ror.utils.ScalaOps.*
+
+final case class Query(value: String) extends AnyVal
+
+object Query {
+
+  /** Where in a query's text something sits, as a half-open range of characters. */
+  final case class TextSpan(start: Int, end: Int)
```
coutoPL: Some of the types belong to the public, but some belong to the private domain.
We should consider making them package-private or move.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  /** Where in a query's text something sits, the way ES reports it: a 1-based line and a 0-based column. */
+  final case class SourceLocation(line: Int, column: Int)
+
+  /** Why a query cannot be run at all, once ROR knows it cannot hold it to the indices the ACL allowed. */
```
coutoPL: Do we need this comment? It documents kind of obvious things

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,224 @@
...
+  object Rejection {
+
+    case object CannotParse extends Rejection
+
+    final case class CannotReadIndexList(failure: LocatedIndexList.ReadingFailure) extends Rejection
+
+    final case class NotReplacedAsIntended(intendedIndexLists: List[String], readIndexLists: List[String])
+        extends Rejection
+
+    implicit val show: Show[Rejection] = Show.show {
```
coutoPL: `Show` instances for the public domain should be defined in the `LogsShowInstances`

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,224 @@
...
+  final case class SourceLocation(line: Int, column: Int)
+
+  /** Why a query cannot be run at all, once ROR knows it cannot hold it to the indices the ACL allowed. */
+  sealed trait Rejection
+
+  object Rejection {
+
+    case object CannotParse extends Rejection
+
+    final case class CannotReadIndexList(failure: LocatedIndexList.ReadingFailure) extends Rejection
```
coutoPL: Cannot read or maybe "cannot extract indices from query"? Or sth like that?

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,224 @@
...
+
+    final case class NotReplacedAsIntended(intendedIndexLists: List[String], readIndexLists: List[String])
```
coutoPL: This is hard to reason about. Probably badly named. 
The public domain should be easy to grasp without reading the implementation.
 
The context is as follows: 
- input: a raw query,  allowed indices
- operations: create new query with index substitution

And knowing that the rejection types should be easy to grasp.

core/src/test/scala/tech/beshu/ror/unit/es/esql/IndexListReplacerTest.scala
```diff
@@ -0,0 +1,699 @@
...
+import org.scalatest.matchers.should.Matchers.*
+import org.scalatest.wordspec.AnyWordSpec
+import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, RequestedIndex}
+import tech.beshu.ror.es.esql.*
+import tech.beshu.ror.es.esql.LocatedIndexList.ReadingFailure
+import tech.beshu.ror.es.esql.LocatedIndexList.ReadingFailure.*
+import tech.beshu.ror.es.esql.Query.SourceLocation
+
+/**
+ * Covers reading the index lists ES reported out of the query text and replacing them in one go, because that is
```
coutoPL: bla bla bla ... we are testing the IndexListReplacer logic, right? The class has a comment with a helper explanation of how it works, right? Do we need to add anything more?

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,224 @@
...
+import cats.data.NonEmptyList
+import cats.syntax.traverse.*
+import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, IndexName, RequestedIndex}
+import tech.beshu.ror.accesscontrol.matchers.PatternsMatcher
+import tech.beshu.ror.es.esql.Query.SourceLocation
+import tech.beshu.ror.implicits.*
+import tech.beshu.ror.syntax.*
+import tech.beshu.ror.utils.ScalaOps.*
+
+final case class Query(value: String) extends AnyVal
```
coutoPL: I read the core changes. I will put my comment here because I think the shape of the Query should be different, but my comment is related to the logic of the index locator and replacer. 

---
IMO, the problem with the current implementation is the fact that the index locator and replacer leak into the ES modules code.

Maybe I'm missing something, but I see it like this:

```
**The problem:** 

An ESQL query could carry indices, and some of the indices may not be allowed. ROR needs to decide whether the indices could be replaced or if the whole request should be forbidden because some required indices are not allowed.
```

ROR ACL is supposed to filter the indices. As a result, we will have an allowed indices set. 
We have an ESQL request, and we can extract a raw query (string) from it. And in general, we have to do the following:
1. extract indices from the query
2. run the ACL processing
3. If successful, get the allowed indices and replace the indices in the query
4. Save the new query in the request

That's why I think we should improve the Query type. IMO, the Query type should:
1. have a smart constructor that will allow us to build the Query instance from a raw query string
2. represent From query, LookupJoin query, and other types of queries (if, in the future, new indices-related query will be added, we could extend the hierarchy)
3. The instance should:
3.1. know where its indices are defined (location) and what are the types of them (and hold the invariants - e.g., a LookupJoin index is one element only and full name; From indices are a list, and there could be wildcards)
3.2. Should be able to return index names (`IndexName` instances, not `string` ones) related to it
3.3. Should be able to return a new query based on the passed allowed indices list (the replace logic - it can be done as a public method of the type or an extension method)
3.4. Should have an extension or method to stringify it

Thanks to this:
* all index-locating and replacing logic is encapsulated in the Query (it doesn't leak to the modules' code)
* the public interface is easy and represents **the problem** defined above

WDYT?

coutoPL (changes requested):

coutoPL:
BTW, @mgoworko, I read the changelog entries and TBH, I have no idea what we fixed :P Knowing the codebase ;) So, there is no way that clients will get it. Please rewrite them - the client's perspective is crucial.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/EsqlLookupJoinSuite.scala
mgoworko: Yes, it was in two places in the test code, I did not catch it. There is now a new `BaseEsqlApiSuite`, next to the rest of the ES|QL tests.

integration-tests/src/test/scala/tech/beshu/ror/integration/suites/EsqlLookupJoinSuite.scala
mgoworko: Done. As described in previous comment, there is a new `BaseEsqlApiSuite` with all the ES|QL
tests moved into it. It runs in both X-Pack modes.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: Yes, it way mixed-up a bit. Now public means "ES module needs it", and everything else is `private[esql]`

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: No, removed. I've already removed almost all agent-generated comments from the code, but left them in the domain, to better document the (quite convoluted) model, which heavily depends on ES internal model. But yes, some of those comments are trivial.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: Renamed to `CannotExtractIndices`

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: There were 2 separate error hierarchies: `RequestClassification.Error` for the first step and
a separate one for the query rewrite. Now there is one `Rejection` with variants:

- `CannotParseQuery` - ES could not parse the query
- `CannotExtractIndices(failure)` - query is parsed, but we could not read the indices from it (failure explains why)
- `SubstitutionNotConfirmed(intended, read)` - ROR rewrote the query, but then it is verified after the rewrite, and ES read something else than intended.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: Done, the `Query` now follows this proposal. One difference from the proposal (point 2). You suggested query types like "a FROM query" and "a LOOKUP JOIN query". But one query can contain both, so the type hierarchy describes the lower level - the index lists inside the query.

es811x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/BaseFilterableEsRequestContext.scala
```diff
@@ -63,24 +63,26 @@ abstract class BaseFilterableEsRequestContext[R <: ActionRequest](
...
-        filteredRequestedIndices = NonEmptyList.of(discoveredIndices.randomNonexistentLocalIndex()),
-        filter = None,
-        fieldLevelSecurity = None
-      )
-      Modified
+      updateToNonexistentIndices(NonEmptyList.of(discoveredIndices.randomNonexistentLocalIndex()))
+    }
+  }
+
+  private def updateToNonexistentIndices(
```
coutoPL, line 75: The name of the private function refers to "non-existent indices". But its implementation doesn't bother. So, it means that the context the name brings with it is not needed (and may be misleading).

es811x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/BaseFilterableEsRequestContext.scala
```diff
@@ -63,24 +63,26 @@ abstract class BaseFilterableEsRequestContext[R <: ActionRequest](
...
+      indices: NonEmptyList[RequestedIndex[ClusterIndexName]]
+  ): ModificationResult = {
+    update(
+      request = actionRequest,
+      filteredRequestedIndices = indices,
+      filter = None,
+      fieldLevelSecurity = None
+    ) match {
+      case ShouldBeInterrupted => ShouldBeInterrupted
+      case _                   => Modified
```
coutoPL, line 85: I don't know if I understand this. Why is the other-than-ShouldBeInterrupted result rewritten to Modified? Maybe we should just remove the match and return the result returned by the update method?

es811x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/EsqlIndicesEsRequestContext.scala
```diff
@@ -76,57 +65,17 @@ class EsqlIndicesEsRequestContext private (
       filter: Option[Filter],
       fieldLevelSecurity: Option[FieldLevelSecurity]
   ): ModificationResult = {
-    requestClassification match {
-      case Right(r @ EsqlRequestClassification.IndicesRelated(tables))
-          if filteredRequestedIndices.toList.toCovariantSet != r.requestedIndices =>
-        updatedRequestWithIndicesNarrowedTo(request, tables, filteredRequestedIndices, filter, fieldLevelSecurity)
+    EsqlRequestHelper.modifyIndicesOf(request, esqlQuery, filteredRequestedIndices) match {
```
coutoPL, line 68: It would be good to modify this part of the code too. 
Now we're trying to modify the indices of the request. IMO, we should try to create a new query instead (with the new, narrowed indices). 

Then, if the query can be created, we do the following with the request:
1. modify the query inside it
2. apply the field-level security to it
3. apply the filter to it

And return the UpdateResponse, which applies the field-level security to the response.

Moreover, we have the following now:
```
EsqlRequestHelper.esqlQueryOf(actionRequest)
```
(can be renamed to `EsqlRequestHelper.extractEsqlQueryFrom(actionRequest)`)

This means it would be good to have also:
```
EsqlRequestHelper.setEsqlQueryTo(actionRequest)
```
(and we could use it in the 1st point mentioned above)

WDYT?

es811x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
```diff
@@ -24,31 +24,29 @@ import org.joor.ReflectException
...
-        setQuery(request, newQuery)
-        IndicesModificationResult.IndicesModified
-      case EsqlQueryRewriteResult.CannotRewriteQuery(reason) =>
-        IndicesModificationResult.CannotModifyIndices(reason)
-    }
+      query: Query,
+      allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
+  ): Either[Rejection, Unit] = {
+    query
+      .narrowedTo(allowedIndices, readerFor(request))
```
coutoPL, line 48: It looks like the reader shouldn't be passed here. It's needed in the smart constructor, so the query instance should hold it.

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
```diff
@@ -0,0 +1,129 @@
...
+  def indices: Set[RequestedIndex[ClusterIndexName]]
+
+  def narrowedTo(
+      allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]],
+      reader: EsqlIndexListsReader
+  ): Either[Rejection, Query]
+
+}
+
+object Query extends Logging {
```
coutoPL, line 39: Shouldn't we use ROR's logging trait?

es811x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
```diff
@@ -24,31 +24,29 @@ import org.joor.ReflectException
...
 import tech.beshu.ror.utils.ScalaOps.*
+import tech.beshu.ror.utils.slf4j.Logging
 
 import java.util.List as JList
+import java.util.function.Predicate as JPredicate
 import scala.jdk.CollectionConverters.*
 import scala.util.{Failure, Success, Try}
 
-object EsqlRequestHelper {
+object EsqlRequestHelper extends Logging {
```
coutoPL, line 40: Shouldn't we use ROR's logging trait?

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
```diff
@@ -0,0 +1,129 @@
...
+      if (allowedIndices.toList.toCovariantSet == indices) Right(this)
+      else {
+        val replaced = IndexListReplacer.replacing(text, indexLists, allowedIndices)
+        reader.indexListsIn(replaced.query) match {
+          case Right(reported) =>
+            replaced
+              .checkedAgainst(reported.map(_.read).filterNot(_.indexListIsEmpty))
+              .flatMap(narrowed => confirmed(located(narrowed, reported)))
+          case Left(cause) =>
+            logger.warn("Elasticsearch cannot parse the ES|QL query ReadonlyREST rewrote", cause)
```
coutoPL, line 77: it would be good to attach the request ID to the logger. I'm not sure if we should attach the request ID to the method or move the logging outside.

coutoPL (changes requested):

coutoPL:
The changelog entries review:
ad1. ok
ad2. "a query without the index parameter could read all metrics-* indices, including ones the user isn't granted." I'm not sure if I understand why we talk about metrics-* here. Could you please explain?
ad3. "When ReadonlyREST forbids an ES|QL query, the log says which part of the query caused it and how to change the query". IMO, we should not add it. It's kind of an improvement, but the end user creates the query, so 1) the admin sees in the logs that there was a problem with some query, but the admin cannot change it 2) the end user doesn't see the logs.

es811x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/BaseFilterableEsRequestContext.scala
mgoworko: renamed to `updateIndicesTo`

es811x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/BaseFilterableEsRequestContext.scala
mgoworko: Yes, the match is removed. It made no sense, it was only a leftover of some previous intermediate version.

es811x/src/main/scala/tech/beshu/ror/es/handler/request/context/types/EsqlIndicesEsRequestContext.scala
mgoworko: Done. The `update` now first narrows the query. When that works, does the three remaining steps. Helper methods renamed accordingly.

es811x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
mgoworko: Logging trait removed (unused)

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
mgoworko: requestId attached, logging left in the same place

core/src/main/scala/tech/beshu/ror/es/esql/IndexListLocator.scala
```diff
@@ -0,0 +1,187 @@
...
+
```
mgoworko, line 44: About PROMQL (related to the changelog entry you quoted as unclear):
- PROMQL is a new ES|QL source command that ES 9.4 added, GA in 9.5. It runs a Prometheus query language expression over time series indices, like that: `PROMQL index=metrics-app step=1m max by (host) (cpu)`                                                                                                                                         
- it was added, so that Prometheus and Grafana clients can query Elasticsearch

LLM reviewing this PR noticed, that the new PROMQL-type query is not explicitly handled by ROR. The PROMQL by default does not contain list of indexes. That is why it is not trivial to limit it using index list rewriting. And that is why in the version you reviewed, I added the security fix, to forbid the PROMQL request, if the use does not have universal permission for all indexes.


After the review, and after looking at it again, I decided to improve it, and ROR now writes the parameter itself. It puts the indices the ACL resolved after the keyword (ES parser tells us per request which indices the query reads):
  - sent:      `PROMQL step=1m max by (host) (cpu)`                                                                                                                                                                 
  - runs as:  `PROMQL index=metrics-app step=1m max by (host) (cpu)`

ROR then re-parses its own rewrite and forbids the query unless ES reads exactly those indices back out of it.

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
```diff
@@ -0,0 +1,136 @@
...
+      allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]
+  )(
+      implicit requestId: RequestId
+  ): Either[Rejection, Query]
+
+}
+
+object Query extends RequestIdAwareLogging {
+
+  private[esql] final case class TextSpan(start: Int, end: Int)
```
coutoPL: IMO, we should move it somewhere else. This class is not used in Query

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  /** A 1-based line and a 0-based column, the way ES reports them. */
+  final case class SourceLocation(line: Int, column: Int)
```
coutoPL: the same this class

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,162 @@
...
+import enumeratum.{Enum, EnumEntry}
+import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, IndexName, RequestedIndex}
+import tech.beshu.ror.accesscontrol.matchers.PatternsMatcher
+import tech.beshu.ror.es.esql.Query.SourceLocation
+import tech.beshu.ror.syntax.*
+import tech.beshu.ror.utils.ScalaOps.*
+
+sealed trait Rejection
+
+object Rejection {
```
coutoPL: I would move it to the Query.scala. WDYT?

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,162 @@
...
+  final case class CannotExtractIndices(failure: ReadingFailure) extends Rejection
+
+  final case class CannotParseRewrittenQuery(intendedIndexLists: List[String]) extends Rejection
+
+  final case class SubstitutionNotConfirmed(intendedIndexLists: List[String], readIndexLists: List[String])
+      extends Rejection
+
+}
+
+sealed trait ReadingFailure
```
coutoPL: It looks like all of the following could be a private domain of the esql package

core/src/main/scala/tech/beshu/ror/es/esql/EsqlIndexListsReader.scala
```diff
@@ -0,0 +1,23 @@
...
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.es.esql
+
+trait EsqlIndexListsReader {
+
+  def indexListsIn(query: String): Either[Throwable, List[ReportedIndexList]]
```
coutoPL: The ReportedIndexList and all its dependent classes should be defined in the `EsqlIndexListsReader` companion object.

coutoPL: Can the List be empty? What does it mean?

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,162 @@
...
+
+object IndexListRead {
+
+  final case class SourceCommand(indexList: String) extends IndexListRead
+
+  final case class LookupJoin(indexList: String) extends IndexListRead
+
+}
+
+final case class ReportedIndexList(read: IndexListRead, writtenAt: SourceLocation, writtenText: String)
```
coutoPL: I don't know if the naming is good ...
1. ReportedIndexList - why "reported"; why list? 
2. `IndexListRead` what does it mean?
3. EsqlIndexListsReader

IMO, we have to improve it.

core/src/main/scala/tech/beshu/ror/es/esql/EsqlIndexListsReader.scala
```diff
@@ -0,0 +1,23 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with ReadonlyREST.  If not, see http://www.gnu.org/licenses/
+ */
+package tech.beshu.ror.es.esql
+
+trait EsqlIndexListsReader {
```
coutoPL: Maybe EsqlQueryIndicesReader?

core/src/main/scala/tech/beshu/ror/es/esql/EsqlIndexListsReader.scala
coutoPL: I have a doubt about the returned list. In the ReportedIndexList type, we have information about the type of query. But IMO, we should not return a list but a shape that is specific to the query. E.g., we know that LookupJOIN query has a source and lookup index, FROM query has a list of indices, etc. 

Maybe the generalisation in this place went too far?

coutoPL (changes requested):

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
mgoworko: Ok, TextSpan is now in domain, package private.

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
mgoworko: Ok, SourceLocation is now in the EsqlQueryIndicesReader companion object

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: Ok, Rejection and ReadingFailure are now in Query.scala.

core/src/main/scala/tech/beshu/ror/es/esql/EsqlIndexListsReader.scala
mgoworko: Yes, the list can be empty, when the query reads no index, for example ROW or SHOW INFO.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: Renamed to follow ES naming conventions when possible: IndexPatternInQuery, IndexPatternRole and EsqlQueryIndicesReader.

core/src/main/scala/tech/beshu/ror/es/esql/EsqlIndexListsReader.scala
mgoworko: The reader now returns QueryIndices(fromSources, lookupJoins). Each field is a list, because the number of indices in each role is not fixex, some AI-generated examples:                                                                              
 - ROW k = "x" | LOOKUP JOIN lk ON k has a lookup index and no source index.                                                                                                                                                
 - FROM src | LOOKUP JOIN lk1 ON k | LOOKUP JOIN lk2 ON k has two lookup indices.                                                                                                                                           
- FROM a, (FROM b | LOOKUP JOIN lk ON k) has two source indices.                                                                                                                                                           
                                                                                                                                                          
I also tried to implement a type hierarchy that shows the full query structure: subqueries, joins etc.. It gives no clear gain. The ACL checks each index separately. The replacer has no need for such information. No code would use the rest of the structure.

core/src/main/scala/tech/beshu/ror/es/esql/EsqlQueryIndicesReader.scala
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Both lists are empty for a query that reads no index, e.g. `ROW` or `SHOW INFO`. Neither has a fixed size: a
+   * source command with subqueries reads from each of them, and a query can have many `LOOKUP JOIN` commands.
+   */
+  final case class QueryIndices(fromSources: List[IndexPatternInQuery], lookupJoins: List[IndexPatternInQuery])
+
+  final case class IndexPatternInQuery(indexPattern: String, writtenAt: SourceLocation, writtenText: String) {
+
+    /** ES reports an empty pattern for a source command of only subqueries. */
+    private[esql] def isEmpty: Boolean = indexPattern.isBlank
```
coutoPL, line 38: I'm not sure if I get it. Why can the index pattern be empty?

core/src/main/scala/tech/beshu/ror/es/esql/EsqlQueryIndicesReader.scala
```diff
@@ -0,0 +1,45 @@
...
+
+object EsqlQueryIndicesReader {
+
+  /**
+   * Both lists are empty for a query that reads no index, e.g. `ROW` or `SHOW INFO`. Neither has a fixed size: a
+   * source command with subqueries reads from each of them, and a query can have many `LOOKUP JOIN` commands.
+   */
+  final case class QueryIndices(fromSources: List[IndexPatternInQuery], lookupJoins: List[IndexPatternInQuery])
+
+  final case class IndexPatternInQuery(indexPattern: String, writtenAt: SourceLocation, writtenText: String) {
```
coutoPL, line 35: I wonder if we shouldn't use the `IndexPattern` class from the ROR domain here?

es94x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
```diff
@@ -101,74 +88,57 @@ object EsqlRequestHelper {
         .get[Any]()
     }
 
-    def createStatementBasedOn(request: CompositeIndicesRequest): Either[ClassificationError, Statement] = {
-      createStatement(request).flatMap(statementWithIndices)
+    override def indicesIn(query: String): Either[Throwable, QueryIndices] = {
+      createStatement(query).map(statement => queryIndicesIn(planOf(statement)))
```
coutoPL, line 92: **Reflection failures escape `indicesIn` instead of becoming a clean rejection.**

Only `createStatement` is inside a `Try`. `planOf` and `queryIndicesIn` do most of the reflection (`plan()`, `collect()`, `indexPattern()`, `indexMode()`, `source()`, `text()`, `getLineNumber()`) with no guard. A `ReflectException` from those calls goes out of a method whose type says `Either[Throwable, ...]`.

The code before this change wrapped the same operations in `Try(indicesFrom(statement))` and mapped a failure to `ClassificationError.IndicesExtractionException`, which became `ModificationResult.ShouldBeInterrupted`.

Scenario: an ES patch release renames `UnresolvedRelation.indexPattern()`. These are ES-internal APIs, which is why the old code guarded them. Each `_query` request then throws out of `EsqlIndicesEsRequestContext.requestedIndicesFrom` during ACL evaluation, and out of `Query.WithIndices.narrowedTo` during `modifyRequest`, instead of getting a clean denial.

Suggested fix: put the body of the `map` in `Try(...).toEither`.

The same code is in all 11 ES modules: `es811x`, `es812x`, `es813x`, `es814x`, `es815x`, `es816x`, `es818x`, `es90x`, `es91x`, `es92x`, `es94x`.

---
*Agent review finding. Please verify it before you act on it.*

core/src/main/scala/tech/beshu/ror/es/esql/IndexListReplacer.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  private def indexListOf(indices: NonEmptyList[ClusterIndexName]): String =
+    indices.toList.map(_.stringify).mkString(",")
+
+  private def textOf(syntax: IndexListSyntax, indexList: String): String = syntax match {
+    case IndexListSyntax.BareIndexList        => indexList
+    case IndexListSyntax.PromqlIndexParameter => s" index=$indexList"
+  }
+
+  private def rewritten(query: String, edits: List[Edit]): String = {
```
coutoPL, line 132: **The overlapping-span guard is not in the new splice.**

`spliceIndexLists`, which this code replaces, rejected overlapping spans with the message "the index lists of the tables overlap in the query text".

Now, if two located spans overlap, the descending fold applies the later edit first, then indexes the shortened text with an offset from the original text. Example: edit A `(30,40)` replaced with a 2-character mask makes a 40-character query 32 characters long; edit B `(10,35)` then calls `text.substring(35)` and throws `StringIndexOutOfBoundsException` out of `narrowedTo`.

`ReadingFailure.SubqueryInSourceCommand` and the empty-pattern filter make this hard to reach today. But the failure is an unguarded crash, not a rejection.

---
*Agent review finding. Please verify it before you act on it.*

es94x/src/main/scala/tech/beshu/ror/es/utils/EsqlRequestHelper.scala
```diff
@@ -16,81 +16,68 @@
...
+    Query.from(getQuery(request), readerFor(request))
+  }
 
-  def classifyEsqlRequest(request: CompositeIndicesRequest): Either[ClassificationError, EsqlRequestClassification] = {
-    createStatement(request) match {
-      case Right(statement: IndicesRelatedStatement) => Right(IndicesRelated(statement.indices))
-      case Right(command: OtherCommand)              => Right(NonIndicesRelated)
-      case Left(error)                               => Left(error)
-    }
+  def setEsqlQueryTo(request: CompositeIndicesRequest, query: Query): Unit = {
```
coutoPL, line 53: **Query parameters stay unchanged when ROR replaces a parameter placeholder.**

`IndexListLocator.checkHoldsReportedIndexPattern` accepts a bound parameter (`?index`, `?1`) as the located index list. `IndexListReplacer` then replaces that placeholder text with literal index names (unit test: `IndexListReplacerTest.scala:505-518`). `setEsqlQueryTo` writes only the query string, thus the request keeps the parameter in `params`, and the query no longer refers to it.

The re-parse in `narrowedTo` uses the same `params`, thus the failure mode is safe (`CannotParseRewrittenQuery` -> 403). But there is no integration test for parameterized ES|QL. A query whose index comes from `?index` can get a rejection on each ES version where the analyzer refuses unused parameters.

---
*Agent review finding. Please verify it before you act on it.*

mgoworko: Verified, that the ES accepts a parameter that the query does not use. Added integration tests for that case.

core/src/main/scala/tech/beshu/ror/es/esql/IndexListReplacer.scala
mgoworko: Confirmed, according to AI analysis it applies to FORK query. The fix removes repeats in QueryIndices.withoutRepeats, and the locator now rejects overlapping spans again. New unit tests and integration tests for FORK query.

core/src/main/scala/tech/beshu/ror/es/esql/EsqlQueryIndicesReader.scala
mgoworko: You are right, the comment was wrong and the empty pattern makes no sense. If ES ever reports an empty list, ROR now rejects the query.

core/src/main/scala/tech/beshu/ror/es/esql/EsqlQueryIndicesReader.scala
mgoworko: I think that we cannot do that. The `IndexPattern` holds one pattern. But the ES parser reports a list, for example a,b. `LocatedIndexList` splits that list into `RequestedIndex` values. To make this more readable, I renamed the field to reportedIndexList.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,128 @@
...
+import enumeratum.{Enum, EnumEntry}
+import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, IndexName, RequestedIndex}
+import tech.beshu.ror.accesscontrol.matchers.PatternsMatcher
+import tech.beshu.ror.es.esql.EsqlQueryIndicesReader.QueryIndices
+import tech.beshu.ror.syntax.*
+import tech.beshu.ror.utils.ScalaOps.*
+
+private[esql] final case class TextSpan(start: Int, end: Int)
+
+private[esql] sealed trait IndexPatternRole extends EnumEntry {
```
coutoPL: Do we need the Enum here?

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  def describe(indexPattern: String): String = this match {
```
coutoPL: TBH, this is weird for me. Hard to get what's going on.

tests-utils/src/main/scala/tech/beshu/ror/utils/elasticsearch/EsqlApiManager.scala
```diff
@@ -29,6 +29,12 @@ class EsqlApiManager(restClient: RestClient, esVersion: String) extends BaseMana
     call(createSqlQueryRequest(selectQuery), new EsqlResult(_))
   }
 
+  def execute(selectQuery: String, params: ujson.Arr): EsqlResult = {
+    val request = createSqlQueryRequest(selectQuery)
+    request.setEntity(new StringEntity(s"""{ "query": "$selectQuery", "params": ${params.render()} }"""))
```
coutoPL: This solution is not clean.
Let's add `params: Option[ujson.Arr]` to createSqlQueryRequest method

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,128 @@
...
+    indexPattern
+      .split(',')
+      .asSafeList
+      .filter(_.nonEmpty)
+      .traverse(RequestedIndex.fromString)
+      .flatMap(NonEmptyList.fromList)
+
+}
+
+private[esql] final case class ReplacedQuery(query: String, intendedIndexLists: List[String]) {
```
coutoPL: I'm not sure if we need this type.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
```diff
@@ -0,0 +1,128 @@
...
+    /** A join reads one index, named in full: ES resolves neither a wildcard nor a remote cluster here. */
+    def parse(span: TextSpan, indexPattern: String): Option[LookupJoinTarget] =
+      requestedIndicesIn(indexPattern).collect {
+        case NonEmptyList(RequestedIndex(index @ ClusterIndexName.Local(_: IndexName.Full), false), Nil) =>
+          LookupJoinTarget(span, index)
+      }
+
+  }
+
+  def requestedIndicesOf(indexLists: NonEmptyList[LocatedIndexList]): Set[RequestedIndex[ClusterIndexName]] =
```
coutoPL: This method and the one below can be moved to the `WithIndices` class

core/src/main/scala/tech/beshu/ror/es/esql/IndexListReplacer.scala
```diff
@@ -0,0 +1,147 @@
...
+
+import cats.data.NonEmptyList
+import tech.beshu.ror.accesscontrol.domain.{ClusterIndexName, RequestedIndex}
+import tech.beshu.ror.accesscontrol.matchers.PatternsMatcher
+import tech.beshu.ror.es.esql.LocatedIndexList.{LookupJoinTarget, SourceCommandIndices}
+import tech.beshu.ror.syntax.*
+
+private[esql] object IndexListReplacer {
+
+  def replacing(
```
coutoPL, line 27: Referring to the comment about the `ReplacedQuery`. Maybe the class should be renamed to `QueryIndicesReplacer`, and the replacing (or replace) method should look like this:

```scala
def replace(query: Query.WithIndices, allowedIndices: NonEmptyList[RequestedIndex[ClusterIndexName]]): Either[Error, Query.WithIndices]
```

WDYT?

Maybe it could even be an extension for `Query.WithIndices`.

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
```diff
@@ -0,0 +1,163 @@
...
+  ): Either[Rejection, Query]
+
+}
+
+object Query extends RequestIdAwareLogging {
+
+  def from(query: String, reader: EsqlQueryIndicesReader)(
+      implicit requestId: RequestId
+  ): Query = {
+    reader.indicesIn(query) match {
```
coutoPL, line 45: This part also smells a little. 

We read the indices and then locate them. It's weird and may be error-prone (how do we know that these params: `query: String, indices: QueryIndices` are related, because we implicitly assume that the QueryIndices are created from the String.

IMO, part of the logic of the IndicesLocator should be moved to the `EsqlQueryIndicesReader` trait. 
The trait should have a method:
```scala
final def indicesIn(query: String): Either[Throwable, List[LocatedIndices]]
```
and an abstract method:
```scala
proteced def queryIndicesFrom(query:String): QueryIndices
```

Rationale:
When we extract indices we should know their location. Maybe the implementation perspective is different because we do it in two steps, but it's just an implementation detail. From the functional point of view, the reader should read the indices with their locations.

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
```diff
@@ -0,0 +1,163 @@
...
+    )(
+        implicit requestId: RequestId
+    ): Either[Rejection, Query] = {
+      if (allowedIndices.toList.toCovariantSet == indices) Right(this)
+      else {
+        val replaced = IndexListReplacer.replacing(text, indexLists, allowedIndices)
+        reader.indicesIn(replaced.query) match {
+          case Right(indices) =>
+            replaced
+              .checkedAgainst(indices)
```
coutoPL: this also smells a little. 
We replaced indices in the query, and we do a check? Why are we not sure about the replacer?

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: `trait IndexPatternRole` deleted.

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: Yes, after checking it again, the IndexPatternRole was redundant, removed. `LocatedIndexList` instances now have the `describe` implementation.

tests-utils/src/main/scala/tech/beshu/ror/utils/elasticsearch/EsqlApiManager.scala
mgoworko: Done (`createSqlQueryRequest(query: String, params: Option[ujson.Arr] = None)`)

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: It contains the rewritten text together with the index lists that we expect that rewrite should produce. Moved to `IndexListReplacer`

core/src/main/scala/tech/beshu/ror/es/esql/domain.scala
mgoworko: The first one inlined in `WithIndices.indices` and deleted. The second one needs to stay here, because it is called by domain classes in this file.

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
mgoworko: Done as described:

```scala
trait EsqlQueryIndicesReader {

  private[esql] final def indicesIn(query: String): Either[ReadError, List[LocatedIndexList]] =
    for {
      indices <- queryIndicesFrom(query).left.map(ReadError.QueryNotParsed.apply)
      indexLists <- IndexListLocator.locatedIn(query, indices).left.map(ReadError.IndicesNotLocated.apply)
    } yield indexLists

  protected def queryIndicesFrom(query: String): Either[Throwable, QueryIndices]
}
```

core/src/main/scala/tech/beshu/ror/es/esql/Query.scala
mgoworko: It was a deliberate part of the solution from the beginning.  The rationale behind it is that we essentially operate on texts, even if we now have much more information about internal structure and positions. The replacer cannot see whether or not the modified query is still even correct. This check (AFAIK cheap, in memory, one parse) can tell what the new text reads and that it is still correct

It would help us diagnose, the error message is
```
    case Rejection.SubstitutionNotConfirmed(intended, read) =>
      s"The ES|QL query has been forbidden. ReadonlyREST rewrote it to read only [${intended.mkString(", ")}], " +
        s"the indices the user is allowed to, but Elasticsearch reads the rewritten query as reading " +
        s"[${read.mkString(", ")}] instead. Since the two disagree, ReadonlyREST cannot tell which indices the " +
        s"query would really read, so it does not run it. Please report this query to the ReadonlyREST team."
```

core/src/main/scala/tech/beshu/ror/es/esql/IndexListReplacer.scala
mgoworko: The method you describe already exists as  `WithIndices.narrowedTo(allowedIndices): Either[Rejection, Query]`. It replaces the index lists, reads the result back and returns the narrowed query. It is a member, not an extension, mainly because it uses private `reader` and `indexLists`. Other reason is that `WithoutIndices` and `Unreadable` have their own `narrowedTo`.

Inside `narrowedTo` the `IndexListReplacer` takes text and located index lists, and returns text. It does not know `Query`, so `IndexListReplacerTest` can test it with plain strings. That is why I would like to keep the name. It replaces index lists in the query text, not the indices of a `Query`.

Issue #114, opened 2016-10-20
Full-blown RegEx support for indices rule


sscarduzio:
The reasonable use case supporting this enhancement is well described by @seuf

> Hi, I have almost the same issue with my rules :
> I want to restrict access to all indices except those matching "confidential"
> for example, user "public" can acces to "foo_bar_" but not "foo_bar_confidential_"
> I've tried to use regex in the indices rules or uri_re, but it's not working:
> 
>    uri_re: "/^metrics(?:(?!_confidential_).)*$/"
> or
> 
>   indices: ["metrics(?:(?!_confidential_))*"]
> Is it possible to add a rule to forbid access to some indices ?

For backwards compatibility, I suggest we keep `indices` as is (supporting only `*`), and create a new rule called `indices_re` which does the same thing, except it uses a RegEx matcher.

Comments?

PR #1355, opened 2026-08-24
[RORDEV-2183] Let the e2e job use the Docker Hub mirror

coutoPL:
The e2e job takes its images from Docker Hub, and not from the cache the other jobs use. It was set
that way because the job pulls the ROR ES and ROR KBN images that two other runs push a moment
before, and a cache can answer with what it saw earlier.

The e2e repo decides that image by image now. It names the few images it takes from the cache, and
the ROR images are not among them, so those two stay on the direct path. The rest of the job no
longer has to.

Merge this after the e2e repo change: https://github.com/beshu-tech/readonlyrest-e2e-tests/pull/114

The change it follows: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1352

## Summary by CodeRabbit

* **Documentation**
  * Clarified Docker Hub mirror behavior across CI jobs.
  * Documented how to disable mirror usage for individual jobs.
  * Clarified which end-to-end test image pulls use or bypass the mirror.

* **Chores**
  * Enabled the default Docker mirror configuration for end-to-end test runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -648,9 +648,6 @@ jobs:
           ROR_GH_TOKEN: ${{ secrets.ROR_GH_TOKEN }}
           DOCKER_REGISTRY_USER: ${{ secrets.DOCKER_HUB_USER }}
           DOCKER_REGISTRY_PASSWORD: ${{ secrets.DOCKER_HUB_RW_TOKEN }}
-          # The only job that must not use the Docker Hub mirror. It pulls the ROR ES and ROR KBN
-          # images that two other runs pushed a moment before, and a cache can answer stale.
-          ROR_DOCKER_HUB_MIRROR: 'false'
           # Do not let corepack ask before it downloads yarn. CI has no operator.
           COREPACK_ENABLE_DOWNLOAD_PROMPT: '0'
         run: |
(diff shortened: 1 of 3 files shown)


sscarduzio (approved):
Small, coherent change: now that beshu-tech/readonlyrest-e2e-tests#114 makes the mirror selective per image (ROR ES/KBN images and the Kind node image stay on the direct Docker Hub path, everything else goes through the cache), this job no longer needs its own blanket `ROR_DOCKER_HUB_MIRROR: 'false'` opt-out. Removing it lets `e2e_tests` pick up the mirror for the pulls that are safe to cache (the digest-pinned gosu/corretto pulls, the ELK stack) while the e2e repo's own per-image list protects the freshness-sensitive ones.

Docs (`ci/CI.md`, `ci/configure-docker.sh`) updated consistently with the code change — no stale claims left about `e2e_tests` being the one job that disables the mirror.

Sequencing is correct and stated in the PR body: this merges after #114, which just merged. No unresolved threads.

`cve_check` is red here, but it is red on every open PR in this repo right now (checked #1356, #1354, #1313, #1300) — the analyzer is hitting Sonatype's paid OSS Index API with `402 Payment Required`, after already finding 0 vulnerabilities on the free pass. Unrelated to this diff, not something this PR can fix. Approving on the real checks.

PR #1356, opened 2026-08-25
[RORDEV-2187] Use one word for an audit destination: output

coutoPL:
The audit code used two words for one thing. The YAML key and the settings decoder said "output". The types, the ES services and the block-level keys said "sink". Two words for one thing make a reader look for two things. There is only one thing.

The code now uses only "output". "Sink" is gone from the sources.

### 1. Use one word for an audit destination: output

This commit renames. It does not change behaviour.

Runtime types:

| Before | After |
|---|---|
| `AuditSink`, `*AuditSink` | `AuditOutput`, `*AuditOutput` |
| `AuditSinkService` | `AuditOutputService` |
| `AuditSinkServiceCreator` | `AuditOutputServiceCreator` |
| `audit.sink` package | `audit.output` package |
| `AUDIT_SINK_MAX_*` | `AUDIT_OUTPUT_MAX_*` |

Settings types:

| Before | After |
|---|---|
| `AuditSettings.AuditSink` | `AuditSettings.AuditOutputConfig` |
| `EsIndexBasedSink` and friends | `EsIndexBasedOutput` and friends |
| `SinkName` | `AuditOutputName` |

Block-level settings keys:

| Before | After |
|---|---|
| `enabled_audit_sinks` | `enabled_audit_outputs` |
| `disabled_audit_sinks` | `disabled_audit_outputs` |

### 2. Make an empty list of audit output names unrepresentable

`EnabledAuditOutputs.Selected` and `EnabledAuditOutputs.AllExcept` held a `Set`. An empty `Set` was legal for the type, so `CoreFactory` looked for one by hand after it built the block. Both now hold a `UniqueNonEmptyList`. The decoder rejects an empty list when it builds the value, and the type keeps the names unique, as the `Set` did.

The cross-block pass keeps the check for unknown output names. That check needs the names of all the outputs, so it must stay where it can see them.

The two messages for an empty list say the same thing, but they no longer name the block, because the audit decoder does not know the block name. The message for "Cannot define both ..." already had this shape.

There were no tests for an empty list. There are now two.

## One name needed thought

`MainSettingsApi` already has an `AuditOutput` type for its response. The settings type is therefore `AuditOutputConfig`, not `AuditOutput`. The two names stay apart.

## Effect on users

None. The two block-level keys are on `develop`, but not on `master`. No release has them. No user must edit a configuration file.

The PR therefore carries the `no-changelog` label.

## Documentation

https://github.com/beshu-tech/readonlyrest-docs/pull/332


coutoPL:
@claude review please

mgoworko (approved):
LGTM

PR #1357, opened 2026-08-26
[RORDEV-2192] Move the CVE check into a composite action

coutoPL:
No changelog entry: this pull request changes CI and build logic only. A client cannot see the change.

## What it does

The `cve_check` job in `ci.yml` was a long list of inline steps. It now lives in `.github/actions/cve-check`. The job is two steps: checkout, and the action. A nightly workflow can call the same action later.

It is a composite action, not a reusable workflow. `continue-on-error` is not an allowed key on a job that calls a workflow, and the CVE check must not block a run.

## The cache was broken

One monthly key held the NVD database and the OSS Index reports together. A GitHub cache entry cannot change. Nothing was thus saved after the first run of each month. But each OSS Index report continued to expire on its own clock. Each run therefore paid Sonatype again for each dependency, and then discarded the answer.

The two caches are now separate, because their costs differ:

| Cache | Key | Who writes it |
|---|---|---|
| NVD H2 database | new each day | a push to `develop` or `master` |
| OSS Index reports | new each ISO week, unique for each run | each run that wrote reports |

The database is large, but free to download. A pull request only reads it. If the refresh on `develop` stops, `nvd.validForHours=24` lets a pull request run repair its own copy. That run is slow, but its results are correct.

The OSS Index cache costs a Sonatype credit for each miss. A unique key for each run lets each run save a newer copy, so a new dependency goes into the cache one time. The new key each week keeps the data recent.

The caller chooses who writes the NVD cache, through the `save-nvd-cache` input. A fork pull request skips both saves: its token is read-only, thus a cache write always fails.

## An OSS Index outage no longer stops the scan

`warnOnlyOnRemoteErrors = true`. Without it, the analyzer throws each remote error, and the task stops with "Analysis failed." It stops before it writes the reports, and before the CVSS gate runs. A 401, a 403, a 429, a 5xx, a timeout, or exhausted Guide credits could thus hide a true finding.

With it, the run completes with NVD data only. The option applies to this analyzer alone. Errors from every other analyzer stay fatal, and the CVSS gate still fails the build on a true finding.

`-Danalyzer.ossindex.request.delay=1` limits the retry loop that warn-only mode makes possible. On a good run it costs about 41 seconds, one time for each project.

## The scan now tells you what it used

The run writes one annotation and one step-summary line: `full`, `nvd-only`, `nvd-only-no-credentials`, `nvd-only-disabled`, or `unknown`. A run that stopped early reports `unknown`, so the label cannot claim that both sources answered.

The summary also carries the result: one row for each module with a finding, and a count for each severity.

| Module | Critical | High | Medium | Low | Other | Vulnerable deps |
|---|--:|--:|--:|--:|--:|--:|
| tests-utils | 24 | 176 | 208 | 21 | 0 | 35 |

The HTML and JSON reports go to the `dependency-check-reports` artifact for 14 days. They upload after a failed scan too, which is when you want them.

## Smaller changes

- `timeout-minutes: 30` on the job. The GitHub default is 6 hours, and a Sonatype outage could use all of it.
- `ROR_CVE_OSS_INDEX=false` runs the scan with NVD only, for local work.
- `.dependency-check-data/` is now in `.gitignore`.

## What is not proved

`actionlint` is not available here. Both YAML files parse, and `bash -n` passes on `ci/run-pipeline.sh`. The action itself needs a real run. The fork guard has no test.

## CI Improvements
* Added automated vulnerability scanning using NVD and OSS Index sources.
* Added HTML and JSON vulnerability reports with clearer scan status summaries.
* Improved scan reliability by continuing across applicable projects and clearly identifying incomplete scans.
* Added caching for vulnerability databases to speed up recurring scans.
* Increased scan timeout and improved handling of temporary OSS Index service errors.

## Configuration
* Added an option to disable OSS Index scanning.
* Excluded local vulnerability databases and cache files from version control.


coutoPL:
@claude review please

coutoPL:
@claude re-review please

coutoPL:
@claude re-review please

coutoPL:
Thanks — finding 2 is real, and it is now fixed in `529bbe297`.

One of the supporting facts does not hold, but it does not save the code. The 24 critical and 176 high findings are all in `tests-utils`, which sets `failBuildOnCVSS = 11` (`tests-utils/build.gradle:79`) and thus never fails. Every shipped module is at zero. So the bug has not fired yet. It fires on the day a real CVE lands in a shipped module, which is the day the report has to be right.

**The gap.** `failBuildOnCVSS = 3` and no `--continue`. Gradle stops scheduling after the first task failure, thus one finding ended the run and the later subprojects were never scanned. The reports then covered a prefix of the repository, and the summary named its sources as if the scan were whole.

**`--continue` on the gradle call.** Each subproject task is independent, thus all of them are scanned and all of them write a report. The build still fails. `--continue` is not `--parallel`, thus the shared H2 directory stays safe.

**A completeness signal for the killed case.** `--continue` does not help when the 30-minute timeout, or a killed JVM, ends the run. Two guards:

- `ci/run-pipeline.sh` sets the mode to `unknown` when the log holds no `BUILD SUCCESSFUL` or `BUILD FAILED` line. Gradle writes that line after `--continue` has run the last task, thus its absence means that nothing ended the run in an orderly way.
- `ci/cve-scan-report.sh` publishes a new `scan_completed`. It is true only when reports exist **and** `ci/run-pipeline.sh` reached its last act of writing the mode file.

**Both cache saves now use `scan_completed`, not `reports_written`.** A half-written NVD database must never reach the shared cache, because every later run restores it. The upload keeps `reports_written`: a part of the reports is still worth reading.

I did not add the count against an expected subproject count. That count comes from the same log that a killed run truncates, thus it cannot see a task that never started. The terminal `BUILD` line can.

Verified under the runner's flags (`bash --noprofile --norc -eo pipefail`), against the reports from a real run:

| Case | result |
| --- | --- |
| full tree, mode file present | `reports_written=true scan_completed=true mode=nvd-only` |
| full tree, **no mode file** (killed) | `reports_written=true scan_completed=false mode=unknown` |
| clean modules, mode file present | `reports_written=true scan_completed=true mode=full` |
| no reports at all | `reports_written=false scan_completed=false mode=unknown` |

Row two is the new behaviour: the reports still upload, no cache is written, and the summary reports unknown sources in place of a complete scan.

coutoPL:
@claude re-review please

coutoPL:
@claude re-review please

mateuszkp96 (approved):
LGTM

PR #1358, opened 2026-08-26
[RORDEV-2183] Put the mirror in front of the container: image pull

coutoPL:
## The problem

The runner pulls the job `container:` image before step 1 starts. `ci/configure-docker.sh` cannot
reach that pull, so it sets neither the mirror nor the login for it.

Ten jobs and their matrices share that image, which makes it the most pulled image of a run. On
2026-08-26 Docker Hub answered a whole run with `429 toomanyrequests`. Sixteen jobs died before
`actions/checkout`.

## The change

`container: image:` can read a job output. The `setup` job now runs `ci/resolve-toolchains-image.sh`
once and publishes the name to pull.

A job pulls by digest, and a digest names the bytes. A cache cannot answer a digest with the wrong
image. So the script asks one question: can `mirror.gcr.io` serve this digest? **Build toolchains
image** writes the digest of its push to the Actions cache, and `setup` sends one HEAD request for
that digest. No request goes to Docker Hub.

| What the script finds | What the run pulls |
|---|---|
| no digest on record | Docker Hub |
| the mirror cannot serve the digest | Docker Hub |
| the mirror serves the digest | the mirror |

Docker Hub is the safe answer in the first two rows, so a wrong choice costs speed only.

The mirror fetches a digest it has never held. So a run keeps the mirror in the hours after a
rebuild, before the mirror knows the new tag. A question about the tag would lose the mirror in that
window, where the recorded digest is newest.

`setup` writes the choice, and the reason for it, to the step summary. The run page then shows which
registry a run used, without opening a log.

## What a fork gains

A fork run has no secrets. It pulled this image anonymously from Docker Hub, where the limit counts
against a shared runner address, so a fork run met the 429 first. The same run now pulls anonymously
from `mirror.gcr.io`, which sets no limit. A fork run reads the base branch, so it finds the digest
a `develop` rebuild saved.

## The rebuild workflow

`build-toolchains-image.yml` holds the weekly rebuild. It takes up to four hours, and no test run
waits for it.

- It keeps a concurrency group of its own. Two rebuilds would push one tag and file two cache
  entries, and `setup` takes the newest, which need not hold the manifest Docker Hub keeps.
- The digest comes from the `docker push` output. An empty digest is an error and the job fails. A
  silent miss would leave the mirror off for every run until the next rebuild.
- A dispatch off `develop` raises a warning. A cache entry belongs to the ref that saved it, so no
  `develop` run and no pull request run can read that one. The dispatch stays allowed: a branch that
  bumps a Gradle dependency needs its own tag and its own build.

## Smaller things

The drift guard in `build_toolchains_image` is gone. `ci/toolchains/image.env` holds the tag now.
`azure-pipelines.yml` keeps a copy, because Azure resolves `container:` at compile time and cannot
read the file there. Its build step sources the file, so that fallback pipeline cannot build a stale
tag.

The cache key and the restore prefix put two hyphens after the tag. One hyphen would let the key of
`…-9.2.1-arm64` answer the prefix of `…-9.2.1`.

Five step ids said nothing. `f`, `m`, `dkey` and `img` are now `flags`, `matrices`, `digest_key` and
`toolchains`. `r` is `release_check`.

## After the merge

Run **Build toolchains image** once, on `develop`. Until a rebuild writes the first digest, every run
pulls from Docker Hub, as it does today.

## Two limits

The choice holds for the whole run. If the mirror stops answering during a run, the jobs that already
took the mirrored name fail. The runner's pull has no fallback of its own.

`ghcr.io` is the other direction. It would remove the digest bookkeeping, and a public package needs
no `credentials:` block at all. It would not retire the mirror, which `buildx`, the test suite and
the toolchains build still need for their own Docker Hub pulls. This change does not settle that
question.

## Tested

`ci/resolve-toolchains-image.sh` ran against the live registries on every path: a digest the mirror
can serve, a digest it cannot, no digest file, the mirror switched off, a name with no tag, and an
empty `TOOLCHAINS_IMAGE`. Each gave the right answer.

`mirror.gcr.io` served a digest it had never held. `beshultd/elasticsearch-readonlyrest:6.0.1-ror-1.54.0`
was asked by digest before any request for its tag, and answered 200. A fabricated digest answered
404.

Both workflows parse. The restore prefix matches the tag's own entries, and no longer matches the key
of a tag that extends it.

## Review

Nine commits answer @sscarduzio's review, one commit per point.

## Summary by CodeRabbit

* **CI Improvements**
  * CI now dynamically verifies and selects the toolchains image source.
  * Mirrored images use digest-pinned references, with Docker Hub authentication only when needed.
  * Improved fallback handling when mirrored images are unavailable, outdated, or unverifiable.
  * Toolchain image digests are reliably persisted across CI runs.

* **Documentation**
  * Updated CI and Docker guidance covering image resolution, authentication, mirroring, and pull behavior.

### DIFF
--- a/.github/workflows/build-toolchains-image.yml
+++ b/.github/workflows/build-toolchains-image.yml
@@ -14,6 +14,13 @@ on:
     - cron: '0 3 * * 1'   # weekly rebuild
   workflow_dispatch:
 
+# One rebuild at a time. Two rebuilds push the same tag and file two cache entries, and setup takes
+# the newest entry, which need not hold the manifest that Docker Hub keeps. A second run waits: a
+# build takes up to four hours, and cancelling one wastes all of it.
+concurrency:
+  group: build-toolchains-image
+  cancel-in-progress: false
+
 permissions:
   contents: read
 
--- a/.github/workflows/build-toolchains-image.yml
+++ b/.github/workflows/build-toolchains-image.yml
@@ -77,7 +77,9 @@ jobs:
           [ -n "$DIGEST" ] || { echo "::error::The push of $TOOLCHAINS_IMAGE printed no digest."; exit 1; }
           echo ">>> Pushed digest: $DIGEST"
           echo "digest=$DIGEST" >> "$GITHUB_OUTPUT"
-          echo "key=toolchains-digest-${TOOLCHAINS_IMAGE##*:}-${GITHUB_RUN_ID}" >> "$GITHUB_OUTPUT"
+          # Two hyphens after the tag. ci.yml restores by prefix, and one hyphen would let a tag
+          # that extends another, such as ...-9.2.1-arm64, answer the prefix of ...-9.2.1.
+          echo "key=toolchains-digest-${TOOLCHAINS_IMAGE##*:}--${GITHUB_RUN_ID}" >> "$GITHUB_OUTPUT"
 
   # ── RECORD_DIGEST ─────────────────────────────────────────────────────────
   # An Ubicloud runner keeps its own Actions cache. A save there is invisible to a GitHub-hosted
--- a/.github/workflows/build-toolchains-image.yml
+++ b/.github/workflows/build-toolchains-image.yml
@@ -1,0 +1,94 @@
+name: Build toolchains image
+
+# Builds the CI toolchains image and pushes it to Docker Hub. Every container: job of ci.yml pulls
+# that image, and it bakes the JDKs, Gradle and this commit's Gradle dependencies.
+#
+# It is not part of CI. A rebuild takes up to four hours, and no test run waits for it. Run it
+# weekly, and by hand after a JDK or Gradle bump.
+#
+# ci/toolchains/image.env holds the tag. ci.yml sources the same file.
+# See "The container: image" in ci/CI.md.
+
+on:
+  schedule:
+    - cron: '0 3 * * 1'   # weekly rebuild
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+
+  # ── BUILD ─────────────────────────────────────────────────────────────────
+  build:
+    # A GitHub schedule fires on the default branch. The guard holds if that branch ever changes.
+    if: github.event_name != 'schedule' || github.ref == 'refs/heads/develop'
+    runs-on: ubicloud-standard-4
+    timeout-minutes: 240
+    # record_digest reads these. A digest is public, so an output carries it safely.
+    outputs:
+      digest: ${{ steps.push.outputs.digest }}
+      key:    ${{ steps.push.outputs.key }}
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
+        with: { fetch-depth: 1, persist-credentials: false }
+      - name: Read the toolchains image tag
+        run: |
+          . ci/toolchains/image.env
+          [ -n "${TOOLCHAINS_IMAGE:-}" ] || { echo "::error::ci/toolchains/image.env sets no tag."; exit 1; }
+          echo "TOOLCHAINS_IMAGE=$TOOLCHAINS_IMAGE" >> "$GITHUB_ENV"
+      - name: docker build & push
+        id: push
+        env:
+          DOCKER_REGISTRY_USER: ${{ secrets.DOCKER_HUB_USER }}
+          DOCKER_REGISTRY_PASSWORD: ${{ secrets.DOCKER_HUB_RW_TOKEN }}
+        run: |
+          set -e
+          # Source before the build. It logs in and it sets ROR_DOCKER_HUB_MIRROR_PREFIX, which the
+          # FROM lines need. The Dockerfile pulls five images from Docker Hub in its FROM lines (four
+          # eclipse-temurin JDKs and the docker CLI), and those pulls are anonymous if the login
+          # comes after them.
+          source ci/configure-docker.sh
+          echo ">>> Building $TOOLCHAINS_IMAGE"
+          # MIRROR puts the pull-through cache in front of those five names. It is empty when the
+          # mirror is off, and the pulls then go to Docker Hub.
+          docker build --build-arg MIRROR="${ROR_DOCKER_HUB_MIRROR_PREFIX:-}" \
+            -f ci/toolchains/JdkToolchains.Dockerfile -t "$TOOLCHAINS_IMAGE" .
+          docker push "$TOOLCHAINS_IMAGE"
+          # The digest of this push, from the local daemon. A run takes the mirror only when the
+          # mirror serves the same digest. See ci/resolve-toolchains-image.sh.
+          DIGEST=$(docker inspect --format '{{range .RepoDigests}}{{println .}}{{end}}' "$TOOLCHAINS_IMAGE" \
+                     | grep "^${TOOLCHAINS_IMAGE%:*}@" | head -1 | sed 's/.*@//')
+          echo ">>> Pushed digest: ${DIGEST:-<none>}"
+          if [ -n "$DIGEST" ]; then
+            echo "digest=$DIGEST" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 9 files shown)


coutoPL:
@claude review please

coutoPL:
@claude re-review please

sscarduzio (approved):
Approve. The mechanism is sound, every path degrades to today's behaviour, and I found no way for it to make a run worse than `develop`. Notes below, ranked. The first one is six lines and I would take it before the merge.

I reproduced the whole resolver against the live registries, and I checked the pieces the PR body could not: the cold-digest behaviour of `mirror.gcr.io`, the `needs: setup` edge on all ten `container:` jobs, and the action pin.

---

## 1. The comparison asks about the tag, and the run pulls a digest

`main()` compares `docker-content-digest` behind the **tag** with the recorded digest, then pins the **digest**. Those are two different questions, and only the second one decides whether the pull works.

A digest reference is immutable. A pull-through cache cannot serve wrong bytes for one. So "is the mirror's tag stale?" stopped mattering the moment commit `4ed0ddc` pinned the name. The question that still matters is "can the mirror serve *this* digest?", and the script never asks it.

The cost is real. The mirror holds a tag→digest answer of its own. A rebuild moves the tag on Docker Hub, `record_digest` files the new digest, and until the mirror's own answer catches up the two disagree. The run then takes Docker Hub and raises a `::warning::` — in the window right after a rebuild, where the recorded digest is freshest and where `mirror.gcr.io/…@<new digest>` would have served correctly.

It would have served correctly. `mirror.gcr.io` fetches a digest it has never held:

```
# sha256:53db202f… = beshultd/elasticsearch-readonlyrest:6.0.1-ror-1.54.0, a 2019 tag.
# Asked by digest first, before any request for the tag.
GET  mirror.gcr.io/v2/beshultd/elasticsearch-readonlyrest/manifests/sha256:53db202f…  -> 200
```

Ask about the digest instead. One request, same cost, and it tests the name the jobs receive:

```bash
mirror_has_digest() {          # $1 repo:tag, $2 digest
  curl -fsS --max-time 15 -o /dev/null -I -H "Accept: ${accept}" \
       "https://${MIRROR_HOST}/v2/${1%:*}/manifests/$2" >/dev/null 2>&1
}

mirror_has_digest "$image" "$pushed" || use_docker_hub "$image" \
  "${MIRROR_HOST} cannot serve ${pushed}."
use_mirror "$image" "$pushed"
```

I checked both directions against the live mirror: the real digest gives 200, a fabricated one gives 404. `mirror_digest`, the `served != pushed` branch and its warning all go away, and so does the post-rebuild window.

## 2. Three ways the mirror switches itself off in silence

Each of these leaves CI green and the 429 waiting for the next busy week:

- `docker inspect .RepoDigests` returns nothing. The rebuild prints `::warning::` and finishes green. **This path has never run** — the Tested section covers `resolve-toolchains-image.sh` alone, and the digest capture is new code on a job that runs weekly, unwatched.
- `actions/cache/save` fails. A warning, and the workflow is green.
- A rebuild dispatched from a feature branch writes an entry that only that branch reads.

Two cheap fixes:

- Make the rebuild loud. The push has already succeeded when the digest is read, so `::error::` and `exit 1` on an empty digest cost nothing and put a red X where a person will see it. Consider taking the digest from `docker push`, which prints `digest: sha256:…`, rather than from `docker inspect`, which can come back empty.
- Make the choice visible on every run. `setup` already prints `[CI] This run skips the mirror: …`. Send that line to `$GITHUB_STEP_SUMMARY` as well, and the state of the feature is on the run page instead of inside a step log.

## 3. `build-toolchains-image.yml` has no `concurrency:` group

Splitting the workflow dropped the group it used to inherit from `ci.yml`, and the commit message names concurrency as a reason for the split. Two dispatches, or a dispatch that meets the Monday cron, now build and push the same tag for up to four hours each and both file a cache entry. `setup` restores the newest by prefix, which need not be the manifest Docker Hub ends up holding.

```yaml
concurrency:
  group: build-toolchains-image
  cancel-in-progress: false
```

## 4. A dispatch from a feature branch is a foot-gun the design now leans on

The `if:` guard constrains `schedule` only. A dispatch from a feature branch pushes to the shared Docker Hub tag and writes a cache entry that no `develop` run and no PR run can read. `CI.md` says to dispatch on `develop`; nothing holds anyone to it. Either extend the guard, or have `record_digest` say `::warning::` when `github.ref != 'refs/heads/develop'`.

## 5. The "always exits 0" contract turns one case into ten broken jobs

`resolve-toolchains-image.sh` has one path that exits 0 and writes no `image=`: an empty `TOOLCHAINS_IMAGE`. I ran it — status 0, no output. `setup` then publishes an empty `toolchains_image`, and ten jobs get `container: image: ''`.

The step above it makes that unreachable in `ci.yml` today. The header does not say so, it says the script always exits 0, so the next caller inherits the trap. There is nothing safe to emit in that case, so let it `exit 1`. The stated principle — a mirror must never stop a run — is the wrong one here, because silence is what stops the run.

## 6. The second copy of the tag is still there

`78b69c6` removed the drift guard because "those copies no longer exist". `azure-pipelines.yml:7` still carries `beshultd/ror-ci-toolchains:jdk-8-11-17-21-gradle-9.2.1`, its first line still says "Single source of truth … bump the tag here", and line 153 greps that file for the tag. `trigger: none` and `pr: none` keep it out of a green run, so this is a comment that now lies, and a fallback pipeline that would build a stale tag if anyone reached for it. Point that comment at `ci/toolchains/image.env`, or delete the pipeline. The line near 40 about the weekly rebuild living in "the GitHub Actions workflow's schedule" is stale for the same reason.

## 7. Smaller things

- `grep "^${TOOLCHAINS_IMAGE%:*}@"` reads the repository name as a regular expression. Harmless for this name, over-matching for a name with dots — which is what the `ghcr.io` direction in the doc would bring.
- `.toolchains-digest` lands untracked in the `setup` workspace and is not in `.gitignore`. Nothing reads the tree in `setup`, so this only bites a developer who runs the resolver by hand.
- The cache prefix `toolchains-digest-<tag>-` matches a future tag that extends this one after a hyphen. The trailing hyphen already rules out `9.2.1` against `9.2.10`. A `v1` / `v1-beta` pair would collide, and the result is a mismatch and a Docker Hub pull. Not worth code.

## 8. The question behind the change

The PR names `ghcr.io` and defers it. What the interim buys: a workflow, a job, a cross-workflow cache side channel with branch-scoping rules, a 110-line resolver, ten jobs newly coupled to a `setup` output, and a documented failure with no fallback. What `ghcr.io` costs: a tag in `image.env`, `docker login ghcr.io` with `GITHUB_TOKEN`, `packages: write`, and a public package. No rate limit, no mirror, no digest bookkeeping, and a public package needs no `credentials:` block at all.

Not a blocker, and the scoping is yours to make. Worth one line in the PR or the ticket saying what holds `ghcr.io` back, so the next person does not re-open this.

## 9. The best part, and it is not in the body

A fork PR now resolves `mirrored=true`, takes empty credentials, and pulls anonymously from `mirror.gcr.io` instead of anonymously from Docker Hub. Fork runs are the ones with no token and the most exposure to the 429. Say so.

---

### Verified

- Six resolver paths against the live registries: match, stale, empty file, missing file, mirror off, tag the mirror lacks. Every one gives the documented answer.
- `mirror.gcr.io` serves this tag anonymously and returns `docker-content-digest`, and that digest equals the one Docker Hub returns today (`sha256:0af2fd0d…`).
- `mirror.gcr.io` serves a cold digest it has never held.
- All ten `container:` jobs carry `setup` in `needs`, parsed from the file rather than read.
- Both workflows parse. `actions/cache@0057852` is v4.3.0, the pin the other three workflows use.
- No `TOOLCHAINS_IMAGE` reference is left dangling in `ci.yml` after the workflow `env:` went away.
- The twelve `github.event_name != 'schedule'` guards are dead once the trigger is gone. Removing them is correct.

coutoPL:
All action points from [this comment](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1358#pullrequestreview-5038879745) are addressed. @claude check it please

PR #1359, opened 2026-08-27
[RORDEV-2190] Rework the audit config model and the output creation

coutoPL:
## Why

The audit code had three problems.

`AuditOutputConfig` had a `Disabled` case. It came from `enabled: false` in the settings file, not from the domain. Every match on an output carried a dead branch.

The names disagreed. The outer type said `Config`, the inner type said `Settings`.

Nothing in the types said which Elasticsearch versions accept a data stream output. The decoder checked the version, then the creation code trusted the result.

## What changed

**One vocabulary.** `Config` describes an audit output. `Settings` is the raw settings document. A name with no suffix is the running thing, for example `AuditOutput`. `AuditingSettingsDecoder` is now `AuditingConfigDecoder`.

**A flat output model.** An output config is `EsIndexBased`, `EsDataStreamBased`, `LogBased` or `RollingFileBased`. Each one holds its own `Config` and has a name. The `Disabled` case is gone.

**Honest output states.** `AuditOutputs` is `Disabled`, `Defaults` or `Configured`. `Configured` holds a `NonEmptyList`, so it cannot be empty. When the settings turn off all the outputs, the result is `Disabled`.

**The ES layer states what it can write.** The new `EsAuditCapabilities` has two cases: `IndexOnly` and `IndexOrDataStream`. Each ES module supplies one. The modules older than ES 7.9 supply `IndexOnly`.

**The capability and the config travel together.** `AuditingTool.AuditSetup` pairs them. `SupportedByAllEsVersions` holds a config whose outputs are index, log or file only. `AnyOutput` holds a config that can also have a data stream. A data stream output on an old ES node is not representable.

**One entry point.** `AuditingTool.create(setup)` replaces the two public create methods. The dispatch on the capability moved from the boot code into `AuditingTool`.

**A seam for the remote audit client.** `MultiNodeRestClient` hides the ES REST client behind `perform`. `RoundRobinClient` is the only implementation today. Each ES module supplies a `RestClientRequestExecutor`. The submit of an audit event is a `Task` now, and it runs on the main scheduler.

**A scoped data stream client.** `dataStreamCreator` returns `Resource[Task, AuditDataStreamCreator]`. The remote client that creates the data stream opens and closes with the resource.

## Behaviour

The settings format does not change. The audit records do not change. The response of the settings API does not change.

One message changed. The error for a `data_stream` output on an old ES node no longer prints the node version.

## Review notes

The 34 ES modules get the same change: build an `EsAuditCapabilities` and give it to `CoreFactory`. The es67x module shows the `IndexOnly` shape, es818x shows the `IndexOrDataStream` shape.

`CoreCreationResult` carries the config two times: `core.auditingConfig` and `auditSetup.config` hold the same value. This is not new, and I did not fix it here. The clean fix moves the live output services into `Core`, and the settings API reads them from there.


coutoPL:
@claude review please

mateuszkp96 (approved):
LGTM

coutoPL:
@claude please verify the latest changes

PR #1360, opened 2026-08-27
[RORDEV-2183] Pull the container: image without credentials

coutoPL:
Every CI run since 10:24 today fails. A job with a `container:` block stops before its first step:

```
The template is not valid. .github/workflows/ci.yml
(Line: 178, Col: 19): Unexpected value '', (Line: 179, Col: 19): Unexpected value ''
```

## Cause

The `credentials:` block held an expression that gives an empty string when the run pulls from the
mirror. GitHub rejects an empty `username:` and stops the run in template validation.

That path had never run. Every earlier run missed the digest cache and took Docker Hub, where the
expression gives the real secrets. A rebuild recorded the digest on `develop` at 10:24, the mirror
path opened, and the template broke.

## Change

The `credentials:` block is gone, and so is the `toolchains_mirrored` output that fed it.

No other shape works. A Docker Hub login against `mirror.gcr.io` fails, so the block must be absent
when the name is mirrored. An empty value does not remove it. YAML cannot drop a key on a condition,
and `credentials:` is a mapping, so no expression reaches it.

The notes also state how Docker Hub counts a pull. The earlier text had it backwards.

## Cost

The pull is anonymous on both paths. `mirror.gcr.io` sets no limit, and it serves the normal path.
Docker Hub counts the fallback pull against the runner's address, not against our account. That path
opens only when a tag has no recorded digest, and the next rebuild closes it.

A fork gains. It never had secrets for this pull, and it now reads the mirror.

## Checks

- `ci.yml` parses. 16 jobs. All ten `container:` blocks carry `image:` only.
- `docker login mirror.gcr.io` refuses `dummyuser`, `_token`, `oauth2accesstoken` and `anonymous`.
- An anonymous manifest request for `beshultd/ror-ci-toolchains` returns 200, so the fallback pull works.
- `ci/resolve-toolchains-image.sh` on all three paths: mirror hit, no digest, empty tag.

## Summary by CodeRabbit

* **CI Improvements**
  * Updated toolchain image retrieval to use anonymous pulls without Docker Hub login credentials.
  * Simplified CI job outputs to report only the resolved toolchain image.
  * Improved support for mirrored images and fork-based workflows by avoiding unnecessary registry authentication.

* **Documentation**
  * Clarified image-pull behavior, token permissions, rate limits, and mirrored registry usage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,8 +74,7 @@ jobs:
       it_linux_matrix:   ${{ steps.matrices.outputs.linux }}
       it_windows_matrix: ${{ steps.matrices.outputs.windows }}
       e2e_modules:       ${{ steps.matrices.outputs.e2e }}
-      toolchains_image:    ${{ steps.toolchains.outputs.image }}
-      toolchains_mirrored: ${{ steps.toolchains.outputs.mirrored }}
+      toolchains_image: ${{ steps.toolchains.outputs.image }}
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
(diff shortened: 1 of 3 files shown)


coutoPL:
@claude review please

Issue #121, opened 2016-10-26
groups: empty user list should not disable groups rule entirely


maitai:
Consider the following scenario:

```
readonlyrest:
  enable: true
  access_control_rules:

  - name: myblock
    type: allow
    indices: ["myindex"]
    groups: ["mygroup"]

  users:
  - username: alice
    auth_key: alice:p455wd
    groups: ["mygroup"] 
```
Only Alice has access to `myindex`. All fine there.

Now let's _empty out_ her user groups to make the index completely private for a moment:
```
  users:
  - username: alice
    auth_key: alice:p455wd
    groups: []              # <-- removed "mygroup" from the list to revoke access
```
**Expected outcome**:
Alice loses access to `myindex`.
**Actual outcome**:
The index can be accessed by Alice _**and the rest of the world**_.

**Reason**: Alice is not added to the internal user list because of her `groups` list beeing empty. With no other users configured, this leads to a completely empty user list as well. No harm until here, but this on the other hand leads to a _RuleNotConfiguredException_ for the `groups` rule ([source](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/dedd1c45c4d593f1e903b11550bf13093a6d0061/src/main/java/org/elasticsearch/plugin/readonlyrest/acl/blocks/rules/impl/GroupsRule.java#L40-L42)) inside the `myblock` ACL block. So as soon as Alice's  groups are empty (or more generally: if all groups of all users are empty), the groups rule disappears from the block leaving the corresponding index unprotected.

**Proposition**: The _RuleNotConfiguredException_ should not be thrown on an empty user list.

sscarduzio:
Got it, makes sense. Thanks @maitai as usual!

PR #120, opened 2016-10-25
Correct plugin version for 2.4.1

atlasian:
For ES_VERSION=2.4.1, v1.12.0-rc1 should be used.

PR #1368, opened 2026-09-04
ci: run the ES matrix, e2e and builds on GitHub-hosted runners

sscarduzio:
## What this PR does

- Moves `it_linux`, `e2e_tests` and `build_ror` from `ubicloud-standard-4` to `ubuntu-latest`. The repo is public, so these runners cost nothing.
- Removes five ES series from the test matrix: 7.3, 7.4, 7.9, 7.11 and 7.14. The customer portal shows no downloads of these builds in two years. The release still builds and publishes them.
- Sets `max-parallel: 29`. The full matrix now starts in one wave.
- Makes `ci/free-host-disk.sh` fail closed. It reclaims disk only on a GitHub-hosted runner, and only when free space is below 40 GB.

## Why

The ES matrix was the largest line of the Ubicloud bill: about $260 per month. A leg is slower here (25-35 min against 21-27 min on Ubicloud). That is the trade.

## Verification

- Full 34-leg matrix on `ubuntu-latest`: run [33912487848](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runs/33912487848), all green.
- Reduced 29-leg matrix: run [33921602759](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runs/33921602759), all green, one wave.


.github/workflows/ci.yml
```diff
@@ -276,7 +280,10 @@ jobs:
       needs.toolchains_verify.result == 'success' &&
       needs.setup.outputs.it_linux_matrix != '[]' &&
       (github.event_name != 'workflow_dispatch' || needs.setup.outputs.action == 'run_all_tests_on_linux')
-    runs-on: ubicloud-standard-4
+    # free for public repos (4cpu/16GB) — the same shape as ubicloud-standard-4, so the shard
+    # count and the heavy-suite gate below carry over unchanged. This is the single biggest line
+    # of the Ubicloud bill (~79k min/month), and it costs nothing here.
+    runs-on: ubuntu-latest
```
sscarduzio: **The 34-leg matrix this PR is really about has never run on a GitHub-hosted runner.**

The green run everyone is looking at is the **10-leg PR subset**:

```
run 33891462996 (this branch)     it_linux legs on ubuntu-latest:  10
run 33885860018 (develop, today)  it_linux legs on ubicloud:       34
                                  total non-skipped jobs:          57
```

`setup` picks `LINUX_IT_PR_READY_SET` (10 modules) for a pull request and `LINUX_IT_FULL_SET` (34) for a push to develop/master. So the path that actually costs the money, and the one that carries the memory-tuned heavy modules (es812x-es815x, es87x-es89x, es91x, es92x, es73x, es74x, es77x-es79x, es711x, es714x, es716x), is unmeasured on the new host.

Two things change at 34 legs and 57 jobs:

- **Concurrency.** The GitHub Free plan allows 20 concurrent jobs **per account**, not per repo, and `sscarduzio/readonlyrest_kbn` draws on the same pool. 57 jobs will serialise into roughly three waves.
- **Memory.** The heavy modules are the ones the `IT_PARALLELISM` comment says host-OOM'd at exit 137 on a 16 GB Ubicloud host. None of them are in the PR subset.

**Action:** before merging, run `workflow_dispatch` with `run_all_tests_on_linux` on this branch and record the wave count, the total wall time, and the worst `avail_mb` per leg. If the full matrix passes, this finding is closed and the PR is proven. If it OOMs or times out, better to know now than on the first push to develop.

.github/workflows/ci.yml
```diff
@@ -276,7 +280,10 @@ jobs:
...
     container: *toolchains_container
     timeout-minutes: 120
     # "it_linux_es94x", not the truncated "it_linux (es94x)" composite, in the graph view. Same
```
sscarduzio, line 282: _Anchored at line 289; the code this is about is at `.github/workflows/ci.yml:294`, which is outside the diff._

**`max-parallel: 20` against a 20-slot pool leaves nothing for the other jobs in the same run.**

The cap was chosen for a 20-job account limit, and `ci/CI.md:7-8` now states that as the reason. But the limit is per **account**, and this workflow needs slots for far more than `it_linux`: 6 `it_win` legs on a develop push, 3 `e2e_tests`, the light checks, and the ROR KBN pre-build that `e2e_prepare` dispatches into `sscarduzio/readonlyrest_kbn` — which also runs on `ubuntu-latest`, on this same account.

So `it_linux` alone can fill the pool and push everything else into a later wave. That matters most for `e2e_tests`, which holds its slot while polling the kbn pre-build it just dispatched. If the pre-build cannot get a runner because `it_linux` owns all 20, the e2e legs wait on a run that has not started.

**Action:** decide the cap deliberately — something near 12-14 leaves room for `it_win`, `e2e` and the kbn pre-build in the first wave — and then correct the sentence in `ci/CI.md`, which currently gives the cap a reason that is the opposite of the constraint.

ci/free-host-disk.sh
```diff
@@ -1,34 +1,76 @@
...
   echo ">>> [host] after reclaim:"
   df -h /
-elif [ "${AGENT_ISSELFHOSTED:-0}" != "1" ]; then
+  exit 0
+fi
+
+# RUNNER_ENVIRONMENT is set by the runner itself: "github-hosted" on GitHub's runners,
+# "self-hosted" on a registered runner (which is how Ubicloud runners appear). AGENT_ISSELFHOSTED
+# stays supported as an explicit opt-out for jobs that set it at job level.
+if [ "${AGENT_ISSELFHOSTED:-0}" = "1" ] || [ "${RUNNER_ENVIRONMENT:-github-hosted}" = "self-hosted" ]; then
```
sscarduzio: **The runner guard fails open: with no environment set, the script takes the destructive path.**

`${RUNNER_ENVIRONMENT:-github-hosted}` treats *absent* as *GitHub-hosted*. `RUNNER_ENVIRONMENT` is set only by the Actions runner, so anywhere else — a developer shell, `act`, a future job on the Ryzen box that copies the `e2e_tests` step without setting `AGENT_ISSELFHOSTED` or `ROR_SHARED_DOCKER_HOST` — the script runs `sudo rm -rf /usr/share/dotnet ...` and `docker system prune -af` against whatever daemon it can reach.

Demonstrated with the levers stubbed out:

```
$ env -i PATH=/usr/bin:/bin bash ci/free-host-disk.sh
>>> [host] no preinstalled toolchain dirs visible from here (container: job) - skipping
WOULD-PRUNE-ALL
WOULD-PRUNE-BUILDER
```

One of the review agents hit this for real while checking the script and removed about 8 GB from its own machine.

The `docker system prune -af` lever is new in this PR, and the Ryzen box is exactly the wrong place for it: `ci/self-hosted-runner.md` records that the same daemon holds the kbn runners' in-flight ELK stacks.

**Action:** invert the default so the destructive path is opt-in, not opt-out:

```bash
[ "${RUNNER_ENVIRONMENT:-}" = "github-hosted" ] || { echo ">>> not a GitHub-hosted runner - skipping"; exit 0; }
```

ci/free-host-disk.sh
```diff
@@ -1,34 +1,76 @@
...
+  echo ">>> self-hosted runner - skipping disk reclaim"
+  exit 0
+fi
+
+echo ">>> [host] disk before reclaim:"
+df -h / || true
+
+# --- lever 1: preinstalled toolchains (bare-runner jobs) ---------------------------------------
```
sscarduzio, line 69: **The reclaim frees space that was never scarce, and costs up to 3.6 minutes a leg.**

From the green run's own logs:

```
e2e_es94x       before: /dev/root 145G  59G  87G avail   after: 24G used, 121G avail   (3m37s)
it_linux_es94x  before: overlay   145G  62G  83G avail   after: 60G used,  85G avail   (18s)
```

ES then reports `net usable_space [81.3gb]` and the leg finishes with 77 GB spare.

The comment on line 576 says the step "reclaims the ~25 GB the ELK docker stack needs". 87 GB were already free before it ran. The premise is a stale memory of the old 14 GB hosted image; the current one is 145 GB.

Before this PR the script was a no-op on these jobs, so this wall time is entirely new: about 1-3.6 min on each e2e leg, and 8-26 s on each of 34 `it_linux` legs on a develop push.

**Action:** gate both levers on a threshold so they run only when the disk is actually tight, for example skip when `df --output=avail -BG /` exceeds 40 GB, and correct the line 576 comment.

.github/workflows/ci.yml
```diff
@@ -645,7 +655,9 @@ jobs:
       needs.it_linux.result == 'success' &&
       (needs.it_windows.result == 'success' || needs.it_windows.result == 'skipped') &&
       (needs.e2e_tests.result == 'success' || needs.setup.outputs.e2e_modules == '[]')
-    runs-on: ubicloud-standard-4
+    # free for public repos (4cpu/16GB) — same shape as ubicloud-standard-4. Gradle deps are baked
```
sscarduzio: **"Same shape as ubicloud-standard-4" is true of the spec sheet and false of the throughput.**

Same 10 legs, this run against the Ubicloud-era run, and the `build_ror` matrix against its 30-day Ubicloud average:

```
it_linux legs   25-35 min   vs  21-27 min
build_es6xx      4.1 min    vs   2.9 min  (+41%)
build_es9xx     10.6 min    vs   6.3 min  (+68%)
build_es7xx     19.4 min    vs  12.3 min  (+58%)
build_es8xx     26.6 min    vs  15.5 min  (+72%)   n=107 for the baseline
```

Inside `build_es8xx` each per-version Gradle invocation reports `BUILD SUCCESSFUL in 37s` where Ubicloud reported 16s.

That is a fine trade for $227 a month, and I am not arguing against the move. But the comment on line 659 — "Gradle deps are baked into the toolchains image, so this job needs no runner-side cache to be fast" — is contradicted by the job taking 72% longer, and the next person tuning `IT_PARALLELISM` or the heavy-suite permits from these comments will assume Ubicloud numbers.

**Action:** replace the three "same shape" comments (lines 283, 574, 658) with the measured figures, or point them at the full-matrix dispatch run from the first finding.

ci/free-host-disk.sh
```diff
@@ -1,34 +1,76 @@
...
+#                        `-a`, never a container we did not start. Selected by ROR_SHARED_DOCKER_HOST=1.
+#   other self-hosted    reclaim nothing. We do not know what else lives on the box.
+#   ephemeral hosted VM  reclaim both levers below. The VM dies after the job anyway.
+#
+# On an ephemeral runner two shapes of job call this, and they need different levers:
+#
+#   * bare runner (e2e_tests, publish-pre-builds): the preinstalled toolchains are visible, so
+#     deleting them is what reclaims space. GitHub's ubuntu image ships ~25 GB of dotnet, android,
+#     ghc, swift and hostedtoolcache that no ROR job uses.
+#   * container: job (it_linux, build_ror): the steps run inside the toolchains image, where those
```
sscarduzio: **The caller lists in these comments are wrong in both directions.**

Line 16 names `build_ror` as a `container:` caller. It is not one:

```
$ grep -n free-host-disk .github/workflows/*.yml
ci.yml:301                  (it_linux)
ci.yml:591                  (e2e_tests)
publish-pre-builds.yml:78
mirror-es-libs.yml:43
```

`build_ror` has no `Free host disk` step at all.

Line 13 names `publish-pre-builds` as a *bare runner* caller. Since #1369 it runs on `[self-hosted, Linux, X64]` with `ROR_SHARED_DOCKER_HOST: '1'`, so it takes the shared path and never touches the toolchain lever.

Same class of error in two neighbours, both introduced by the rebase over #1369:

- `.github/workflows/ci.yml:60` says "The two workflows that stay on Ubicloud (mirror-es-libs.yml, publish-pre-builds.yml)". Both are on `[self-hosted, Linux, X64]` on this branch.
- `ci/CI.md:11` says those workflows "stay on `ubicloud-standard-4` ... and are moving to a self-hosted runner". They already moved.

This is the file whose entire value is the comment explaining which lever applies where, so a wrong caller list is worse here than elsewhere.

**Action:** correct all four sentences to name the real callers and the real runners.

sscarduzio (commented):
## Verdict: the change is right, the evidence for it is not yet there. One thing to run before merge, three fixes, two decisions.

**Disclosure: I wrote this PR**, so check the evidence rather than the authority. I also read @coderabbitai's two rounds first and do not repeat them — see the note at the end.

Moving these jobs to free runners is correct and worth about $227/month. Nothing below argues against it.

### The one that matters

**The green run is the 10-leg PR subset, not the 34-leg matrix this PR exists for.** A push to develop asks for 34 `it_linux` legs and 57 jobs total, against a 20-concurrent-job limit that applies to the whole account and is shared with `readonlyrest_kbn`. The 24 heavy modules outside the PR subset — including the ones the `IT_PARALLELISM` comment says host-OOM'd at exit 137 — have never run on this host. One `workflow_dispatch` with `run_all_tests_on_linux` on this branch settles it.

### Verified as sound

- The runner move itself works: 10/10 `it_linux`, 3/3 `e2e`, 3/3 `it_win`, 4/4 `build_es*` green on the rebased run.
- Memory holds on the legs that did run: worst `avail_mb` 2262 (es818x) against 2422 on Ubicloud, no sample under 1 GB, no exit 137.
- The prune does **not** touch the job's own container or image — `docker system df` after it shows the toolchains image and job container still active — and it runs as step 2, before any testcontainer starts.
- The #1369 merge in `free-host-disk.sh` is semantically correct: shared-host first (it is self-hosted too and the skip would otherwise swallow it), then skip, then the hosted levers.
- Removing the workflow-level `AGENT_ISSELFHOSTED` exposes no self-hosted job in `ci.yml`: `upload_pre_ror`, `release_ror` and `publish_mvn` have no `Free host disk` step, and the first two carry `ROR_SHARED_DOCKER_HOST` anyway.
- No cache regression: the moved Linux jobs use no `actions/cache`, so they add no writer or reader to the 10 GB repo cache (which is 11.5 GB active, all Windows and NVD entries).
- No Docker Hub rate-limit exposure: every Docker Hub pull in the moved jobs goes through `mirror.gcr.io` or an authenticated login.

### Findings

| # | Where | What | |
|---|---|---|---|
| 1 | `ci.yml:286` | The 34-leg develop matrix is unmeasured on this runner | **run before merge** |
| 2 | `ci.yml:294` | `max-parallel: 20` fills a 20-slot account pool, starving e2e and the kbn pre-build | decide |
| 3 | `free-host-disk.sh:43` | Guard fails open — unset `RUNNER_ENVIRONMENT` selects the destructive path | fix |
| 4 | `free-host-disk.sh:51` | Reclaim frees space nobody needed, costs up to 3.6 min/leg | fix |
| 5 | `ci.yml:658` | "Same shape" is false — `build_ror` is 41-72% slower | fix |
| 6 | `free-host-disk.sh:16` | Caller lists wrong in `free-host-disk.sh`, `ci.yml:60` and `CI.md:11` | fix |
| 7 | `mirror-es-libs.yml:43` | nit: the step is a guaranteed no-op | nit |

### On @coderabbitai's round

Both of its findings stand and I confirmed them rather than restating them inline. Its `df -h /` point deserves more weight than "minor": lines 32 and 36 are the **only** unguarded commands in the file (I audited all of them), and they sit in the shared-host branch — which is now the release path on the Ryzen box, so a failing `df` fails a release, not a test. Its documentation point is finding 6 here, with the specific wrong facts named.

---

### Findings outside the PR's changed files

**`.github/workflows/mirror-es-libs.yml:43`** — `.github/workflows/mirror-es-libs.yml` is not among the PR's changed files

nit: **this step is now a guaranteed no-op.**

The job sets `AGENT_ISSELFHOSTED: '1'` at line 37 and runs on a self-hosted runner, so `ci/free-host-disk.sh` takes the skip branch and returns immediately, every time.

Unlike `publish-pre-builds.yml`, this job does not set `ROR_SHARED_DOCKER_HOST`, so it does not get the conservative prune either — it reclaims nothing at all, on the box where disk is the binding constraint.

Pre-existing since #1369, not caused by this PR, but this PR is what makes the script's behaviour legible enough to notice.

**Action:** either delete the step, or add `ROR_SHARED_DOCKER_HOST: '1'` so it gets the same safe prune as `publish-pre-builds.yml`.

.github/workflows/ci.yml
sscarduzio: Still open, and it is the last thing before merge. Dispatching `run_all_tests_on_linux` on this branch now so the full 34-leg matrix runs on the new host. I will report the wave count, wall time and worst `avail_mb`.

.github/workflows/ci.yml
sscarduzio: Closed by decision, not by code. The author wants the account pool saturated and will move the Kibana runs elsewhere, so `max-parallel: 20` stays as it is. `ci/CI.md` now states that the 20-job limit is per account, shared with `readonlyrest_kbn`, and that filling it is deliberate.

ci/free-host-disk.sh
sscarduzio: Fixed in df1fab8e1. The guard now requires `RUNNER_ENVIRONMENT=github-hosted` instead of defaulting to it, so unknown means skip. Verified: `env -i PATH=/usr/bin:/bin bash ci/free-host-disk.sh` prints `not a GitHub-hosted runner (RUNNER_ENVIRONMENT='unset') - skipping disk reclaim` and exits 0. All five paths tested: laptop, shared box, labelled self-hosted, hosted with plentiful disk, hosted with tight disk.

ci/free-host-disk.sh
sscarduzio: Fixed in df1fab8e1. Both levers are now gated on free space, default 40 GB, overridable with `ROR_DISK_RECLAIM_THRESHOLD_GB`. On the current 145 GB image the script prints `enough free space - skipping reclaim` and exits, so the 3m37s per e2e leg goes away. The `~25 GB` claim in the ci.yml comment is gone too.

.github/workflows/ci.yml
sscarduzio: Fixed in df1fab8e1. The three "same shape" comments now carry the measured numbers: `it_linux` 25-35 min against 21-27, and `build_ror` 4.1/19.4/26.6/10.6 against a 30-day Ubicloud average of 2.9/12.3/15.5/6.3. Same vCPU and RAM, not the same throughput.

ci/free-host-disk.sh
sscarduzio: Fixed in df1fab8e1. The header now lists the real callers (`it_linux`, `e2e_tests`, `publish-pre-builds.yml`, `mirror-es-libs.yml`) and says explicitly that `build_ror` does not call it. The `ci.yml` env comment and `ci/CI.md` no longer claim those two workflows are on Ubicloud. Also gave `mirror-es-libs.yml` a `ROR_SHARED_DOCKER_HOST` so its step does the conservative prune instead of nothing — that was the nit in the review body.

.github/workflows/ci.yml
sscarduzio: Closed. Dispatched `run_all_tests_on_linux` on this branch: run 33912487848, **all 34 `it_linux` legs green**, run conclusion `success`.

Numbers for the record:

```
legs            34/34 success, 0 failures
avg leg         30.6 min   (min 24.4, max 37.2)
run wall clock  63 min     (19:43 -> 20:46)
waves           2  — 20 legs started 19:43, the other 14 from 20:10-20:15
```

Two waves, exactly as predicted, because the 20-concurrent-job limit is per account. The heavy modules that were absent from the PR subset — es812x-es815x, es87x-es89x, es91x, es92x, es73x, es74x, es77x-es79x, es711x, es714x, es716x — all passed, with no exit 137.

For comparison the same matrix takes ~54 min on Ubicloud, so the trade is about +9 min of wall time on develop pushes, for $227/month.

sscarduzio: Settled by running it. Dispatched the full matrix on this branch: run [33912487848](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runs/33912487848) - **all 34 legs green on `ubuntu-latest`, 63 min, in two waves.**

That is what led to the leg reduction in this PR. The 34+6+3 shape did not fit the 40-slot account limit, so the run split into two waves; dropping the five ES series with no downloads in two years gives 29+6+3, and run [33921602759](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runs/33921602759) started all 29 legs simultaneously and finished green in one wave.

.github/workflows/ci.yml
sscarduzio: Fixed. `max-parallel` is 29, not 20.

The premise was wrong in my original finding too: the cap was chosen for a 20-job account limit, but this account is GitHub **Pro**, so the limit is 40. Proved it from the run data - 26 jobs started inside one minute, which is impossible under a 20 cap, and exactly 20 `it_linux` legs started, which was our own `max-parallel`. `ci/CI.md` now states 40 and explains the 29 + 6 + 3 = 38 shape that leaves two slots for the ROR KBN pre-build.

ci/free-host-disk.sh
sscarduzio: Fixed. The guard is now `[ "${RUNNER_ENVIRONMENT:-}" != "github-hosted" ]` - absent means skip, not "assume GitHub-hosted". A developer shell, `act`, or an unlabelled self-hosted job reclaims nothing.

The same fail-closed treatment now covers the measurement, after @coderabbitai pointed out the other half: an unreadable `df` or a malformed threshold used to fall through to the levers as well.

ci/free-host-disk.sh
sscarduzio: Fixed. The reclaim is gated on free space: `ROR_DISK_RECLAIM_THRESHOLD_GB` (default 40). On the current `ubuntu-latest` image there are 83-87 GB free before anything runs, so the levers are skipped and the 3m37s per e2e leg is not spent. It only fires when the disk is genuinely tight.

.github/workflows/ci.yml
sscarduzio: Fixed in the ci.yml comments and, as of 27abe4094, in `ci/CI.md` too - it was the one place still repeating "same shape as ubicloud-standard-4" without the caveat.

All three now say the spec sheet matches and the throughput does not, with the measured numbers: `it_linux` 25-35 min against 21-27 on Ubicloud, `build_ror` 40-70% slower. Stated as the deliberate trade rather than glossed over.

ci/free-host-disk.sh
sscarduzio: Fixed. The comment no longer names `build_ror` as a `container:` caller. It now describes the two shapes that actually call the script - the bare runner (`e2e_tests`), where the preinstalled toolchains are visible, and the `container:` job (`it_linux`), where only the bind-mounted Docker daemon is reachable - and stops enumerating callers that drift.

ci/free-host-disk.sh
```diff
@@ -1,34 +1,90 @@
...
+# daemon, so "unknown" has to mean "skip", not "go ahead".
+if [ "${RUNNER_ENVIRONMENT:-}" != "github-hosted" ] || [ "${AGENT_ISSELFHOSTED:-0}" = "1" ]; then
+  echo ">>> not a GitHub-hosted runner (RUNNER_ENVIRONMENT='${RUNNER_ENVIRONMENT:-unset}') - skipping disk reclaim"
+  exit 0
+fi
+
+avail_gb=$(df --output=avail -BG / 2>/dev/null | tail -1 | tr -dc '0-9' || echo 0)
+echo ">>> [host] disk before reclaim: ${avail_gb}GB free (threshold ${ROR_DISK_RECLAIM_THRESHOLD_GB}GB)"
+df -h / || true
+if [ "${avail_gb:-0}" -ge "$ROR_DISK_RECLAIM_THRESHOLD_GB" ]; then
```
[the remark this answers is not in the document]
sscarduzio: Fixed in 27abe4094, and thank you - this was the half of the fail-closed argument I had missed. An unreadable `df` left `avail_gb` empty and a malformed threshold made the numeric test error, and both fell through to the destructive levers.

Both now skip:

```text
df broken, sane threshold    -> cannot trust the free-space check - skipping
df fine, malformed threshold -> cannot trust the free-space check - skipping
df fine, sane threshold      -> enough free space - skipping reclaim
```

(The `df -h /` telemetry calls already carried `|| true`.)

.github/workflows/ci.yml
```diff
@@ -90,10 +96,17 @@ jobs:
...
+          # specific to one of them would now ship unnoticed. Revisit if a customer appears.
+          LINUX_IT_FULL_SET='["es94x","es92x","es91x","es90x","es818x","es816x","es815x","es814x","es813x","es812x","es811x","es810x","es89x","es88x","es87x","es85x","es84x","es83x","es82x","es81x","es80x","es717x","es716x","es710x","es78x","es77x","es72x","es70x","es67x"]'
           LINUX_IT_PR_READY_SET='["es94x","es90x","es818x","es816x","es810x","es80x","es717x","es710x","es70x","es67x"]'
           LINUX_IT_PR_DRAFT_SET='["es94x","es818x","es717x","es67x"]'
-          WIN_IT_FULL_SET='["es94x","es92x","es91x","es90x","es818x","es816x","es815x","es814x","es813x","es812x","es811x","es810x","es89x","es88x","es87x","es85x","es84x","es83x","es82x","es81x","es80x","es717x","es716x","es714x","es711x","es710x","es79x","es78x","es77x","es74x","es73x","es72x","es70x"]'
+          WIN_IT_FULL_SET='["es94x","es92x","es91x","es90x","es818x","es816x","es815x","es814x","es813x","es812x","es811x","es810x","es89x","es88x","es87x","es85x","es84x","es83x","es82x","es81x","es80x","es717x","es716x","es710x","es78x","es77x","es72x","es70x"]'
```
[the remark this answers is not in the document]
sscarduzio: Fixed in 27abe4094. You were right about the Windows count, and the same drift was in two more places, so I checked every documented count against the set `setup` actually emits:

```text
LINUX_IT_FULL_SET  29   ci.yml said "full 34"
WIN_IT_FULL_SET    28   ci.yml and CI.md said "full 33"
WIN_IT_MASTER...    6   CI.md said "7 on branches"
```

All three corrected, in both `ci.yml` and `ci/CI.md`.

ci/CI.md
```diff
@@ -1,9 +1,32 @@
...
 
-CI runs on GitHub Actions: `.github/workflows/ci.yml`. Linux jobs run on **Ubicloud**
-runners (`ubicloud-standard-4` = 4 vCPU / 16 GB) inside the `beshultd/ror-ci-toolchains`
-image; Windows jobs run on GitHub-hosted `windows-2025`. `ci/toolchains/image.env` holds that
-image's tag, and every workflow that needs it sources that file.
+CI runs on GitHub Actions: `.github/workflows/ci.yml`. The Linux **test and build** jobs run on
+**GitHub-hosted** `ubuntu-latest` runners inside the `beshultd/ror-ci-toolchains` image; Windows
+jobs run on GitHub-hosted `windows-2025`. The **release** jobs are the exception and stay on the
+shared self-hosted box: `upload_pre_ror`, `release_ror` and `publish_mvn` here, plus the standalone
+`mirror-es-libs.yml`. (`publish-pre-builds.yml` moved back to `ubuntu-latest` — see PR #1372.)
```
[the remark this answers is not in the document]
sscarduzio: Right, and my mistake. I forward-referenced #1372, which moves `publish-pre-builds.yml` to `ubuntu-latest` but has not merged - so the doc described a state that does not exist on this branch.

Fixed in 88e1fcb3f: it now lists `publish-pre-builds.yml` alongside the other self-hosted workflows, which is what this branch actually leaves.

.github/workflows/ci.yml
```diff
@@ -383,7 +405,7 @@ jobs:
 
   # ── TEST: Windows integration tests (GitHub-hosted) ───────────────────────
   # Matrix from setup: 6-version subset on develop/master/epic, 3 on ready PRs, none on draft
-  # PRs (job skipped), full 33 on manual run_all_tests_on_windows.
+  # PRs (job skipped), full 28 on manual run_all_tests_on_windows.
```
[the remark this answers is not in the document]
sscarduzio: Not taking this one - I think it reads a wrapped sentence as two. Lines 406-408 are one sentence:

```text
# Matrix from setup: 6-version subset on develop/master/epic, 3 on ready PRs, none on draft
# PRs (job skipped), full 28 on manual run_all_tests_on_windows.
```

Read across the line break it is "...3 on ready PRs, **none on draft PRs (job skipped)**, full 28 on manual...". The "(job skipped)" qualifies the draft case, and `WIN_IT_PR_DRAFT_SET='[]'` in `setup` is exactly that: an empty matrix means the job does not run. There is no contradiction with line 407.

Applying the suggested diff would delete the draft-PR coverage from the comment and leave it describing less than it does today, so I am leaving it as is. Happy to reword if you can show me a reading where it is ambiguous.

.github/workflows/ci.yml
```diff
@@ -54,7 +54,13 @@ env:
   # full-matrix run (es80x/es810x/es818x/es90x, exit 137) — those legs run suites es94x excludes,
   # so their memory peaks are higher. Fleet wall time is set by the slowest leg anyway: k=4.
   IT_PARALLELISM: '4'
-  AGENT_ISSELFHOSTED: '1'
+  # AGENT_ISSELFHOSTED is deliberately NOT set here any more. Its only reader is
```
coutoPL: Please remove this comment.

  It explains the change, not the code. A reader of ci.yml sees only that AGENT_ISSELFHOSTED is absent — the normal state. The comment tells the story of a line that
  no longer exists. Six months from now nobody will know if it is still true.

  Two more problems:

  1. It rots. It names mirror-es-libs.yml, publish-pre-builds.yml, and three release jobs. Each rename makes the comment wrong, and nothing fails when it does.  
  2. It hides the rule. The real rule is one line: the disk reclaim runs only on GitHub-hosted runners. That line is buried in a paragraph about history.

  What to do here:

  - Move the history to the commit message and the PR description. git log -S AGENT_ISSELFHOSTED finds it there, forever, with the diff next to it.
  - Put the rule in ci/free-host-disk.sh, at the guard that enforces it. That is the one place a reader must see it, and the code there cannot drift from it.

  And a proposal for all of us, because we have no written convention yet:

  ▎ A comment explains the current state, and only when the state is not obvious. It does not explain the change that made it. History belongs to the commit message 
  ▎ and the PR.
  ▎
  ▎ Write it in Simplified Technical English (ASD-STE100), and follow Zinsser's four principles: simplicity, brevity, clarity, humanity. In practice:
  ▎
  ▎ - One idea per sentence. Short sentences.
  ▎ - Active voice. Say who does what.
  ▎ - Use the same word for the same thing each time.
  ▎ - Cut every word that does nothing.
  ▎ - Write for a person, not for a specification.

  Let us add that to CLAUDE.md. Then this discussion does not repeat on the next PR.

Review other comments in the PR and check if they follow the rule.

.github/workflows/ci.yml
```diff
@@ -266,8 +279,8 @@ jobs:
           path: '**/TEST*.xml'
           if-no-files-found: ignore
 
-  # ── TEST: Linux integration tests (Ubicloud) ──────────────────────────────
-  # Matrix comes from setup: full 34-version set on develop/master/epic + manual run, the
+  # ── TEST: Linux integration tests (GitHub-hosted) ─────────────────────────
```
coutoPL: > (GitHub-hosted) 

this information should not be added in this comment.

.github/workflows/ci.yml
```diff
@@ -266,8 +279,8 @@ jobs:
...
+  # Matrix comes from setup: full 29-version set on develop/master/epic + manual run, the
```
coutoPL: This comment is low quality - it may rot easily. It doesn't bring anything important to the context. It should be removed.

Check other places too.

.github/workflows/ci.yml
```diff
@@ -276,15 +289,24 @@ jobs:
...
     # <leg>_<module> shape as it_win_* and e2e_*, so the three test families sort together.
     name: it_linux_${{ matrix.ES_MODULE }}
     strategy:
       fail-fast: false
-      max-parallel: 20
+      # 29 = the full Linux matrix, so every leg starts in one wave. The account allows 40
+      # concurrent jobs (Pro); the other long jobs of a develop push are 6 it_windows and 3
+      # e2e, so 29 + 6 + 3 = 38 fits with two slots left for the ROR KBN pre-build that
+      # e2e_prepare dispatches into readonlyrest_kbn on this same account.
+      max-parallel: 29
```
coutoPL: This is in the comment description.
> The repo is public, so these runners cost nothing.

Now, you write about Pro and set `max-parallel: 29`. Is it Pro or Free? 
If Pro, the savings calculations may be wrong, because I'm not sure if you took into consideration the price of Pro plan.

.github/workflows/ci.yml
```diff
@@ -383,7 +405,7 @@ jobs:
 
   # ── TEST: Windows integration tests (GitHub-hosted) ───────────────────────
```
coutoPL: this information should not be added in this comment.
Remove it and check other places

.github/workflows/mirror-es-libs.yml
```diff
@@ -35,6 +35,10 @@ jobs:
       GRADLE_USER_HOME: ${{ github.workspace }}/.gradle-home
       GRADLE_OPTS: '-Dorg.gradle.java.installations.auto-download=true'
       AGENT_ISSELFHOSTED: '1'
+      # Shared box: let ci/free-host-disk.sh do its conservative prune (dangling layers and build
```
coutoPL: low quality comment. to be removed

ci/CI.md
```diff
@@ -1,9 +1,32 @@
...
+shared self-hosted box: `upload_pre_ror`, `release_ror` and `publish_mvn` here, plus the standalone
+`mirror-es-libs.yml` and `publish-pre-builds.yml`.
+
+`ubuntu-latest` is 4 vCPU / 16 GB, the same shape as the `ubicloud-standard-4` these jobs used to
+run on, but **not the same throughput**: measured on the 10-module PR set, an `it_linux` leg takes
+25-35 min here against 21-27 min on Ubicloud, and the `build_ror` legs are 40-70% slower. That is
+the deliberate trade — this was the single biggest line of the Ubicloud bill and it now costs
+nothing. `ci/toolchains/image.env` holds that image's tag, and every workflow that needs
+it sources that file. The repo is public, so GitHub-hosted runners are free. The concurrency limit
+is 40 jobs **per account** (GitHub Pro), shared with `readonlyrest_kbn`. A develop push has 29
```
coutoPL: the rotting problem

ci/CI.md
```diff
@@ -17,8 +40,8 @@ directory contain the build logic; the workflow only orchestrates.
 | `required_checks` | audit build, cross-Scala compile, format, license | pushes + PRs |
 | `unit_tests_linux` | `core:test` and friends | pushes + PRs |
 | `optional_checks` | non-blocking checks (matrix; today: `cve_check` OWASP dependency-check, needs `NVD_API_KEY`) — failures annotate the run but never block it | pushes + PRs |
-| `it_linux` | integration tests, one job per ES version | 10-version subset on PRs, full 34 on develop/master/epic and manual |
-| `it_windows` | integration tests on native-Windows ES | 3 on PRs, 7 on branches, full 33 on manual |
+| `it_linux` | integration tests, one job per ES version | 10-version subset on PRs, full 29 on develop/master/epic and manual |
```
coutoPL: the rotting problem

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: Agreed, and removed. The block described a line that is not there, which is the worst kind of comment: nothing fails when it stops being true.

The rule it carried now sits in `ci/free-host-disk.sh`, at the guard that enforces it:

> Do not set AGENT_ISSELFHOSTED workflow-wide: it turns this script off in every job that inherits it.

That is the one place a reader must see it, and it cannot drift from the code below it. The history is in the commit.

On your proposal: I agree with it and I have applied it to all seven other PRs in this set, not just this one. I have not put it in CLAUDE.md yet, because that binds everyone and the wording should be yours, not mine. Open a PR with it and I will follow it.

.github/workflows/ci.yml
sscarduzio: Removed. `runs-on` two lines below already says the runner.

.github/workflows/ci.yml
sscarduzio: Removed, here and at the `it_windows` header. `setup` computes both matrices, so the counts had two homes and only one of them fails when they diverge.

I checked the other places. Also gone in this PR: the measured leg timings on three `runs-on` comments, the disk-reclaim measurements in `ci/free-host-disk.sh`, and the caller list in that same header, which was a list that goes stale in silence.

.github/workflows/ci.yml
sscarduzio: Good catch, and the comment deserved the doubt. I removed the plan claim rather than correct it: I cannot read the plan from the API, so I would only be swapping one unverifiable claim for another.

To answer directly: the savings are not affected either way. The repo is public, so GitHub-hosted minutes are free on Free and on Pro. The plan changes only the concurrent-job ceiling (20 against 40). If the account is Free, `max-parallel: 29` is not wrong, the extra legs just queue instead of starting in one wave, so it costs wall time and not money.

The comment now states the rule and tells the reader to check the real limit before raising the number.

.github/workflows/ci.yml
sscarduzio: Removed. Same as the Linux header, and I sweeped the rest of the file.

.github/workflows/mirror-es-libs.yml
sscarduzio: Removed. It is now one line on the variable itself:

```yaml
ROR_SHARED_DOCKER_HOST: '1'   # shared box: ci/free-host-disk.sh prunes conservatively
```

What the flag does belongs in the script that reads it, and that script explains it.

ci/CI.md
sscarduzio: Fixed. That paragraph carried a measured throughput comparison, a leg count, a Windows count and the plan claim. All of it was true on one day, and a reader had no way to know which day.

It now gives the rule: the concurrency limit is per account and `readonlyrest_kbn` shares it, `max-parallel` tracks the size of the full matrix, and you check the account limit against every long job a develop push starts before you raise either. The excluded ES series point at `LINUX_IT_FULL_SET` in `ci.yml` instead of listing themselves a second time.

ci/CI.md
sscarduzio: Fixed. The `When` column said "10-version subset on PRs, full 29 on develop/master/epic" and the Windows row carried three more numbers. `setup` owns those sets, so the table now says which kind of set applies and lets the workflow hold the counts.

sscarduzio:

What changed in each:

|---|---|---|

.github/workflows/ci.yml
```diff
@@ -266,25 +268,27 @@ jobs:
...
     # <leg>_<module> shape as it_win_* and e2e_*, so the three test families sort together.
     name: it_linux_${{ matrix.ES_MODULE }}
     strategy:
       fail-fast: false
-      max-parallel: 20
+      # Keep this equal to the number of legs in LINUX_IT_FULL_SET, so the whole matrix starts
+      # in one wave. Check the account's concurrent-job limit before you raise it: it_windows and
+      # e2e_tests run at the same time, and e2e_prepare dispatches one more job into
+      # readonlyrest_kbn on the same account.
+      max-parallel: 29
```
coutoPL: on the free runner, the limit is 20:
https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners

.github/workflows/ci.yml
coutoPL: Drop `max-parallel` here. `LINUX_IT_FULL_SET` has exactly 29 legs, so the cap never
  binds — unbounded is already the default. It only starts acting when someone adds a
  30th module, and then it throttles silently. Same for the `99` at line 578.

  Keep `it_windows`' `10` — 28 legs against it, so that one is a real throttle.

coutoPL (commented):
one comment

.github/workflows/ci.yml
sscarduzio: You are right, and the "throttles silently" part is the real argument. I checked all five caps against their matrices:

```
it_linux    cap 29  vs LINUX_IT_FULL_SET = 29 legs   -> never binds   REMOVED
e2e_tests   cap 99  vs E2E_FULL_SET      =  3 legs   -> never binds   REMOVED
it_windows  cap 10  vs WIN_IT_FULL_SET   = 28 legs   -> real throttle  kept
upload_pre_ror / release_ror  cap 2                  -> the self-hosted pool sets it, kept
```

Both removed caps were worse than useless. They did nothing until a 30th module appeared, and then they would have halved the fleet with no message and no failing check.

That also disposes of the Pro-or-Free question underneath this thread. With no cap there is no number to justify, and the account limit becomes what it always was: a real ceiling that queues the surplus, not something the workflow has to guess.

`ci/CI.md` now carries the rule rather than a description of the removed cap:

> `it_linux` and `e2e_tests` set no `max-parallel`: a cap equal to the number of legs never binds, and it starts throttling silently the day someone adds one more module.

Pushed.

coutoPL (approved):
LGTM

coutoPL:
I have changed the base branch. Let's merge into master and then merge it back into develop.

PR #1370, opened 2026-09-04
ci: raise the release_ror timeout to 600 minutes

sscarduzio:
## The problem

Release 1.71.0 lost its ES 8.x artifacts. The `release_es8xx` leg was killed at the 180-minute limit.

## The numbers

Measured release legs, 1 Aug to 4 Sep:

```text
es6xx   19-26 min
es7xx   83-109 min
es8xx  118-152 min
es9xx   51-85 min
```

The old limit was inside the normal range for `es8xx`.

## Also

`ci/self-hosted-runner.md` said `release_ror` takes about 16 min per leg. It now shows the measured values above, and the observed end-to-end time of a release: 118 to 207 min.

## Summary by CodeRabbit

* **Documentation**
  * Updated self-hosted runner guidance with clearer recommendations for job placement and pool sizing.
  * Added instructions for reviewing current release job timing and runner occupancy.
  * Replaced outdated timing and cost estimates with guidance for sizing capacity based on the slowest job.

* **Chores**
  * Increased the release workflow timeout limit to accommodate longer-running release jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -843,7 +843,12 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    timeout-minutes: 180
+    # 600, matching upload_pre_ror. These legs run on the self-hosted box now, which is a
+    # Ryzen 7 3700X — release_es8xx averaged 139 min on ubicloud-standard-4 and was killed by the
+    # old 180 bound on its first self-hosted run (1.71.0), leaving that release without its ES 8.x
+    # artifacts. The bound exists to catch a hung job, not to cap a slow one, and a self-hosted
+    # minute costs nothing, so give it the same headroom the pre-build path already has.
+    timeout-minutes: 600
     env: *shared_docker_host
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -843,9 +843,8 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    # Catches a hung job; it does not cap a slow one. A killed leg leaves the release without
-    # that ES major's artifacts, and a self-hosted minute costs nothing, so keep it generous
-    # and equal to upload_pre_ror above.
+    # Generous on purpose: a killed leg leaves the release without that ES major's artifacts,
+    # and a self-hosted minute costs nothing.
     timeout-minutes: 600
     env: *shared_docker_host
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -843,11 +843,9 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    # 600, matching upload_pre_ror. These legs run on the self-hosted box now, which is a
-    # Ryzen 7 3700X — release_es8xx averaged 139 min on ubicloud-standard-4 and was killed by the
-    # old 180 bound on its first self-hosted run (1.71.0), leaving that release without its ES 8.x
-    # artifacts. The bound exists to catch a hung job, not to cap a slow one, and a self-hosted
-    # minute costs nothing, so give it the same headroom the pre-build path already has.
+    # Catches a hung job; it does not cap a slow one. A killed leg leaves the release without
+    # that ES major's artifacts, and a self-hosted minute costs nothing, so keep it generous
+    # and equal to upload_pre_ror above.
     timeout-minutes: 600
     env: *shared_docker_host
     strategy:
--- a/ci/self-hosted-runner.md
+++ b/ci/self-hosted-runner.md
@@ -8,11 +8,17 @@ Jobs that need a self-hosted runner:
 
 | Workflow | Job | Shape |
 |---|---|---|
-| `ci.yml` | `upload_pre_ror` | 4-leg matrix, ~45–57 min/leg, pre-release only |
-| `ci.yml` | `release_ror` | 4-leg matrix, ~16 min/leg, release only |
+| `ci.yml` | `upload_pre_ror` | 4-leg matrix, 8–95 min/leg (median 15/43/54/26 for es6/7/8/9xx), pre-release only |
+| `ci.yml` | `release_ror` | 4-leg matrix, 19–152 min/leg (median 22/93/138/61 for es6/7/8/9xx), release only |
 | `ci.yml` | `publish_mvn` | seconds, after `release_ror` |
 | `publish-pre-builds.yml` | `publish` | manual, long build-and-push |
 | `mirror-es-libs.yml` | `mirror` | manual, short |
+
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -773,7 +773,12 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    timeout-minutes: 180
+    # 600, matching upload_pre_ror. These legs run on the self-hosted box now, which is a
+    # Ryzen 7 3700X — release_es8xx averaged 139 min on ubicloud-standard-4 and was killed by the
```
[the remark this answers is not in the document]
sscarduzio: Confirmed and fixed in e61d8e42b. I pulled the numbers rather than trusting either figure - n=10 release legs, n=26 upload legs, 1 Aug to 4 Sep:

```text
release_ror     es6xx  19-26   median  22
                es7xx  83-109  median  93
                es8xx 118-152  median 138
                es9xx  51-85   median  61

upload_pre_ror  es6xx   8-36   median  15
                es7xx  31-85   median  43
                es8xx  41-95   median  54
                es9xx  15-63   median  26
```

So `release_ror` was understated by about 10x, and `upload_pre_ror` was wrong the other way - it claimed "~45-57 min/leg" for all four.

It matters exactly where you say it does: this is the doc that sizes the pool these jobs now run on. The note under the table now says to size on the slowest leg, not the median, and points at the 180-minute bound this PR raises.

ci/self-hosted-runner.md
```diff
@@ -8,12 +8,18 @@ Jobs that need a self-hosted runner:
...
+| `ci.yml` | `release_ror` | 4-leg matrix, 19–152 min/leg (median 22/93/138/61 for es6/7/8/9xx), release only |
 | `ci.yml` | `publish_mvn` | seconds, after `release_ror` |
 | `publish-pre-builds.yml` | `publish` | manual, long build-and-push |
 | `mirror-es-libs.yml` | `mirror` | manual, short |
 
+Those are measured, not estimates: n=10 release legs and n=26 upload legs from the 1 Aug - 4 Sep
+window. Size the pool on the **slowest** leg, not the median - `release_es8xx` reaches 152 min on a
+good day, and one run hit the 180-minute job bound and was killed, which is what PR #1370 raises.
+A two-slot pool means the four release legs run two at a time, so a release occupies the box for
+roughly `es8xx + es7xx` back to back.
```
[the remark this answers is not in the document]
sscarduzio: Right - that was an assumption about scheduling, not an observation. `max-parallel: 2` caps concurrency; it does not pin which legs share a slot.

Replaced with the measured end-to-end time of the four `release_ror` legs, first leg start to last leg finish, over the last five develop releases:

```text
118.4  127.7  144.4  151.8  207.1 min
```

The 207.1 is the run where `release_es8xx` was killed at the 180-minute bound - the case this PR fixes. Thanks.

ci/self-hosted-runner.md
```diff
@@ -8,12 +8,27 @@ Jobs that need a self-hosted runner:
 
 | Workflow | Job | Shape |
 |---|---|---|
-| `ci.yml` | `upload_pre_ror` | 4-leg matrix, ~45–57 min/leg, pre-release only |
-| `ci.yml` | `release_ror` | 4-leg matrix, ~16 min/leg, release only |
+| `ci.yml` | `upload_pre_ror` | 4-leg matrix, 8–95 min/leg (median 15/43/54/26 for es6/7/8/9xx), pre-release only |
```
coutoPL: Take a look at [this comment](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1368#discussion_r3948461591) from the other PR>

This info will soon become out of date. We should not add info like this to the code or comments. It should land in the PR or/and commit description. 

-- 
the comment applies to the all comments from this PR

coutoPL (changes requested):

ci/self-hosted-runner.md
sscarduzio: Applied, and you are right that it was the worst case of it. That table carried per-leg medians and ranges, and below it five end-to-end release times. Every one of them was measured on a single date, and nothing in the file said which.

There was a real tension to resolve, so I want to be explicit about how. The runbook exists to answer one question, "how many runner slots do I need", and that question needs a number. Deleting the numbers and leaving nothing would have made the file useless. So the file now keeps the rule and gives the reader the command that produces a current answer:

> **Size the pool on the slowest leg, not the average, and measure before you change it.** Leg times differ by more than an order of magnitude between ES majors, and they move when the build changes.

followed by the `gh run list` invocation that prints the current per-leg times. The `Shape` column now says only what a job is.

I also added the placement rule that this whole set of PRs turned on, because it is the thing a reader most needs and it was in no file at all:

> **Do not move a job here if another run polls it.** The pool has few slots. A queued job holds up every runner that waits for its result, and those runners may be paid.

The `timeout-minutes: 600` comment now says what the bound is for, and no longer retells the run that hit the old one.

Applied the same rule to the other six PRs in this set. Summary in the PR conversation.

sscarduzio:

What changed in each:

|---|---|---|

.github/workflows/ci.yml
```diff
@@ -773,7 +773,10 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    timeout-minutes: 180
+    # Catches a hung job; it does not cap a slow one. A killed leg leaves the release without
```
coutoPL: This comment is unnecessary.
See: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1377

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: I read `docs/dev/code-style.md` from #1377 and trimmed the comment, but I did not remove it. Your own doc is the reason:

> Comment the things a reader cannot get from the code:
> - The non-obvious: **why this value**, why this order, which constraint forbids the simpler form.

`timeout-minutes: 600` is a bare number. Nothing else in the file says why it is not 180, and 180 is what a reader would otherwise pick as reasonable. That is the "why this value" case your doc names.

What was genuinely wrong with my version, by that same doc: the first sentence explained what a timeout is for, which is self-describing, and the last named `upload_pre_ror`, which is a cross-reference that goes stale on a rename. Both are gone. It now reads:

    # Generous on purpose: a killed leg leaves the release without that ES major's artifacts,
    # and a self-hosted minute costs nothing.
    timeout-minutes: 600

Two lines, both answering why the number is large.

If you still want it at zero lines, say so and I will drop it. You own the convention. I would just rather the next person not shorten this back to 180 and lose an ES major from a release to find out why.

Thanks for writing #1377. It settles this properly instead of us relitigating it on each PR.

(Apologies for the truncated first version of this reply — my tooling ate it at the code fence.)

coutoPL (approved):
LGTM

coutoPL:
@sscarduzio I changed the target branch - it should be master. Please rebase the solution.

PR #1371, opened 2026-09-04
ci: give the master merge-back the reduced matrix

sscarduzio:
## The problem

A PR merged into master starts two runs: the push to master, and the merge-back push to develop about one minute later. Both run the full Linux and Windows matrices on the same tree. Between 1 Aug and 4 Sep this happened 4 times. Three pairs started less than 60 s apart and competed for the same job slots.

## The change

`setup` compares the develop tree with the tip of master. When they are equal, the develop push takes the reduced PR matrix: 10 Linux legs instead of 34, and 3 Windows legs instead of 6.

## Safety

- A release version (no `-pre` in `pluginVersion`) always keeps the full matrix.
- A failed `gh api` call does not stop the run. The step falls back to the full matrix.
- No job is skipped. No release gate changes.

## Summary by CodeRabbit

- **Chores**
  - Improved CI reliability by centralizing merge-back detection and handling lookup failures more consistently.
  - Release-version pushes continue to use the full validation matrix.
  - Eligible pre-release merge-backs continue to use the reduced validation matrix.
  - Linux and Windows merge-back validation now use the same computed merge-back status.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,13 +77,29 @@ jobs:
       - id: flags
         env:
           REF: ${{ github.head_ref || github.ref_name }}
+          IS_DEVELOP_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/develop' }}
         run: |
+          source ci/ci-lib.sh
+
           echo "is_master=${{ github.ref == 'refs/heads/master' }}"   >> "$GITHUB_OUTPUT"
           echo "is_develop=${{ github.ref == 'refs/heads/develop' }}" >> "$GITHUB_OUTPUT"
+
           # Treat nested */epic/* branches like top-level epic/* branches.
-          case "$REF" in *epic/*)  echo "is_epic=true"  >> "$GITHUB_OUTPUT";; *) echo "is_epic=false"  >> "$GITHUB_OUTPUT";; esac
+          case "$REF" in
+            *epic/*) echo "is_epic=true"  >> "$GITHUB_OUTPUT";;
+            *)       echo "is_epic=false" >> "$GITHUB_OUTPUT";;
+          esac
+
+          MERGE_BACK=no
+          [ "$IS_DEVELOP_PUSH" = "true" ] && is_merge_back_of_master && MERGE_BACK=yes
+
+          case "$MERGE_BACK" in
+            yes) echo "is_merge_back=true"  >> "$GITHUB_OUTPUT";;
+            *)   echo "is_merge_back=false" >> "$GITHUB_OUTPUT";;
+          esac
       - id: matrices
         env:
+          IS_MERGE_BACK: ${{ steps.flags.outputs.is_merge_back || false }}
           # Deliberately also true for epic/** refs — they get the same full matrices as master/develop.
           IS_MASTER_OR_DEVELOP: ${{ github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master' || contains(github.head_ref || github.ref_name, 'epic/') }}
           IS_PR_DRAFT: ${{ github.event.pull_request.draft == true }}
@@ -104,6 +120,8 @@ jobs:
 
           if [ "$IS_PR_DRAFT" = "true" ]; then
             echo "linux=$LINUX_IT_PR_DRAFT_SET" >> "$GITHUB_OUTPUT"
+          elif [ "$IS_MERGE_BACK" = "true" ]; then
+            echo "linux=$LINUX_IT_PR_READY_SET" >> "$GITHUB_OUTPUT"
           elif [ "$MANUAL_RUN_ACTION" = "run_all_tests_on_linux" ] || [ "$IS_MASTER_OR_DEVELOP" = "true" ]; then
             echo "linux=$LINUX_IT_FULL_SET" >> "$GITHUB_OUTPUT"
           else
@@ -112,6 +130,8 @@ jobs:
 
           if [ "$IS_PR_DRAFT" = "true" ]; then
             echo "windows=$WIN_IT_PR_DRAFT_SET" >> "$GITHUB_OUTPUT"
+          elif [ "$IS_MERGE_BACK" = "true" ]; then
+            echo "windows=$WIN_IT_PR_READY_SET" >> "$GITHUB_OUTPUT"
           elif [ "$MANUAL_RUN_ACTION" = "run_all_tests_on_windows" ]; then
             echo "windows=$WIN_IT_FULL_SET" >> "$GITHUB_OUTPUT"
           elif [ "$IS_MASTER_OR_DEVELOP" = "true" ]; then
@@ -843,9 +863,7 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    # Generous on purpose: a killed leg leaves the release without that ES major's artifacts,
-    # and a self-hosted minute costs nothing.
-    timeout-minutes: 600
+    timeout-minutes: 180
     env: *shared_docker_host
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -863,7 +863,9 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    timeout-minutes: 180
+    # Generous on purpose: a killed leg leaves the release without that ES major's artifacts,
+    # and a self-hosted minute costs nothing.
+    timeout-minutes: 600
     env: *shared_docker_host
     strategy:
       fail-fast: false
--- a/ci/ci-lib.sh
+++ b/ci/ci-lib.sh
@@ -23,6 +23,22 @@ gradle_property() {
   printf '%s\n' "$value"
 }
 
+# True when HEAD carries the same tree as master's tip and pluginVersion is a -pre version.
+# Fetches master from origin; needs no token on a public repo. False on any failure.
+is_merge_back_of_master() {
+  local master_tree here_tree plugin_version
+  git fetch --quiet --no-tags --depth=1 origin master 2>/dev/null || return 1
+  master_tree=$(git rev-parse --verify --quiet 'FETCH_HEAD^{tree}') || return 1
+  here_tree=$(git rev-parse --verify --quiet 'HEAD^{tree}') || return 1
+  echo ">>> HEAD tree $here_tree, master tree $master_tree" >&2
+  [ "$here_tree" = "$master_tree" ] || return 1
+  plugin_version=$(gradle_property pluginVersion) || return 1
+  case "$plugin_version" in
+    *-pre*) return 0 ;;
+    *) echo ">>> release version $plugin_version - keeping the full matrix" >&2; return 1 ;;
+  esac
+}
+
 docker_image_exists() {
   # This answer decides whether we skip an expensive rebuild, so it must come from Docker Hub and
   # never from a cache: a cache can hold a stale answer for a tag we pushed a moment ago. `docker


.github/workflows/ci.yml
```diff
@@ -102,6 +133,8 @@ jobs:
 
           if [ "$IS_PR_DRAFT" = "true" ]; then
             echo "linux=$LINUX_IT_PR_DRAFT_SET" >> "$GITHUB_OUTPUT"
+          elif [ "${IS_MERGE_BACK:-false}" = "true" ]; then
```
sscarduzio: The reduced matrix also reduces the coverage behind the develop publish and release gate.

The merge-back arm fires on a `push` to develop. That is exactly the run that publishes:

```
ci.yml:712  determine_ci_type   needs.it_linux.result == 'success'
ci.yml:743  upload_pre_ror      needs.determine_ci_type.result == 'success'
ci.yml:799  release_ror         needs.it_linux.result == 'success'  (+ contents: write, tag push)
```

With `IS_MERGE_BACK=true`, `it_linux` is `LINUX_IT_PR_READY_SET` - 10 of 34 modules - and `it_windows` is 3 of 6.

The step comment says develop "keeps an independent signal and its release gate still needs every job to pass". That is true of the *jobs* and false of the *coverage*. The tree test also proves nothing about master's run: it does not check that master passed, or even finished. So a break that only the 24 dropped legs catch fails master and still publishes from develop.

There is a second-order case. The guard fires when develop's tree equals master's tip, which is when develop carries master's released non-`-pre` `pluginVersion` - so the develop *release* run itself can take the reduced matrix.

Decide one of these, and say which in the comment: keep the full matrix when `determine_ci_type` would fire, or add `needs.setup.outputs.is_merge_back != 'true'` to `upload_pre_ror` and `release_ror`, or accept it and write in the comment that master's run is the coverage of record for a merge-back.

sscarduzio (commented):
**COMMENT.** One finding needs a decision before merge: the reduced matrix also reduces the coverage behind the develop publish and release gate.

Verified as sound: the tree comparison is the right key (the two commits are different objects, so a commit-id check finds nothing), it replays correctly on all 4 duplicate pairs in the window, and it stays quiet on `fbd2eb906` and `73c326792` where develop had its own commits. `${IS_MERGE_BACK:-false}` handles the skipped-step empty string. Comparing against master's current tip rather than the merged commit fails safe: a moved master gives the full matrix. The step is `push`-only, so `workflow_dispatch` cannot reach the merge-back arm. No script injection: no `${{ }}` reaches the `run:` block, and `permissions: contents: read` at workflow level keeps `github.token` read-only.

Index:
1. `ci.yml:136` - the develop release gate drops to 10 of 34 ES modules. **Decision needed before merge.**
2. `ci.yml:106` - a failing `gh api` fails the whole run. Extends @coderabbitai.
3. `ci.yml:74` - dead job output.
4. `ci.yml:87` - the comment counts work the change does not save, and #1368 makes the number wrong.

Not in this diff: the PR title fails this repo's own `conventions` check - `the title must start with the Jira key`. It is red on this PR now.

.github/workflows/ci.yml
```diff
@@ -83,8 +84,38 @@ jobs:
...
+      # This does not skip the run. It gives the merge-back the same reduced matrix a ready PR
+      # gets, so develop keeps an independent signal and its release gate still needs every job to
+      # pass. Waiting for master's run instead would couple the two runs and buy little: master is
+      # covering the full matrix at that moment anyway.
+      - id: merge_back
+        if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          MASTER_SHA=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/heads/master" -q .object.sha)
```
sscarduzio: A failing `gh api` here fails the whole CI run. This extends @coderabbitai's finding on this line with the reproduction and a narrower fix.

GitHub runs `shell: bash` as `bash --noprofile --norc -e -o pipefail {0}`. Reproduced with those exact flags:

```
--- case A: gh api exits non-zero (rate limit / 5xx) ---
gh: HTTP 502
step exit=1      -> setup fails -> the whole run fails, release path included

--- case B: gh api succeeds but returns empty ---
>>> develop tree deadbeef, master tree
value=false
step exit=0      -> full matrix. Safe.
```

So only case A needs handling; the empty-response path already fails safe to the full matrix. Every develop push now depends on two GitHub API calls, and a transient 502 or a rate limit fails a run that has nothing wrong with it.

Add `|| true` to both assignments and let the empty value fall through to the existing `else` branch.

.github/workflows/ci.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       it_linux_matrix:   ${{ steps.matrices.outputs.linux }}
       it_windows_matrix: ${{ steps.matrices.outputs.windows }}
       e2e_modules:       ${{ steps.matrices.outputs.e2e }}
+      is_merge_back:     ${{ steps.merge_back.outputs.value }}
```
sscarduzio: This job output has no consumer.

```
$ grep -rn "is_merge_back" .github/ ci/
.github/workflows/ci.yml:74:      is_merge_back:     ${{ steps.merge_back.outputs.value }}
```

One hit: the declaration. The `matrices` step reads `steps.merge_back.outputs.value` directly at line 118.

It is also a trap for whoever adds the first consumer. On any non-develop push the step is skipped, so the output is `''`, not `'false'` - a future `needs.setup.outputs.is_merge_back == 'false'` would be false everywhere it matters.

Delete the line, or emit `${{ steps.merge_back.outputs.value || 'false' }}` and use it.

.github/workflows/ci.yml
```diff
@@ -83,8 +84,38 @@ jobs:
           echo "is_develop=${{ github.ref == 'refs/heads/develop' }}" >> "$GITHUB_OUTPUT"
           # Treat nested */epic/* branches like top-level epic/* branches.
           case "$REF" in *epic/*)  echo "is_epic=true"  >> "$GITHUB_OUTPUT";; *) echo "is_epic=false"  >> "$GITHUB_OUTPUT";; esac
+      # A PR merged into master is pushed to master, and the merge-back into develop is pushed a
```
sscarduzio: The comment bills work this change does not save, and one of its numbers is about to be wrong.

It says both runs take "34 Linux legs, 6 Windows legs and 3 e2e legs each". The `matrices` step gains a merge-back arm for `linux` (line 136) and `windows` (line 147) only. The `e2e` branch is untouched and still emits `E2E_FULL_SET` on a merge-back, so e2e does not shrink.

Separately, `ci/github-hosted-runners` (#1368) reduces `LINUX_IT_FULL_SET` to 29 modules, which makes "34" wrong once either lands.

I also need to correct my own PR description: it claims this conflicts with #1368. It does not.

```
$ git merge-tree --write-tree --messages origin/ci/github-hosted-runners pr-1371-review
exit=0, 0 CONFLICT lines
Auto-merging .github/workflows/ci.yml
```

They touch different hunks of the same file and merge cleanly. Fix the comment to drop "and 3 e2e legs" (or add the e2e arm), and stop quoting a leg count that another open PR changes. `ci/CI.md:20` also still says "full 34 on develop/master/epic".

.github/workflows/ci.yml
```diff
@@ -83,8 +84,38 @@ jobs:
...
+      # gets, so develop keeps an independent signal and its release gate still needs every job to
+      # pass. Waiting for master's run instead would couple the two runs and buy little: master is
+      # covering the full matrix at that moment anyway.
+      - id: merge_back
+        if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          MASTER_SHA=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/heads/master" -q .object.sha)
+          MASTER_TREE=$(gh api "repos/$GITHUB_REPOSITORY/git/commits/$MASTER_SHA" -q .tree.sha)
```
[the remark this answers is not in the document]
sscarduzio: Fixed in 0567c5073. Both `gh api` calls take `|| true`, and an empty `MASTER_TREE` never equals `HERE_TREE`, so a failed lookup lands on the full matrix. I reproduced the abort first under GitHub's exact flags (`bash --noprofile --norc -e -o pipefail`) to confirm it was the non-zero-exit path and not the empty-output path. Thanks for catching it.

.github/workflows/ci.yml
sscarduzio: Fixed in 0567c5073. The step now reads `gradle.properties` the way `determine_ci_type` does: no `-pre` in `pluginVersion` means this run can tag and release, so it keeps the full matrix and gives up the saving. A `-pre` version still takes the reduced set and its pre-build upload does go out on that; the comment now says so plainly instead of claiming the gate is unaffected. Verified all four paths under the workflow's own shell flags.

.github/workflows/ci.yml
sscarduzio: Fixed in 0567c5073, together with @coderabbitai's thread on the line above.

.github/workflows/ci.yml
sscarduzio: Fixed in 0567c5073 - the output is gone.

.github/workflows/ci.yml
sscarduzio: Fixed in 0567c5073. The leg count is gone from the comment (#1368 changes it) and "3 e2e legs" with it, because e2e is not reduced. I also corrected the PR description: it claimed a conflict with #1368 that does not exist - `git merge-tree --write-tree` is clean.

sscarduzio:

.github/workflows/ci.yml
```diff
@@ -83,8 +83,56 @@ jobs:
...
+          # empty MASTER_TREE never equals HERE_TREE, so a failed lookup lands on the full matrix,
+          # which is the safe side.
+          MASTER_SHA=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/heads/master" -q .object.sha || true)
+          MASTER_TREE=$(gh api "repos/$GITHUB_REPOSITORY/git/commits/${MASTER_SHA:-missing}" -q .tree.sha || true)
+          HERE_TREE=$(git rev-parse "HEAD^{tree}")
+          echo ">>> develop tree $HERE_TREE, master tree ${MASTER_TREE:-<lookup failed>}"
+
+          # A develop push is also the run that publishes. determine_ci_type reads gradle.properties
+          # the same way: no `-pre` in pluginVersion means release_ror runs, tags the repo and pushes
+          # to S3, Docker Hub and Maven Central. That artifact must not be signed off by 10 of 34 ES
```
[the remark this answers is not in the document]
sscarduzio: Right - I warned about this 26 lines above and then did it anyway. Fixed in 345731048: it now says "the reduced PR set" and lets the set define itself.

.github/workflows/ci.yml
```diff
@@ -83,8 +83,56 @@ jobs:
...
+          # from either call would abort setup and fail the whole run, release path included. An
+          # empty MASTER_TREE never equals HERE_TREE, so a failed lookup lands on the full matrix,
+          # which is the safe side.
+          MASTER_SHA=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/heads/master" -q .object.sha || true)
+          MASTER_TREE=$(gh api "repos/$GITHUB_REPOSITORY/git/commits/${MASTER_SHA:-missing}" -q .tree.sha || true)
+          HERE_TREE=$(git rev-parse "HEAD^{tree}")
+          echo ">>> develop tree $HERE_TREE, master tree ${MASTER_TREE:-<lookup failed>}"
+
+          # A develop push is also the run that publishes. determine_ci_type reads gradle.properties
+          # the same way: no `-pre` in pluginVersion means release_ror runs, tags the repo and pushes
```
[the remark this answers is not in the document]
sscarduzio: Correct. `release_ror` uploads to S3, pushes the Docker image and tags the repo; Maven Central is `publish_mvn`, gated on `is_master`, so it never runs on a develop push.

Fixed in 345731048, and I named the master-side publish explicitly so the omission does not read as an oversight.

.github/workflows/ci.yml
```diff
@@ -83,8 +83,57 @@ jobs:
...
+      # This does not skip the run. It gives the merge-back the same reduced matrix a ready PR
+      # gets, so develop keeps an independent signal and its release gate still needs every job to
+      # pass. Waiting for master's run instead would couple the two runs and buy little: master is
+      # covering the full matrix at that moment anyway.
+      - id: merge_back
+        if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # `|| true` on both: the step runs under `bash -e -o pipefail`, so a rate limit or a 502
```
coutoPL: We have a `ci-lib.sh` script. WDYT about putting the "merge back" detection code in a new function in the script?

.github/workflows/ci.yml
```diff
@@ -83,8 +83,57 @@ jobs:
...
+      #
+      # Detect the merge-back by comparing trees, not commit ids: the two commits are different
+      # objects (master gets the squash, develop gets a merge commit) but they carry the same tree.
+      # A develop push whose tree already IS master's tip adds no code of its own.
+      #
+      # This does not skip the run. It gives the merge-back the same reduced matrix a ready PR
+      # gets, so develop keeps an independent signal and its release gate still needs every job to
+      # pass. Waiting for master's run instead would couple the two runs and buy little: master is
+      # covering the full matrix at that moment anyway.
+      - id: merge_back
```
coutoPL: Why the new step? It looks like the "flags" step is the right place for this.

sscarduzio:

What changed in each:

|---|---|---|

.github/workflows/ci.yml
```diff
@@ -110,6 +161,8 @@ jobs:
 
           if [ "$IS_PR_DRAFT" = "true" ]; then
             echo "windows=$WIN_IT_PR_DRAFT_SET" >> "$GITHUB_OUTPUT"
+          elif [ "${IS_MERGE_BACK:-false}" = "true" ]; then
```
coutoPL: Can we put the fallback close to the  `IS_MERGE_BACK` env definition? Like this:
```
IS_MERGE_BACK: ${{ steps.merge_back.outputs.value || false }}
```

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: Done. `is_merge_back_of_master()` now lives in `ci/ci-lib.sh`, and it owns all three rules that were spread across the step: the tree comparison, the release exception, and the fail-safe on a failed lookup.

The workflow is now:

```bash
IS_MERGE_BACK=false
if [ "$IS_DEVELOP_PUSH" = "true" ]; then
  source ci/ci-lib.sh
  if is_merge_back_of_master; then IS_MERGE_BACK=true; fi
fi
```

I tested the function against the live repository, five cases: different tree, equal tree with a `-pre` version, equal tree with a release version, a failed `gh` lookup, and each of those under `bash -e -o pipefail`. It returns true only in the second case, and the step exits 0 in all five, so an unknown answer keeps the full matrix.

.github/workflows/ci.yml
sscarduzio: You are right. It computes a flag, so it belongs with the flags. Folded in.

The only thing the separate step gave me was the `if:` guard, and that is now one `IS_DEVELOP_PUSH` env test in the script. The result: one step fewer, one job output fewer, and the branch conditions all in one place.

.github/workflows/ci.yml
sscarduzio: Applied exactly as written:

```yaml
IS_MERGE_BACK: ${{ steps.flags.outputs.is_merge_back || false }}
```

The two `"${IS_MERGE_BACK:-false}"` reads in the body are now plain `"$IS_MERGE_BACK"`. The default had two homes and could disagree with itself; now it has one.

sscarduzio:
@coutoPL a question about the base branch, so I do not guess.

You moved #1370 to `master`. #1371 and #1375 are still on `develop`. Do you want the whole CI set on `master`?

If yes, I will move both. I checked what that costs first, because a plain rebase is the wrong tool here:

- Both branches sit on `develop`, so they carry `develop` history. #1370 had **13 commits, only 5 of them mine** — the rest were #1369, #1374, four `develop`←`master` merges, and two other people's PRs. Rebasing all of it onto `master` would have pushed that work into `master` as a side effect of a timeout change. I rebuilt the branch from `master` and cherry-picked only my five instead.
- GitHub reports both of these as conflicting with `master`, but that compares the whole branches, including the inherited history. I tested the real question in a scratch worktree, and the answer is better than it looks:

```
ci/skip-already-green-tree   5 own commits  -> all apply CLEAN onto master
ci/prebuild-tag-race         3 own commits  -> all apply CLEAN onto master
```

So both move with no conflict resolution and no `develop` work leaking across. Say the word and I will do them the same way I did #1370.

One thing to decide with it: #1371 keys the merge-back detection on a tree comparison against `master`. On a `master` base that logic is still correct, but it never fires there — only a `develop` push can be a merge-back. That is harmless, and worth knowing before you pick.

sscarduzio:
Thanks — `docs/dev/branching.md` in #1379 answers it. Case 4, "a pipeline, build or publishing change", covers this PR, so it is now on `master`.

Rebuilt from `origin/master` and cherry-picked only this branch's own commits, the same way as #1370. Nothing from `develop` came with it. `MERGEABLE`, CI running.

Two notes while the rule is fresh:

The caveat I raised above resolves itself under this rule. This PR compares the tree against `master` to spot a merge-back, so on a `master` base the check can never fire. That is fine now, because the doc says a `master` merge is followed by a merge back into `develop`, which is where the check does its work.

By the same rule, #1374 went to the wrong branch. It is a CVE-scan suppression, which is case 3, and it merged to `develop`. That is why `cve_check` still fails on `master` runs: the suppression is not there. Worth a small PR to `master`, unless you would rather it ride the next merge back.

.github/workflows/ci.yml
```diff
@@ -77,13 +77,26 @@ jobs:
       - id: flags
         env:
           REF: ${{ github.head_ref || github.ref_name }}
+          IS_DEVELOP_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/develop' }}
+          GH_TOKEN: ${{ github.token }}
         run: |
           echo "is_master=${{ github.ref == 'refs/heads/master' }}"   >> "$GITHUB_OUTPUT"
```
coutoPL, line 84: Let's go it like this:
```
          source ci/ci-lib.sh

          echo "is_master=${{ github.ref == 'refs/heads/master' }}"   >> "$GITHUB_OUTPUT"
          echo "is_develop=${{ github.ref == 'refs/heads/develop' }}" >> "$GITHUB_OUTPUT"

          # Treat nested */epic/* branches like top-level epic/* branches.
          case "$REF" in
            *epic/*) echo "is_epic=true"  >> "$GITHUB_OUTPUT";;
            *)       echo "is_epic=false" >> "$GITHUB_OUTPUT";;
          esac

          MERGE_BACK=no
          [ "$IS_DEVELOP_PUSH" = "true" ] && is_merge_back_of_master && MERGE_BACK=yes

          case "$MERGE_BACK" in
            yes) echo "is_merge_back=true"  >> "$GITHUB_OUTPUT";;
            *)   echo "is_merge_back=false" >> "$GITHUB_OUTPUT";;
          esac
```

Comments are not needed.

ci/ci-lib.sh
```diff
@@ -23,6 +23,41 @@ gradle_property() {
   printf '%s\n' "$value"
 }
 
+# True when this commit is the merge-back of master into develop.
```
coutoPL: The comment should describe only what the function does. Not the whole story about it's existance and usage. see `@docs/dev/coding-style.md`

ci/ci-lib.sh
```diff
@@ -23,6 +23,41 @@ gradle_property() {
...
+# False for a release. A develop push with no `-pre` in pluginVersion runs release_ror, which
+# uploads to S3, pushes the Docker image and tags the repo. That artifact must not be signed off by
+# a reduced matrix. A `-pre` version reaches only upload_pre_ror, and master is running the full
+# matrix on the identical tree at that moment.
+#
+# False when a lookup fails. The caller reduces the matrix on true, so an unknown answer has to mean
+# "keep the full matrix". Needs GH_TOKEN, GITHUB_REPOSITORY and a checkout.
+is_merge_back_of_master() {
+  local master_sha master_tree here_tree plugin_version
+  master_sha=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/heads/master" -q .object.sha 2>/dev/null) || return 1
```
coutoPL: IMO, the method should be gh-independent. We don't need GitHub to do that. Let's base it on Git only. No vendor lock-in.

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: Applied as written. The step sources `ci/ci-lib.sh` once, sets the flag with one AND-chain, and carries no comments.

Tested the chain under the runner's `bash -e -o pipefail`: a false `[ ... ]` or a false `is_merge_back_of_master` leaves `MERGE_BACK=no` and the step exits 0.

ci/ci-lib.sh
sscarduzio: Trimmed to what the function returns and what it needs:

    # True when HEAD carries the same tree as master's tip and pluginVersion is a -pre version.
    # Fetches master from origin; needs no token on a public repo. False on any failure.

The story is in the commit message.

ci/ci-lib.sh
sscarduzio: Agreed, and done. No `gh`, no token, no `GITHUB_REPOSITORY`:

    git fetch --quiet --no-tags --depth=1 origin master
    master_tree=$(git rev-parse --verify --quiet 'FETCH_HEAD^{tree}')
    here_tree=$(git rev-parse --verify --quiet 'HEAD^{tree}')

The checkout is `fetch-depth: 1`, so the fetch is what brings master in. On a public repo it needs no credentials, so `persist-credentials: false` stays.

Tested against origin: different tree -> no; master's tree with a `-pre` version -> yes; master's tree with a release version -> no.

.github/workflows/ci.yml
```diff
@@ -843,9 +863,7 @@ jobs:
       )
     runs-on: *self_hosted_runner
     container: *toolchains_container
-    # Generous on purpose: a killed leg leaves the release without that ES major's artifacts,
-    # and a self-hosted minute costs nothing.
-    timeout-minutes: 600
+    timeout-minutes: 180
```
coutoPL: Why has it changed?

sscarduzio: My mistake. #1370 merged after I took a copy of this file, and I copied the whole file back over `master` instead of applying only this PR's hunks. That reverted `timeout-minutes` to 180.

Restored to 600 with its comment. The diff against `master` now touches only the `setup` job.

coutoPL (approved):
LGTM

PR #1372, opened 2026-09-05
ci: take the ES pre-build off the self-hosted pool

sscarduzio:
## The problem

#1369 moved `publish-pre-builds.yml` to the self-hosted pool. That pool has two slots.

Other jobs wait for this workflow. Three paid `ubicloud-standard-8` legs in `readonlyrest_kbn` poll it, for up to 120 minutes. When release jobs hold both slots, the pre-build waits in the queue, and the paid legs wait with it. One run waited 82 minutes in the queue for 0.3 minutes of work.

## The change

- `publish-pre-builds.yml` runs on `ubuntu-latest`. The repo is public, so this costs nothing.
- Removes `AGENT_ISSELFHOSTED` and `ROR_SHARED_DOCKER_HOST` from this workflow. On a hosted VM they stop the disk reclaim, and the build runs out of disk.
- Documents the job-started hook that the self-hosted runners need, in `ci/self-hosted-runner.md`.

## The rule

A job that other paid jobs wait for must run on a hosted runner. `release_ror`, `upload_pre_ror` and `publish_mvn` have no waiters, so they stay self-hosted.

## Not in this PR

- No `concurrency` group. GitHub keeps one pending run per group and cancels the others. The pollers treat a cancelled run as a failure.
- The shared-tag race. #1375 fixes it.

## Summary by CodeRabbit

- **CI/Infrastructure**
  - The pre-build publishing workflow now runs on hosted Ubuntu runners with a shorter maximum runtime.
  - Publishing jobs no longer rely on self-hosted runner configuration or shared Docker host settings.
  - Workflow guidance and comments were streamlined to reflect the hosted runner environment.

- **Documentation**
  - Added setup guidance for preventing workspace permission issues on self-hosted runners.
  - Expanded security guidance for runner recovery permissions and updated troubleshooting instructions.

### DIFF
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -41,18 +41,26 @@ defaults:
 jobs:
   publish:
     name: Publishing
-    # Self-hosted: a long build-and-push job, IO-bound, one at a time. Needs a registered
-    # self-hosted runner — see ci/self-hosted-runner.md.
-    runs-on: [self-hosted, Linux, X64]
+    # NOT self-hosted, even though this is a long IO-bound build. Two paid jobs in other runs wait
+    # on this workflow: the e2e legs of readonlyrest_kbn's pipeline (ubicloud-standard-8) and this
+    # repo's own e2e_prepare. They hold their runner for as long as the pre-build takes to FINISH,
+    # queue time included. The self-hosted pool is two slots on one box, so a release or upload job
+    # occupying both puts this workflow behind them and the waiters pay for the queue:
+    #
+    #   created 17:23:41, started 18:45:19  ->  81.6 min queued, 0.3 min of work
+    #
+    # Measured over the runs on either side of #1369: 0.0-0.6 min queued on ubicloud, and 0.0, 0.0,
+    # 18.2, 19.4, 81.6 min queued self-hosted. The minutes this workflow itself costs are small - it
+    # is 1-13 min of work per run - so moving it back buys the waiters far more than it spends.
+    # readonlyrest_kbn's twin of this workflow runs on ubuntu-latest for the same reason.
+    #
+    # The rest of what #1369 moved stays self-hosted: nothing waits on release_ror, upload_pre_ror,
+    # publish_mvn or mirror-es-libs.
+    runs-on: ubuntu-latest
     timeout-minutes: 600
     env:
       GRADLE_USER_HOME: ${{ github.workspace }}/.gradle-home
       GRADLE_OPTS: '-Dorg.gradle.java.installations.auto-download=true'
-      AGENT_ISSELFHOSTED: '1'
-      # This job talks to the box's docker daemon directly, and that daemon also serves the
-      # readonlyrest_kbn runners. No host-wide `docker system prune -a`, no removing containers that
-      # are not ours. See ci/run-pipeline.sh and ci/publish-ror-plugins.sh.
-      ROR_SHARED_DOCKER_HOST: '1'
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 0, persist-credentials: false }
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -34,7 +34,7 @@ on:
 permissions:
   contents: read
 
-# NO concurrency group here, deliberately. See the note in ci/self-hosted-runner.md.
+# NO concurrency group here, deliberately. Why, and what that leaves open:
 #
 # The obvious guard would be `concurrency: {group: ..., cancel-in-progress: false}`, and it is a
 # trap. GitHub keeps ONE pending run per group and cancels any previously pending one, so a third
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -34,19 +34,21 @@ on:
 permissions:
   contents: read
 
-# The self-hosted pool this workflow used to run on had two slots, and that was the only thing
-# serialising it. On ubuntu-latest it is unbounded, and the publish is not safe to run twice at
-# once: publishEsRorPreBuildDockerImage pushes the SHARED tag <esVersion>-ror-<pluginVersion>, and
-# retag_dev_image then freezes "this commit's image" FROM that shared tag (ci/ci-lib.sh:207-230).
-# Interleave A-push, B-push, A-retag and <esVersion>-ror-<shaA> points at B's image, so an e2e leg
-# tests the wrong build while believing it is commit-pinned.
+# NO concurrency group here, deliberately. See the note in ci/self-hosted-runner.md.
 #
-# Keyed on the versions and the branch actually built, not on github.ref: two dispatches for the
-# same branch and the same ES versions are the pair that collides. cancel-in-progress is false -
-# the loser waits, because cancelling a half-finished push is what leaves the shared tag dangling.
-concurrency:
-  group: publish-pre-builds-${{ inputs.target_branch || github.ref }}-${{ inputs.es_versions }}
-  cancel-in-progress: false
+# The obvious guard would be `concurrency: {group: ..., cancel-in-progress: false}`, and it is a
+# trap. GitHub keeps ONE pending run per group and cancels any previously pending one, so a third
+# dispatch cancels the second. Two repos wait on this workflow by run id
+# (wait_for_es_prebuild_images), and its `*)` branch treats `cancelled` exactly like `failure`:
+# "its images will never be published", exit 6. A group would turn a queued dispatch into a hard
+# failure for a waiter that did nothing wrong - the failure mode this PR exists to remove.
+#
+# That leaves a real, PRE-EXISTING race: publishEsRorPreBuildDockerImage pushes the shared tag
+# <esVersion>-ror-<pluginVersion>, and retag_dev_image then freezes "this commit's image" from it
+# (ci/ci-lib.sh:207-230). Interleave A-push, B-push, A-retag and <esVersion>-ror-<shaA> points at
+# B's image. Two self-hosted slots already allowed that; hosted runners widen the window.
+# The fix belongs in ci-lib.sh - push the immutable source tag directly instead of pushing a shared
+# tag and copying it - not in a workflow-level queue that breaks the callers.
 
 defaults:
   run:
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -34,6 +34,20 @@ on:
 permissions:
   contents: read
 
+# The self-hosted pool this workflow used to run on had two slots, and that was the only thing
+# serialising it. On ubuntu-latest it is unbounded, and the publish is not safe to run twice at
+# once: publishEsRorPreBuildDockerImage pushes the SHARED tag <esVersion>-ror-<pluginVersion>, and
+# retag_dev_image then freezes "this commit's image" FROM that shared tag (ci/ci-lib.sh:207-230).
+# Interleave A-push, B-push, A-retag and <esVersion>-ror-<shaA> points at B's image, so an e2e leg
+# tests the wrong build while believing it is commit-pinned.
+#
+# Keyed on the versions and the branch actually built, not on github.ref: two dispatches for the
+# same branch and the same ES versions are the pair that collides. cancel-in-progress is false -
+# the loser waits, because cancelling a half-finished push is what leaves the shared tag dangling.
+concurrency:
+  group: publish-pre-builds-${{ inputs.target_branch || github.ref }}-${{ inputs.es_versions }}
+  cancel-in-progress: false
+
 defaults:
   run:
     shell: bash
@@ -57,7 +71,8 @@ jobs:
     # The rest of what #1369 moved stays self-hosted: nothing waits on release_ror, upload_pre_ror,
     # publish_mvn or mirror-es-libs.
     runs-on: ubuntu-latest
-    timeout-minutes: 600
+    # ubuntu-latest is capped at 6 h, so this is the real ceiling, not a choice.
+    timeout-minutes: 360
     env:
       GRADLE_USER_HOME: ${{ github.workspace }}/.gradle-home
       GRADLE_OPTS: '-Dorg.gradle.java.installations.auto-download=true'
@@ -112,9 +127,10 @@ jobs:
           echo ">>> Publishing pre-build docker images for: $BUILD_ROR_ES_VERSIONS"
           ci/run-pipeline.sh
 
-      # configure-docker.sh logs in with the Docker Hub RW token. On an ephemeral VM that credential
-      # died with the VM; here it would stay in the runner user's ~/.docker/config.json, where every
-      # later job on the box can read it.
+      # configure-docker.sh logs in with the Docker Hub RW token. The VM is ephemeral again, so the
+      # credential dies with it either way. The step stays: it costs nothing, and it is what keeps
+      # this job correct if it ever moves back to a shared host, where the token would otherwise sit
+      # in the runner user's ~/.docker/config.json for every later job on the box to read.
       - name: Log out of Docker Hub
         if: always()
         run: docker logout || true
--- a/.github/workflows/publish-pre-builds.yml
+++ b/.github/workflows/publish-pre-builds.yml
@@ -34,21 +34,16 @@ on:
 permissions:
   contents: read
 
-# NO concurrency group here, deliberately. Why, and what that leaves open:
+# Do NOT add a `concurrency` group to this workflow.
 #
-# The obvious guard would be `concurrency: {group: ..., cancel-in-progress: false}`, and it is a
-# trap. GitHub keeps ONE pending run per group and cancels any previously pending one, so a third
-# dispatch cancels the second. Two repos wait on this workflow by run id
-# (wait_for_es_prebuild_images), and its `*)` branch treats `cancelled` exactly like `failure`:
-# "its images will never be published", exit 6. A group would turn a queued dispatch into a hard
-# failure for a waiter that did nothing wrong - the failure mode this PR exists to remove.
+# GitHub keeps one pending run per group and cancels any run already pending. Two repos wait on
+# this workflow by run id (wait_for_es_prebuild_images), and that waiter treats `cancelled` the
+# same as `failure`. A group would therefore turn a queued dispatch into a hard failure for a
+# waiter that did nothing wrong.
 #
-# That leaves a real, PRE-EXISTING race: publishEsRorPreBuildDockerImage pushes the shared tag
-# <esVersion>-ror-<pluginVersion>, and retag_dev_image then freezes "this commit's image" from it
-# (ci/ci-lib.sh:207-230). Interleave A-push, B-push, A-retag and <esVersion>-ror-<shaA> points at
-# B's image. Two self-hosted slots already allowed that; hosted runners widen the window.
-# The fix belongs in ci-lib.sh - push the immutable source tag directly instead of pushing a shared
-# tag and copying it - not in a workflow-level queue that breaks the callers.
+# Concurrent runs that publish the same plugin version share the tag
+# <esVersion>-ror-<pluginVersion>. Keep that safe in ci/ci-lib.sh, by pushing each commit's
+# immutable tag in the build that produced it, not with a workflow-level queue.
 
 defaults:
   run:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-pre-builds.yml
```diff
@@ -41,18 +41,26 @@ defaults:
...
+    #   created 17:23:41, started 18:45:19  ->  81.6 min queued, 0.3 min of work
+    #
+    # Measured over the runs on either side of #1369: 0.0-0.6 min queued on ubicloud, and 0.0, 0.0,
+    # 18.2, 19.4, 81.6 min queued self-hosted. The minutes this workflow itself costs are small - it
+    # is 1-13 min of work per run - so moving it back buys the waiters far more than it spends.
+    # readonlyrest_kbn's twin of this workflow runs on ubuntu-latest for the same reason.
+    #
+    # The rest of what #1369 moved stays self-hosted: nothing waits on release_ror, upload_pre_ror,
+    # publish_mvn or mirror-es-libs.
+    runs-on: ubuntu-latest
```
sscarduzio, line 63: The comment this PR deletes said "a long build-and-push job, IO-bound, **one at a time**". The two self-hosted slots were what made that true. `ubuntu-latest` is unbounded, and this workflow has no `concurrency:` key:

```
$ grep -c concurrency .github/workflows/publish-pre-builds.yml
0
$ grep -rln concurrency .github/workflows/
ci.yml  build-toolchains-image.yml  pr-conventions.yml  actionstrings_gen.yml
```

The publish is not concurrency-safe. `ci/ci-lib.sh:207-222`:

```bash
if [ "$FORCE_REBUILD_NORM" != "true" ] && docker_image_exists "${ES_DEV_IMAGE_REPO}:${SOURCE_TAG}"; then
...
  ./gradlew publishEsRorPreBuildDockerImage ...   # pushes the SHARED mutable CANONICAL_TAG
  retag_dev_image "$CANONICAL_TAG" "$SOURCE_TAG"  # freezes "this commit's image" FROM that shared tag
```

`CANONICAL_TAG` is `<esVersion>-ror-<pluginVersion>`, identical across branches. Interleave A-push, B-push, A-retag and `<esVersion>-ror-<shaA>` points at B's image. The caller alias `<esVersion>-ror-<tag>` is then derived from it (`ci-lib.sh:224-230`), so an e2e leg tests the wrong build while believing it is commit-pinned.

The race is pre-existing, but it was bounded to 2 by the pool this PR leaves. The workflow is dispatched per non-master PR and per develop push by the e2e repo's `prepare-dev-images`, and again by `readonlyrest_kbn`.

Add a `concurrency:` group keyed on the built commit (not the ref) in this PR, or say why the race is acceptable unbounded.

.github/workflows/publish-pre-builds.yml
```diff
@@ -41,18 +41,26 @@ defaults:
...
     timeout-minutes: 600
```
sscarduzio: `timeout-minutes: 600` cannot take effect. Confirming @coderabbitai: GitHub-hosted jobs are capped at 6 hours, so the job dies at 360 regardless.

Leaving 600 reads as a deliberate ten-hour budget and hides the real bound from whoever next debugs a killed job.

Set it to 360, or drop the line and let the default apply.

ci/self-hosted-runner.md
```diff
@@ -97,3 +97,65 @@ gh api repos/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runners \
 - Shared host, so the release scripts must not sweep the whole Docker daemon. `ROR_SHARED_DOCKER_HOST=1`
```
sscarduzio, line 96: _Anchored at line 97; the code this is about is at `ci/self-hosted-runner.md:14`, which is outside the diff._

This file still lists the workflow this PR moves off the pool, and the PR edits this file.

```
line 14: | `publish-pre-builds.yml` | `publish` | manual, long build-and-push |
```

under the heading "Jobs that need a self-hosted runner". Line 97-99 also still says `ROR_SHARED_DOCKER_HOST=1` "is set on these jobs", which this PR makes false for this workflow.

The diff on this file is a pure append, so the table was never revisited. This is the one doc an operator uses to size the pool, and it now contradicts the workflow.

Delete the table row and scope the `ROR_SHARED_DOCKER_HOST` sentence to `ci.yml`.

ci/self-hosted-runner.md
```diff
@@ -97,3 +97,65 @@ gh api repos/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runners \
...
+EOS
+chmod 0755 /usr/local/sbin/reclaim-runner-workspace
+
+# The runner user has no sudo. Grant exactly this one command, nothing else.
+echo 'runner ALL=(root) NOPASSWD: /usr/local/sbin/reclaim-runner-workspace' \
+  > /etc/sudoers.d/50-runner-workspace
+chmod 0440 /etc/sudoers.d/50-runner-workspace
+visudo -c -q
+
+cat > /home/runner/actions-runner/job-started-hook.sh <<'EOS'
```
sscarduzio: Two flaws in the install block I added here.

**1. The sudoers file is validated after it is installed.** The order is `echo ... > /etc/sudoers.d/50-runner-workspace`, then `chmod`, then `visudo -c -q`. Under `set -e` a bad line leaves a broken drop-in in place and then aborts - and a broken file in `/etc/sudoers.d` can break `sudo` for everyone on that container.

Write to a temp file, run `visudo -c -q -f` on it, and only then move it into place.

**2. The `.env` line is appended unconditionally**, so re-running the block on a runner that already has the hook duplicates it:

```bash
echo 'ACTIONS_RUNNER_HOOK_JOB_STARTED=...' >> /home/runner/actions-runner/.env
```

The block is written to be re-runnable per `$N`, so guard it with `grep -q ... || echo ...`.

Separately, on the wording at the end of this section: "It adds no real privilege either way: `runner` is already in the `docker` group, which is root-equivalent on this box." That is accurate, but it is stated as a reason not to worry when it is the reason to worry. Every runner container shares a profile with `security.privileged: "true"` and no uid map, so container root is host root, and `gh-ror-kbn-*` runs pull-request jobs on the same host as the ES release secrets (`PGP_SECRET_KEY_B64`, `MAVEN_REPO_PASSWORD`, `DOCKER_HUB_RW_TOKEN`, the `ROR_S3_*` pair). Rewrite that sentence as an explicit risk acceptance naming the real boundary, so the next reader does not take "root-equivalent anyway" to mean "therefore fine".

sscarduzio (commented):
**COMMENT.** One finding needs a decision before merge: the change removes a serialization the workflow depends on and puts nothing in its place.

Verified as sound: the `readonlyrest_kbn` twin really does run on `ubuntu-latest`, so the precedent holds. Nothing `needs:` `release_ror`, `upload_pre_ror`, `publish_mvn` or `mirror-es-libs`, so leaving those self-hosted is right. Dropping both env vars is correct and necessary - `AGENT_ISSELFHOSTED` flips `ci/free-host-disk.sh:22` back to the hosted-toolchain branch and `ROR_SHARED_DOCKER_HOST` restores `docker system prune -fa` at `ci/run-pipeline.sh:276`, which is what an ephemeral VM wants. I executed the documented install block from `ci/self-hosted-runner.md` against a throwaway root: the nested heredocs parse under `bash -s`, the files land 0755 and 0440, and the sudoers line passes `visudo -c -q -f`.

Index:
1. `publish-pre-builds.yml:59` - the deleted comment's "one at a time" was load-bearing; there is no `concurrency:` key. **Decision needed before merge.**
2. `publish-pre-builds.yml:60` - `timeout-minutes: 600` cannot happen. Confirms @coderabbitai.
3. `ci/self-hosted-runner.md:14` - this file still lists the workflow the PR just moved.
4. `ci/self-hosted-runner.md` - two flaws in the install block I wrote.

Not in this diff: the PR title fails this repo's `conventions` check - `the title must start with the Jira key`.

.github/workflows/publish-pre-builds.yml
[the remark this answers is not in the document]
sscarduzio: Fixed in 2a11b8618 - `timeout-minutes: 360`, with a comment saying it is the platform ceiling rather than a choice. Thanks.

ci/self-hosted-runner.md
```diff
@@ -97,3 +97,65 @@ gh api repos/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runners \
...
   is set on these jobs and downgrades `docker system prune -af --volumes` to dangling layers and
   build cache only. Without it, a retry would kill the Kibana runners' in-flight ELK stacks.
+
+## Required: the job-started hook
+
+The `container:` jobs in `ci.yml` run as root, and on a machine that survives the job they leave
+the whole workspace owned by root. The runner service runs as `runner`, so the next job that is
+**not** a container job dies in `actions/checkout`:
+
+```
```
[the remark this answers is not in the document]
sscarduzio: Fixed in 2a11b8618 - the fence is `text` now. Thanks.

ci/self-hosted-runner.md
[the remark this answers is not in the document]
sscarduzio: Fixed in 2a11b8618, and applied to the live runners. The hook is now `/usr/local/sbin/job-started-hook.sh`, root-owned, with `.env` left in the runner directory pointing at it. Two more reasons beyond the requirement: `./svc.sh` self-update rewrites the runner directory, and a hook the `runner` user can edit is a foothold that survives between jobs. Good catch.

.github/workflows/publish-pre-builds.yml
sscarduzio: Fixed in 2a11b8618. Added a `concurrency:` group keyed on the branch and ES versions actually built, not on `github.ref`, with `cancel-in-progress: false` - cancelling a half-finished push is what would leave the shared tag dangling.

.github/workflows/publish-pre-builds.yml
sscarduzio: Fixed in 2a11b8618 - 360, matching @coderabbitai's thread.

ci/self-hosted-runner.md
sscarduzio: Fixed in 2a11b8618. The `publish-pre-builds.yml` row is gone from the table, and the `ROR_SHARED_DOCKER_HOST` sentence is scoped to the `ci.yml` jobs with a line saying why this workflow no longer sets it.

ci/self-hosted-runner.md
sscarduzio: Fixed in 2a11b8618. The sudoers drop-in is written to a temp file, validated with `visudo -c -q -f`, then installed - so a bad line cannot land. The `.env` append is guarded by `grep -q`. I also rewrote the privilege paragraph as an explicit risk acceptance naming the real boundary, instead of presenting it as a reason not to worry.

.github/workflows/publish-pre-builds.yml
```diff
@@ -34,25 +34,48 @@ on:
...
+# once: publishEsRorPreBuildDockerImage pushes the SHARED tag <esVersion>-ror-<pluginVersion>, and
+# retag_dev_image then freezes "this commit's image" FROM that shared tag (ci/ci-lib.sh:207-230).
+# Interleave A-push, B-push, A-retag and <esVersion>-ror-<shaA> points at B's image, so an e2e leg
+# tests the wrong build while believing it is commit-pinned.
+#
+# Keyed on the versions and the branch actually built, not on github.ref: two dispatches for the
+# same branch and the same ES versions are the pair that collides. cancel-in-progress is false -
+# the loser waits, because cancelling a half-finished push is what leaves the shared tag dangling.
+concurrency:
+  group: publish-pre-builds-${{ inputs.target_branch || github.ref }}-${{ inputs.es_versions }}
```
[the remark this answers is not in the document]
sscarduzio: You are right that the group and the checkout disagreed, and chasing it showed the group should not be there at all. Removed in e2f68059e.

Keying on the branch was wrong in a second way: the tag that actually collides, `<esVersion>-ror-<pluginVersion>`, does not carry the branch, so a branch key lets the cross-branch collision straight through.

The race is real and pre-existing - two self-hosted slots already allowed the interleaving. The fix belongs in `ci/ci-lib.sh`, pushing the immutable source tag directly instead of pushing a shared tag and copying it. The workflow now carries a comment saying that, so the next reader does not reach for a concurrency group again.

.github/workflows/publish-pre-builds.yml
```diff
@@ -34,25 +34,48 @@ on:
...
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
sscarduzio: Your premise is right and it is why the group is now gone (e2f68059e), but `queue: max` is not the fix: `concurrency` accepts only `group` and `cancel-in-progress`, so there is no queue input to set.

The consequence of the one-pending-run rule is worse than a lost queue slot here. Two repos wait on this workflow by run id, and `wait_for_es_prebuild_images` treats `cancelled` exactly like `failure` - its `*)` branch prints "its images will never be published" and returns 6. A third dispatch cancelling the second would turn a queued dispatch into a hard failure for a waiter that did nothing wrong, which is the failure mode this PR exists to remove. Thanks for pushing on it.

ci/self-hosted-runner.md
```diff
@@ -95,5 +94,85 @@ gh api repos/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runners \
...
+echo ">>> workspace ownership reclaimed for the runner user"
+EOS
+chown root:root /usr/local/sbin/job-started-hook.sh
+chmod 0755 /usr/local/sbin/job-started-hook.sh
+
+# .env itself must stay in the runner directory - that is where the runner reads it. Guarded, so
+# re-running this block on a runner that already has the hook does not add a second line.
+grep -q ACTIONS_RUNNER_HOOK_JOB_STARTED /home/runner/actions-runner/.env || \
+  echo 'ACTIONS_RUNNER_HOOK_JOB_STARTED=/usr/local/sbin/job-started-hook.sh' \
+    >> /home/runner/actions-runner/.env
```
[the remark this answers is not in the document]
sscarduzio: Valid, and a real hole in the doc. Fixed in e2f68059e.

A name-only `grep -q` passes on a runner whose variable still points at the old runner-owned hook and leaves that path in place. The block now deletes any existing assignment, appends the root-owned path, and removes the old file - which is what I actually ran on `gh-ror-es-1` and `gh-ror-es-2`. The doc was describing something weaker than the migration. Thanks.

.github/workflows/publish-pre-builds.yml
```diff
@@ -34,25 +34,50 @@ on:
 permissions:
   contents: read
 
+# NO concurrency group here, deliberately. See the note in ci/self-hosted-runner.md.
```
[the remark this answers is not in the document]
sscarduzio: Right - the note is inline immediately below, so the pointer sent readers to a file that never got it. Fixed in 37f076163: the line now reads "Why, and what that leaves open:" and the note follows.

I kept it inline rather than moving it to `ci/self-hosted-runner.md`, because it is about this workflow's concurrency and the shared-tag race in `ci/ci-lib.sh`, not about the runner host. Thanks.

sscarduzio:

sscarduzio:

What changed in each:

|---|---|---|

coutoPL (commented):
please challenge this change against:
https://github.com/sscarduzio/readonlyrest_kbn/pull/1030
https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1376

If still makes sense, these two above should be merged first. 
The target branch of the pipeline chages should be master. After the merge of each pipeline change, we should merge it back to the develop branch,

.github/workflows/publish-pre-builds.yml
```diff
@@ -34,25 +34,38 @@ on:
...
-    timeout-minutes: 600
+    # NOT self-hosted, even though this is a long IO-bound build. Other runs wait on this workflow:
+    # the e2e legs of readonlyrest_kbn's pipeline and this repo's own e2e_prepare. Each waiter holds
+    # its runner until this workflow FINISHES, queue time included, and those runners are paid. The
+    # self-hosted pool has few slots, so one release job can put this workflow behind it and the
+    # waiters pay for the whole queue. A hosted runner starts at once and costs nothing here.
+    #
+    # The rule: a workflow that another run polls does not belong in a small fixed pool. Jobs that
+    # nothing waits on - release_ror, upload_pre_ror, publish_mvn, mirror-es-libs - do.
+    runs-on: ubuntu-latest
```
coutoPL, line 63: I'm not sure if this is a good idea. 
The Free runner parallel sits are limited to 20.  This workflow is used by the ROR KBN pipeline (and in the ROR KBN repo, we have a corresponding workflow too). It should be done as fast as possible.

Consider leaving it in the self-hosted runner.
If we can't afford it, we should review all the jobs are done on the self-hosted runner first. And decide which of them are important. IMO, this one is pretty important, and we have jobs on the runner that are not so important for us.

sscarduzio: You are right, and my own measurement says so. I went and got the numbers rather than argue.

All 60 completed runs of this workflow, queue time and work time split by where the job landed:

    self-hosted      n=52   QUEUE med=0.1  max=81.6    WORK med=3.0  max=7.5
    ubicloud         n= 8   QUEUE med=0.4  max= 0.6    WORK med=4.3  max=4.8

Two things fall out, and both favour your position:

1. **Self-hosted is faster at the actual work** — 3.0 min against 4.3. Your "it should be done as fast as possible" argument is correct, and my PR would have made this workflow slower on every single run.
2. **Self-hosted normally queues less**, not more: 0.1 min against 0.4.

Self-hosted wins 51 times out of 52. The whole case for this PR rests on one outlier: a single 81.6-minute queue on 2026-09-04 17:23. I built a permanent change on one bad afternoon, and the PR description presented that as the normal shape. That was wrong of me.

I could not pin the cause of that one event — the run list does not reach back far enough to show what held both slots. It falls on the same day the two ES runners were wedged by root-owned workspaces, which is now fixed by the job-started hook in `ci/self-hosted-runner.md`, so at least one failure mode behind it is already gone.

**Closing this PR.** The pre-build stays on the self-hosted box.

Your third point is the one worth acting on, and I agree with it: the pool has 2 slots for this repo (`gh-ror-es-1`, `gh-ror-es-2`), and `release_ror` and `upload_pre_ror` sit on it with `max-parallel: 2`, so a release can occupy both. That is the real exposure — not the runner type. Two ways to remove it, and I would rather you pick:

- **Add a third ES runner.** `incus copy` from `gh-base`, about ten minutes. A release then cannot take every slot.
- **Rank the jobs, as you suggest.** Nothing waits on `release_ror`, `upload_pre_ror`, `publish_mvn` or `mirror-es-libs`; two repos wait on this pre-build. If something has to yield, it is those, not this.

Say which and I will do it.

sscarduzio:
Closing, per the measurement in the review thread.

Self-hosted beats hosted on this workflow 51 runs out of 52: work 3.0 min against 4.3, queue 0.1 min against 0.4. Moving it to `ubuntu-latest` would have made every run slower to remove one bad afternoon.

The real exposure is the 2-slot pool, not the runner type. `release_ror` and `upload_pre_ror` sit on the same two runners with `max-parallel: 2`, so a release can take both while two repos wait on this build. That is worth fixing — by adding a third ES runner, or by ranking the jobs, per @coutoPL's suggestion. Neither needs this PR.

The job-started hook from this branch is already installed on both runners and documented, so that part is not lost.

sscarduzio:
For the record, the same failure mode on the same pool, one day after this PR closed:

```
09:19:28  beshu-tech/readonlyrest-e2e-tests#123 dispatches the ES pre-build (run 34334047105)
09:19:28  the pre-build job queues: both runners hold release_ror legs (es8xx, es7xx) since 08:31
10:03:37  a slot frees, the pre-build starts
10:05:26  the e2e waiter gives up at its 45 min timeout          -> the whole 13-leg matrix is skipped
10:09:13  the pre-build finishes: 6 min of work, 44 min in the queue
```

Two more `ROR ES Publish Pre-Builds` runs this morning took 73 and 76 minutes end to end for the same reason, and `release_es6xx` is still queued behind them.

The measurement I posted above still holds: self-hosted is faster and normally queues less. But the tail is not one bad afternoon — it is every time a release runs, because `release_ror` takes both slots for one to two hours and two repos wait on this pre-build.

@coutoPL you offered two fixes and I asked you to pick: a third ES runner on the box, or ranking the jobs so the release cannot take every slot. Today's evidence says one of them is needed. Which one?

coutoPL:
@sscarduzio, the third runner is the simplest solution, I think. WDYT?

sscarduzio:
Done, with Simone's OK: `gh-ror-es-3` is registered and online, with the job-started hook and the shared-box marker from #1380. `release_ror` and `upload_pre_ror` keep `max-parallel: 2`, so a release can hold at most two of the three slots.

It worked within a minute of coming up:

```
11:01  gh-ror-es-3 takes "ROR ES pre-build run-34340903697-1" while es-1 and es-2 still hold release_es8xx and release_es7xx
```

Three more pre-builds and `release_es6xx` were queued behind those two; they now drain through the third slot instead of waiting for the release to end. The runbook (`ci/self-hosted-runner.md`, in #1380) records the rule.

PR #1373, opened 2026-09-05
ci: repeat the Sonatype staging-repository request

sscarduzio:
## The problem

Three of the last four real Maven publishes failed with HTTP 401 from `ossrh-staging-api`. The fourth publish used the same secrets and passed. So the 401 is a temporary service error.

## The change

The retry repeats only `initializeSonatypeStagingRepository`. This is the first task of the chain, and nothing is uploaded before it. Later failures are not repeated, because a second attempt would create a second staging repository.

Four attempts, 30 s between them, doubling. A dead credential still fails after four attempts.

## Verification

The predicate matches the three real 401 logs. It does not match the clean publish, a 403, or the same message from a later task.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when publishing packages by retrying eligible Sonatype staging initialization failures.
  * Retries are limited to authorization and rate-limit errors, with up to four attempts and increasing delays.
  * Prevented automated publishing runs from waiting for interactive input.

### DIFF
--- a/ci/ci-lib.sh
+++ b/ci/ci-lib.sh
@@ -133,9 +133,15 @@ is_docker_registry_error() {
 }
 
 # A --retry-if function for `./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository`.
-# True only when the run died in initializeSonatypeStagingRepository, which is the FIRST task of
-# that chain: it asks ossrh-staging-api for a staging repository and nothing has been uploaded yet.
-# Repeating from there cannot leave a second staging repository behind or half-release a version.
+# True only when the run died in initializeSonatypeStagingRepository's create-staging-repository
+# call AND the status is a transient one. That task is the FIRST of the chain: it asks
+# ossrh-staging-api for a staging repository and nothing has been uploaded yet, so repeating from
+# there cannot leave a second staging repository behind or half-release a version.
+#
+# Not every failure of that task is repeated, deliberately. `No response body` and the profile
+# lookup ("Failed to find staging profile for package group") do not match the second grep, so they
+# fail on the first attempt - a wrong MAVEN_STAGING_PROFILE_ID answers the same way every time, and
+# waiting three more times to say so helps nobody.
 #
 # Every later failure - an upload, the close, the release - is deliberately not repeated. Those run
 # against a staging repository that already exists, so a second attempt would open another one.
--- a/ci/ci-lib.sh
+++ b/ci/ci-lib.sh
@@ -132,6 +132,22 @@ is_docker_registry_error() {
     "$1"
 }
 
+# A --retry-if function for `./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository`.
+# True only when the run died in initializeSonatypeStagingRepository, which is the FIRST task of
+# that chain: it asks ossrh-staging-api for a staging repository and nothing has been uploaded yet.
+# Repeating from there cannot leave a second staging repository behind or half-release a version.
+#
+# Every later failure - an upload, the close, the release - is deliberately not repeated. Those run
+# against a staging repository that already exists, so a second attempt would open another one.
+#
+# 401 is in the list because it is measured, not assumed: the same three secrets produced three
+# 401s (runs 33210133390, 33724419080, 33725765043) and one clean publish (33854480199). An
+# expired credential answers the same way every time, so a retry still fails and still reports it.
+is_sonatype_staging_init_error() {
+  grep -Fq "Execution failed for task ':initializeSonatypeStagingRepository'" "$1" &&
+    grep -Eq 'Failed to create staging repository.*status code (401|408|429|5[0-9][0-9])' "$1"
+}
+
 # Force-remove every container belonging to THIS CI job, scoped by the ror.ci-job=$ROR_CI_JOB_ID label so we
 # never touch a sibling CI job sharing the self-hosted Docker daemon. Single source of truth for "kill
 # this CI job's containers" — used by run-pipeline.sh's SIGTERM trap, the pipeline's always() cleanup
--- a/ci/ci-lib.sh
+++ b/ci/ci-lib.sh
@@ -146,9 +146,9 @@ is_docker_registry_error() {
 # Every later failure - an upload, the close, the release - is deliberately not repeated. Those run
 # against a staging repository that already exists, so a second attempt would open another one.
 #
-# 401 is in the list because it is measured, not assumed: the same three secrets produced three
-# 401s (runs 33210133390, 33724419080, 33725765043) and one clean publish (33854480199). An
-# expired credential answers the same way every time, so a retry still fails and still reports it.
+# 401 is in the list. ossrh-staging-api answers 401 intermittently on credentials that work
+# minutes later, so a 401 here does not prove the secret is wrong. An expired credential answers
+# the same way every time, so a repeat still fails and still reports it.
 is_sonatype_staging_init_error() {
   grep -Fq "Execution failed for task ':initializeSonatypeStagingRepository'" "$1" &&
     grep -Eq 'Failed to create staging repository.*status code (401|408|429|5[0-9][0-9])' "$1"
--- a/ci/ci-lib.sh
+++ b/ci/ci-lib.sh
@@ -149,9 +149,16 @@ is_docker_registry_error() {
 # 401 is in the list. ossrh-staging-api answers 401 intermittently on credentials that work
 # minutes later, so a 401 here does not prove the secret is wrong. An expired credential answers
 # the same way every time, so a repeat still fails and still reports it.
+#
+# ONLY the codes that prove the server made nothing. 401 and 429 are refusals: the request is
+# rejected before a staging repository exists, so a repeat cannot open a second one. A timeout
+# (408) or a server error (5xx) can arrive AFTER the POST took effect, and the plugin retries only
+# transitions, never the create - see AbstractTransitionNexusStagingRepositoryTask. Repeating those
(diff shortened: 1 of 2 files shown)


sscarduzio (commented):
**COMMENT.** Nothing here blocks. One consistency fix and two comment corrections.

Verified as sound, and I tried hard to break the guard and could not. The literal in the predicate matches the plugin: `NexusClient.createStagingRepository` throws through `failure("create staging repository", response)`, formatted as `Failed to $action, server at $baseUrl responded with status code ${response.code()}` - one line, both substrings, right order. `initializeSonatypeStagingRepository` really is the first task of the chain, and its other failure modes (`No response body`, profile lookup) do not match the regex, so they still fail fast. `ci-lib.sh` is sourced unconditionally at `run-pipeline.sh:3`, so `retry_with_backoff` is in scope. The prefix-assignment-in-a-subshell propagates the exit status correctly under `#!/bin/bash -ex`, and `ROR_RETRY_ATTEMPTS=4` produces exactly 4 invocations - measured against a stub. Output is `tee`'d, so the gradle log still reaches the CI log.

Index:
1. `run-pipeline.sh:258` - this call site drops the `</dev/null` both twins carry.
2. `run-pipeline.sh:253` - nit: two numbers in the comment.

Not in this diff: the PR title fails this repo's `conventions` check - `the title must start with the Jira key`.

ci/run-pipeline.sh
```diff
@@ -248,7 +248,15 @@ if [[ $ROR_TASK == "publish_maven_artifacts" ]]; then
...
       echo ">>> Publishing audit module artifacts to sonatype repo"
-      ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository
+      # ossrh-staging-api answers the staging-repository request with 401 intermittently, on
+      # credentials that work minutes later: 3 of the last 4 real publishes died that way, each
+      # after 19 s. See is_sonatype_staging_init_error for why only that first task is repeated.
+      # Longer gaps than the default 15 s: this is a release, and the service needs time to settle.
+      (
+        export ROR_RETRY_ATTEMPTS=4 ROR_RETRY_DELAY_SECONDS=30
+        retry_with_backoff --retry-if is_sonatype_staging_init_error \
+          ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository
```
sscarduzio: This call site drops the `</dev/null` that both existing `retry_with_backoff --retry-if` sites carry.

```
ci-lib.sh:229            ./gradlew publishEsRorPreBuildDockerImage "-PesVersion=$ES_VERSION" </dev/null; then
publish-ror-plugins.sh:138  ./gradlew ":${module}:pushRorDockerImage" ... "-PreusePackagedZip" </dev/null; then
run-pipeline.sh:258      ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository
```

Two of three. Under `--retry-if` the command runs inside `"$@" 2>&1 | tee "$log"` and inherits the job's stdin, which is the classic way a retried command's second attempt behaves differently from its first - gradle reads stdin for prompts.

Append `</dev/null`, matching the twins, or add one line saying why this one differs.

ci/run-pipeline.sh
```diff
@@ -248,7 +248,15 @@ if [[ $ROR_TASK == "publish_maven_artifacts" ]]; then
       echo ">>> Skipping publishing audit module artifacts"
     else
       echo ">>> Publishing audit module artifacts to sonatype repo"
-      ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository
+      # ossrh-staging-api answers the staging-repository request with 401 intermittently, on
+      # credentials that work minutes later: 3 of the last 4 real publishes died that way, each
+      # after 19 s. See is_sonatype_staging_init_error for why only that first task is repeated.
```
sscarduzio: nit: two numbers in this comment are off.

**"each after 19 s"** - one of the three was 20 s:

```
33210133390: BUILD FAILED in 20s
33724419080: BUILD FAILED in 19s
33725765043: BUILD FAILED in 19s
```

**"Longer gaps than the default 15 s"** is true but understates it. The delay doubles (`ci-lib.sh:121`), so `ROR_RETRY_DELAY_SECONDS=30` gives 30 / 60 / 120 s - about 3.5 minutes of waiting across 4 attempts, not 30 s. That is the number a reader needs when judging whether this belongs in a release job.

Say "19-20 s" and "30 s doubling, so about 3.5 min across 4 attempts".

ci/run-pipeline.sh
sscarduzio: Fixed in ee9a7ece2 - `</dev/null` added, matching `ci-lib.sh:229` and `publish-ror-plugins.sh:138`.

ci/run-pipeline.sh
sscarduzio: Fixed in ee9a7ece2 - "19-20 s", and the doubling spelled out as 30 / 60 / 120 s, about 3.5 min across 4 attempts.

ci/ci-lib.sh
```diff
@@ -132,6 +132,22 @@ is_docker_registry_error() {
     "$1"
 }
 
+# A --retry-if function for `./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository`.
+# True only when the run died in initializeSonatypeStagingRepository, which is the FIRST task of
```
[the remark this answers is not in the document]
sscarduzio: Right - the header was wider than the code. Fixed: it now says the run must have died in the create-staging-repository call **and** carry a transient status, and it names the two init failures that are deliberately not repeated (`No response body`, and the profile lookup) with the reason - a wrong `MAVEN_STAGING_PROFILE_ID` answers the same way every time, so failing on the first attempt is the point.

Re-ran the predicate against the six on-disk logs afterwards; behaviour is unchanged - RETRY on the three real 401s, no-retry on the clean publish and both controls.

ci/run-pipeline.sh
```diff
@@ -248,7 +248,18 @@ if [[ $ROR_TASK == "publish_maven_artifacts" ]]; then
       echo ">>> Skipping publishing audit module artifacts"
     else
       echo ">>> Publishing audit module artifacts to sonatype repo"
-      ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository
+      # ossrh-staging-api answers the staging-repository request with 401 intermittently, on
+      # credentials that work minutes later: 3 of the last 4 real publishes died that way, each
```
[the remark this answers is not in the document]
sscarduzio: Agreed, and my own cleanliness pass flagged the same duplication. Fixed: the call site no longer carries the ratio or the run ids. It keeps only what is specific to it - this is a release, so it can afford to wait, and 30 doubling gives about 3.5 min across 4 attempts - and points at `is_sonatype_staging_init_error` for the measurement.

sscarduzio:

sscarduzio:

What changed in each:

|---|---|---|

coutoPL (commented):
Are you sure they return 401 as a transient error? 
Are you sure the credentials didn't change? 

This is really weird. Maybe we can find proof of that (similar case reported by someone else) somewhere in the internet?

ci/ci-lib.sh
```diff
@@ -132,6 +132,28 @@ is_docker_registry_error() {
...
+#
+# Every later failure - an upload, the close, the release - is deliberately not repeated. Those run
+# against a staging repository that already exists, so a second attempt would open another one.
+#
+# 401 is in the list. ossrh-staging-api answers 401 intermittently on credentials that work
+# minutes later, so a 401 here does not prove the secret is wrong. An expired credential answers
+# the same way every time, so a repeat still fails and still reports it.
+is_sonatype_staging_init_error() {
+  grep -Fq "Execution failed for task ':initializeSonatypeStagingRepository'" "$1" &&
+    grep -Eq 'Failed to create staging repository.*status code (401|408|429|5[0-9][0-9])' "$1"
```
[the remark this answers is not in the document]
sscarduzio: Good catch, and you are right about the mechanism. I narrowed the predicate rather than removing the retry.

I checked the plugin first: `io.github.gradle-nexus.publish-plugin:2.0.0-rc-1` has a retrier, but it wraps transitions only — `AbstractTransitionNexusStagingRepositoryTask$transitionStagingRepo$retrier$1$1`. There is no retrier around the create. So nothing upstream protects a repeated POST, exactly as you say.

The list now holds only the codes that prove the server made nothing:

```
status 401 -> retry? YES     refusal, no repository exists yet
status 429 -> retry? YES     refusal, no repository exists yet
status 503 -> retry? no      can arrive AFTER the POST took effect
```

408 and 5xx are out. Either can land after the request took effect, which is the orphan case you describe.

This costs nothing in practice: 401 is the only status we have measured on this task, and it is the failure the PR was opened for. The comment now records why the list must stay short.

I did not go the other way you offered — find and reuse an existing staging repository. The plugin does expose `FindStagingRepository`, so it is possible, but it is a much larger change for a failure we have not observed.

sscarduzio:
No no I agree 401 is not something to retry on. Complete bullshit.

sscarduzio:
Closing. Retrying a 401 is wrong: 401 means the request was not authorized, and repeating an unauthorized request is not a fix. Retrying it also hides the real fault behind three more attempts and a delay.

For the record, so the underlying problem is not lost: `publish_mvn` failed on 3 of 5 master runs with `Failed to create staging repository ... status code 401` on the same secrets that published cleanly on the fourth. That still needs an explanation — a token close to expiry, a wrong `MAVEN_STAGING_PROFILE_ID`, or ossrh-staging-api rejecting a valid credential. The answer belongs at the credential, not in a retry loop.

No code from this branch is on develop.

PR #1375, opened 2026-09-05
ci: push the commit-identity image with the build that produced it

sscarduzio:
## The problem

`publishEsRorPreBuildDockerImage` pushed the shared tag `<esVersion>-ror-<pluginVersion>`. Then `ci-lib.sh` copied that tag to `<esVersion>-ror-<gitSha>`.

The shared tag has no branch in its name. Two publishes on the same plugin version can overwrite it between the push and the copy. Then the commit tag points to the other run's image. Every caller treats that tag as fixed to one commit.

## The change

Gradle pushes both tags in one `buildx` command. The build that makes the image writes the commit tag. The skip check now needs both tags before it skips a rebuild.

## Verification

Forced rebuild from this branch: run [33980510711](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/actions/runs/33980510711). Both tags resolve to the same manifests.

## Summary by CodeRabbit

- **Chores**
  - Prebuilt Docker image publishing now supports both shared version tags and immutable commit-based tags.
  - Images are published together, improving consistency during concurrent builds.
  - Build checks now recognize either the commit-specific or shared version image before deciding whether a rebuild is needed.
  - Removed the separate post-publish retagging step, reducing potential tag conflicts and simplifying image publication.

### DIFF
--- a/build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
+++ b/build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
@@ -36,6 +36,15 @@ def dockerImageNamespace = project.property('dockerImageNamespace')
 def releaseDockerImageLatest = "${dockerImageNamespace}/elasticsearch-readonlyrest:${esVersion}-ror-latest"
 def releaseDockerImageVersion = "${dockerImageNamespace}/elasticsearch-readonlyrest:${esVersion}-ror-${pluginVersion}"
 def preBuildDockerImageVersion = "${dockerImageNamespace}/elasticsearch-readonlyrest-dev:${esVersion}-ror-${pluginVersion}"
+// The commit-identity image, when the caller supplies one. ci/ci-lib.sh passes the full reference,
+// which it builds from the same dockerImageNamespace, so the two cannot drift.
+//
+// It is pushed by the SAME buildx invocation as the tag above, so "this commit's image" is written
+// by the build that produced it. It used to be copied from preBuildDockerImageVersion after the
+// push, and that was a race: <esVersion>-ror-<pluginVersion> carries no branch, so a concurrent
+// publish on the same plugin version could overwrite it in between, leaving the commit tag pointing
+// at the other run's image - which every caller then treats as commit-pinned.
+def preBuildDockerImageSource = project.hasProperty('preBuildSourceImage') ? project.property('preBuildSourceImage') : null
 
 // Shared docker CLI task setup + the multi-arch `buildx build ... --push` invocation (latest/version and dev
 // images differ only in their -t tags). Keeps the three buildx tasks below from repeating the platform list,
@@ -217,7 +226,11 @@ tasks.register('pushRorPreBuildDockerImage', Exec) {
     dependsOn prepareDockerBuildContext, prepareMultiplatformBuilder
     dockerCliDefaults(delegate)
     environment 'SOURCE_DATE_EPOCH', sourceDateEpoch
-    args = buildxPushArgs([preBuildDockerImageVersion])
+    // Both tags in one push. Without -PpreBuildSourceImage this is exactly what it was, so a local
+    // or manual invocation keeps working.
+    args = buildxPushArgs(preBuildDockerImageSource
+            ? [preBuildDockerImageVersion, preBuildDockerImageSource]
+            : [preBuildDockerImageVersion])
 }
 
 tasks.register('localRorDockerImage', Exec) {
--- a/build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
+++ b/build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
@@ -36,14 +36,6 @@ def dockerImageNamespace = project.property('dockerImageNamespace')
 def releaseDockerImageLatest = "${dockerImageNamespace}/elasticsearch-readonlyrest:${esVersion}-ror-latest"
 def releaseDockerImageVersion = "${dockerImageNamespace}/elasticsearch-readonlyrest:${esVersion}-ror-${pluginVersion}"
 def preBuildDockerImageVersion = "${dockerImageNamespace}/elasticsearch-readonlyrest-dev:${esVersion}-ror-${pluginVersion}"
-// The commit-identity image, when the caller supplies one. ci/ci-lib.sh passes the full reference,
-// which it builds from the same dockerImageNamespace, so the two cannot drift.
-//
-// It must be pushed by the SAME buildx invocation as the tag above. Callers treat this tag as
-// commit-pinned, and only one push can guarantee that. The shared tag
-// <esVersion>-ror-<pluginVersion> carries no branch, so any route that copies this tag from the
-// shared one lets a concurrent publish on the same plugin version supply the image instead.
-def preBuildDockerImageSource = project.hasProperty('preBuildSourceImage') ? project.property('preBuildSourceImage') : null
 
 // Shared docker CLI task setup + the multi-arch `buildx build ... --push` invocation (latest/version and dev
 // images differ only in their -t tags). Keeps the three buildx tasks below from repeating the platform list,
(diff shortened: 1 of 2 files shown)


ci/ci-lib.sh
```diff
@@ -209,16 +209,18 @@ publish_ror_es_prebuild_plugin() {
...
+    #
+    # Both tags go up in the SAME buildx push: the shared <esVersion>-ror-<pluginVersion> and this
+    # commit's immutable source tag. The source tag used to be copied from the shared one after the
+    # push, and that was a race - the shared tag carries no branch, so a concurrent publish on the
+    # same plugin version could overwrite it in between and leave the commit tag pointing at the
+    # other run's image, which every caller then treats as commit-pinned.
     if ! retry_with_backoff --retry-if is_docker_registry_error \
-         ./gradlew publishEsRorPreBuildDockerImage "-PesVersion=$ES_VERSION" </dev/null; then
+         ./gradlew publishEsRorPreBuildDockerImage "-PesVersion=$ES_VERSION" \
+         "-PpreBuildSourceImage=${ES_DEV_IMAGE_REPO}:${SOURCE_TAG}" </dev/null; then
```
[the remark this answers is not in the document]
sscarduzio: Correct, and it is a regression my change introduced. Fixed.

The old code did not need this check: it pushed the shared tag first and copied it to the source tag afterwards, so the source tag was the **last** write and its presence implied both existed. Pushing them together in one `buildx --push -t A -t B` removes that ordering - one build, two registry writes, either of which can fail alone - so the check now asks for both tags before it skips.

`SHARED_TAG` comes back for that check only; the shell names it, gradle builds it. Thanks for catching it.

sscarduzio:

What changed in each:

|---|---|---|

sscarduzio:
@coutoPL same base-branch question as on #1371 — do you want this one on `master` too?

This branch has 3 commits of its own, and I confirmed they cherry-pick onto `master` with no conflict. GitHub shows the PR as conflicting only because the branch carries inherited `develop` history, which should not go to `master` anyway.

I will move it the same way I moved #1370 if you want it there.

coutoPL:
@sscarduzio I've described it in the linked PR

sscarduzio:
Moved to `master` per `docs/dev/branching.md` case 4 (a pipeline, build or publishing change) — thanks for writing it down in #1379.

Rebuilt from `origin/master` with only this branch's own three commits, same as #1370. No `develop` history came across. `MERGEABLE`, CI running.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -36,6 +36,14 @@ def dockerImageNamespace = project.property('dockerImageNamespace')
 def releaseDockerImageLatest = "${dockerImageNamespace}/elasticsearch-readonlyrest:${esVersion}-ror-latest"
 def releaseDockerImageVersion = "${dockerImageNamespace}/elasticsearch-readonlyrest:${esVersion}-ror-${pluginVersion}"
 def preBuildDockerImageVersion = "${dockerImageNamespace}/elasticsearch-readonlyrest-dev:${esVersion}-ror-${pluginVersion}"
+// The commit-identity image, when the caller supplies one. ci/ci-lib.sh passes the full reference,
```
coutoPL: the comment doesn't follow the style-guide.md

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -36,6 +36,14 @@ def dockerImageNamespace = project.property('dockerImageNamespace')
...
+// which it builds from the same dockerImageNamespace, so the two cannot drift.
+//
+// It must be pushed by the SAME buildx invocation as the tag above. Callers treat this tag as
+// commit-pinned, and only one push can guarantee that. The shared tag
+// <esVersion>-ror-<pluginVersion> carries no branch, so any route that copies this tag from the
+// shared one lets a concurrent publish on the same plugin version supply the image instead.
+def preBuildDockerImageSource = project.hasProperty('preBuildSourceImage') ? project.property('preBuildSourceImage') : null
```
coutoPL: 1. It could be more generic. E.g. `additionalRorEsImageTag`
2. IMO, it'd be better add it to `def buildxPushArgs`

ci/ci-lib.sh
```diff
@@ -194,7 +194,9 @@ publish_ror_es_prebuild_plugin() {
   ROR_VERSION=$(gradle_property pluginVersion) || return 1
   GIT_SHA=$(git rev-parse --short HEAD)
 
-  local CANONICAL_TAG="${ES_VERSION}-ror-${ROR_VERSION}"
+  # Gradle pushes both of these in one buildx invocation. The shell still names the shared one for
+  # the skip check below - it does not build it.
+  local SHARED_TAG="${ES_VERSION}-ror-${ROR_VERSION}"
```
coutoPL: why do you think this name is better than `CANONICAL_TAG`?

ci/ci-lib.sh
```diff
@@ -204,21 +206,27 @@ publish_ror_es_prebuild_plugin() {
   local FORCE_REBUILD_NORM
   FORCE_REBUILD_NORM=$(echo "${FORCE_REBUILD:-false}" | tr '[:upper:]' '[:lower:]')
 
-  if [ "$FORCE_REBUILD_NORM" != "true" ] && docker_image_exists "${ES_DEV_IMAGE_REPO}:${SOURCE_TAG}"; then
-    echo ">>> Sources unchanged (image for this commit already published), skipping build"
+  # BOTH tags, not just the source one. `buildx --push -t A -t B` is one build but two registry
+  # writes, and it can publish one and fail on the other. Neither tag implies the other, so a check
+  # that asks for only one can skip the rebuild and leave the other stale or missing for good.
+  if [ "$FORCE_REBUILD_NORM" != "true" ] &&
+     docker_image_exists "${ES_DEV_IMAGE_REPO}:${SOURCE_TAG}" &&
```
coutoPL: are you sure we need these two?!

coutoPL (changes requested):

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
sscarduzio: The block is gone. The property moved into `buildxPushArgs` with a one-line comment (see the next thread).

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
sscarduzio: Both done. The property is `additionalImageTag`, and `buildxPushArgs` applies it:

    // -PadditionalImageTag adds one more tag to the same push.
    if (project.hasProperty('additionalImageTag')) cmd += ['-t', project.property('additionalImageTag')]

`pushRorPreBuildDockerImage` is back to `buildxPushArgs([preBuildDockerImageVersion])`. Any push task can now carry one more tag.

ci/ci-lib.sh
sscarduzio: It is not better, and it is gone. The shell no longer names that tag at all — gradle builds `<esVersion>-ror-<pluginVersion>` itself, and after the next thread the skip check does not read it either.

ci/ci-lib.sh
sscarduzio: No. Reverted to one check, on the commit tag.

I looked for a reader of the `<esVersion>-ror-<pluginVersion>` dev tag before deciding: nothing in CI reads it. The e2e waiter probes `${REPO}:${ES}-ror-${RUN_TAG}` (`prebuild-images-lib.sh:109`), and the kbn scripts name no dev tag. The commit tag is what every consumer waits on, so it is the only one the skip check needs.

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
```diff
@@ -50,6 +50,8 @@ def buildxPushArgs = { List<String> tags ->
     def platforms = project.hasProperty('dockerPlatforms') ? project.property('dockerPlatforms') : 'linux/amd64,linux/arm64'
     def cmd = ['buildx', 'build', '--platform', platforms] + commonBuildArgs
     tags.each { cmd += ['-t', it] }
+    // -PadditionalImageTag adds one more tag to the same push.
```
coutoPL: This comment violates the coding-style.md rule: `The non-obvious: why this value, why this order, which constraint forbids the simpler form.`

coutoPL (changes requested):

build-base/src/main/groovy/readonlyrest.docker-image-conventions.gradle
sscarduzio: Right — the line said what the `if` below it says. Removed.

coutoPL (approved):
LGTM

PR #1380, opened 2026-09-09
ci: e2e legs on GitHub-hosted runners, and a disk reclaim that detects its runner

sscarduzio:
## What

1. The e2e legs run on `ubuntu-latest` again, and `max-parallel: 99` is gone.
2. `ci/free-host-disk.sh` detects the runner instead of being told. No job declares runner facts any more.

## Why

**1.** #1368 moved every test leg to GitHub-hosted runners. #1376 merged two hours later, rewrote the e2e stage, and carried the old `runs-on: ubicloud-standard-4` back on the new test-leg job. Not a git conflict — new lines. Measured on `master` on 8 September: 21 legs, 487 minutes on Ubicloud, about $36 a month. The repo is public, so `ubuntu-latest` costs nothing.

**2.** From @coutoPL's review: a wrong `runs-on` value must not switch the disk reclaim off in silence, and a script that depends on the runner must detect the runner.

| Runner | Detected by | Reclaim |
|---|---|---|
| shared self-hosted box | `/etc/ror-shared-docker-host`, written at provisioning | dangling layers and build cache only |
| GitHub-hosted VM | `RUNNER_ENVIRONMENT=github-hosted`, set by the runner | toolchains and the whole daemon, when the disk is tight |
| anything else | — | nothing |

Inside a `container:` job the host toolchains are not on the job's filesystem, but the bind-mounted docker socket controls the host daemon, so a throwaway container with the host root mounted deletes them, behind the same guard.

`ROR_SHARED_DOCKER_HOST` and `AGENT_ISSELFHOSTED` are removed from every workflow. `ci-lib.sh` gains `is_shared_docker_host`; `run-pipeline.sh` and `publish-ror-plugins.sh` use it.

## Runners

The pool has three ES runners now: `gh-ror-es-3` was added today with the job-started hook, and all three carry the marker. `release_ror` and `upload_pre_ror` keep `max-parallel: 2`, so a release never takes every slot and the pre-build two repos wait on always finds one. Today a release held both old slots for 2½ hours and a pre-build queued 44 minutes for 6 minutes of work.

## Verified

- `grep -rE 'ROR_SHARED_DOCKER_HOST|AGENT_ISSELFHOSTED' .github ci` returns nothing.
- Decision table with `docker`/`df`/`sudo` shimmed: unknown → skip; hosted, wide disk → skip; hosted, tight, container → docker-socket path; unreadable `df` → skip.
- Shared-box path run for real on `gh-ror-es-3`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -641,12 +641,12 @@ jobs:
       needs.e2e_build_es_images.result == 'success' &&
       needs.setup.outputs.e2e_modules != '[]' &&
       (github.event_name != 'workflow_dispatch' || needs.setup.outputs.action == 'run_e2e_tests')
-    runs-on: ubicloud-standard-4
+    # Not a container: job, so ci/free-host-disk.sh below can see the preinstalled toolchains.
+    runs-on: ubuntu-latest   # free for public repos (4cpu/16GB)
     timeout-minutes: 120
     name: e2e_${{ matrix.module }}
     strategy:
       fail-fast: false
-      max-parallel: 99
       matrix: ${{ fromJSON(needs.e2e_matrix.outputs.matrix) }}
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -641,7 +641,6 @@ jobs:
       needs.e2e_build_es_images.result == 'success' &&
       needs.setup.outputs.e2e_modules != '[]' &&
       (github.event_name != 'workflow_dispatch' || needs.setup.outputs.action == 'run_e2e_tests')
-    # No container: the disk-reclaim step below must see the host's preinstalled toolchains.
     runs-on: ubuntu-latest   # free for public repos (4cpu/16GB)
     timeout-minutes: 120
     name: e2e_${{ matrix.module }}
@@ -782,11 +781,6 @@ jobs:
     runs-on: &self_hosted_runner [self-hosted, Linux, X64]
     container: *toolchains_container
     timeout-minutes: 600
-    # The job container gets the HOST docker socket, and that daemon also serves the readonlyrest_kbn
-    # runners on the same box. Tell the pipeline not to use host-wide docker cleanups. See
-    # cleanup_docker_and_build in ci/publish-ror-plugins.sh.
-    env: &shared_docker_host
-      ROR_SHARED_DOCKER_HOST: '1'
     strategy:
       fail-fast: false
       # At most 2 legs at a time: two runners, ~4 GB each, and the box is shared with other repos.
@@ -846,7 +840,6 @@ jobs:
     # Generous on purpose: a killed leg leaves the release without that ES major's artifacts,
     # and a self-hosted minute costs nothing.
     timeout-minutes: 600
-    env: *shared_docker_host
     strategy:
       fail-fast: false
       # Same bound as upload_pre_ror — see the note there.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - id: flags
         env:
           REF: ${{ github.head_ref || github.ref_name }}
@@ -167,6 +169,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - name: Verify baked Gradle home
         run: |
           set -e
@@ -198,6 +202,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - name: ${{ matrix.ROR_TASK }}
         uses: ./.github/actions/cve-check
         with:
@@ -228,6 +234,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - name: ${{ matrix.ROR_TASK }}
         env:
           ROR_TASK: ${{ matrix.ROR_TASK }}
@@ -252,6 +260,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - name: core_tests
         env:
           ROR_TASK: core_tests
@@ -760,6 +770,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - id: release_check
         run: |
           IS_RELEASE=true
@@ -791,6 +803,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - name: ${{ matrix.ROR_TASK }}
         env:
           ROR_S3_ENDPOINT_URL: ${{ vars.ROR_S3_ENDPOINT_URL }}
@@ -848,6 +862,8 @@ jobs:
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
         with: { fetch-depth: 1, persist-credentials: true }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
       - name: ${{ matrix.ROR_TASK }}
         env:
           ROR_S3_ENDPOINT_URL: ${{ vars.ROR_S3_ENDPOINT_URL }}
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -641,12 +641,12 @@ jobs:
       needs.e2e_build_es_images.result == 'success' &&
       needs.setup.outputs.e2e_modules != '[]' &&
       (github.event_name != 'workflow_dispatch' || needs.setup.outputs.action == 'run_e2e_tests')
-    runs-on: ubicloud-standard-4
+    # Not a container: job, so ci/free-host-disk.sh below can see the preinstalled toolchains.
```
coutoPL: Are you sure the comment follow the coding-style.md?

sscarduzio: Not quite — it named `ci/free-host-disk.sh`, and the style doc says not to name other files. Now:

    # No container: the disk-reclaim step below must see the host's preinstalled toolchains.

The fact itself stays, because it is the non-obvious one: every other test job runs inside the toolchains container, and this one cannot, or the reclaim step two lines down finds nothing to reclaim.

.github/workflows/ci.yml
```diff
@@ -641,12 +641,12 @@ jobs:
       needs.e2e_build_es_images.result == 'success' &&
       needs.setup.outputs.e2e_modules != '[]' &&
       (github.event_name != 'workflow_dispatch' || needs.setup.outputs.action == 'run_e2e_tests')
-    runs-on: ubicloud-standard-4
+    # No container: the disk-reclaim step below must see the host's preinstalled toolchains.
```
coutoPL: This comment shows that `runs-on` and `container:` are coupled to `ci/free-host-disk.sh`. I do not
  want to keep that coupling in my head each time I change a runner. A wrong `runs-on` value would
  disable the reclaim silently.

  The script depends on the runner, so the script must detect the runner. It must not read env vars
  that we set by hand in the job. `ROR_SHARED_DOCKER_HOST` and `AGENT_ISSELFHOSTED` describe the
  machine, not the job:

  - `ROR_SHARED_DOCKER_HOST`: replace it with a marker file on the self-hosted box, for example
    `/etc/ror-shared-docker-host`, written when we provision the runner.
  - `AGENT_ISSELFHOSTED`: delete it. The runner sets `RUNNER_ENVIRONMENT`, the script already reads
    it, and the script is the only reader of `AGENT_ISSELFHOSTED` in this repo.

  The container case is also solvable. GitHub bind-mounts `/var/run/docker.sock` into a `container:`
  job, and that socket controls the host daemon. The script can therefore delete the host toolchains
  from inside a container:

      docker run --rm -v /:/host alpine sh -c 'rm -rf /host/usr/share/dotnet /host/opt/hostedtoolcache'

  Keep the `RUNNER_ENVIRONMENT=github-hosted` guard on that branch, because it removes system
  directories.

  Then `run: ci/free-host-disk.sh` goes into any job, with no env block and no comment, and this
  comment is not necessary.

sscarduzio: Done, folded into this PR as you described it.

- **Shared box**: `/etc/ror-shared-docker-host`, written when the runner is provisioned (runbook updated, and the file is on `gh-ror-es-1`, `-2` and the new `-3` now). `ci-lib.sh` exposes `is_shared_docker_host`, and the three prune sites — `free-host-disk.sh`, `run-pipeline.sh`, `publish-ror-plugins.sh` — read it. `ROR_SHARED_DOCKER_HOST` is gone from every workflow.
- **Hosted VM**: `RUNNER_ENVIRONMENT=github-hosted` only. `AGENT_ISSELFHOSTED` is gone; the script was its only reader.
- **Container job**: when the toolchain directories are not on this filesystem but the daemon is reachable, a throwaway `alpine` with the host root mounted deletes them, behind the same github-hosted guard.
- The e2e leg has no comment and no env block any more. `run: ci/free-host-disk.sh` goes into any job as is.

Tested the decision table with shims (unknown runner → skip; hosted with a wide disk → skip; hosted, tight, container → docker-socket path; unreadable `df` → skip), and the shared-box path for real on `gh-ror-es-3`.

.github/workflows/ci.yml
```diff
@@ -782,15 +781,10 @@ jobs:
...
-    # The job container gets the HOST docker socket, and that daemon also serves the readonlyrest_kbn
-    # runners on the same box. Tell the pipeline not to use host-wide docker cleanups. See
-    # cleanup_docker_and_build in ci/publish-ror-plugins.sh.
-    env: &shared_docker_host
-      ROR_SHARED_DOCKER_HOST: '1'
     strategy:
       fail-fast: false
-      # At most 2 legs at a time: two runners, ~4 GB each, and the box is shared with other repos.
-      # GitHub queues the rest — this bound is really set by how many self-hosted runners are registered.
+      # One below the number of self-hosted runners, so a release never takes every slot: the ES
```
coutoPL: we don't need this comment, right?

.github/workflows/ci.yml
coutoPL, line 1: Do you confirm we have `ci/free-host-disk.sh` called in each job?

coutoPL: the same question for other workflows

ci/ci-lib.sh
```diff
@@ -5,6 +5,10 @@
 # lib is sourced rather than run (`source ci/ci-lib.sh && reap_ci_job_containers` resolved it to ".").
 CI_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 
+# True on a box whose Docker daemon and system directories other runners share. The runner
+# provisioning writes the marker (ci/self-hosted-runner.md); no job sets it.
+is_shared_docker_host() { [ -e /etc/ror-shared-docker-host ]; }
```
coutoPL: Make sure we use this function instead of checking the `/etc/ror-shared-docker-host` file

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: Right. Both `max-parallel: 2` comments are gone; `ci/self-hosted-runner.md` carries the rule.

.github/workflows/ci.yml
sscarduzio: Checked every job. It is called where a job builds or pulls ES images, and nowhere else:

| Workflow | Job | Calls it | Why |
|---|---|---|---|
| ci.yml | `it_linux`, e2e legs | yes | ES images and data dirs |
| ci.yml | `build_ror` | **yes, added in this push** | builds four ES images on a 4-vCPU VM and had no reclaim |
| ci.yml | `upload_pre_ror`, `release_ror`, `publish_mvn` | no | self-hosted; `run-pipeline.sh` reclaims between legs through the same `is_shared_docker_host` |
| ci.yml | `setup`, checks, unit tests, `it_windows` | no | no docker work (`it_windows` is not bash) |
| mirror-es-libs.yml | `mirror` | yes | |
| publish-pre-builds.yml | `publish` | yes | |
| build-toolchains-image.yml | | no | builds one image on `ubicloud-standard-4`; not touched here |

The script is safe in any job now, so nothing stops adding it elsewhere if a job starts to pull images.

ci/ci-lib.sh
sscarduzio: Done. `is_shared_docker_host` now lives in `ci/runner-detect.sh` (functions only), and `ci-lib.sh` and `free-host-disk.sh` both source it. No script tests the file path directly:

```
ci/runner-detect.sh:6:    is_shared_docker_host() { [ -e /etc/ror-shared-docker-host ]; }
ci/free-host-disk.sh:27:  if is_shared_docker_host; then
ci/run-pipeline.sh:276:   if is_shared_docker_host; then
ci/publish-ror-plugins.sh:9:  if is_shared_docker_host; then
```

The separate file is deliberate: sourcing `ci-lib.sh` runs `gradle_property dockerImageNamespace || exit 1` at top level, and a disk-reclaim script must not depend on parsing `gradle.properties`.

ci/free-host-disk.sh
```diff
@@ -1,36 +1,30 @@
...
 #
-#   shared self-hosted   the system directories belong to the box, so never touch them. Docker
-#                        garbage from earlier runs is ours to reclaim — but only the parts no
-#                        other repo's runner can be using: dangling layers and build cache, never
-#                        `-a`, never a container we did not start. Selected by ROR_SHARED_DOCKER_HOST=1.
-#   ephemeral hosted VM  reclaim both levers below, but only when the disk is actually tight.
-#                        The VM dies after the job anyway. Selected by RUNNER_ENVIRONMENT=github-hosted.
-#   anything else        reclaim nothing. A developer shell, `act`, or an unlabelled self-hosted
-#                        job must never lose its docker images to this script.
+#   shared self-hosted box   /etc/ror-shared-docker-host exists (written when the runner is
```
coutoPL: we write here about internal of the some script.

.github/workflows/ci.yml
coutoPL: we should put it in any job. Moreover, we should write in ci.md that it should be added to any new job. The script should recognize if the cleanup should be done or not. We should not add it depending on the runner. That's why we added it.

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: Agreed, and done. Every Linux job in `ci.yml` and in `build-toolchains-image.yml` now calls `ci/free-host-disk.sh` right after checkout — the light jobs and the self-hosted release jobs included. The script decides; no job does. `ci/CI.md` states the rule for new jobs:

> **Every job calls `ci/free-host-disk.sh` right after checkout**, including a new one. The script detects the runner and decides whether to reclaim disk; a job never decides that for it.

The only jobs without it are `it_windows` and `unit_tests_windows`: they run PowerShell on `windows-2025`, and there is no bash there to run the script.

ci/free-host-disk.sh
sscarduzio: Right. The header now says what this script does and nothing about how a runner is provisioned — that lives in `ci/self-hosted-runner.md`.

.github/workflows/ci.yml
sscarduzio: Correction to my table: my first pass missed `e2e_matrix`, `e2e_order_kbn_images` and `e2e_build_es_images` — their job keys carry a digit and my patch matched letters only. All three call the script now, and the last copy of the container comment in `e2e_build_es_images` is gone. Every Linux job with a checkout has the step; verified job by job.

coutoPL (approved):
LGTM

coutoPL:
@sscarduzio should we port this change to the ROR KBN repo? If so, please do that

PR #1381, opened 2026-09-10
ci: prove the Sonatype credentials before release day

sscarduzio:
## The problem

`publish_mvn` looks healthy on master and is not. It runs only when the run is a release, and inside it `ci/run-pipeline.sh` skips when `pluginVersion == publishedPluginVersion` and the artifacts already exist. Every green `publish_mvn` on master in the last two weeks took that skip:

| run | result | what the log says |
|---|---|---|
| 34358272805 | green | `>>> Skipping publishing audit module artifacts` |
| 34275922988 | green | `>>> Skipping publishing audit module artifacts` |
| 34126271209 | green | `>>> Skipping publishing audit module artifacts` |
| 33724419080 | **red** | `>>> Publishing audit module artifacts to sonatype repo` → 401 |
| 33725765043 | **red** | `>>> Publishing audit module artifacts to sonatype repo` → 401 |

Every attempt that was real failed. The last artifact that reached Central is 1.70.3.

## The 401 is not a flake

```
Execution failed for task ':initializeSonatypeStagingRepository'.
> Failed to create staging repository, server at
  https://ossrh-staging-api.central.sonatype.com/service/local/
  responded with status code 401, body:
```

The body is empty. That API answers exactly the same way to a request with no credentials and to one with wrong credentials:

```
$ curl -s -o /dev/null -w '%{http_code}\n' -X POST \
    https://ossrh-staging-api.central.sonatype.com/service/local/staging/profiles/deadbeef/start
401
$ curl -s -o /dev/null -w '%{http_code}\n' -u bogus:bogus -X POST \
    https://ossrh-staging-api.central.sonatype.com/service/local/staging/profiles/deadbeef/start
401
```

So the server rejected who we said we were. A retry sends the same identity and gets the same answer. That is why #1373 was the wrong fix.

## Why nobody has noticed

`MAVEN_REPO_USER` and `MAVEN_REPO_PASSWORD` were rotated on **4 September**, after the last failure. Nothing has exercised the new pair since: `gradle.properties` holds `pluginVersion=1.71.0-pre6` against `publishedPluginVersion=1.70.3`, so the job takes the version-mismatch branch and skips.

The next release is the first test of the new credentials.

## What this PR adds

`verify_sonatype_credentials` — one job, one HTTPS call, on every master push, on a free hosted runner.

- `GET /service/local/staging/profiles` with the publish credentials. A 200 proves the pair is accepted.
- Then it checks that `MAVEN_STAGING_PROFILE_ID` is a profile the account owns. `build.gradle` hands that id straight to the call that failed, so a stale id from the old OSSRH console fails the same way.
- On a 401 the message says where to fix it: the Doppler project `ror_ci`, config `prd`, not GitHub.
- It prints no credential and no response body.

Verified both failure paths locally:

```
$ MAVEN_REPO_USER= MAVEN_REPO_PASSWORD=x ci/verify-sonatype-credentials.sh
>>> ERROR: MAVEN_REPO_USER is empty.
$ MAVEN_REPO_USER=bogus MAVEN_REPO_PASSWORD=bogus ci/verify-sonatype-credentials.sh
>>> ERROR: The staging API rejected the credentials with HTTP 401.
```

The 200 path needs the real secrets, so the first master run after merge is what proves it. **That run is the point of the PR**: if it goes red, the credentials rotated on 4 September do not work either, and we learn it now instead of during a release.

## Why a clock, and not a push

The check first rode on pushes to `master`. It should not: over the 62 days GitHub still retains runs for, `master` took **14 pushes, and 12 of them fall inside one 13-day burst** of pipeline work.

```
$ gh run list --branch master --event push --workflow ci.yml --limit 200
2 2026-08-28   1 2026-09-01   2 2026-09-03   1 2026-09-04
1 2026-09-07   4 2026-09-08   2 2026-09-09   1 2026-09-10
```

Nothing between 10 July and 27 August. For 49 of those 62 days the check would not have run once — worse than no check, because it would have looked healthy by staying silent. It is a daily cron now, in its own workflow, which is also a stable thing for a bot to watch.

## Decision for the reviewer

The job **fails** the run rather than warning. That makes master red when the publish credential is dead. I think that is right — a warning is how we got here — but it is one line to change if you would rather it were `continue-on-error: true`.


.github/workflows/ci.yml
```diff
@@ -906,6 +906,30 @@ jobs:
           source ci/configure-docker.sh
           ci/run-pipeline.sh
 
+  # ── VERIFY_SONATYPE_CREDENTIALS (every master push) ───────────────────────
+  # publish_mvn runs only on a release, and it skips when the version is already published.
+  # A broken publish credential is therefore invisible until the run that must not fail.
+  # This job exercises the credential on every master push, for a few seconds of a free runner.
+  verify_sonatype_credentials:
```
coutoPL: I understand that the job responsibility is to let us know early that the credentials have been rotated. IMO, putting it in the ci.yml and triggering on push to the master branch is not a good choice. Because:

1. The master branch is not updated so often (the recent changes, when we work heavily on the pipeline improvements, are rather an exception, not a rule).
2. If my understanding is right, it means that this job should have a separate workflow.

So, let's create one. It should be triggered automatically, e.g., every day. 
The question is: how to notify the ROR team that it failed. Maybe we can use #ror-plugin-ci Slack's channel and ROROCOP v2 notifications (the bot could track the workflow). CC @ton77v

coutoPL (changes requested):

.github/workflows/ci.yml
sscarduzio: Agreed on both points, and the numbers are stronger than "not so often". Over the 62 days GitHub still retains runs for, `master` took **14 pushes, and 12 of them fall inside one 13-day burst** — 28 Aug to 10 Sep, which is exactly the pipeline work you call the exception:

```
$ gh run list --branch master --event push --workflow ci.yml --limit 200
2 2026-08-28   1 2026-09-01   2 2026-09-03   1 2026-09-04
1 2026-09-07   4 2026-09-08   2 2026-09-09   1 2026-09-10
```

Nothing between 10 July and 27 August. For 49 of those 62 days the job would not have run once, which is worse coverage than no check at all — it would have looked healthy by being silent.

Moved to its own workflow in 82d8c582: `.github/workflows/verify-publish-credentials.yml`, daily cron at 06:00 UTC plus `workflow_dispatch`. Removed from `ci.yml`.

## On the notification question

I did not want to leave a scheduled job that nobody watches, since that is the same silent-failure shape we are removing. So for now a failed run **opens one issue and reuses it** — it comments on the open one rather than opening a new issue every morning, which is how a daily job trains people to ignore it. That needs no new secret and works today.

It is deliberately one step and easy to remove:

```yaml
      - name: Report the failure where someone will see it
        if: failure()
        run: ci/report-credential-failure.sh
```

@ton77v — the workflow is now a separate, stable name (`Verify publish credentials`), so ROROCOP v2 can track it directly. If you wire it to #ror-plugin-ci, delete that step and the script with it; the comment above it says so.

One thing worth knowing while you decide, because it changes the urgency: `publish_mvn` has not made a real publish since 1.70.3. Every green run on master took the silent skip, and all three real attempts died on the same 401. `MAVEN_REPO_USER` and `MAVEN_REPO_PASSWORD` were rotated on 4 September, after the last of those, and nothing has exercised the new pair since. The first run of this workflow tells us whether the rotation fixed it.

.github/workflows/verify-publish-credentials.yml
```diff
@@ -0,0 +1,51 @@
+name: Verify publish credentials
+
+# The one thing a release cannot do without, checked on a day when nothing is at stake.
```
coutoPL: this comment violates coding style (see commens seting)

.github/workflows/verify-publish-credentials.yml
```diff
@@ -0,0 +1,51 @@
...
+  contents: read
+  issues: write
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  sonatype:
+    name: Sonatype
```
coutoPL: better name needed

.github/workflows/verify-publish-credentials.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest   # free for public repos; the job is one HTTPS call
```
coutoPL: dummy comment (see coding style)

.github/workflows/verify-publish-credentials.yml
```diff
@@ -0,0 +1,51 @@
...
+        with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
+      - name: verify_sonatype_credentials
+        env:
+          MAVEN_REPO_USER: ${{ secrets.MAVEN_REPO_USER }}
+          MAVEN_REPO_PASSWORD: ${{ secrets.MAVEN_REPO_PASSWORD }}
+          MAVEN_STAGING_PROFILE_ID: ${{ secrets.MAVEN_STAGING_PROFILE_ID }}
+        run: ci/verify-sonatype-credentials.sh
+      # A scheduled workflow that nobody watches is the failure mode this job exists to remove.
```
coutoPL: the same here

coutoPL (changes requested):

.github/workflows/verify-publish-credentials.yml
sscarduzio: Right — that header was the story of the change: which job skips, how often `master` is pushed, and how few days a push-triggered check would have covered. `docs/dev/code-style.md` puts that in the PR description, and it is there now. f7eaf773.

What stays says what the workflow proves, and why a clock rather than an event:

```yaml
# Proves that the Maven Central credentials still work. The publish job runs only for a release,
# and it skips when the version is already published, so a credential that stops working stays
# invisible until a release needs it. This asks the staging API on a clock instead. It publishes
# nothing.
```

The script header had the same fault, so it got the same treatment even though you did not flag it.

.github/workflows/verify-publish-credentials.yml
sscarduzio: Agreed — `sonatype` says where the job goes, not what it does.

It is `verify_sonatype_credentials` now, and the redundant `name:` is gone with it, so the check reads like the other jobs in this repo rather than carrying a second, different label. f7eaf773.

.github/workflows/verify-publish-credentials.yml
sscarduzio: Fixed in f7eaf773. `# free for public repos; the job is one HTTPS call` repeated what the line above already said, and `ubuntu-latest` needs no defence.

.github/workflows/verify-publish-credentials.yml
sscarduzio: Fixed in f7eaf773. That comment argued for the design rather than describing the code, which is the same fault as the file header. The step name says what the step does, and the reuse rule — one issue, commented on rather than a new issue every morning — sits in the script that enforces it.

.github/workflows/verify-publish-credentials.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  verify_sonatype_credentials:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
+        with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
+      - name: verify_sonatype_credentials
```
coutoPL: Shouldn't the name be human-readable here?

ci/verify-sonatype-credentials.sh
coutoPL, line 1: It would be good to have one script (e.g., `verify-sonatype-credentials.sh`) and two functions in it: `verify_credentials` and `report_credentials_failure`. The script could be sourced in the proper places, and the proper function could be called.

Now, we have the scripts separated, and `ci/report-credential-failure.sh` doesn't have good context attached. They are connected, so should be defined in the same place.

coutoPL (changes requested):

.github/workflows/verify-publish-credentials.yml
sscarduzio: Right. The step is now "Ask the staging API whether it accepts the credentials", and the job carries `name: Verify the Sonatype credentials`. d85a836

ci/verify-sonatype-credentials.sh
sscarduzio: Agreed, they belong together. `ci/report-credential-failure.sh` is gone. `ci/verify-sonatype-credentials.sh` is now a sourced lib (like `ci/e2e-tests-lib.sh`) with `verify_credentials` and `report_credentials_failure`, and the workflow sources it in both steps. Behaviour is the same; I checked the 401 path and the temp-file cleanup against the real API. d85a836

coutoPL (approved):
LGTM

sscarduzio:
Merged — and it cannot run yet. Flagging it here so it does not sit silent, which is the exact failure this was written to remove.

`schedule:` fires only from the repository's default branch, and that is `develop`, not `master`:

```
$ gh workflow run verify-publish-credentials.yml --ref master
HTTP 404: workflow verify-publish-credentials.yml not found on the default branch
```

`workflow_dispatch` is blocked by the same rule, so it cannot even be started by hand. My mistake on the base branch: the job began inside `ci.yml` on master pushes, and I did not revisit that choice when it became a scheduled workflow.

#1382 is the merge back — one commit, the squash of this PR. Once it lands the cron is live and the credentials rotated on 4 September finally get exercised.

Issue #129, opened 2016-11-29
Build for ES v2.4.2


reftel:
Hi,

ES 2.4.2 is out. Could you please release a build for it?

sscarduzio:
Hi @reftel a build of the latest version of the plugin is available now for 2.4.2

reftel:
Thanks! =)

PR #1383, opened 2026-09-12
Notify the e2e tests repo when a release publishes the plugin images

sscarduzio:
The bootstrap sweep in `beshu-tech/readonlyrest-e2e-tests` boots 34 released ELK versions against
the released `<elk>-ror-latest` plugin images. PR #129 there moved the sweep off its nightly cron
and onto a `repository_dispatch` of type `ror-plugins-released`. Nothing sends that event. This PR
sends the ES half of it.

## What it does

A new job, `notify_e2e_tests`, runs `ci/notify-e2e-tests.sh`, which sends
`POST /repos/beshu-tech/readonlyrest-e2e-tests/dispatches` with
`{"event_type":"ror-plugins-released","client_payload":{...}}`. The payload carries the plugin
version, the repo, the ref, the sha and the URL of the run that sent it. The receiving workflow
reads no key of it, so it is there for the person who reads the sweep's log.

## Why this job

`release_ror` is the only job that pushes `<elk>-ror-latest`: the tag comes from
`pushRorDockerImage`, which `release_ror_docker_image()` calls, and that call happens only in
`publish_one_version` with `mode = release`. A pre-build or a PR build never reaches it.

`notify_e2e_tests` declares `needs: [release_ror]` and no `if:`. The default gate is then "every
`release_ror` leg succeeded", which is exactly when the tags moved. A run that is not a release
skips `release_ror`, so this job is skipped too. "Re-run failed jobs" after a fixed leg re-runs this
job as well.

It runs on `ubuntu-latest`: one HTTP call needs no gradle, no docker and no self-hosted minute.

## The notification cannot fail the release

When this job starts, the zips are on S3, the images are on Docker Hub and the git tags are pushed.
A red run would say the release failed, and someone would repeat it. So the script has no `set -e`,
and every failure path prints a warning, names the manual command that starts the sweep, and exits
0. The warning also says that a 404 means the token cannot write to the e2e repo, because the
GitHub API hides that case behind "not found".

## The secret

`ROR_GH_TOKEN`, the cross-repo PAT this repo already uses to dispatch the ROR KBN pre-build and to
clone the e2e repo. No new secret.

**One thing for you to confirm:** the token needs write access to
`beshu-tech/readonlyrest-e2e-tests` for the dispatch endpoint. I can see it reads that repo (the
e2e clone in `ci/e2e-tests-lib.sh`) and writes to the ROR KBN repo, but I cannot read the token, so
I cannot prove it writes to the e2e repo. If it does not, the first release after this merge prints
the warning above and stays green, and the token needs the scope.

## Out of scope

The ROR KBN release pipeline must send the same event for the Kibana `-ror-latest` images. That is a
PR in `readonlyrest_kbn`.

## Branch

`docs/dev/branching.md` sends a pipeline change to `master`. This repo has no `master` ref today
(`git ls-remote --heads` lists `develop` and the feature branches only), so this targets `develop`,
the default branch. Merge it back to `master` in the usual way if that branch comes back.

## Verified

- `.github/workflows/ci.yml` parses with a duplicate-key-rejecting YAML loader; 18 jobs, the new one
  among them.
- `actionlint` reports the same 11 pre-existing alias findings as on `develop`, and none in the new
  job.
- `bash -n ci/notify-e2e-tests.sh`, plus a run with no token and a run with a bad token: both print
  the warning and exit 0. The bad-token run proves the request shape reaches the API (401 Bad
  credentials, not 422).
- Not verified: a real dispatch. Sending one would start 34 legs in the e2e repo, so the first real
  release is the first real test.

## Summary by CodeRabbit

* **New Features**
  * Added automatic notifications to the end-to-end test repository after successful plugin releases.
  * Release notifications include version and workflow metadata to support downstream test runs.
  * Notification failures provide warnings and manual follow-up instructions without failing the release.

* **Documentation**
  * Documented the release notification workflow, payload, permissions, and configuration requirements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -959,3 +959,22 @@ jobs:
       - name: Remove the PGP signing key
         if: always()
         run: rm -f .travis/secret.pgp
+
+  # ── NOTIFY_E2E_TESTS (after a successful release) ─────────────────────────
+  # The bootstrap sweep in the e2e tests repo boots every released ELK version against the
+  # `<elk>-ror-latest` images. release_ror is what moves those tags, so the sweep starts from this
+  # event instead of a nightly cron that asks the same question 30 times between two releases.
+  # No `if:`, on purpose: the default gate is "every release_ror leg succeeded", which is exactly
+  # when the tags moved. A skipped or failed release_ror skips this job too.
+  notify_e2e_tests:
+    needs: [ release_ror ]
+    runs-on: ubuntu-latest   # one HTTP call; no gradle, no docker, so it costs no self-hosted minute
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
+        with: { fetch-depth: 1, persist-credentials: false }
+      - name: Free host disk
+        run: ci/free-host-disk.sh
+      - name: Notify the e2e tests repo
+        env:
+          ROR_GH_TOKEN: ${{ secrets.ROR_GH_TOKEN }}
+        run: ci/notify-e2e-tests.sh
--- a/ci/CI.md
+++ b/ci/CI.md
@@ -52,6 +52,7 @@ directory contain the build logic; the workflow only orchestrates.
 | `e2e_tests` | Cypress e2e suite, one job per ES version | pushes + PRs (not drafts) |
 | `build_ror` | builds all plugin zips + bytecode-reuse guard | PRs |
 | `determine_ci_type` → `upload_pre_ror` / `release_ror` / `publish_mvn` | release pipeline | develop/master pushes + manual `release_without_testing` |
+| `notify_e2e_tests` | tells the e2e tests repo that the released images are published | every `release_ror` leg green |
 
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
coutoPL, line 1: IMO, we don't need this.
[See comment](https://github.com/sscarduzio/readonlyrest_kbn/pull/1039#discussion_r4003619999).

coutoPL (changes requested):
BTW, you didn't add the no-changelog label to the PR and you didn't add the changelog entry (it's ok obviously). But the PR convention workflow didn't catch that. Please check it and fix.

.github/workflows/ci.yml
sscarduzio: Agreed. The bootstrap sweep moves to a weekly cron in the e2e repo (#129), so no release event is needed. Closing this PR.

sscarduzio:
Closed: the e2e bootstrap sweep runs on a weekly cron (readonlyrest-e2e-tests #129), as coutoPL proposed. No release event is needed.

PR #1387, opened 2026-09-17
[RORDEV-2227] Skip build and republish for ES versions already tagged

coutoPL:
## Problem

A release rebuilt every ES version in a major, even versions a previous run already published. The tag check ran only before the S3 upload and the Docker push. It never ran before the build. One new ES version in a major still triggered a full rebuild of every other version in it.

## Fix

The release now checks git tags before it builds anything. It fetches every release tag for the current plugin version once, per ES major. It then skips the build, the repackage, and the publish for each version that tag already covers.

Function names now state what each step does: `publish_es_major`, `publish_module`, `publish_version_artifacts`, `push_ror_docker_image`.

## Observable result

A release that adds one ES version to a major no longer rebuilds the rest. The log prints one line per skipped version. A major with nothing new to publish runs no build and no repackage task. It still runs one gradle configure for each module, because the module reports its own ES versions, so a no-op rerun of ES8 costs 18 configures and 17 tag queries.

`ci/CI.md` now states that the tag is the publish record, and how to publish one version again: delete its tag on origin and run the release again. There is no force flag.

## Review round

Correctness:

- A registry that cannot answer no longer reads as "Elastic never published this image". `docker manifest inspect` returns the same status for both, so `docker_image_state` reads the message. A registry outage used to skip the image push, return success, and write the tag, which then blocked every later rebuild.
- The S3 upload reads the HTTP status. `curl -f` exits 0 on 301 and 307, which S3 answers for a wrong region and for a young bucket, so a version could get its tag with no zip on S3. A wrong argument count now exits 2 instead of 0.
- The version list of a module comes from a file, not from gradle stdout. One polluted line made `base_version` junk, dropped every other version from the list, and printed "nothing to build" in green. `printEsVersionsForModule`, `printNewestEsVersionForModule` and `printAllSupportedEsVersions` write their answer, as `printEsModules` already did.
- The tag query retries with backoff again. A transient `git ls-remote` failure cost a whole module attempt, with a cleanup and no pause.

Cleanups:

- `origin_tags <glob>` in `ci-lib.sh` is the one parser of `git ls-remote` output. `remote_tag_state` reads it, and `release_tag` sits beside it as the one owner of the tag shape.
- The tag list passes as a file path. A variable that held 265 tags appeared in full on every `bash -x` line that carried it: 9,459 characters, now 82.
- `pending_versions` lost its dead `mode` parameter. An empty tag file already keeps every version, which is what `upload_pre` hands over.
- One retry layer, not three. The upload repeats through `retry_with_backoff`, and the image push keeps its own registry-error retry. One broken push cost up to 27 multi-arch builds, now 9.

Logging:

- Every CI script sends its messages to standard error, through one `ci_log` in `ci/log.sh`. Standard output carries a value and nothing else, so a caller can capture a function's answer. Two scripts wrote a value and a message to the same stream.
- `ci/run-pipeline.sh` and `ci/ci-lib.sh` no longer read `GITHUB_ACTIONS` or write `::error::`. A failed step carries its own name and its own exit code.
- `docs/dev/code-style.md` gains one rule: state the contract, not the mechanism.


ci/publish-ror-plugins.sh
```diff
@@ -147,53 +205,43 @@ release_ror_docker_image() {
...
+    echo "Usage: publish_es_major <es major> <upload_pre|release>"
     return 1
   fi
   local es_major=$1 mode=$2
   local ror_version
   ror_version=$(gradle_property pluginVersion) || return 1
 
+  local origin_tags=""
+  if [ "$mode" = release ]; then
+    origin_tags=$(origin_release_tags "$ror_version") || return 1
```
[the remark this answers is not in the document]
coutoPL: Fixed at 2291b4330: the tag snapshot now refreshes at the top of every publish_module attempt, instead of once for the whole major. A retry now sees tags an earlier attempt in this run already wrote.

ci/publish-ror-plugins.sh
```diff
@@ -47,11 +47,52 @@ list_es_module_versions() {
...
+    if grep -qFx "$git_tag" <<< "$origin_tags"; then
+      echo ">>> $git_tag is already on origin. Skipping ES $version." >&2
+    else
+      printf '%s\n' "$version"
+    fi
+  done
+}
+
+# Builds the module's base version once and verifies bytecode reuse for the newest version.
+# Then it repackages and publishes each ES version that origin has not tagged yet.
```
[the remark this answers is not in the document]
coutoPL: Fixed at 2291b4330: the publish_module doc comment now says the tag skip applies in release mode only, and that upload_pre publishes every version regardless.

ci/publish-ror-plugins.sh
```diff
@@ -126,7 +192,7 @@ publish_module_versions() {
 }
 
 # Pushes the ES+ROR Docker image for one version
-release_ror_docker_image() {
+push_ror_docker_image() {
   local es_version=$1 module=$2
 
   if docker manifest inspect "docker.elastic.co/elasticsearch/elasticsearch:${es_version}" >/dev/null 2>&1; then
```
sscarduzio: A registry error on `docker manifest inspect` takes the same path as "Elastic never published this ES version": the image push is skipped, the function returns 0, and line 172 writes the tag. From now on that tag also stops every later rebuild of the image.

```
$ docker manifest inspect docker.elastic.co/elasticsearch/elasticsearch:9.99.99
no such manifest: ...                                   exit 1
$ docker manifest inspect docker.elastic.invalid/elasticsearch/elasticsearch:9.4.2
failed to configure transport: ... no such host        exit 1
```

Line 198 tests only the exit status. docker.elastic.co unreachable for one minute during a release -> `v<ror>_es<es>` is pushed with no `elasticsearch-readonlyrest:<es>-ror-<ror>` image -> every re-run prints "every version is already published" and an operator must delete the tag on origin by hand.

Pre-existing, but before this PR the tag only skipped the upload; now it is the only input to the build decision. Fix: capture the stderr of `docker manifest inspect`; take the skip path only when it matches `no such manifest|manifest unknown`, and return non-zero on any other error so the existing retries repeat it.

ci/publish-ror-plugins.sh
```diff
@@ -147,47 +213,32 @@ release_ror_docker_image() {
...
-    local tag_state
-    tag_state=$(remote_tag_state "$TAG") || return 1
-    if [ "$tag_state" = present ]; then
-      echo "$TAG is already on origin, so ES $es_version is published. Skipping."
-      return 0
-    fi
-  fi
-
   # publish always - even if this is not a release
   if ! ci/upload-files-to-s3.sh "$zip" "${zip}.sha512" "${ror_version}/"; then
```
sscarduzio: `ci/upload-files-to-s3.sh` cannot tell a failed upload from a redirected one, so a version can get its tag with no zip on S3, and this PR then never rebuilds it.

`ci/s3-uploader.sh:139-151`: the upload runs `curl -f` with no `-L`. `-f` fails only on status >= 400; a local test server answering 301 or 307 gives exit 0. S3 answers `301 PermanentRedirect` for a wrong region and `307 TemporaryRedirect` for a new bucket. With `S3_UPLOADER_DEBUG=1` the script drops `-f` altogether, so even a 403 exits 0.

Wrong `ROR_S3_REGION` in the repo variables, or a debug run -> `publish_version_artifacts` returns 0 -> line 172 tags -> the download page 404s and a re-release says "Nothing to build".

Pre-existing; the PR widens what the tag suppresses. Fix in `ci/s3-uploader.sh`: `--fail-with-body` in both branches, plus `--write-out '%{http_code}'` and accept only 200/201/204.

ci/publish-ror-plugins.sh
```diff
@@ -68,6 +110,28 @@ publish_module_versions() {
...
   echo ">>> Module $module: base ES $base_version, ${#versions[@]} version(s): ${versions[*]}"
 
+  # Publish newest-to-oldest so the most recent version is available first.
+  mapfile -t versions < <(printf '%s\n' "${versions[@]}" | tac)
+
+  # Fresh on every attempt: a retry of this module must see the tags an earlier attempt already
+  # wrote, or it rebuilds and re-uploads a version this run already published.
+  local origin_tags=""
+  if [ "$mode" = release ]; then
+    origin_tags=$(origin_release_tags "$ror_version") || return 1
```
sscarduzio: A transient failure of `git ls-remote` now costs one full module attempt, with `rm -rf` of the build dirs and no pause, where before this PR it was retried with a 15 s and a 30 s backoff.

`origin_release_tags` returns 1 -> line 120 returns 1 -> `publish_es_major:264-277` runs `cleanup_docker_and_build` and calls `publish_module` again at once. On `origin/master` the query sat in `publish_one_version` (line 158), inside the per-version loop with `sleep $((attempt * 15))`. The outer loop has no sleep, so a 10-second GitHub hiccup consumes all three attempts and the leg goes red.

Fix: in `origin_release_tags`, run the query through the helper the file already uses: `refs=$(retry_with_backoff git ls-remote --tags origin "refs/tags/v${ror_version}_es*")`.

ci/publish-ror-plugins.sh
```diff
@@ -47,10 +47,52 @@ list_es_module_versions() {
...
+
+# Builds the module's base version once and verifies bytecode reuse for the newest version.
+# Then it repackages and publishes each ES version. Release mode skips any version origin
+# already tagged. upload_pre publishes every version regardless.
 #   $1 mode (upload_pre|release)  $2 ror_version  $3 module
-publish_module_versions() {
+publish_module() {
   local mode=$1 ror_version=$2 module=$3
   local dist_dir="${module}/build/distributions"
   local es_jars_dir
```
sscarduzio: _Anchored at line 98; the code this is about is at `ci/publish-ror-plugins.sh:104`, which is outside the diff._

The early return at line 130 turns a polluted gradle stdout into a green skip of unpublished versions.

`list_es_module_versions` reads lines 1 and 2 of raw gradle stdout. The twin in this file, `list_es_modules` (lines 33-41), refuses to do that: "read THAT, never gradle stdout (configuration-time build-script logging can pollute it even under --quiet)", added in #1290 after #1287 hit real pollution. Nothing validates `base_version` or `versions`. With one leading junk line (`WARN x` / `8.1.0` / `8.1.0 8.10.0 8.11.0`) and 8.1.0 already tagged: `base_version=WARN x`, `versions=(8.1.0)`, `pending` is empty, the function prints "every version is already published. Nothing to build." and returns 0. 8.10.0 and 8.11.0 are never built. Before this PR the same pollution failed loudly in `buildRorPluginZip -PesVersion=WARN x`.

Fix: after line 104 reject any `base_version` or `versions[@]` entry that does not match `^[0-9]+\.[0-9]+\.[0-9]+$` (return 1, print the raw output), or give `printEsVersionsForModule` the same file handshake `printEsModules` has.

ci/publish-ror-plugins.sh
```diff
@@ -47,10 +47,52 @@ list_es_module_versions() {
...
+    return 1
+  }
+  printf '%s\n' "$refs" | awk '{print $2}' | sed -e 's#^refs/tags/##' -e 's/\^{}$//' | sort -u
+}
+
+# Prints the versions from $4.. that still need publishing, one per line. Release mode skips any
+# version whose tag is already in $3. upload_pre never tags, so it keeps every version there and
+# ignores $3.
+#   $1 mode  $2 ror_version  $3 origin_tags  $4.. versions
+pending_versions() {
```
sscarduzio, line 57: Decide: there is no way to force a rebuild of one tagged version in a release. The only escape hatch is `git push origin :refs/tags/v<ror>_es<es>` by hand, and no document in this repo says so.

`FORCE_REBUILD` exists only for the pre-build path (`ci-lib.sh:273`, `publish-pre-builds.yml:28`). `release.yml` has one input, `confirm_version`; `pending_versions` reads no override; `ci/CI.md` has no line that names the tag as the publish record. A zip found corrupt, or built from a bad commit -> re-dispatch -> "Nothing to build" -> the bad zip stays.

Either add `FORCE_REBUILD` here (same name as the pre-build path, wired to a `workflow_dispatch` input of `release.yml`) and one paragraph in `ci/CI.md` on the tag as the publish record and how to republish one version, or tell me the tag deletion is the intended procedure and document that instead.

ci/publish-ror-plugins.sh
```diff
@@ -47,10 +47,52 @@ list_es_module_versions() {
...
+  local mode=$1 ror_version=$2 origin_tags=$3
+  shift 3
+  local versions=("$@") version git_tag
+
+  if [ "$mode" != release ]; then
```
sscarduzio: The `mode` parameter and this early return are dead: an empty `origin_tags` gives the same result.

`pending_versions release 1.2.3 "" 8.1.0 8.0.0` prints both versions; `grep -qFx` against an empty string never matches. `publish_module` already branches on mode at line 119 and passes `""` for `upload_pre`, so the rule "upload_pre never filters" is enforced twice.

Fix: drop `$1 mode` and lines 76-79; signature `$1 ror_version  $2 origin_tags  $3.. versions`. Keep the mode branch at line 119 only.

ci/publish-ror-plugins.sh
```diff
@@ -47,10 +47,52 @@ list_es_module_versions() {
...
-# then repackages and publishes every supported ES version.
+release_tag() {
+  printf 'v%s_es%s\n' "$1" "$2"
+}
+
+# Prints, newline-separated, every release tag of $2 that exists on origin. One query for the
+# whole ES major, not one per version — with 17 versions in the ES8 major, that is 17 round
+# trips down to 1.
+#   $1 ror_version
+origin_release_tags() {
```
sscarduzio: `origin_release_tags` is a twin of `remote_tag_state` (`ci-lib.sh:302-314`): two parsers of the same `git ls-remote --tags origin refs/tags/...` output, in two files. And `tag()` (`ci-lib.sh:322`) calls `remote_tag_state` again for every version that `pending_versions` already proved absent, one round trip per tagged version, so the "17 round trips down to 1" in the comment above holds for the check only.

Fix: one primitive in `ci-lib.sh` next to `remote_tag_state`, `origin_tags <refs/tags glob>`, that prints the matching tag names (the body of today's `origin_release_tags`). Then `remote_tag_state` becomes `[ -n "$(origin_tags "$git_tag")" ]` and `origin_release_tags` becomes `origin_tags "v${ror_version}_es*"`. Move `release_tag` beside them.

ci/publish-ror-plugins.sh
```diff
@@ -47,10 +47,52 @@ list_es_module_versions() {
   ./gradlew ":${module}:printEsVersionsForModule" --quiet </dev/null
 }
 
-# Builds the module's base version once, verifies bytecode reuse for the newest version,
-# then repackages and publishes every supported ES version.
+release_tag() {
+  printf 'v%s_es%s\n' "$1" "$2"
+}
+
+# Prints, newline-separated, every release tag of $2 that exists on origin. One query for the
```
sscarduzio: nit: the comment names `$2`, the function takes only `$1`; and since 2291b433 the query runs once per `publish_module` attempt, so "17 round trips down to 1" is a story of an older revision: for ES8 that is 17 modules, 110 versions, 17 queries per rerun.

Fix: `# Prints, one per line, every release tag of ror_version $1 that origin holds. One query per publish_module attempt serves every version of the module (110 in ES8 against one per version).`

ci/publish-ror-plugins.sh
```diff
@@ -94,7 +155,7 @@ publish_module_versions() {
 
     local attempt published=0
     for attempt in 1 2 3; do
```
sscarduzio: Pre-existing, not for this PR unless you want it: this loop is a copy of `retry_with_backoff` (`ci-lib.sh:112`, same 15 s / 30 s delays), and the nesting is hidden: `push_ror_docker_image` retries registry errors 3x inside, this loop 3x, `publish_es_major` 3x, so one broken push can cost up to 27 multi-arch Docker builds.

If you take it: `retry_with_backoff publish_version_artifacts "$mode" "$ror_version" "$module" "$version" "$zip" || { echo "ERROR: publish of $module ES $version failed"; return 1; }`, then decide whether the per-version retry is still wanted given the inner one.

ci/publish-ror-plugins.sh
```diff
@@ -68,6 +110,28 @@ publish_module_versions() {
...
+  fi
+
+  # Checking is a git query. Building and repackaging costs minutes. Filtering here, before any
+  # build, skips that cost for every version already published.
+  local pending_output
+  pending_output=$(pending_versions "$mode" "$ror_version" "$origin_tags" "${versions[@]}") || return 1
+  local -a pending=()
+  [ -n "$pending_output" ] && mapfile -t pending <<< "$pending_output"
+
+  if [ "${#pending[@]}" -eq 0 ]; then
```
sscarduzio, line 119: Decide: the PR text says a major with nothing new "returns at once, with no build step at all". No build or repackage task runs, but one gradle configure per module still runs before this line: line 104 `list_es_module_versions` is `./gradlew :<module>:printEsVersionsForModule --quiet`, a full multi-project configure, and `publish_es_major:252` runs `printEsModules` once more. A no-op rerun of ES8 is 18 gradle configures plus 17 `git ls-remote` calls; the cost is unmeasured here, but one configure is tens of seconds warm and minutes cold on the shared box.

Either reword the PR text and `ci/CI.md` to "no build or repackage task runs; one gradle configure per module still does", or hoist the version query: let `printEsModules` also write each module's versions under `build/es-modules/`, so `publish_es_major` reads one file and passes `base_version` and `versions` to `publish_module`.

sscarduzio (commented):
**Verdict: COMMENT.** The skip logic is right; what needs work is what happens around it when something fails, because the tag is now the only input to the build decision.

Verified as sound (with stubs, a local bare origin and the real `origin` tags of 1.71.0): `origin_release_tags` strips peeled `^{}` refs and dedups; the `refs/tags/v<ror>_es*` glob cannot match another ror version (`v1.7.0` vs `v1.71.0`); `grep -qFx` keeps `8.1`, `8.1.0`, `8.10.0` apart; `upload_pre` ignores the tag list; an unreachable remote fails closed; a base version already tagged still builds the base zip and repackages only the pending versions; a failed tag push on one version makes the retry republish that version only (the 2291b433 fix holds); the tag is written last (S3, Docker, then tag), and git history shows every pipeline since 28e89fff2 did the same, so no existing tag predates its zip; no old function name survives anywhere in the repo, workflows or docs; the Kibana pipeline reads only the `-dev` pre-build images, which this PR does not touch.

The red `cve_check` is netty CVE-2026-89044 in `core`, not this PR.

Inline, in order of weight:
1. `push_ror_docker_image:198` — a registry error is read as "no base image", the version gets tagged without an image, and the tag now blocks the rebuild. **Fix in this PR.**
2. `publish_version_artifacts:222` — the S3 uploader accepts 3xx (and, in debug mode, anything) as success. Pre-existing; the tag now makes it permanent.
3. `publish_module:120` — a transient `ls-remote` failure lost its backoff in the move; it now burns a whole module attempt with cleanup. **Fix in this PR.**
4. `publish_module:104` — the version list is read from raw gradle stdout, and the new early return turns pollution into a green skip.
5. `pending_versions:71` — **decision:** no way to force a rebuild of one tagged version; not documented.
6. `pending_versions:76` — dead `mode` parameter.
7. `origin_release_tags:58` — twin of `remote_tag_state`; `tag()` still re-queries per version.
8. `:54` — nit, stale comment.
9. `:157` — pre-existing, optional: hand-rolled retry, up to 27 nested Docker builds.
10. `:130` — **decision:** "returns at once" is 18 gradle configures for ES8; reword or hoist.
11. `:126` — under `-x` the tag list floods the log; `local -; set +x`.

Not inline, for a follow-up: nothing reconciles tags, S3 and Docker Hub after a release, and nothing checks that the sources did not change between a tag's commit and HEAD when the same `pluginVersion` is re-released (all 1.69.1 tags sit on five commits with identical sources today, by discipline). A `verify_release` job at the end of `release.yml` would catch both.

_Deep review: whole-file read of `ci/publish-ror-plugins.sh`, `ci/ci-lib.sh`, `ci/s3-uploader.sh`, `release.yml`; five lenses, every finding executed or refuted._

ci/publish-ror-plugins.sh
```diff
@@ -68,6 +110,28 @@ publish_module_versions() {
...
+  # wrote, or it rebuilds and re-uploads a version this run already published.
+  local origin_tags=""
+  if [ "$mode" = release ]; then
+    origin_tags=$(origin_release_tags "$ror_version") || return 1
+  fi
+
+  # Checking is a git query. Building and repackaging costs minutes. Filtering here, before any
+  # build, skips that cost for every version already published.
+  local pending_output
+  pending_output=$(pending_versions "$mode" "$ror_version" "$origin_tags" "${versions[@]}") || return 1
```
sscarduzio: Under `bash -x` (`ci/run-pipeline.sh` line 1 is `#!/bin/bash -ex`) the tag list is echoed to the job log five times per module, so the one-line-per-skipped-version log this PR advertises is buried.

```
$ bash -c 'source ci/ci-lib.sh; source ci/publish-ror-plugins.sh; set -x; origin_tags=$(origin_release_tags 1.71.0); pending_output=$(pending_versions release 1.71.0 "$origin_tags" 8.18.1 8.18.0)' 2>&1 | wc -l
1874
```

Two versions, 1,874 xtrace lines: the 530-line `refs=` assignment, the 265-line `origin_tags=` assignment, the `pending_versions ... "$origin_tags"` call and `local ... origin_tags=$3`. The two `Skipping ES` lines are at 1865 and 1872. For ES8 that is about 32,000 lines per rerun.

Fix: make `local -` then `set +x` the first two statements of `origin_release_tags` and `pending_versions` (bash 4.4+, xtrace comes back on return), or pass the tag list as a temp file path instead of its content.

ci/ci-lib.sh
```diff
@@ -84,16 +88,37 @@ is_merge_back_of_master() {
...
 
+# Asks a registry about one image. Prints `present` or `absent`, and fails when it cannot ask, so
+# no caller reads an unreachable registry as an image that does not exist.
+#
+# Pass a name that carries no mirror: a cache can hold a stale answer for a tag we pushed a moment ago.
+docker_image_state() {
+  local image=$1 log
+
+  log=$(mktemp) || return 1
+  if docker manifest inspect "$image" >"$log" 2>&1; then
```
sscarduzio: The probe has no retry, so one transient error from docker.elastic.co now costs a whole module attempt: cleanup, two compiles, the base build, the repackage and the S3 upload of that version, again.

```
docker_image_state: docker manifest inspect  (once)           ci-lib.sh:103
push_ror_docker_image: ... || return 4                         publish-ror-plugins.sh:172
publish_module -> return 1 -> publish_es_major: cleanup_docker_and_build, attempt 2
```

Harness, probe answers `dial tcp: i/o timeout` for one version: 3 module attempts, each with `verifyRepackageBytecodeNewest`, `buildRorPluginZip`, the repackage and the upload, then the leg fails with no tag. The push two lines below the probe is wrapped in `retry_with_backoff --retry-if is_docker_registry_error`; the probe is not, and `is_docker_registry_error` names exactly these messages as transient. Before this PR the same error read as "absent" and skipped the push (round-1 #1); the new failure path is right, its cost is not.

Fix: run the probe through `retry_with_backoff --retry-if is_docker_registry_error` inside `docker_image_state` (capture the output instead of the temp file, or keep the file and retry the whole function body).

ci/mem-telemetry.sh
```diff
@@ -14,8 +14,14 @@
 #   ci/mem-telemetry.sh start <logfile>    prints the sampler PID; runs until stopped or 3h
 #   ci/mem-telemetry.sh stop <pid>
 #   ci/mem-telemetry.sh report <logfile>   worst-pressure sample + end-of-run OOM forensics
+#
+# `start` prints the sampler PID to standard output, because its caller captures it. Every other
+# message of this script is a diagnostic, and goes to standard error.
 set -u
 
+# shellcheck source=ci/log.sh
+source "$(dirname "${BASH_SOURCE[0]}")/log.sh"
```
sscarduzio: This PR and #1389 (approved, yours) conflict, and the resolved file sources a path that does not exist.

```
$ git merge-tree --write-tree <master+1387> pr-1389
CONFLICT (content): Merge conflict in ci/telemetry/mem-telemetry.sh   (the usage-comment hunk)
$ git show <tree>:ci/telemetry/mem-telemetry.sh | grep source
source "$(dirname "${BASH_SOURCE[0]}")/log.sh"        -> ci/telemetry/log.sh; ci/log.sh stays where it is
```

Whichever merges second, every instrumented job (8 of them after #1389) prints `ci/telemetry/log.sh: No such file or directory` on start, stop and report, and the usage branch calls an undefined `ci_log` (exit 127). #1389 also moves master's `ci/resource-sampler.sh`, which still logs with plain `echo`, so "every CI script sends its messages through one `ci_log`" is false after the two merge.

Fix: merge #1389 first, rebase this branch, point the source line at `../log.sh`, and convert the `echo` lines of `ci/telemetry/resource-sampler.sh` (keep its `::group::` and `GITHUB_STEP_SUMMARY` lines on stdout).

ci/ci-lib.sh
```diff
@@ -321,11 +364,11 @@ tag() {
...
-    echo "Git tag $GIT_TAG already exists on remote, skipping."
+    ci_log "Git tag $GIT_TAG is already on origin. This run skips it."
     return 0
   fi
 
-  echo "Tagging as $GIT_TAG"
+  ci_log "Tagging as $GIT_TAG."
   git config --global push.default matching
   git config --global user.email "support@readonlyrest.com"
   git config --global user.name "CI"
```
sscarduzio, line 385: _Anchored at line 374; the code this is about is at `ci/ci-lib.sh:377`, which is outside the diff._

`git push origin "$GIT_TAG"` lost its retry in the move. On master, `tag` ran inside `publish_one_version`, which the per-version loop repeated three times with 15 s and 30 s pauses. Now a failed push returns 1 from `tag`, `publish_module:144` returns 1, and the module attempt restarts: cleanup, configure, tag query, verify, base build, the repackage, the upload and the multi-arch push of that version, again.

One GitHub 5xx on a tag push → about 5 to 10 extra minutes and a build-dir wipe; three in one module → the leg fails. The artifact is unchanged, the cost is not.

Fix: `retry_with_backoff git push origin "$GIT_TAG"`.

ci/ci-lib.sh
```diff
@@ -321,11 +364,11 @@ tag() {
 
   state=$(remote_tag_state "$GIT_TAG") || return 1
```
sscarduzio, line 376: Round-1 #7, second half, still open: `tag()` asks origin again for every version that `pending_versions` proved absent seconds before, one `git ls-remote` per published version, up to 265 per full release. `remote_tag_state` has this one reader.

The check can only fire when a tag appears during the attempt. `release.yml:24-26` puts every publishing run of a branch in one concurrency group with `cancel-in-progress: false`, and the matrix legs cover different majors, so the only way is a tag pushed by hand mid-run. In that case the version was already uploaded and pushed a second time, and the log says "already on origin. This run skips it", the same wording as a normal skip.

Decide: delete `remote_tag_state` and this check (the push itself fails on an existing tag: `! [rejected] ... already exists`, verified against a bare origin), or keep it and make the message say what it now means: the tag appeared during this attempt and the artifacts of this version were published twice.

ci/CI.md
```diff
@@ -111,6 +111,27 @@ Both workflows share two composite actions rather than two copies:
 | `resolve-toolchains-image` | picks the mirror or Docker Hub, and returns the image | `ci_setup`, `release_setup` |
 | `verify-toolchains-image` | proves the baked Gradle home works | `toolchains_verify` (a whole job, to fail fast before the CI matrices), `discover` in `release.yml` (a step, being its first container job) |
 
+### The tag is the publish record
+
+A release tags every version it publishes, on origin: `v<pluginVersion>_es<esVersion>`. The next
+release reads those tags before it builds anything. It then skips the build, the repackage, the S3
+upload and the Docker push of every version a tag already covers, so a release that adds one ES
+version to a major rebuilds nothing else. It still runs one gradle configure for each module, which
```
sscarduzio: "a release that adds one ES version to a major rebuilds nothing else" is false for the module that owns the new version.

`publish-ror-plugins.sh:118-125` run `verifyRepackageBytecodeNewest` (two full compiles) and `buildRorPluginZip -PesVersion=$base_version` whenever `pending` is not empty, also when the base version itself is tagged, because the repackage starts from the base zip. Harness, only 8.18.2 pending: `VERIFY`, `BUILD zip base 8.18.0`, `REPACKAGE 8.18.2`.

An operator who deletes one tag expects seconds and gets the base compile plus the bytecode check of that module. Fix: "rebuilds only that module's base and the new version; every other module builds nothing", here and in the `publish_module` header (`publish-ror-plugins.sh:72-74`, which also lists the verify after the base build; it runs before).

ci/CI.md
```diff
@@ -111,6 +111,27 @@ Both workflows share two composite actions rather than two copies:
...
+is how a module reports the ES versions it owns.
+
+There is no force flag. `FORCE_REBUILD` applies to the pre-build images, not to a release.
+
+To publish one version again, after a corrupt zip or a build from the wrong commit, delete its tag
```
sscarduzio, line 125: The section states the delete-tag recipe and not the four facts an operator needs with it:

- **Which commit a re-run builds.** `release.yml:38` sets `RELEASE_SHA` to the workflow_run head or `github.sha`. "Re-run jobs" on the original run builds the original commit; a new dispatch builds the tip of master. The section names "a build from the wrong commit" as a reason to republish and does not say which re-run gives which commit, so the operator can rebuild the same wrong commit and tag it again.
- **A failed or cancelled leg needs no tag deletion.** Re-run the job; the run resumes from the tags. Only `.claude/skills/ror-release/SKILL.md:28` says so today.
- **A tag can exist with no Docker image behind it.** `push_ror_docker_image:173-178` returns 0 when Elastic has no base image, and `:144` writes the tag. If Elastic publishes the image later, every later release skips the version, and `elasticsearch-readonlyrest:<es>-ror-<ror>` never appears unless someone deletes the tag. Pre-existing behaviour; the doc is new.
- **Every push to master starts a release run**, a CI-only merge included, and that run reads the tags. A registry that does not answer now fails the leg instead of a silent skip plus tag.

Fix: add those four sentences here.

ci/CI.md
```diff
@@ -111,6 +111,27 @@ Both workflows share two composite actions rather than two copies:
...
+git push origin :refs/tags/v1.71.0_es8.18.1
+```
+
+The release then counts that version as unpublished: it builds it, uploads it over the old
+artifact, and writes the tag again. Every other version of the major stays skipped. Delete only the
+tags of the versions you want rebuilt.
+
 ### Two traps in `workflow_run`
 
 **`github.ref` and `github.sha` point at the default branch**, not at the commit CI tested. Every
```
sscarduzio, line 152: _Anchored at line 137; the code this is about is at `ci/CI.md:222`, which is outside the diff._

The table lists `printEsMajors`, `printTestMatrices` and `isPreReleaseVersion`, and it is the one place that states the "read the file, not stdout" rule. After this PR four more tasks write files that scripts must read:

```
printEsModules                build/es-modules/es<major>x.txt         publish-ror-plugins.sh:35
printEsVersionsForModule      <module>/build/es-modules/versions.txt  publish-ror-plugins.sh:45
printNewestEsVersionForModule <module>/build/es-modules/newest-version.txt  e2e-tests-lib.sh
printAllSupportedEsVersions   build/es-modules/es-versions.txt        actionstrings/run.sh
```

A new caller that follows the table parses stdout of the four unlisted tasks, and the bug this PR fixes comes back. Fix: add the four rows.

ci/publish-ror-plugins.sh
```diff
@@ -44,67 +44,105 @@ list_es_modules() {
...
   { read -r base_version; read -r -a versions; } < <(list_es_module_versions "$module")
   if [ -z "$base_version" ]; then
-    echo "ERROR: no versions for module $module"
+    ci_log "Module $module has no ES version."
     return 1
   fi
 
-  echo ""
-  echo ">>> Module $module: base ES $base_version, ${#versions[@]} version(s): ${versions[*]}"
+  ci_log "Module $module builds from base ES $base_version, and publishes ${#versions[@]} versions: ${versions[*]}."
```
sscarduzio: nit: this line counts the versions the run is about to skip. Harness, one pending: `publishes 3 versions: 8.18.0 8.18.1 8.18.2`, then `skips ES 8.18.1`, then two publishes. On a no-op rerun it is followed by "builds nothing".

Fix: "owns N versions" here, and one line with the pending count after line 111.

ci/publish-ror-plugins.sh
```diff
@@ -44,67 +44,105 @@ list_es_modules() {
...
+# over: that mode writes no tag.
+#   $1 ror_version  $2 file of origin tags  $3.. versions
+pending_versions() {
+  local ror_version=$1 origin_tags_file=$2
+  shift 2
+  local versions=("$@") version git_tag
+
+  for version in "${versions[@]}"; do
+    git_tag=$(release_tag "$ror_version" "$version")
+    if grep -qFx "$git_tag" "$origin_tags_file"; then
```
sscarduzio, line 70: nit: an unreadable tags file reads as "no tags". `grep -qFx` exits 2 on a missing file, the `if` takes the else branch, and every version is pending:

```
$ pending_versions 1.71.0 /nonexistent/tags.txt 8.18.1
grep: /nonexistent/tags.txt: No such file or directory
8.18.1        rc=0
```

Line 101 always creates the file today, so no input reaches this. The rule at `ci-lib.sh:347` is the one this would break: "a query failure read as 'absent' publishes the whole release again". Fix: `[ -r "$origin_tags_file" ] || { ci_log "..."; return 1; }` at the top of the function; the call site already has `|| return 1`.

ci/ci-lib.sh
```diff
@@ -290,23 +316,40 @@ publish_ror_es_prebuild_plugin() {
...
   if [ -n "$IMAGE_TAG" ]; then
     if ! retag_dev_image "$SOURCE_TAG" "${ES_VERSION}-ror-${IMAGE_TAG}"; then
-      echo "Failed to tag prebuild Docker image as ${ES_DEV_IMAGE_REPO}:${ES_VERSION}-ror-${IMAGE_TAG}"
+      ci_log "Cannot tag the ROR pre-build image as ${ES_DEV_IMAGE_REPO}:${ES_VERSION}-ror-${IMAGE_TAG}."
       return 6
     fi
   fi
 }
 
+release_tag() {
```
sscarduzio, line 332: nit: the description calls this "the one owner of the tag shape", and the function has no comment. The shape is also written in `ci/CI.md:116` and `.claude/skills/ror-release/SKILL.md:28`; the rule this PR adds to `docs/dev/code-style.md` asks a function comment to state what it returns.

Fix: one line: the tag of one published version; CI.md "The tag is the publish record" and the release skill quote this shape.

build-base/src/main/groovy/readonlyrest.plugin-common-conventions.gradle
```diff
@@ -358,29 +358,38 @@ tasks.register('packageRorPlugin') {
...
 
-// Prints two lines to stdout: the base ES version (EsVersions.oldest) on line 1, then all supported
-// versions oldest-first space-separated on line 2. Invoke with --quiet to suppress Gradle noise.
+// Prints the base ES version on line 1, and every supported version on line 2, space-separated.
+// Writes the same two lines to build/es-modules/versions.txt. A script reads that file: build-script
+// logging can pollute stdout even under --quiet.
 // Usage: ./gradlew :es818x:printEsVersionsForModule --quiet
 tasks.register('printEsVersionsForModule') {
     group = 'help'
     description = "Prints base version and all supported versions for ${project.name} to stdout"
```
sscarduzio: nit: `description` says "to stdout" here and at line 385, and both `doLast` blocks now write the file a script reads. `./gradlew tasks` hides the one output that matters.

Fix: append "and writes build/es-modules/versions.txt" and "... newest-version.txt".

ci/upload-es-artifacts.sh
```diff
@@ -18,16 +18,19 @@
...
-echo ">>> ($0) UPLOADING ES ARTIFACTS ..."
+# shellcheck source=ci/log.sh
+source "$(dirname "${BASH_SOURCE[0]}")/log.sh"
+
+ci_log "($0) UPLOADING ES ARTIFACTS ..."
 
 VERSIONS_INPUT="${ES_VERSIONS_TO_UPLOAD:-}"
 
 # Not an error here: this script is runnable by hand, and "show me what it would do" should not fail.
 # It is announced loudly all the same — a silent no-op is exactly what used to be mistaken for a
```
sscarduzio: nit: "It is announced loudly all the same" described the `::warning::` annotation on the run page. The diff replaces it with one `[CI]` line in the log and keeps `exit 0`, so a direct run with an empty list ends green with nothing on the summary. `mirror-es-libs.yml:66-68` guards the same state, so only a direct call reaches this branch.

Fix: exit non-zero here, or reword the comment to say the workflow makes the no-op loud, not this script.

ci/upload-cypress-artifacts-to-s3.sh
```diff
@@ -23,6 +23,9 @@
 set -uo pipefail
```
sscarduzio, line 21: _Anchored at line 23; the code this is about is at `ci/upload-cypress-artifacts-to-s3.sh:13`, which is outside the diff._

nit: "the gateway authorizes each prefix separately — the publishing creds are 403ed here" contradicts `ci/CI.md:457-459` in the same PR: "That was never true". The script reads `ROR_S3_ACCESS_KEY_ID` / `ROR_S3_SECRET_ACCESS_KEY` (lines 34-35), the pair `release_ror` uses.

Fix: delete the sentence and point to the one-credential-set rule in CI.md.

docs/dev/code-style.md
```diff
@@ -24,7 +24,7 @@ elif [ "${AGENT_ISSELFHOSTED:-0}" != "1" ]; then
```
sscarduzio, line 24: _Anchored at line 24; the code this is about is at `docs/dev/code-style.md:21`, which is outside the diff._

nit: the "Good" example models the log format and the variable this PR removes: `echo ">>> [host] ..."` is now `ci_log "[host] ..."` (`free-host-disk.sh:61`), and `AGENT_ISSELFHOSTED` exists nowhere in `ci/` or `.github/` (`ci/runner-detect.sh:9` uses `RUNNER_ENVIRONMENT`).

Fix: rewrite the example with `is_shared_docker_host` and `ci_log`.

sscarduzio (commented):
**Verdict: COMMENT.** Round 2. Ten of the eleven round-1 findings are fixed and verified; the eleventh is half done (below). Two things to fix before merge, one decision, then this is ready.

Since round 1 (2291b433 → ad6b8b41, 7 commits, 27 files), verified with the real `ci-lib.sh` and `publish-ror-plugins.sh` under `bash -ex` in a stubbed release harness, `origin_tags` against the real origin (265 tags), `docker manifest inspect` against real registries, `s3-uploader.sh` against a fake S3, and `git merge-tree` against master, develop and #1389:

- `origin_tags`: a permanent `ls-remote` failure propagates its status through `PIPESTATUS`, with and without `pipefail`, inside `$(...)` and into a file; a transient one is retried 15 s / 30 s. `retry_with_backoff` writes only to stderr, so the awk pipe sees ls-remote output alone. Two versions under `bash -x`: 57 lines, longest 82 characters (round 1: 1,874).
- `docker_image_state`: missing tag on docker.elastic.co, Docker Hub and ghcr → `absent`; missing repository, unreachable host, refused connection → failure, no tag written.
- `s3-uploader.sh`: 200/201/204 → 0; 301, 307, 403, dropped connection (`000`) → 1 with the body on stderr; `S3_UPLOADER_DEBUG=1` keeps stdout empty; 4 arguments → 2. The DGP form POST answers 204. No caller passes fewer than 5 arguments.
- Gradle file handshake: `rm -f` precedes gradle, gradle stdout goes to stderr, a polluted line never reaches `base_version`. No script or workflow still parses the stdout of the three tasks.
- `RETURN` trap in `publish_module`: fires on that function's return only, temp dir removed on every path.
- Module retry refreshes the snapshot; `upload_pre` makes 0 tag queries and writes 0 tags; one permanently broken push = 9 docker builds (3 × 3), matches the description.
- The `ci_log` conversion: I built the table of every changed script → its callers → stdout captured? Every captured value (`gradle_property`, `es_majors_matrix`, `is_pre_release_version`, `fetch.sh`, `e2e_elk_version_for_module`, `mem-telemetry start`, the flake-ledger table) still arrives on stdout alone. No workflow greps a reworded phrase or depends on a dropped `::error::`. `git merge-tree` PR → master and (master + PR) → develop: clean.
- The merge of this PR starts a release run under the new code (every master push does, see run 35337221887). master holds 1.71.0 and origin the 265 tags of exactly the versions the 34 modules declare, so the expected result is 34 lines "has every version on origin" and zero uploads. That run is the first execution outside a sandbox; watch it, and cancel a leg whose log shows `Uploading` for 1.71.0.

Inline, in order of weight:
1. `ci-lib.sh:103` — `docker_image_state` has no retry: one registry hiccup now costs a full module rebuild. **Fix before merge.**
2. `mem-telemetry.sh:23` — conflicts with #1389 (approved, same author); the merged file sources `ci/telemetry/log.sh`, which does not exist. **Rebase after #1389 merges.**
3. `ci-lib.sh:377` — the tag push lost its retry in the move.
4. `ci-lib.sh:365` — round-1 #7, second half: `tag()` still queries origin once per version. **Decide.**
5. `CI.md:119` — "rebuilds nothing else" is false for the module that owns the new version.
6. `CI.md:124` — the section lacks four facts an operator needs on release day.
7. `CI.md:222` — the gradle-task table misses the four tasks that now write files.
8. nits: `publish-ror-plugins.sh:93` and `:64`, `ci-lib.sh:325`, the two gradle `description` strings, `upload-es-artifacts.sh:29`, `upload-cypress-artifacts-to-s3.sh:13`, `code-style.md:21`.

PR description: "fetches every release tag … once, per ES major" → once per module attempt (the "Observable result" paragraph already says 17 queries); "One retry layer, not three" → two layers (3 × 3 = 9 is your own arithmetic).

Later, not this PR: nothing exercises `pending_versions`, `origin_tags` or `docker_image_state` before release day; my harness (a bare repo with tags as origin, stub gradlew/docker/uploader, `tag` overridden) fits in `ci/utils/` and `required_checks`. `is_pre_release_version` still reads gradle stdout (fails loud, not silent). `upload_using_aws_s3_uploader_to_key` has no caller and its header names `ROR_<STORE>_STORE_*` variables that do not exist. `release.yml:202` says "17 versions" (17 modules, 110 versions, no tests). `integration-tests/build.gradle:187` justifies `logger.lifecycle` with stdout parsing that no longer exists.

ci/mem-telemetry.sh
coutoPL: ok, will do it

sscarduzio (approved):
**Approve** at 4ee49b0e. Every round-2 finding is closed in the code and the docs, verified against the diff since ad6b8b41:

- `docker_image_state` runs the probe through `retry_with_backoff --retry-if is_docker_registry_error`; the tee'd output lands in the log file and the "no such manifest" answer still reads as `absent` on the first attempt.
- The tag push retries. The pre-check in `tag()` stays, with a message that says what it now means; `tag()` has one caller, in release mode, so no other path meets an existing tag legitimately.
- `pending_versions` rejects an unreadable tag list. The module log line says "owns N" and "publishes M of them".
- #1389 is merged into the branch; `mem-telemetry.sh` sources `../log.sh`, `resource-sampler.sh` takes its directory from the caller and logs through `ci_log`, and the two actions pass the directory, wrap the series in the `::group::` marker that `step-cost.py` keys on, and put the totals on the run page.
- CI.md: the "rebuilds nothing else" sentence is correct now, the four operator facts are there, and the gradle table lists every task that writes a file. Gradle descriptions, the three stale comments and the style-guide example are fixed.

Two lines of the PR description still say the old thing: "fetches every release tag … once, per ES major" (it is once per module attempt) and "One retry layer, not three" (two layers, 3 × 3). Fix the text when you merge, nothing in the code.

Reminder from round 2: the merge starts a release run under this code, the first real one. Expected: 34 lines "has every version on origin", zero uploads. Watch it.

PR #1389, opened 2026-09-18
[RORDEV-2220] One action pair starts and stops the job samplers

coutoPL:
This pull request cleans up the code of #1388, which merged without a review.

## What

`host-telemetry-start` and `host-telemetry-report` replace the four telemetry steps a job declared:
two inline `run:` steps for the resource sampler, and the `mem-telemetry-*` action pair. One pair of
steps now starts and stops both samplers, and the caller passes one `id`.

The actions own the file layout under `RUNNER_TEMP`, so no job names a log directory any more.

The three telemetry scripts move to `ci/telemetry`.

## Effect

- The memory sampler runs on all eight instrumented jobs, not on `it_linux` alone. Each job gets a
  `mem-telemetry-<id>` artifact and a `docker stats` call every 10 s.
- `resource-sampler.sh report` says that the sampler did not start, instead of printing a zero total.


ci/telemetry/resource-sampler.sh
```diff
@@ -43,6 +43,7 @@ case "${1:-}" in
     echo "resource sampler started: pid $(cat "$dir/pid"), $(nproc) cpus"
     ;;
   report)
+    if [ ! -s "$dir/samples" ]; then echo "no resource samples: the sampler did not start"; exit 0; fi
```
sscarduzio: nit: this guard returns before the `kill` on the next line, so a sampler that is alive with an empty series is never stopped, and the message says it did not start.

Local run: empty `samples`, a live PID in `pid` → `no resource samples: the sampler did not start`, exit 0, `kill -0 $PID` still succeeds after.

The runner kills orphans at job end, so nothing leaks past the job. Fix: move the `kill "$(cat "$dir/pid" 2>/dev/null)"` line above the guard, so `report` always stops what `start` started before it decides what to print.

.github/actions/host-telemetry-start/action.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  The file layout under RUNNER_TEMP is the contract between the two actions: the memory sampler
+  writes host-telemetry/mem-telemetry-<id>.log with its PID file next to it, and the resource sampler
+  owns host-telemetry/resource-sampler. The id is in the name, so the caller chooses the id, not the
+  shape.
+
+inputs:
+  id:
+    description: >-
+      What tells this job's telemetry from another's once the artifacts are downloaded side by side:
```
sscarduzio: nit: the description says the id is "the job name, plus the matrix leg", and two of the eight callers pass a different prefix.

```
ci.yml:541  name: build_es_${{ matrix.ES_MODULE }}   ci.yml:557  id: e2e_build_es_${{ matrix.ES_MODULE }}
ci.yml:584  name: e2e_${{ matrix.ES_MODULE }}        ci.yml:606  id: e2e_tests_${{ matrix.ES_MODULE }}
```

Run 35340491330 has `mem-telemetry-e2e_build_es_es94x` next to the job `build_es_es94x`. A reader who follows this description looks for `mem-telemetry-build_es_es94x` and finds nothing. All names stay unique, so this is naming only.

Fix: pass the job name in those two jobs (`build_es_${{ matrix.ES_MODULE }}` and `e2e_${{ matrix.ES_MODULE }}`), or say here that the prefix is free.

ci/telemetry/mem-telemetry.sh
```diff
@@ -11,9 +11,9 @@
 #   * docker stats for all containers (the ES/deps containers are siblings on the host daemon)
```
sscarduzio, line 12: _Anchored at line 11; the code this is about is at `ci/telemetry/mem-telemetry.sh:2`, which is outside the diff._

nit: the header still scopes the script to "the integration CI legs" (line 2) and "the job container's PID namespace (the gradle/JVM stack)" (line 10). This PR runs it on eight jobs, and `e2e_tests` has no job container: its `P` lines list `Runner.Worker`, `dockerd`, host processes.

Fix: say "every Linux CI job, inside the job container where the job has one, on the VM otherwise", and drop the gradle/JVM parenthesis.

sscarduzio (dismissed):
**Verdict: approve.** The action pair does what the description says, and the run of this PR proves it.

Verified as sound (run 35340491330, the eight instrumented jobs, the scripts executed locally):
- Every instrumented job has the docker CLI and the daemon socket inside the toolchains container, so `mem-telemetry.sh` writes real container lines in all eight; no report printed `docker-stats-unavailable`.
- `runner.temp` maps into the job container: the upload step found the log at `/home/runner/work/_temp/host-telemetry/…` and uploaded it. One `mem-telemetry-<id>` artifact exists for every instrumented job, and no name collides with another action's artifact.
- The PID file holds the PID only. `setsid` makes the collector its own group leader, so `kill -- -PID` also stops the `docker events` stream. `RUNNER_TEMP` is emptied at job start, and `start` truncates the sample file, so no leftover series can reach a later job.
- `resource-sampler.sh report` says "the sampler did not start" for a missing and for an empty series, exit 0.
- `step-cost.py` keys on the `resource samples` group marker and the `resource totals:` line, not on the step name; its parser read the four logs of this run.
- No reference to the old paths or the old action names remains anywhere in the repo. Both scripts keep mode 100755.

Inline, all nits: the kill order in `report`, the `id` description against two callers, the stale header of `mem-telemetry.sh`.

Two points for later, not this PR:
- `stop` kills the group at once, so about one report in six ends on a block without container lines (the sample in flight is cut). A TERM trap in `_collect` that lets `sample_once` finish would close that. Pre-existing.
- The `P` lines copy 90 characters of `ps args` into a public artifact. `s3-uploader.sh` puts the secret key in argv (`openssl … -hmac "AWS4$aws_sk"`, 70 characters, fits). Today no instrumented container job uploads to S3, so the exposure is nil; a future upload step in one of them has a small chance per upload of writing the key into `mem-telemetry-<id>`. The fix belongs in `s3-uploader.sh`: pass the key through the environment and give openssl `-macopt hexkey:` instead of `-hmac`.

sscarduzio (approved):
Re-approved at fa8a8d7f. The three nits are done: `kill` now runs before the empty-series guard (verified in the file), the e2e ids equal the job names and the `e2e_build_es_` exception is stated where the id is defined, and the header of `mem-telemetry.sh` describes where the sampler runs now. Merge when CI is green.

